PR #1, opened 2020-06-11
Remove non-kotlin files and README text.

zach-klippenstein:
Resolves https://github.com/square/workflow/issues/1187.

Corresponds to https://github.com/square/workflow/pull/1198.


zach-klippenstein:
Not sure why the gradle verification is failing, the JAR is clearly present.

zach-klippenstein:
- Fixed the github workflow to not specify path filter and not try to run mkdocs.
- Fixed `lint_docs.sh` to use the correct paths and filters.
- Fixed lint issues with the now-linted README in the terminal sample.

PR #12, opened 2020-06-16
Introduce runningSideEffect API

zach-klippenstein:
First part of Kotlin implementation of https://github.com/square/workflow/issues/1021.

Kotlin counterpart to https://github.com/square/workflow/pull/1174.

This implementation intentionally does not run the side effect coroutine
on the `workerContext` `CoroutineContext` that is threaded through the
runtime for testing infrastructure. Initially, workers ran in the same
context as the workflow runtime. The behavior of running workers on a
different dispatcher by default (`Unconfined`) was introduced in
https://github.com/square/workflow/pull/851 as an optimization to reduce
the overhead for running workers that only perform wiring tasks with
other async libraries. This was a theoretical optimization, since running
on the `Unconfined` dispatcher inherently involves less dispatching work,
but the overhead of dispatching wiring coroutines was never actually shown
to be a problem. Additionally, because tests often need to be in full
control of dispatchers, the ability to override the context used for
workers was introduced in https://github.com/square/workflow/pull/940,
which introduced `workerContext`. I am dropping that complexity here
because it adds a decent amount of complexity to worker/side effect
machinery without any proven value. It is also complexity that needs to
be documented, and is probably just more confusing than anything. The
old behavior for workers is maintained for now to reduce the risk of this
change, but side effects will always run in the workflow runtime's
context. This is nice and simple and unsurprising and easy to reason
about.

## Checklist

- [x] Unit Tests
- ~~UI Tests~~ (will be covered by existing UI tests when workers are migrated)

### DIFF
--- a/workflow-core/src/main/java/com/squareup/workflow/RenderContext.kt
+++ b/workflow-core/src/main/java/com/squareup/workflow/RenderContext.kt
@@ -116,6 +116,31 @@ interface RenderContext<StateT, in OutputT : Any> {
     key: String = "",
     handler: (T) -> WorkflowAction<StateT, OutputT>
   )
+
+  /**
+   * Ensures [sideEffect] is running with the given [key].
+   *
+   * The first render pass in which this method is called, [sideEffect] will be launched in a new
+   * coroutine that will be scoped to the rendering [Workflow]. Subsequent render passes that invoke
+   * this method with the same [key] will not launch the coroutine again, but let it keep running.
+   * Note that if a different function is passed with the same key, the side effect will *not* be
+   * restarted, the new function will simply be ignored. The next render pass in which the
+   * workflow does not call this method with the same key, the coroutine running [sideEffect] will
+   * be
+   * [cancelled](https://kotlinlang.org/docs/reference/coroutines/cancellation-and-timeouts.html).
+   *
+   * The coroutine will run with the same [CoroutineContext][kotlin.coroutines.CoroutineContext]
+   * that the workflow runtime is running in. The side effect coroutine will not be started until
+   * _after_ the first render call than runs it returns.
+   *
+   * @param key The string key that is used to distinguish between side effects.
+   * @param sideEffect The suspend function that will be launched in a coroutine to perform the
+   * side effect.
+   */
+  fun runningSideEffect(
+    key: String,
+    sideEffect: suspend () -> Unit
+  )
 }
 
 /**
--- a/workflow-runtime/src/main/java/com/squareup/workflow/internal/RealRenderContext.kt
+++ b/workflow-runtime/src/main/java/com/squareup/workflow/internal/RealRenderContext.kt
@@ -33,6 +33,7 @@ import kotlinx.coroutines.channels.SendChannel
 class RealRenderContext<StateT, OutputT : Any>(
   private val renderer: Renderer<StateT, OutputT>,
   private val workerRunner: WorkerRunner<StateT, OutputT>,
+  private val sideEffectRunner: SideEffectRunner,
   private val eventActionsChannel: SendChannel<WorkflowAction<StateT, OutputT>>
 ) : RenderContext<StateT, OutputT>, Sink<WorkflowAction<StateT, OutputT>> {
 
@@ -50,6 +51,13 @@ class RealRenderContext<StateT, OutputT : Any>(
       worker: Worker<T>,
       key: String,
       handler: (T) -> WorkflowAction<StateT, OutputT>
+    )
+  }
+
+  interface SideEffectRunner {
+    fun runningSideEffect(
+      key: String,
+      sideEffect: suspend () -> Unit
     )
   }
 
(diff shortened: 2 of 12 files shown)


zach-klippenstein:
This PR is ready for review.

workflow-runtime/src/main/java/com/squareup/workflow/internal/RealRenderContext.kt
```diff
@@ -53,6 +54,13 @@ class RealRenderContext<StateT, OutputT : Any>(
     )
   }
 
+  interface SideEffectRunner {
```
zach-klippenstein, line 57: Can this be internal?

workflow-runtime/src/main/java/com/squareup/workflow/internal/WorkflowNode.kt
```diff
@@ -282,6 +308,16 @@ internal class WorkflowNode<PropsT, StateT, OutputT : Any, RenderingT>(
     return WorkerChildNode(worker, key, workerChannel, handler = handler)
   }
 
+  private fun createSideEffectNode(
+    key: String,
+    sideEffect: suspend () -> Unit
+  ): SideEffectNode {
+    // workerContext is guaranteed to not contain a Job.
+    val scope = this + CoroutineName("sideEffect[$key] for $id") + workerContext
```
zach-klippenstein, line 316: Coroutine name should override any name in worker context - switch order and add tests for name and also base/worker context order, including test for default worker dispatcher (switch to unconfined?) and overridden dispatcher.

workflow-runtime/src/main/java/com/squareup/workflow/internal/WorkflowNode.kt
```diff
@@ -239,6 +261,10 @@ internal class WorkflowNode<PropsT, StateT, OutputT : Any, RenderingT>(
     // Tear down workflows and workers that are obsolete.
     subtreeManager.commitRenderedChildren()
     workers.commitStaging { it.channel.cancel() }
+    // Side effect jobs are launched lazily, since they can send actions to the sink, and must not
+    // be started until after context is frozen.
```
zach-klippenstein, line 265: "and can only be started after"

workflow-runtime/src/main/java/com/squareup/workflow/internal/WorkflowNode.kt
```diff
@@ -239,6 +261,10 @@ internal class WorkflowNode<PropsT, StateT, OutputT : Any, RenderingT>(
...
+    sideEffects.forEachStaging { it.job.start() }
```
zach-klippenstein, line 266: Add tests for multiple side effects (both not started until after render) and to ensure side effects with different keys are handled independently.
- both started in same render pass
- staggered
  - overlapping
  - non-overlapping

workflow-runtime/src/main/java/com/squareup/workflow/internal/SideEffectNode.kt
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.workflow.internal
+
+import com.squareup.workflow.internal.InlineLinkedList.InlineListNode
+import kotlinx.coroutines.Job
+
+/**
+ * TODO write documentation
```
zach-klippenstein, line 22: TODO

workflow-runtime/src/main/java/com/squareup/workflow/internal/RealRenderContext.kt
zach-klippenstein: Nope – it's a parameter to a public constructor of a public type.

workflow-core/src/main/java/com/squareup/workflow/RenderContext.kt
```diff
@@ -116,6 +116,31 @@ interface RenderContext<StateT, in OutputT : Any> {
...
+   * this method with the same [key] will not launch the coroutine again, but let it keep running.
+   * Note that if a different function is passed with the same key, the side effect will *not* be
+   * restarted, the new function will simply be ignored. The next render pass in which the
+   * workflow does not call this method with the same key, the coroutine running [sideEffect] will
+   * be
+   * [cancelled](https://kotlinlang.org/docs/reference/coroutines/cancellation-and-timeouts.html).
+   *
+   * The coroutine will run with the same [CoroutineContext][kotlin.coroutines.CoroutineContext]
+   * that the workflow runtime is running in. The runtime also provides the option to pass
+   * additional context elements to be used only for side effects.
```
zach-klippenstein, line 134: This is false, the runtime only provides the worker context override internally for testing infra, it's not exposed publicly.

workflow-runtime/src/main/java/com/squareup/workflow/internal/WorkflowNode.kt
```diff
@@ -63,7 +67,7 @@ internal class WorkflowNode<PropsT, StateT, OutputT : Any, RenderingT>(
   private val idCounter: IdCounter? = null,
   initialState: StateT? = null,
   private val workerContext: CoroutineContext = EmptyCoroutineContext
-) : CoroutineScope, WorkerRunner<StateT, OutputT> {
+) : CoroutineScope, WorkerRunner<StateT, OutputT>, SideEffectRunner {
```
rjrjr, line 70: Out of scope (ahem) for this PR, but: I forgot that this class implements `CoroutineScope` directly. Isn't that discouraged these days?

Have we already had this conversation?

zach-klippenstein: You are very right. I already fixed it in one of the other 300 branches I put up this weekend but i forget which one 😅

PR #24, opened 2020-06-18
Introduce helpers for working with Sinks of WorkflowActions from suspend functions.

zach-klippenstein:
- `Sink.sendAndAwaitApplication()`
- `Flow.collectToSink()`

These helpers will be used to implement Workers using side effects (#12).
GUWT workers are described in https://github.com/square/workflow/issues/1021.

## Checklist

- [x] Unit Tests
- ~~UI Tests~~

### DIFF
--- a/workflow-core/src/main/java/com/squareup/workflow/Sink.kt
+++ b/workflow-core/src/main/java/com/squareup/workflow/Sink.kt
@@ -15,6 +15,11 @@
  */
 package com.squareup.workflow
 
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.suspendCancellableCoroutine
+import kotlin.coroutines.resume
+
 /**
  * An object that receives values (commonly events or [WorkflowAction]).
  * [RenderContext.actionSink] implements this interface.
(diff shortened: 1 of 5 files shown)


workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
```diff
@@ -57,6 +55,12 @@ interface WorkflowAction<StateT, out OutputT : Any> {
     nextState = mutator.state
   }
 
+  @Suppress("DEPRECATION")
+  @Deprecated("Implement Updater.apply")
+  fun Mutator<StateT>.apply(): OutputT? {
+    throw UnsupportedOperationException()
+  }
+
```
zach-klippenstein, line 63: Moved this method below the non-deprecated one so that the non-deprecated one is what gets resolved for `WorkflowAction.apply` links in kdoc.

workflow-core/src/main/java/com/squareup/workflow/Sink.kt
```diff
@@ -47,3 +52,61 @@ fun <T1, T2> Sink<T1>.contraMap(transform: (T2) -> T1): Sink<T2> {
...
+
+/**
+ * Sends [action] to this [Sink] and suspends until after [action]'s [WorkflowAction.apply] method
+ * has been invoked. Since a [Sink] may be backed by an unbounded buffer, this method can be used
+ * to apply backpressure to the caller when there are a lot events being sent to the workflow
+ * runtime.
+ *
+ * If this coroutine is cancelled before the action gets applied, the action will not be applied.
+ *
+ * This method is intended to be used from [RenderContext.runningSideEffect].
```
rjrjr, line 93: sample use?

zach-klippenstein: Not sure what value a sample would add: there's really only one possible way to use this method.

workflow-core/src/main/java/com/squareup/workflow/Sink.kt
```diff
@@ -47,3 +52,61 @@ fun <T1, T2> Sink<T1>.contraMap(transform: (T2) -> T1): Sink<T2> {
...
+ */
+@ExperimentalWorkflow
+suspend fun <StateT, OutputT : Any> Sink<WorkflowAction<StateT, OutputT>>.sendAndAwaitApplication(
+  action: WorkflowAction<StateT, OutputT>
+) {
+  suspendCancellableCoroutine<Unit> { continuation ->
+    val resumingAction = action<StateT, OutputT>({ "sendAndAwaitExecution($action)" }) {
```
steve-the-edwards, line 100: Nit: Don't you want the name to be `sendAndAwaitApplication`?

workflow-core/src/test/java/com/squareup/workflow/SinkTest.kt
```diff
@@ -0,0 +1,149 @@
...
+    ExperimentalWorkflow::class
+)
+class SinkTest {
+
+  private val sink = RecordingSink()
+
+  @Test fun `collectToActionSink sends action`() {
+    runBlockingTest {
+      val flow = MutableStateFlow(1)
+      val collector = launch {
```
steve-the-edwards, line 41: Nit: collector is slightly overloaded here given the coroutine context.  Maybe sinkCollector? or collectorJob?

Issue #26, opened 2020-06-19
No id set on the view that replaces a WorkflowViewStub after update() is called


bnvinay92:
Not having an id breaks constraint layout. Android's `ViewStub` allows you to declare an inflatedId which seems to solve this issue.

bnvinay92:
Fixed in https://github.com/square/workflow-kotlin/pull/28

PR #27, opened 2020-06-19
Preserve view id when replacing WorkflowViewStubs

bnvinay92:
Fixes https://github.com/square/workflow-kotlin/issues/26

Was not sure when it would be useful for a workflowViewStub and the view that replaces it to have different view ids, so this PR preserves it.

## Checklist

- ~Unit Tests~
- ~UI Tests~

### DIFF
--- a/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/WorkflowViewStub.kt
+++ b/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/WorkflowViewStub.kt
@@ -22,7 +22,8 @@ import android.view.ViewGroup
 
 /**
  * A placeholder [View] that can replace itself with ones driven by workflow renderings,
- * similar to [android.view.ViewStub].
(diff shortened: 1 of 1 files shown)


workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/WorkflowViewStub.kt
```diff
@@ -95,7 +96,10 @@ class WorkflowViewStub @JvmOverloads constructor(
       is ViewGroup -> viewEnvironment[ViewRegistry].buildView(rendering, viewEnvironment, parent)
           .also { buildNewViewAndReplaceOldView(parent, it) }
       else -> viewEnvironment[ViewRegistry].buildView(rendering, viewEnvironment, actual.context)
-    }.also { actual = it }
+    }.also {
+      it.id = actual.id
+      actual = it
```
rjrjr, line 101: We tried this first at first. It…didn't work out so well.

PR #29, opened 2020-06-19
Introduce ImposterWorkflow to properly support operator workflows and GUWT.

### LINKED ISSUE #23: Introduce KeyedWorkflow
```kotlin
/**
 * This is an optional interface that [Workflow]s can implement.
 * When implemented by a [Workflow], the [identifier] property
 * will be used to dedup the worker in addition to the workflow's
 * concrete type and string key from its parent.
 */
interface IdentifiableWorkflow : Workflow<*, *, *> {
  /** Must implement equals and hashcode. */
  val identifier: Any
}
```

Note that this is not being done on Swift. Workflows in swift all get converted into the `AnyWorkflow` type, which already contains the reified type of the actual workflow. That is effectively the functionality that this is adding.

This is required to fully support workflow operators. Workflow operators have a problem: a given operator will always produce a workflow of the same concrete type, but needs to somehow communicate to the runtime that two operator workflows with different upstream workflows are distinct. It also needs to ensure that if the same operator is applied to the same upstream workflow in two different places, the runtime still considers them equivalent. We're currently using a hack that solves the first problem, but not the second: operators are inline functions that create anonymous functions. Because they're inline, each call site causes the compiler to generate a different concrete class. However this means that if you call the same `workflow.operator` combination in, for example, two different switch cases, the workflows will be considered distinct and will be stopped and restarted.

It's also required to implement GUWT Workers (https://github.com/square/workflow/issues/1021), which are effectively another type of operator workflow.

This `IdentifiableWorkflow` solution allows operator workflows to be real concrete types (not inline) and retain all the semantics you'd expect of a normal workflow. Also, because this is functionality that is only required for infrastructure-level workflows (e.g. operators and GUWT Workers), putting this functionality in a separate interface means not changing the main API of `Workflow` to introduce additional surface area that is potentially confusing for developers using workflows normally – out of sight, out of mind. Note that this interface should still be public, to be available to build third-party infrastructure.

zach-klippenstein:
Fixes #23, although ended up going with a slightly different approach than what was described in that issue.
Instead of making an interface that gives the workflow a key, removed the hard-coded reliance on the workflow's
`KClass` in `WorkflowId` into a strongly-typed `WorkflowIdentifier` type that is public in the core moddule,
along with an extension property to get the identifier for a workflow. `WorkflowIdentifier` is aware of wrapper
workflows via the `ImposterWorkflow` interface (instead of `KeyedWorkflow`). `WorkflowIdentifier` is
effectively a linked-list of workflow classes (although the classes may also be represented as strings, to
support rehydration from snapshots).

## Checklist

- [x] Unit Tests
- [x] UI Tests


workflow-core/src/main/java/com/squareup/workflow/WorkflowIdentifier.kt
```diff
@@ -0,0 +1,158 @@
...
+  init {
+    require((type == null) xor (typeString == null)) {
+      "Either type or type string must be passed, not both (type=$type, typeString=$typeString)"
+    }
+  }
+
+  /**
+   * The fully-qualified name of [type]. Computed lazily.
+   */
+  private val typeString: String by lazy(PUBLICATION) { typeString ?: type!!.java.name }
```
zach-klippenstein, line 65: I don't think this laziness is worth it, but highlights an issue with this approach. Whereas the old code would look up the actual classes for each restored workflow, this code just stores the string instead. Since comparisons with string identifiers require turning classes into names, this means that any time a tree is restored from a snapshot, all workflow classes will have their names reflected on every render pass. I haven't benchmarked this, but it seems less than ideal.

zach-klippenstein: We should probably go back to just looking up classes for now, until we measure this change.

zach-klippenstein: Went back to eagerly resolving classes (i.e. the old behavior). If we want to change this later, it should be a separate change anyway.

zach-klippenstein:
I think we might actually want to make the `StatefulWorkflow` that is returned by `StatelessWorkflow.asStatefulWorkflow` an `ImpostorWorkflow` as well, since it needs to assume the identity of the `StatelessWorkflow` subclass. Should `ImpostorWorkflow` basically just merge into `Workflow` itself?

workflow-core/src/main/java/com/squareup/workflow/ImpostorWorkflow.kt
```diff
@@ -0,0 +1,37 @@
...
+ */
+package com.squareup.workflow
+
+/**
+ * Optional interface that [Workflow]s should implement if they need the runtime to consider their
+ * identity to include a child workflow's identity. Two [ImpostorWorkflow]s with the same concrete
+ * class, but different [realIdentifier]s will be considered different workflows by the runtime.
+ *
+ * This is intended to be used for helper workflows that implement things like operators by wrapping
+ * and delegating to other workflows, and need to be able to express that the identity of the
```
rjrjr, line 24:
```suggestion
 * and delegating to other workflows, and which need to be able to express that the identity of the
```

workflow-core/src/main/java/com/squareup/workflow/WorkflowIdentifier.kt
```diff
@@ -0,0 +1,140 @@
...
+import okio.BufferedSink
+import okio.BufferedSource
+import okio.EOFException
+import kotlin.LazyThreadSafetyMode.PUBLICATION
+import kotlin.reflect.KClass
+
+/**
+ * Represents a [Workflow]'s "identity" and is used by the runtime to determine whether a workflow
+ * is the same as a workflow as was rendered in a previous render pass, in which case its state
+ * should be re-used, or if it's a new workflow and needs to be started.
```
rjrjr, line 27:
```suggestion
 * is the same as one that was rendered in a previous render pass, in which case its state
 * should be re-used; or if it's a new workflow and needs to be started.
```

workflow-core/src/main/java/com/squareup/workflow/WorkflowIdentifier.kt
```diff
@@ -0,0 +1,140 @@
...
+ *
+ * The identity consists primarily of the workflow's concrete type (i.e. the class that implements
```
rjrjr, line 29:
```suggestion
 * A workflow's identity consists primarily of its concrete type (i.e. the class that implements
```

workflow-core/src/main/java/com/squareup/workflow/WorkflowIdentifier.kt
```diff
@@ -0,0 +1,140 @@
...
+ * the [Workflow] interface). Two workflows of the same concrete type are considered identical.
+ *
+ * However, if a workflow class implements [ImpostorWorkflow], the identifier will also include
+ * that workflow's [ImpostorWorkflow.realIdentifier].
+ *
+ * @constructor
+ * The workflow's type must be passed _either_ as a [KClass] or a string, but only one must be
```
rjrjr, line 36: Is this stale? The `type` param below isn't nullable.

workflow-core/src/main/java/com/squareup/workflow/WorkflowIdentifier.kt
```diff
@@ -0,0 +1,140 @@
...
+ * non-null. The [KClass] should be passed in if it's available, to avoid reflection calls upfront,
+ * but the class's name may be passed instead if the [KClass] is not available, eg. when restoring
+ * this class from its serialized form.
+ *
+ * The workflow type must be final, since a non-final class or interface means two different
+ * workflow types could be considered identical.
+ *
+ * @param type The [KClass] of the [Workflow] this identifier identifies.
+ * @param typeString The fully-qualified name of the [KClass] of the workflow this identifier
```
rjrjr, line 45: Not a param.

workflow-core/src/main/java/com/squareup/workflow/WorkflowIdentifier.kt
zach-klippenstein: yep

rjrjr:
> I think we might actually want to make the `StatefulWorkflow` that is returned by `StatelessWorkflow.asStatefulWorkflow` an `ImpostorWorkflow` as well, since it needs to assume the identity of the `StatelessWorkflow` subclass. Should `ImpostorWorkflow` basically just merge into `Workflow` itself?

Agree on the first point. I'm not sure what merging `ImposterWorkflow` into `Workflow` would actually look like. Are you saying to just add a `val id: WorkflowId` to the `Workflow` interface, with a default implementation? Seems simpler.

zach-klippenstein:
I was kind of vague on purpose. In the runtime we convert a Workflow to a StatefulWorkflow early, but need to pass around the original ID of the original workflow. Turns out this isn't actually as important as I thought though, so I didn't explore this idea more. That's not a great solution either though because we can never trust the Workflow completely to combine its own id with its proxied one.

PR #32, opened 2020-06-22
Introduce TreeSnapshot.

### LINKED ISSUE #31: Combined tree snapshots should have their own type
This type should be exposed in the public API as the output alongside renderings, and the input to the runtime. To support testing, we can also provide a helper to create a tree snapshot given the snapshot of just the foot workflow.

zach-klippenstein:
This is a concrete type that represents a complete snapshot of the entire workflow
tree, vs the snapshot of a single workflow's state. This makes the testing and runtime
API more typesafe, since it's impossible to accidentally pass a single workflow's
snapshot when a tree is expected. It also makes the internal code more clean because
it's clear which type of snapshot is expected where.

Closes #31 and unblocks https://github.com/square/workflow/issues/1192.

Open to suggestions on the name.

## Checklist

- [x] Unit Tests
- [x] UI Tests


workflow-runtime/src/main/java/com/squareup/workflow/TreeSnapshot.kt
```diff
@@ -0,0 +1,116 @@
...
+   */
+  internal val workflowSnapshot: Snapshot? by lazy(NONE) {
+    workflowSnapshot?.takeUnless { it.bytes.size == 0 }
+  }
+
+  /**
+   * Writes this [Snapshot] and all its children into a [BufferedSink]. The component snapshots can
+   * be restored with [readFrom].
+   */
+  fun writeTo(sink: BufferedSink) {
```
zach-klippenstein, line 50: Should probably just make this and `readFrom` deal with `ByteString`s?

zach-klippenstein: Done.

workflow-runtime/src/main/java/com/squareup/workflow/TreeSnapshot.kt
```diff
@@ -0,0 +1,122 @@
...
+    /**
+     * A [TreeSnapshot] that has a null [workflowSnapshot] and no [childTreeSnapshots].
+     */
+    val NONE = TreeSnapshot(Snapshot.EMPTY, emptyMap())
+
+    /**
+     * Returns a [TreeSnapshot] that only contains a [Snapshot] for the root workflow, and no child
+     * snapshots.
+     */
+    fun forRootOnly(rootSnapshot: Snapshot?): TreeSnapshot =
```
rjrjr, line 79: Annotate as for testing only? Maybe that'd be too much nannying.

zach-klippenstein: I don't think that's necessary. If you want to pass in a root-only snapshot in prod, it will be extremely clear that's what you're doing.

zach-klippenstein:
Made the `childSnapshots` parameter to `TreeSnapshot` a provider function that gets cached by a `lazy` property. This way it's more typesafe (map value type also changed from `Snapshot` to `TreeSnapshot`), but the parsing of the snapshot tree overall stays lazy.

rjrjr:
Nice update, still LG

PR #39, opened 2020-06-23
Fixes WorkflowFragment under Jetpack navigation.

### LINKED ISSUE #14: WorkflowFragment layout does not render after navigating away and back using Jetpack navigation
When using the **AndroidX Navigation** library to navigate to a second fragment from a `WorkflowFragment` and navigating back (back press), the `WorkflowLayout` does not render.

I added some breakpoints in `WorkflowLayout` and it seems like the `onViewAttachedToWindow` in `takeWhileAttached` is never invoked when navigating back to the `WorkflowFragment`.

I've [extended the `hello-workflow-fragment` sample](https://github.com/ychescale9/workflow-kotlin/tree/yc/jetpack-nav/samples/hello-workflow-fragment/src/main/java/com/squareup/sample/helloworkflowfragment) to demonstrate this by setting up a simple navigation graph: clicking on a button in `HelloWorkflowFragment` navigates to the `SecondFragment` using `root.findNavController().navigate(R.id.action_helloFragment_to_secondFragment)`.

rjrjr:
Closes #14.

Verified via manual test against https://github.com/ychescale9/workflow-kotlin/tree/yc/jetpack-nav/samples/hello-workflow-fragment/src/main/java/com/squareup/sample/helloworkflowfragment

Tracking the lack of proper testing in #40.

## Checklist

- [ ] Unit Tests
- [ ] UI Tests
- [ ] I have made corresponding changes to the documentation

### DIFF
--- a/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/WorkflowFragment.kt
+++ b/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/WorkflowFragment.kt
@@ -62,14 +62,14 @@ abstract class WorkflowFragment<PropsT, OutputT : Any> : Fragment() {
   protected abstract val viewEnvironment: ViewEnvironment
 
   /**
-   * Called from [onViewStateRestored], so it should be safe for implementations
+   * Called from [onCreateView], so it should be safe for implementations
    * to call [getActivity].
    */
   protected abstract fun onCreateWorkflow(): Config<PropsT, OutputT>
 
   /**
    * Provides subclasses with access to the products of the running [Workflow].
-   * Safe to call after [onViewStateRestored].
+   * Safe to call after [onCreateView].
    */
   protected val runner: WorkflowRunner<OutputT> get() = _runner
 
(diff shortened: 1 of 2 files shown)


rjrjr:
Contrary to the discussion in the ticket, I didn't make `View.takeWhileAttached` smarter. So far we've been consistent about ensuring that `LayoutRunner` et al are hooked up before views are attached, and that seems like something it would be good to keep. The dumb `View.takeWhileAttached` accidentally enforces that.

workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/WorkflowFragment.kt
```diff
@@ -78,14 +78,9 @@ abstract class WorkflowFragment<PropsT, OutputT : Any> : Fragment() {
...
-  override fun onViewStateRestored(savedInstanceState: Bundle?) {
-    super.onViewStateRestored(savedInstanceState)
-
-    _runner = WorkflowRunner.startWorkflow(this, ::onCreateWorkflow)
-
-    (view as WorkflowLayout).start(runner.renderings, viewEnvironment)
+    return WorkflowLayout(inflater.context).also { newView ->
+      _runner = WorkflowRunner.startWorkflow(this, ::onCreateWorkflow)
+      newView.start(runner.renderings, viewEnvironment)
+    }
```
zach-klippenstein, line 84: Could you add a comment explaining why this is necessary?

rjrjr:
```
    // https://github.com/square/workflow-kotlin/issues/14
    // We're careful to start up the workflow runtime before the view is attached,
    // since that's what LayoutRunner promises. When we're sloppy about that, we
    // break things like Jetpack Navigation and nested fragments.
```

rjrjr:
Oops, found some more broken kdoc.

samples/hello-workflow-fragment/src/main/res/layout/hello_goodbye_layout.xml
```diff
@@ -15,16 +14,24 @@
...
-      android:layout_height="match_parent"
-      android:gravity="center"
+      android:layout_height="0dp"
+      android:layout_weight="1"
       android:autoSizeTextType="uniform"
-    />
+      android:gravity="center"
+      />
+
+  <EditText
```
rjrjr, line 31: Argh, didn't mean to merge this. So tired.

workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/WorkflowFragment.kt
zach-klippenstein: Awesome, thanks!

PR #45, opened 2020-06-25
Implement TracingDiagnosticListener as a WorkflowInterceptor.

zach-klippenstein:
Two commits, first commit keeps it implementing the old `WorkflowDiagnosticListener` and just moves the golden value test over, and then the second commit removes the `WorkflowDiagnosticListener` interface and converts call sites.

There aren't many interesting changes in the golden value text file, as you'd expect. The only change is actually that `onPropsChanged` for the root and for the root workflow are called in slightly different places, and the worker IDs are different. Everything else is exactly the same. The changes to it in the second commit just update the name.

## Checklist

- [x] Unit Tests
- [x] UI Tests


samples/dungeon/app/src/main/java/com/squareup/sample/dungeon/MainActivity.kt
```diff
@@ -34,7 +34,7 @@ class MainActivity : AppCompatActivity() {
       WorkflowRunner.Config(
           workflow = component.timeMachineWorkflow,
           props = "simple_maze.txt",
-          diagnosticListener = TracingDiagnosticListener(traceFile)
+          interceptors = listOf(TracingWorkflowInterceptor(traceFile))
```
rjrjr, line 37: SOOOoooo much better.

workflow-runtime/src/test/java/com/squareup/workflow/internal/RealRenderContextTest.kt
```diff
@@ -140,6 +140,7 @@ class RealRenderContextTest {
     val handler = context.onEvent<String> { expectedUpdate }
     assertTrue(eventActionsChannel.isEmpty)
 
+    context.freeze()
```
rjrjr, line 143: Seems unrelated. Opportunistic cleanup?

zach-klippenstein: Nope, required because above change makes `onEvent` delegate to `send`, which will throw if we don't freeze the context before sending to it. I also want to make `onEvent` an extension method on the context.

PR #46, opened 2020-06-26
Migrate WorkflowTester to renderWorkflowIn.

zach-klippenstein:
This exposed a few actual bugs in `RenderIdempotencyChecker`, the fact that
we still need `workerContext` for `renderWorkflowIn` (until GUWT workers are implemented),
and a bug in the sample `TimeMachineWorkflow` machinery.

It also introduced an explicit behavior change for `WorkflowTester`: `sendProps`
no longer triggers new render passes when props are sent that are equivalent to
the previous props. This is expected, since props is now a `StateFlow`, which implies
`distinctUntilChanged`-like behavior. This only required fixing one test that
was using the old behavior as a trick to trigger a new render pass with the same
state. If this affects actual tests, they can be fixed by wrapping in a workflow
that accepts some arbitrary changing value, and sending new props that change that
value instead of sending the same underlying props again.

## Checklist

- [x] Unit Tests
- ~~UI Tests~~


workflow-testing/src/main/java/com/squareup/workflow/testing/RenderIdempotencyChecker.kt
```diff
@@ -41,12 +41,17 @@ object RenderIdempotencyChecker : WorkflowInterceptor {
...
+    val recordingContext = RecordingRenderContext(context)
+    proceed(props, state, recordingContext)
+
+    // The second render pass should not actually invoke any real behavior.
+    recordingContext.startReplaying()
+    return proceed(props, state, recordingContext)
+        .also {
+          // After the verification render pass, any calls to the context _should_ be passed
+          // through, to allow the real context to run its usual post-render behavior.
+          recordingContext.stopReplaying()
```
rjrjr, line 53: So hacky.

workflow-testing/src/main/java/com/squareup/workflow/testing/WorkflowTester.kt
```diff
@@ -280,18 +267,87 @@ fun <T, PropsT, StateT, OutputT : Any, RenderingT>
...
+}
+
+@OptIn(ExperimentalStdlibApi::class)
+private fun <T> unwrapCancellationCause(block: () -> T): T {
+  try {
+    return block()
+  } catch (e: CancellationException) {
+    throw generateSequence(e as Throwable) { e.cause }
+        // Stop the sequence if an exception's cause is itself.
+        .scanReduce { acc, cause ->
```
rjrjr, line 347: acc?

rjrjr: accumulated?

zach-klippenstein: `error`

workflow-testing/src/main/java/com/squareup/workflow/testing/RenderIdempotencyChecker.kt
zach-klippenstein: Yes, and now all the hacks are entirely contained in this one class!

PR #48, opened 2020-06-26
Remove WorkflowDiagnosticListener.

zach-klippenstein:
This change removes the core `WorkflowDiagnosticListener` interface and wiring. Note that it does not attempt to replace the existing functionality with anything. `TracingDiagnosticListener` was already ported to `WorkflowInterceptor`, and everything else is stuff that should never have lived in the runtime module anyway, and can be added back in later if we need it.


zach-klippenstein:
@rjrjr Feel free to merge if you approve.

PR #64, opened 2020-06-28
Rename EmittedOutput to WorkflowOutput, move to workflow-core.

zach-klippenstein:
This simplifies the API of `WorkflowAction.applyTo`, removes the need for
`OutputHolder` in the core tests and `MaybeOutput` in `workflow-runtime`.
`EmittedOutput` is left as a deprecated typealias to ease migration.

I was initially reluctant to do something like this because I didn't want
Workflow to provide yet another generic `Optional` type. I forgot that the
testing library was already doing that though, and I think that using a
nullable "holder" type is a gross enough API that it won't likely be used
outside of the intended uses.


workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
```diff
@@ -32,21 +32,16 @@ interface WorkflowAction<StateT, out OutputT> {
...
+      private set
 
     /**
      * Sets the value the workflow will emit as output when this action is applied.
      * If this method is not called, there will be no output.
      */
     fun setOutput(output: O) {
-      this.output = output
-      isOutputSet = true
+      this.output = WorkflowOutput(output)
```
rjrjr, line 43: OMG so much better.

PR #68, opened 2020-06-28
Emit output before next render pass.

### LINKED ISSUE #54: Root runWorkflowLoop has inconsistent output behavior compared to its WorkflowNode children
In `RealWorkflowLoop#runWorkflowLoop`, the order of operations after an output for the root workflow's runWorkflowLoop is:

1. `output = select { .. }`
2. `doRender`
3. `output?.let { onOutput }`

However, child workflows have their outputs immediately executed when the WorkflowAction is run.

* From `WorkflowNode#tick` we call `WorkflowNode#applyAction`, but notice that we call `output?.let(emitOutputToParent)` immediately.

This difference in behavior manifests in the root workflow performing an extra render pass before the root workflow runner can receive an output and potentially end the workflow. 

This extra render pass can cause a crash if the workflow does not expect to have render called after emitting an output (because the presumption is that the output stop the workflow from running).

---

To work around this, I can wrap my root workflow `R` in another workflow `P` that has 2 states: Running and CancelledAfterOutput. Initially, `P` will be in the Running state, and call renderChild on `R`. But after 1 output from `R`, `P` transitions to the CancelledAfterOutput state, which doesn't call renderChild on `R`. This "fixes" the errant extra render pass.

zach-klippenstein:
Two commits:
1. Add a passing test for existing behavior.
2. Change behavior and test.

Fixes #54.

## Checklist

- [x] Unit Tests

### DIFF
--- a/workflow-runtime/src/test/java/com/squareup/workflow/RenderWorkflowInTest.kt
+++ b/workflow-runtime/src/test/java/com/squareup/workflow/RenderWorkflowInTest.kt
@@ -391,8 +391,8 @@ class RenderWorkflowInTest {
     assertEquals(
         listOf(
             "rendering({no output})",
-            "rendering(output)",
-            "output(output)"
+            "output(output)",
+            "rendering(output)"
         ),
         events
     )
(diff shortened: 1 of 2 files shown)


workflow-runtime/src/test/java/com/squareup/workflow/RenderWorkflowInTest.kt
```diff
@@ -363,5 +363,9 @@ class RenderWorkflowInTest {
     assertFalse(scope.isActive)
   }
 
+  @Test fun `output is emitted before next render pass`() {
+    TODO()
```
rjrjr, line 367: Ayup.

zach-klippenstein:
Yep, this PR is a draft for a reason 😅

rjrjr:
> Yep, this PR is a draft for a reason 😅

D'oh! Not used to looking for that yet.

Issue #55, opened 2020-03-26
Kotlin: Annotate workflow-ui-* as experimental


rjrjr:
The UI classes are robust, but we're still messing with the API a bit and ramping up their production use. Meanwhile, core is basically done. Let's decouple their releases.

PR #69, opened 2020-06-28
Remove unnecessary data classes from public, non-experimental APIs.

### LINKED ISSUE #57: Either remove data classes, or make them not data classes
Data classes are bad for library binary compatibility (see Jake's [blog post](https://jakewharton.com/public-api-challenges-in-kotlin/)). The only data class in our core public API is `WorkflowTestParams` (and some diagnostic stuff, but that's going to remain unstable even in 1.0).

UI data classes are another matter.

zach-klippenstein:
The only such data classes were `WorkflowTestParams` and `RenderingAndSnapshot`.
The former has no need to be a data class, as it's just passed as configuration
to the test function, and the latter is effectively just an alias for `Pair` and
the shape is fundamental to the workflow runtime design so it won't change.

There are still some data classes in the UI modules, but #55 will mark all those
as `ExperimentalWorkflowApi` before releasing 1.0.0.

Fixes #57.


zach-klippenstein:
I've been carefully to avoid introducing new ones, and most of our public APIs are just interfaces anyway.

PR #70, opened 2020-06-28
Group top-level properties and functions in core module into multifile classes.

zach-klippenstein:
This doesn't affect Kotlin consumers in any way, but cleans up the JVM classes we
generate to be simply `Workflows` and `Snapshots` instead of a whole bunch of
`FooKt` classes.

The API diff is the only really interesting diff here.


workflow-core/src/main/java/com/squareup/workflow/ExperimentalWorkflowApi.kt
```diff
@@ -13,6 +13,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+@file:JvmMultifileClass
+@file:JvmName("Workflows")
```
rjrjr, line 17: TIL

PR #77, opened 2020-06-29
Renames WorkflowAction.nextState to state

### LINKED ISSUE #9: Rename Kotlin action nextState to state!
It's so confusing. Downside is bad effects on inline `action {}`s. That might be the upside as well.

rjrjr:
Undoing a crime against nature and common sense.

Closes #9

### DIFF
--- a/samples/containers/app-poetry/src/main/java/com/squareup/sample/poetryapp/PoemsBrowserWorkflow.kt
+++ b/samples/containers/app-poetry/src/main/java/com/squareup/sample/poetryapp/PoemsBrowserWorkflow.kt
@@ -61,7 +61,7 @@ object PoemsBrowserWorkflow :
   }
 
   private fun choosePoem(index: SelectedPoem) = action("goToPoem") {
-    nextState = index
+    state = index
   }
 
   private val clearSelection = choosePoem(-1)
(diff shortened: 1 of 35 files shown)


rjrjr:
Oops, I was behind `main`. Updating…

rjrjr:
@zach-klippenstein Any idea what that dokka failure means?

zach-klippenstein:
```
e: /home/runner/work/workflow-kotlin/workflow-kotlin/workflow-runtime/src/test/java/com/squareup/workflow/RenderWorkflowInTest.kt: (377, 15): Unresolved reference: nextState
```

It looks like a compilation error, but I'm not sure why the assemble task didn't fail. Does it build locally?

rjrjr:
```
gw clean build connectedCheck
```

Is there more to run these days?

rjrjr:
Did find a lot of stale `nextState` in deprecation messages. Fixing…

rjrjr:
But there is no `nextState` in the class it's complaining about. :/

zach-klippenstein:
Yea, everything builds fine for me locally too, including `dokka`. Let's see if the rebuild passes I guess.

rjrjr:
Found it, in workflow-runtime. Don't understand how top level build missed that.

rjrjr:
Or a rebase weirdness? I don't know, tired, it works now.

zach-klippenstein:
I must be tired too, still confused.

forrestbice:
@rjrjr thanks for making this change! 🙌  Excited for this to make it to PoS

Blisse:
Perhaps silly, but would've been happy to also see immutable val called `currentState` too.

PR #78, opened 2020-06-29
Cleanup the RenderTestResult interface after introducing WorkflowOutput.

zach-klippenstein:
After #49 made outputs nullable, and introduced new `RenderTestResult`
methods to handle nullable outputs (`verifyActionState`, `verifyActionOutput`,
and `verifyNoActionOutput`), #64 came along and introduced a dedicated
type for nullable outputs, `WorkflowOutput`. This type means that we can
revert the `RenderTestResult` changes made in #49 and return to the simpler
API with only `verifyAction` and `verifyActionResult` methods. This change
does exactly that.


zach-klippenstein:
cc @bencochran

PR #81, opened 2020-06-29
Gives WorkflowAction access to props.

### LINKED ISSUE #56: Pass props to WorkflowActions
It's come up a couple of times, at least in Kotlin, that `WorkflowAction`s get the state of the current workflow but not its props. Since we consider props to be part of the workflow's overall "state", it seems unintuitive that we only pass part of that to actions.

The reason for _not_ doing this was, initially, to discourage side effects from being performed in actions. However, we realized that is too restrictive (see [this Slack discussion](https://workflow-community.slack.com/archives/CHB9VKA10/p1572625798145200)).

_Note for Swift this means passing the `Workflow` value itself into the action, since there's no separate props type._

_Kotlin version of square/workflow#919._

rjrjr:
Closes #56.

### DIFF
--- a/workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
+++ b/workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
@@ -24,7 +24,7 @@ import com.squareup.workflow.WorkflowAction.Updater
 /**
  * An atomic operation that updates the state of a [Workflow], and also optionally emits an output.
  */
-interface WorkflowAction<StateT, out OutputT> {
+interface WorkflowAction<in PropsT, StateT, out OutputT> {
   @Deprecated("Use Updater")
   class Mutator<S>(var state: S)
 
@@ -34,7 +34,10 @@ interface WorkflowAction<StateT, out OutputT> {
    *
    * @param state the state that the workflow should move to. Default is the current state.
    */
-  class Updater<S, in O>(var state: S) {
+  class Updater<out P, S, in O>(
+    val props: P,
+    var state: S
+  ) {
     internal var output: WorkflowOutput<@UnsafeVariance O>? = null
       private set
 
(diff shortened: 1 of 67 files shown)


workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
```diff
@@ -73,12 +76,12 @@ interface WorkflowAction<StateT, out OutputT> {
...
+    fun <PropsT, StateT, OutputT> noAction(): WorkflowAction<PropsT, StateT, OutputT> =
+      NO_ACTION as WorkflowAction<Any?, StateT, OutputT>
 
     /**
-     * Convenience function that returns a [WorkflowAction] that will just set the state to [newState]
-     * (without considering the current state) and optionally emit an output.
+     * Convenience function that returns a [WorkflowAction] that will just set the state to
+     * [newState] (without considering the current state) and optionally emit an output.
      */
     @Deprecated(
```
rjrjr, line 86: @zach-klippenstein  We should delete these, they're so dangerous. Is that tracked?

zach-klippenstein: I don't think so. We don't have any issues for removing deprecated APIs afaik.

rjrjr: These are the only ones I really care about. After this merges I'll put up a PR to snip them. Not worth tracking.

zach-klippenstein: I thought they were still extremely popular.

rjrjr: They are, but they're also often extremely a bad idea, and misused more often than not. And easy enough to recreate on your own w/o us sanctioning them.

rjrjr:
> The only reason that the render context needs to be aware of the props type is so that type inference can infer the type of actions when sent to action sinks and stuff, right?

And for the actions tied to any worker or workflow output.

workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
rjrjr: https://github.com/square/workflow-kotlin/pull/85

PR #84, opened 2020-06-29
Introduce unsnapshottableIdentifier() to allow transient WorkflowIdentifiers.

### LINKED ISSUE #79: Figure out how to support GUWT worker identity with ImpostorWorkflow
In https://github.com/square/workflow/issues/961#issuecomment-649776503 I missed the fact that, even without adding keys to Workflow, we still need a way to express a worker's type through ImpostorWorkflow. We currently use KType for worker identity, but that type is not serializable, and definitely not deserializable. 

We don't want to take on the major change of adding a non-snapshottable _workflow type_, which leaves us with the following options:

1. Write our own (de)serializer for KType. This could be hard to do right, there are a lot of edge cases. It's also a waste of effort, since we have no actual state to snapshot anyway.
2. Figure out some dirty hack for making WorkflowIdentifier accept non-serializable types and avoid snapshotting.

Since 1 is wasted work, I think 2 is better, even though i am not optimistic we'll be able to come up with a solution that isn't hacky. As long as give the hack enough friction, it is unlikely to wreak a lot of havoc.

zach-klippenstein:
This fixes #79 and unblocks the ability to implement GUWT Workers as Workflows.
`WorkflowIdentifier` can now indicate that it is not serializable by returning
null from `toByteString()`, in which case neither it nor the snapshot of the
workflow it identifies will be serialized. Because these identifiers do not need
to be serialized, they can contain arbitrary values such as `KType`s, which is
how Workers distinguish themselves for example. Workers also do not need to
be snapshotted, so this works out.

### DIFF
--- a/workflow-core/api/workflow-core.api
+++ b/workflow-core/api/workflow-core.api
@@ -172,12 +172,12 @@ public final class com/squareup/workflow/WorkflowIdentifier {
 	public static final field Companion Lcom/squareup/workflow/WorkflowIdentifier$Companion;
 	public fun equals (Ljava/lang/Object;)Z
 	public fun hashCode ()I
+	public final fun toByteStringOrNull ()Lokio/ByteString;
 	public fun toString ()Ljava/lang/String;
-	public final fun write (Lokio/BufferedSink;)V
 }
 
 public final class com/squareup/workflow/WorkflowIdentifier$Companion {
-	public final fun read (Lokio/BufferedSource;)Lcom/squareup/workflow/WorkflowIdentifier;
+	public final fun parse (Lokio/ByteString;)Lcom/squareup/workflow/WorkflowIdentifier;
 }
 
 public final class com/squareup/workflow/WorkflowOutput {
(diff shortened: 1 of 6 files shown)


workflow-core/api/workflow-core.api
```diff
@@ -168,16 +168,16 @@ public final class com/squareup/workflow/WorkflowAction$Updater {
...
 }
 
-public final class com/squareup/workflow/WorkflowIdentifier {
+public abstract class com/squareup/workflow/WorkflowIdentifier {
 	public static final field Companion Lcom/squareup/workflow/WorkflowIdentifier$Companion;
-	public fun equals (Ljava/lang/Object;)Z
-	public fun hashCode ()I
+	protected abstract fun getProxiedIdentifier ()Lcom/squareup/workflow/WorkflowIdentifier;
+	protected abstract fun getTypeName ()Ljava/lang/String;
+	public abstract fun maybeToByteString ()Lokio/ByteString;
```
zach-klippenstein, line 175: This makes me nervous, these shouldn't be publicly-visible.

workflow-core/src/main/java/com/squareup/workflow/WorkflowIdentifier.kt
```diff
@@ -31,61 +33,82 @@ import kotlin.reflect.KClass
...
-  private val typeString: String by lazy(PUBLICATION) { type.java.name }
+  private val typeName: String by lazy(PUBLICATION) {
+    if (type is KClass<*>) type.java.name else type.toString()
+  }
 
   /**
-   * Returns a description of this identifier including the name of its workflow type and any
-   * [proxiedIdentifier]s. Computes [typeString] if it has not already been computed.
+   * If this identifier is snapshottable, returns the serialized form of the identifier.
+   * If it is not snapshottable, returns null.
```
rjrjr, line 74: This doesn't strike me as hacky at all.

workflow-core/src/main/java/com/squareup/workflow/WorkflowIdentifier.kt
```diff
@@ -31,61 +33,82 @@ import kotlin.reflect.KClass
...
    */
-  override fun toString(): String =
-    generateSequence(this) { it.proxiedIdentifier }
-        .joinToString { it.typeString }
-        .let { "WorkflowIdentifier($it)" }
+  fun maybeToByteString(): ByteString? {
```
rjrjr, line 76: `toByteStringOrNull`?

zach-klippenstein: Done.

PR #85, opened 2020-06-29
Deletes deprecated enterState and modifyState

rjrjr:
A tiny bit of convenience, a lot of landmines.


rjrjr:
Oops, we're using `emitOutput` in a test.

rjrjr:
I'll close it then.

PR #87, opened 2020-06-30
Make RealRenderContext internal.

zach-klippenstein:
The only reason it wasn't already was that it used to be that we needed to
access it from the testing module. That is no longer true.

### DIFF
--- a/workflow-runtime/src/main/java/com/squareup/workflow/internal/RealRenderContext.kt
+++ b/workflow-runtime/src/main/java/com/squareup/workflow/internal/RealRenderContext.kt
@@ -24,7 +24,7 @@ import com.squareup.workflow.Workflow
 import com.squareup.workflow.WorkflowAction
 import kotlinx.coroutines.channels.SendChannel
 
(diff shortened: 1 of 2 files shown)


workflow-runtime/src/main/java/com/squareup/workflow/internal/RealRenderContext.kt
```diff
@@ -24,7 +24,7 @@ import com.squareup.workflow.Workflow
 import com.squareup.workflow.WorkflowAction
 import kotlinx.coroutines.channels.SendChannel
 
-class RealRenderContext<out PropsT, StateT, OutputT>(
+internal class RealRenderContext<out PropsT, StateT, OutputT>(
```
rjrjr, line 27: Wow.

zach-klippenstein: ABI golf: what's the smallest change you can make to reduce your public ABI as much as possible.

zach-klippenstein:
Test failed due to #88, re-running.

PR #89, opened 2020-06-30
Update RenderTester to support ImpostorWorkflow.

zach-klippenstein:
- Added an overload of `RenderTester.expectWorkflow()` which takes a `WorkflowIdentifier`.
 - Added extension properties on `KClass<Workflow>` and `KClass<ImpostorWorkflow>` to create
   identifiers from the classes.
 - Added a method to `WorkflowIdentifier` that compares identifiers by "leaf real identifiers"
   to support testing.

### DIFF
--- a/workflow-core/src/main/java/com/squareup/workflow/WorkflowIdentifier.kt
+++ b/workflow-core/src/main/java/com/squareup/workflow/WorkflowIdentifier.kt
@@ -21,6 +21,7 @@ package com.squareup.workflow
 import okio.Buffer
 import okio.ByteString
 import okio.EOFException
+import org.jetbrains.annotations.TestOnly
 import kotlin.LazyThreadSafetyMode.PUBLICATION
 import kotlin.reflect.KAnnotatedElement
 import kotlin.reflect.KClass
@@ -60,6 +61,11 @@ class WorkflowIdentifier internal constructor(
   private val type: KAnnotatedElement,
   private val proxiedIdentifier: WorkflowIdentifier? = null
 ) {
+  init {
+    require(
+        type is KClass<*> || (type is KType && type.classifier is KClass<*>)
+    ) { "Expected type to be either a KClass or a KType with a KClass classifier, but was $type" }
+  }
 
   /**
    * The fully-qualified name of the type of workflow this identifier identifies. Computed lazily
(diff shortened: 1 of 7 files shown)


workflow-core/src/main/java/com/squareup/workflow/WorkflowIdentifier.kt
```diff
@@ -115,6 +131,25 @@ class WorkflowIdentifier internal constructor(
...
+   *  - where the "leaf real identifier" is the last [ImpostorWorkflow.realIdentifier] in the chain
+   *    of [ImpostorWorkflow] identifiers that does not identify an [ImpostorWorkflow] itself,
+   *  - [actual]'s leaf identifier is the same type as, or a subtype of, this identifier's leaf
+   *    identifier.
+   *
+   * This predicate can be used for unit tests to assert that a workflow rendered a particular
+   * child workflow, given a supertype of that child workflow.
+   */
+  @TestOnly
+  fun matchesActualIdentifierForTest(actual: WorkflowIdentifier): Boolean {
```
rjrjr, line 147: Is this not redundant with the public `val Workflow.identifier`, now that you've changed that to be based on `realIdentifier`? I guess the `isAssignableFrom` check is different than what the runtime normally needs to do?

workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTester.kt
```diff
@@ -220,9 +223,70 @@ interface RenderTester<PropsT, StateT, OutputT, RenderingT> {
   /**
    * Specifies that this render pass is expected to render a particular child workflow.
    *
+   * ## Expecting impostor workflows
+   *
+   * Identifiers are compared using [WorkflowIdentifier.matchesActualIdentifierForTest]. If the
+   * workflow-under-test renders an [ImpostorWorkflow][com.squareup.workflow.ImpostorWorkflow],
+   * the expected [identifier] must match the leaf real identifier of the `ImpostorWorkflow`.
+   *
+   * For example, given the `mapRendering` workflow operator that returns an `ImpostorWorkflow`
```
rjrjr, line 232: Documentation nit: hopefully most folks won't need to know the impostor workflow details, so maybe you want to put the example above this section. Then this heading can be something more like `## Implementation details: imposter workflows`.

If you want to just get this merged, I'm happy to take an editorial pass later.

workflow-core/src/main/java/com/squareup/workflow/WorkflowIdentifier.kt
zach-klippenstein: It's completely different. The regular runtime check does an exact match - the types must be exactly the same and the entire identifier chain must be equal recursively. This only looks at the last identifier and is an instanceOf vs ==.

Issue #88, opened 2020-06-30
UI tests flaky on SDK 24 emulator


zach-klippenstein:
Tests fail regularly with the following error:
```
com.squareup.sample.TicTacToeEspressoTest > canGoBackInModalView[test(AVD) - 7.0] FAILED 
	kotlin.reflect.jvm.internal.KotlinReflectionInternalError: Introspecting local functions, lambdas, anonymous functions and local variables is not yet fully supported in Kotlin reflection
	at kotlin.reflect.jvm.internal.EmptyContainerForLocal.fail(EmptyContainerForLocal.kt:41)
```

They usually pass on the next re-run.

PR #93, opened 2020-07-01
Introduces Snapshot / Parcelable integration.

### LINKED ISSUE #43: Add Parcelable integration for Snapshot
We have something very convenient internally, open source it.

rjrjr:
Closes #43


rjrjr:
Test failed due to #88, re-running.

rjrjr:
> I guess it's not practical to write a unit test for this eh?

It is, but only if we introduce Robolectric. I can't take that on today.

PR #95, opened 2020-07-01
Improve the RenderTester API.

zach-klippenstein:
- Cleans up the testing API exposed by `WorkflowIdentifier` and makes the type
  checking more precise.
- Adds descriptions to expectations to allow for more readable error messages.
- Adds `RenderTester` methods to match using arbitrary predicate functions.
- Allows Workflow and side effect expectations to be non-exact – such expectations
  are optional, not required to be exclusive, and can match multiple times.

### DIFF
--- a/workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
+++ b/workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
@@ -176,7 +176,13 @@ interface WorkflowAction<in PropsT, StateT, out OutputT> {
     ): WorkflowAction<PropsT, StateT, OutputT> =
       action({ "emitOutput($name, $output)" }) { setOutput(output) }
 
-    private val NO_ACTION = action<Any, Any, Any>({ "noAction" }) { }
+    private val NO_ACTION = object : WorkflowAction<Any?, Any?, Any?> {
+      override fun toString(): String = "WorkflowAction.noAction()"
+
+      override fun Updater<Any?, Any?, Any?>.apply() {
+        // Noop
+      }
+    }
   }
 }
 
--- a/workflow-testing/src/main/java/com/squareup/workflow/testing/RealRenderTester.kt
+++ b/workflow-testing/src/main/java/com/squareup/workflow/testing/RealRenderTester.kt
@@ -30,6 +30,16 @@ import com.squareup.workflow.identifier
 import com.squareup.workflow.testing.RealRenderTester.Expectation.ExpectedSideEffect
 import com.squareup.workflow.testing.RealRenderTester.Expectation.ExpectedWorker
 import com.squareup.workflow.testing.RealRenderTester.Expectation.ExpectedWorkflow
+import com.squareup.workflow.testing.RenderTester.ChildWorkflowMatch
+import com.squareup.workflow.testing.RenderTester.ChildWorkflowMatch.Matched
+import com.squareup.workflow.testing.RenderTester.RenderChildInvocation
+import kotlin.reflect.KClass
+import kotlin.reflect.KType
+import kotlin.reflect.full.allSupertypes
+import kotlin.reflect.full.isSuperclassOf
+import kotlin.reflect.full.isSupertypeOf
+
+private const val WORKFLOW_INTERFACE_NAME = "com.squareup.workflow.Workflow"
 
 @OptIn(ExperimentalWorkflowApi::class)
 internal class RealRenderTester<PropsT, StateT, OutputT, RenderingT>(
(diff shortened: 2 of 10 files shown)


workflow-testing/src/main/java/com/squareup/workflow/testing/RealRenderTester.kt
```diff
@@ -206,30 +277,48 @@ internal class RealRenderTester<PropsT, StateT, OutputT, RenderingT>(
     }
   }
 
-  private inline fun <reified T : ExpectedWorkflow<*, *>> consumeExpectedChildWorkflow(
-    predicate: (T) -> Boolean,
+  private fun Workflow<*, *, *>.hasUnitRenderingType(): Boolean {
+    // Can't use Kotlin reflection because of what seems to be a bug when the class is anonymous.
```
rjrjr, line 281: Dark magiks

workflow-testing/src/main/java/com/squareup/workflow/testing/RealRenderTester.kt
```diff
@@ -105,6 +130,12 @@ internal class RealRenderTester<PropsT, StateT, OutputT, RenderingT>(
     return this
   }
 
+  // TODO unit tests
+  override fun disallowUnexpectedChildren(): RenderTester<PropsT, StateT, OutputT, RenderingT> {
```
zach-klippenstein, line 134: @rjrjr @bencochran @dhavalshreyas  Thoughts about the name for this? Maybe it should be something like `expectNoOtherChildren`?

workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTester.kt
```diff
@@ -350,6 +433,58 @@ interface RenderTester<PropsT, StateT, OutputT, RenderingT> {
...
+   */
+  enum class SideEffectMatch {
+    /** Indicates that the side effect does not match the expectation. */
+    NOT_MATCHED,
+
+    /**
+     * Indicates that the side effect matches the expectation, but may also match other expectations.
+     */
+    MATCHED
+  }
```
zach-klippenstein, line 487: This enum used to have three values, but now that it's back down to two, it could probably be replaced by an enum without losing much expressiveness.

workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTester.kt
```diff
@@ -350,6 +433,58 @@ interface RenderTester<PropsT, StateT, OutputT, RenderingT> {
...
+   * @param props The props value passed to `renderChild`.
+   * @param outputType The [KType] of the workflow's output type.
+   * @param renderingType The [KType] of the workflow's rendering type.
+   * @param renderKey The string key passed to `renderChild`.
+   */
+  class RenderChildInvocation(
+    val workflow: Workflow<*, *, *>,
+    val props: Any?,
+    val outputType: KTypeProjection,
+    val renderingType: KTypeProjection,
```
zach-klippenstein, line 451: These types are not trivial to pull out of a generic `Workflow` instance, and by providing them here we allow tests to provide default rendering values for any workflows that have a certain rendering type, without having to know the details about the specific workflows (e.g. don't require exact expectations for all child workflows that have `Unit` rendering types).

workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTester.kt
```diff
@@ -296,19 +343,42 @@ interface RenderTester<PropsT, StateT, OutputT, RenderingT> {
...
+   * @param exactMatch If true, then the test will fail if any other matching expectations are also
+   * exact matches, and the expectation will only be allowed to match a single child workflow.
+   * If false, the match will only be used if no other expectations return exclusive matches (in
+   * which case the first match will be used), and the expectation may match multiple children.
+   * @param matcher A function that determines whether a given [RenderChildInvocation] matches this
+   * expectation by returning a [ChildWorkflowMatch]. If the expectation matches, the function
+   * must include the rendering and optional output for the child workflow.
+   */
+  fun expectWorkflow(
+    description: String,
```
zach-klippenstein, line 378: `description` is required for this overload intentionally, since we can't derive a human-readable description from a predicate function.

workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTester.kt
```diff
@@ -350,6 +433,58 @@ interface RenderTester<PropsT, StateT, OutputT, RenderingT> {
...
   fun render(block: (rendering: RenderingT) -> Unit = {}): RenderTestResult<PropsT, StateT, OutputT>
+
+  /**
+   * Describes a call to
+   * [RenderContext.renderChild][com.squareup.workflow.RenderContext.renderChild].
+   *
+   * @param workflow The child workflow that is being rendered.
+   * @param props The props value passed to `renderChild`.
+   * @param outputType The [KType] of the workflow's output type.
+   * @param renderingType The [KType] of the workflow's rendering type.
```
zach-klippenstein, line 444: Kdoc should mention that these types won't have any meaningful value when the workflow is created with inline builders, and link the bug

workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
```diff
@@ -176,7 +176,13 @@ interface WorkflowAction<in PropsT, StateT, out OutputT> {
     ): WorkflowAction<PropsT, StateT, OutputT> =
       action({ "emitOutput($name, $output)" }) { setOutput(output) }
 
-    private val NO_ACTION = action<Any, Any, Any>({ "noAction" }) { }
+    private val NO_ACTION = object : WorkflowAction<Any?, Any?, Any?> {
```
zach-klippenstein, line 179: I wanted more control over the toString to be able to assert on it in tests.

workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTester.kt
```diff
@@ -365,25 +500,36 @@ interface RenderTester<PropsT, StateT, OutputT, RenderingT> {
...
-    "Use WorkflowOutput",
-    ReplaceWith("WorkflowOutput<OutputT>", "com.squareup.workflow.WorkflowOutput")
-)
-typealias EmittedOutput<OutputT> = WorkflowOutput<OutputT>
+/* ktlint-disable parameter-list-wrapping */
+fun <PropsT, StateT, OutputT, RenderingT>
+    RenderTester<PropsT, StateT, OutputT, RenderingT>.expectSideEffect(key: String):
+    RenderTester<PropsT, StateT, OutputT, RenderingT> =
+/* ktlint-enable parameter-list-wrapping */
+  expectSideEffect("side effect with key \"$key\"") {
```
zach-klippenstein, line 533: Pass exactMatch explicitly

workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTester.kt
zach-klippenstein: Added a note about this in the kdoc.

workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTester.kt
rjrjr: Worth saying so in the kdoc?

zach-klippenstein: It does – there's a whole section about exactly this.

PR #97, opened 2020-07-05
Use inner types for RenderContext to reduce type boilerplate.

zach-klippenstein:
Renamed `RenderContext` to `BaseRenderContext`, and gave both `StatelessWorkflow`
and `StatefulWorkflow` their own verions of `RenderContext`. These are inner classes,
so they can use the workflow's type parameters. This removes the type parameters from
the `RenderContext` reference in the `render` parameter list. Both types implement
`BaseWorkflowContext` however, so code that doesn't know about a particular workflow
instance can use this base type and convert it to a specific `RenderContext` later,
given a `Stateful/StatelessWorkflow` instance.

### DIFF
--- a/samples/hello-terminal/todo-terminal-app/src/main/java/com/squareup/sample/hellotodo/TodoWorkflow.kt
+++ b/samples/hello-terminal/todo-terminal-app/src/main/java/com/squareup/sample/hellotodo/TodoWorkflow.kt
@@ -26,7 +26,7 @@ import com.squareup.sample.helloterminal.terminalworkflow.TerminalWorkflow
 import com.squareup.sample.hellotodo.EditTextWorkflow.EditTextProps
 import com.squareup.sample.hellotodo.TodoWorkflow.TodoList
 import com.squareup.sample.hellotodo.TodoWorkflow.TodoList.Companion.TITLE_FIELD_INDEX
-import com.squareup.workflow1.RenderContext
+import com.squareup.workflow1.BaseRenderContext
 import com.squareup.workflow1.Snapshot
 import com.squareup.workflow1.StatefulWorkflow
 import com.squareup.workflow1.WorkflowAction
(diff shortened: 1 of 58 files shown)


zach-klippenstein:
Discussed offline about landing this in 1.0:

> @zach-klippenstein: wondering if we should try to land this for 1.0
@rjrjr: Nix, 1.0 is too big already.
@zach-klippenstein: Normally I’d agree, but we’re already adding Props to actions/contexts everywhere and this change is just going to make that simpler.
@rjrjr: Ah. That’s pretty compelling, then.
@rjrjr: So you’re thinking just go ahead and make the breaking change, not StatefulWorkflow2.
@zach-klippenstein: yep. we’re already using a separate package name, no need to rename classes
@rjrjr: Sold, no brainer combined w/the props change

workflow-core/src/main/java/com/squareup/workflow/StatelessWorkflow.kt
```diff
@@ -76,7 +82,16 @@ abstract class StatelessWorkflow<in PropsT, out OutputT, out RenderingT> :
     statefulWorkflow
 }
 
-/* ktlint-disable parameter-list-wrapping */
+/**
+ * Creates a `RenderContext` from a [BaseRenderContext] for the given [StatelessWorkflow].
+ */
+@Suppress("UNCHECKED_CAST", "FunctionName")
+fun <PropsT, OutputT, RenderingT> RenderContext(
```
zach-klippenstein, line 89: Should these factory functions be inline?

zach-klippenstein: They're small, but I can't think of a good reason to inline them.

samples/hello-terminal/todo-terminal-app/src/main/java/com/squareup/sample/hellotodo/TodoWorkflow.kt
```diff
@@ -76,14 +76,14 @@ class TodoWorkflow : TerminalWorkflow,
...
-    context: RenderContext<TerminalProps, TodoList, ExitCode>
+    context: RenderContext
   ): TerminalRendering {
 
     context.runningWorker(props.keyStrokes) { onKeystroke(it) }
 
     return TerminalRendering(buildString {
       @Suppress("UNCHECKED_CAST")
-      appendln(state.renderTitle(props, context as RenderContext<TerminalProps, TodoList, Nothing>))
+      appendln(state.renderTitle(props, context))
```
rjrjr, line 86: whoa

zach-klippenstein: This is actually unrelated, `renderTitle` should always have specified a wildcard rendering type.

zach-klippenstein:
> What are you thinking about the analogous Updater change?

I'm not. The `action{}` builder should cover enough cases I think that I think it's reasonable to say if you _really_ want to extend `WorkflowAction` manually, you have to specify the types yourself.

PR #98, opened 2020-07-06
Tweaks kdoc on snapshotState.

rjrjr:
Should we also deprecate `Snapshot.EMPTY`?


rjrjr:
> Can't hurt to deprecate, if we find another use case we can undeprecate after 1.0.

https://github.com/square/workflow-kotlin/pull/107/files

PR #101, opened 2020-07-07
Update the BuilderBinding docs to match the constructor

SalvatoreT:
`builder` is now `viewConstructor`, and the parameters are slightly different.

### DIFF
--- a/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/BuilderBinding.kt
+++ b/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/BuilderBinding.kt
@@ -28,15 +28,14 @@ import kotlin.reflect.KClass
  * [ViewFactory] by delegating to a [BuilderBinding], like this:
  *
  *    class MyView(
- *      context: Context,
- *      attributeSet: AttributeSet?
+ *      context: Context
  *    ) : FrameLayout(context, attributeSet) {
  *      private fun update(rendering:  MyRendering) { ... }
  *
  *      companion object : ViewBuilder<MyScreen>
  *      by BuilderBinding(
  *          type = MyScreen::class,
- *          builder = { _, initialRendering, context, _ ->
+ *          viewConstructor = { initialRendering, _, context, _ ->
  *            MyView(context).apply {
  *              layoutParams = ViewGroup.LayoutParams(MATCH_PARENT, MATCH_PARENT)
  *              bindShowRendering(initialRendering, ::update)


workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/BuilderBinding.kt
```diff
@@ -36,8 +36,8 @@ import kotlin.reflect.KClass
  *      companion object : ViewBuilder<MyScreen>
  *      by BuilderBinding(
  *          type = MyScreen::class,
- *          builder = { _, initialRendering, context, _ ->
- *            MyView(context).apply {
+ *          viewConstructor = { initialRendering, _, context, _ ->
+ *            MyView(context, null).apply {
```
rjrjr, line 40: Why the explicit `null`?

SalvatoreT: The `MyView` class has an `AttributeSet?` and I wasn't sure where to get a non-null from.

```
class MyView(
  context: Context,
  attributeSet: AttributeSet?
) 
```

rjrjr: Oh, good catch. How about giving `attributeSet` a default value, so:

```kotlin
 *    class MyView(
 *      context: Context,
 *      attributeSet: AttributeSet? = null
 *    ) : FrameLayout(context, attributeSet) {
```

SalvatoreT:
updated with @rjrjr's feedback

workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/BuilderBinding.kt
zach-klippenstein: I like @SalvatoreT's suggestion of removing it entirely – if this view is only used by this binding, it never uses it, right? It would keep the sample simpler too.

rjrjr:
Yeah, that's fine

SalvatoreT:
I removed the `AttributeSet?` from the example constructor.

```diff
diff --git a/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/BuilderBinding.kt b/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/BuilderBinding.kt
index 826262ac..2a29226e 100644
--- a/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/BuilderBinding.kt
+++ b/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/BuilderBinding.kt
@@ -28,8 +28,7 @@ import kotlin.reflect.KClass
  * [ViewFactory] by delegating to a [BuilderBinding], like this:
  *
  *    class MyView(
- *      context: Context,
- *      attributeSet: AttributeSet? = null
+ *      context: Context
  *    ) : FrameLayout(context, attributeSet) {
  *      private fun update(rendering:  MyRendering) { ... }
  *
```

zach-klippenstein:
UI tests are flaking hard, this is a comment-only change and so the tests can't be related. Merging anyway.

PR #106, opened 2020-07-08
Revert "Extend ViewGroup in stock Kotlin Container classes"

### LINKED ISSUE #105: WorkflowLayout, maybe others, not honoring layout params
Or something like that. Reported by @bnvinay92. 

> I think there might be a bug in how `WorkflowLayout` is being laid out. I bumped from 0.23.2 to 1.0.0-alpha-2 and it caused the WorkflowFragment's root WorkflowLayout view to be match parent even though the container is wrap_content. I did a git bisect and the commit responsible for this is `e41659bc` (Extend ViewGroup in stock Kotlin Container classes).

Details: https://workflow-community.slack.com/archives/CHTFPR277/p1594221734173500

https://github.com/square/workflow/pull/1193

rjrjr:
Our minimal measure / layout code is too minimal, and broke things.

Closes #105.

This reverts commit e41659bc024631c6793dcbae3aa5c0652c0a8a72.

### DIFF
--- a/workflow-ui/backstack-android/api/backstack-android.api
+++ b/workflow-ui/backstack-android/api/backstack-android.api
@@ -12,15 +12,13 @@ public final class com/squareup/workflow/ui/backstack/BackStackConfig$Companion 
 	public synthetic fun getDefault ()Ljava/lang/Object;
 }
 
-public class com/squareup/workflow/ui/backstack/BackStackContainer : android/view/ViewGroup {
+public class com/squareup/workflow/ui/backstack/BackStackContainer : android/widget/FrameLayout {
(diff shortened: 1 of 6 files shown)


rjrjr:
Forgot to update the api files, as usual.

rjrjr:
> Forgot to update the api files, as usual.

Huh. No `apiCheck` failures. @zach-klippenstein, is that expected?

zach-klippenstein:
`apiCheck` should fail if the revert didn't also revert the API files, but I expect it would have.

rjrjr:
D'oh.

PR #108, opened 2020-07-08
Override setBackground() like we do setVisibility() in WorkflowViewStub.

afollestad:
This allows stubs to have a background color, for an example, that also applies to all displayed content.

### DIFF
--- a/workflow-ui/core-android/api/core-android.api
+++ b/workflow-ui/core-android/api/core-android.api
@@ -180,6 +180,7 @@ public final class com/squareup/workflow1/ui/WorkflowViewStub : android/view/Vie
 	public final fun getActual ()Landroid/view/View;
 	public final fun getInflatedId ()I
 	public final fun getReplaceOldViewInParent ()Lkotlin/jvm/functions/Function2;
+	public fun setBackground (Landroid/graphics/drawable/Drawable;)V
 	public final fun setInflatedId (I)V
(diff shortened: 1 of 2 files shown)


afollestad:
Ah, I see the workflow-ui/core-android/build/api/core-android.api file needs to be updated for the build to totally pas

zach-klippenstein:
> Ah, I see the workflow-ui/core-android/build/api/core-android.api file needs to be updated for the build to totally pass

You just need to run `./gradlew apiDump`.

workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/WorkflowViewStub.kt
```diff
@@ -160,6 +173,7 @@ class WorkflowViewStub @JvmOverloads constructor(
         .also { newView ->
           if (inflatedId != NO_ID) newView.id = inflatedId
           newView.visibility = visibility
+          newView.background = background
```
rjrjr, line 176: Oops, you're always stomping the newView's background, even if the stub's background is unset.

rjrjr:
@afollestad I've lost track. Besides rebasing, is this in its final form?

afollestad:
> @afollestad I've lost track. Besides rebasing, is this in its final form?

Yep everything is looking good in SPOS!

PR #110, opened 2020-07-09
Reimplemented Workers using side effects (GUWT).

### LINKED ISSUE #82: Implement GUWT Workers
Just realized this wasn't actually filed. Blocked by #79.

This should use the final `RenderContext` syntax documented in https://github.com/square/workflow/issues/1019.

### LINKED ISSUE #92: Add a renderTester method for expecting typed workers
Eg `expectTypedWorker`. It's technically a bit weird because it's basically leaking an implementation detail of a private type, but it's been asked for internally, I think it's probably a super common use case, and the fact that that behavior is a private implementation detail is weird to begin with.

I think it will also become less weird after we get implement GUWT.

zach-klippenstein:
Issues:
 - Fixes #82, which is the Kotlin half of https://github.com/square/workflow/issues/1021.
 - Fixes #92.
 - Fixes https://github.com/square/workflow/issues/1197.


workflow-testing/src/test/java/com/squareup/workflow/WorkerStressTest.kt
```diff
@@ -16,7 +16,7 @@ import kotlin.test.Test
 import kotlin.test.assertEquals
 import kotlin.test.fail
 
-private const val WORKER_COUNT = 1000
+private const val WORKER_COUNT = 500
```
zach-klippenstein, line 19: Workers are a lot slower now.

workflow-core/src/main/java/com/squareup/workflow/Worker.kt
```diff
@@ -117,6 +122,14 @@ import kotlin.reflect.typeOf
  */
 interface Worker<out OutputT> {
 
+  /**
+   * Should be overridden in subclasses that have their own type parameters, to allow the runtime to
+   * make use of the value of those type parameters to compare workers. Two workers of the same
+   * concrete [Worker] class will be considered equivalent if and only if their [outputType]s are
+   * also equivalent. If this property returns null, the worker will be treated as a `Worker<*>`.
+   */
+  val outputType: KType? get() = null
```
zach-klippenstein, line 131: This was originally intended to be used to create the workflow identifier without reflecting on the worker itself, but that didn't work out since we're forced to just use `typeOf<W>()` (where `W` is the reified type of the worker) in `runningWorker` to get the worker type. We can't create the `KType` from just the worker `KClass` and this `outputType` (e.g. do something like `worker::class.createType(KTypeProjection.covariant(worker.outputType))`) because we don't have access to the full reflection library from the core module. 

This should be removed since it's not used.

workflow-core/src/main/java/com/squareup/workflow/RenderContext.kt
```diff
@@ -196,12 +185,56 @@ fun <PropsT, StateT, OutputT> RenderContext<PropsT, StateT, OutputT>.runningWork
...
+@OptIn(ExperimentalStdlibApi::class)
+/* ktlint-disable parameter-list-wrapping */
+inline fun <T, reified W : Worker<T>, PropsT, StateT, OutputT>
+    RenderContext<PropsT, StateT, OutputT>.runningWorker(
+  worker: W,
+  key: String = "",
+  noinline handler: (T) -> WorkflowAction<PropsT, StateT, OutputT>
+) {
+/* ktlint-enable parameter-list-wrapping */
+  runningWorker(worker, typeOf<W>(), key, handler)
```
zach-klippenstein, line 212: This isn't the API I'd ideally like to have, but since we don't have access to the full reflection library in this module, it's the best we can do. If we did, we could create the full worker type from the worker's class and the worker's `outputType` property, or maybe using the reified `T` type parameter. But since we don't, all we have is `typeOf()`, which means we need to reify the entire worker type.

However, while less flexible, it has the nice property that the type that is used to compare the worker will _always_ be _exactly_ the type for the worker that the IDE is showing you when you pass it to `runningWorker`. There's no mystery or hidden implementation details anymore. You might need to pass a few more `key`s in, but that's probably a good idea anyway.

workflow-core/src/main/java/com/squareup/workflow/LifecycleWorker.kt
```diff
@@ -30,6 +31,8 @@ import kotlinx.coroutines.suspendCancellableCoroutine
  */
 abstract class LifecycleWorker : Worker<Nothing> {
```
zach-klippenstein, line 32: Will deprecate this in a follow-up PR, since it's better implemented by a plain side effect now.

workflow-core/src/test/java/com/squareup/workflow/WorkerTest.kt
```diff
@@ -1,41 +1,111 @@
...
-    class Worker2 : Worker<Nothing> {
-      override fun run(): Flow<Nothing> = emptyFlow()
-    }
-
-    assertTrue(Worker1().doesSameWorkAs(Worker1()))
-    assertFalse(Worker1().doesSameWorkAs(Worker2()))
-  }
-
   @Test fun `createSideEffect workers are equivalent`() {
     val worker1 = Worker.createSideEffect {}
```
zach-klippenstein, line 38: Will deprecate this method in a follow-up PR, since side effects are a primary concept now.

workflow-core/src/main/java/com/squareup/workflow/LifecycleWorker.kt
zach-klippenstein: #114

workflow-core/src/test/java/com/squareup/workflow/WorkerTest.kt
zach-klippenstein: #114

workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTesterWorkers.kt
```diff
@@ -0,0 +1,232 @@
...
+/* ktlint-disable parameter-list-wrapping */
+fun <PropsT, StateT, OutputT, RenderingT>
+    RenderTester<PropsT, StateT, OutputT, RenderingT>.expectWorker(
+  workerType: KType,
+  key: String = "",
+  assertWorker: (Worker<*>) -> Unit = {},
+  output: WorkflowOutput<*>? = null,
+  description: String = ""
+): RenderTester<PropsT, StateT, OutputT, RenderingT> =
+/* ktlint-enable parameter-list-wrapping */
```
rjrjr, line 156: Dead code and TODOs in here, known?

workflow-testing/src/test/java/com/squareup/workflow/testing/WorkerRenderExpectationsTest.kt
```diff
@@ -0,0 +1,234 @@
+package com.squareup.workflow.testing
```
rjrjr, line 1: copyright. May be missing in other files too.

workflow-testing/src/test/java/com/squareup/workflow/testing/WorkerRenderExpectationsTest.kt
```diff
@@ -0,0 +1,234 @@
...
+import com.squareup.workflow.runningWorker
+import com.squareup.workflow.stateless
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.emptyFlow
+import kotlin.reflect.typeOf
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
+
+// TODO add more failure tests
```
rjrjr, line 16: Want to track this and get this merged?

workflow-core/api/workflow-core.api
```diff
@@ -227,6 +222,7 @@ public final class com/squareup/workflow/Workflows {
 	public static synthetic fun renderChild$default (Lcom/squareup/workflow/RenderContext;Lcom/squareup/workflow/Workflow;Ljava/lang/String;Lkotlin/jvm/functions/Function1;ILjava/lang/Object;)Ljava/lang/Object;
 	public static final fun rendering (Lcom/squareup/workflow/Workflow$Companion;Ljava/lang/Object;)Lcom/squareup/workflow/Workflow;
 	public static final fun runningWorker (Lcom/squareup/workflow/RenderContext;Lcom/squareup/workflow/Worker;Ljava/lang/String;)V
+	public static final fun runningWorker (Lcom/squareup/workflow/RenderContext;Lcom/squareup/workflow/Worker;Lkotlin/reflect/KType;Ljava/lang/String;Lkotlin/jvm/functions/Function1;)V
```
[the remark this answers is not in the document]
zach-klippenstein: I was actually thinking about migrating a few more methods on these interfaces to extension functions now that we are committed to doing the migration incrementally. Keeping them as small as possible seems good.

workflow-core/src/main/java/com/squareup/workflow/Worker.kt
```diff
@@ -179,7 +178,7 @@ interface Worker<out OutputT> {
    * }
    * ```
    */
-  fun doesSameWorkAs(otherWorker: Worker<*>): Boolean = otherWorker::class == this::class
+  fun doesSameWorkAs(otherWorker: Worker<*>): Boolean = true
```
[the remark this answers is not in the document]
zach-klippenstein: Good point. The thinking was that the old default, comparing classes, would always be true now since the runtime itself is ensuring that worker types match. However, we ended up only comparing on the declared type of the worker, which means two workers that the runtime thinks are equivalent could still be different classes, so it would be safer to leave the class checks in here.

That said, by defaulting to relying on concrete types again, we lose the nice simplicity for testing. Not sure what the best thing to do is.

workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTesterWorkers.kt
zach-klippenstein: Probably not

workflow-testing/src/test/java/com/squareup/workflow/testing/WorkerRenderExpectationsTest.kt
zach-klippenstein: Yea, got a few other PRs stacking up on this. I'll file.

zach-klippenstein: #117

workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTesterWorkers.kt
```diff
@@ -0,0 +1,232 @@
...
+    RenderTester<PropsT, StateT, OutputT, RenderingT>.expectWorker(
+  doesSameWorkAs: WorkerT,
+  key: String = "",
+  output: WorkflowOutput<WorkerOutputT>? = null,
+  description: String = ""
+): RenderTester<PropsT, StateT, OutputT, RenderingT> = expectWorker(
+/* ktlint-enable parameter-list-wrapping */
+    workerType = typeOf<WorkerT>(),
+    key = key,
+    assertWorker = { doesSameWorkAs.doesSameWorkAs(it) },
```
zach-klippenstein, line 95: Oops, this is the opposite of what this method is supposed to do according to the documentation.

workflow-core/api/workflow-core.api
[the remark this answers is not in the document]
zach-klippenstein: #118

zach-klippenstein:
Just found a pretty bad bug that blocks the release, but I want to fix separately: #120

workflow-core/src/main/java/com/squareup/workflow/Worker.kt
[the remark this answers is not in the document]
zach-klippenstein: Reverted this default implementation, and added a bunch of tests around it.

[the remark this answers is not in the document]
zach-klippenstein: Additional documentation added to `runningWorker` and `doesSameWorkAs`.

Issue #51, opened 2020-06-02
Ensure all dependencies are freshened before releasing v1.0.0


zach-klippenstein:
Solving square/workflow#1044 would make this a lot easier, but it's not necessarily a requirement.

PR #111, opened 2020-07-10
Start using the refreshVersions Gradle plugin to manage dependencies.

zach-klippenstein:
https://github.com/jmfayard/refreshVersions

This change migrates dependency versions to the versions.properties file,
which will be resolved into actual dependency strings by the plugin. New
versions can be checked for by running `./gradlew refreshVersions`.

Note that **this commit does not change any version numbers**, it simply
moves them to be defined in a different location. It also deletes some
stale dependency definitions that weren't actually used.

Prep for #51.


build.gradle.kts
```diff
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+//import de.fayard.OrderBy
```
rjrjr, line 16: Planning to delete here and below?

versions.properties
```diff
@@ -0,0 +1,54 @@
+## suppress inspection "SpellCheckingInspection" for whole file
+##
+## Dependencies and Plugin versions with their available updates
+## Generated by $ ./gradlew refreshVersions
```
rjrjr, line 4: What kind of control do we have when there are specific versions we don't want to bump? I guess we just don't use the `_`?

build.gradle.kts
zach-klippenstein: Oh, i still want to try to get this working so we don't have to keep manually sorting that file.

versions.properties
zach-klippenstein: Yep. But the tooling is very safe – you run `./gradlew refreshVersions`, and it just adds a bunch of comments to that file telling you what the latest versions are. You have to actually edit it to update.

build.gradle.kts
zach-klippenstein: Deleted instead, don't want to fight with gradle more than absolutely necessary today.

zach-klippenstein:
UI test failed with #88:
```
com.squareup.sample.TicTacToeEspressoTest > canGoBackInModalView[test(AVD) - 7.0] FAILED 
	kotlin.reflect.jvm.internal.KotlinReflectionInternalError: Introspecting local functions, lambdas, anonymous functions and local variables is not yet fully supported in Kotlin reflection
	at kotlin.reflect.jvm.internal.EmptyContainerForLocal.fail(EmptyContainerForLocal.kt:41)
Tests on test(AVD) - 7.0 failed: Instrumentation run failed due to 'Process crashed.'
```

zach-klippenstein:
Failure was a cache failure, merging.

PR #114, opened 2020-07-10
Deprecate older Worker helpers that are now easier to do by using side effects directly.

zach-klippenstein:
# DON"T MERGE UNTIL #110 DOES


zach-klippenstein:
I think this might be too aggressive. These functions are still useful when a Worker type is required by some other API. Abandoning.

PR #127, opened 2020-07-13
Extract public interface methods with default implementations into extension methods.

### LINKED ISSUE #118: Convert non-essential methods from interfaces to extension functions
I was actually thinking about migrating a few more methods on these interfaces to extension functions now that we are committed to doing the migration incrementally. Keeping them as small as possible seems good.

- [x] `RenderContext.onEvent`
- [x] `RenderContext.makeActionSink`
- [x] `RenderTester.expectWorkflow(KClass)`
- [x] `RenderTester.expectWorkflow(WorkflowIdentifier)`

_Originally posted by @zach-klippenstein in https://github.com/square/workflow-kotlin/pull/110_

zach-klippenstein:
Closes #118.

### DIFF
--- a/workflow-testing/src/test/java/com/squareup/workflow1/testing/RealRenderTesterTest.kt
+++ b/workflow-testing/src/test/java/com/squareup/workflow1/testing/RealRenderTesterTest.kt
@@ -715,7 +715,7 @@ class RealRenderTesterTest {
       renderChild(TestImpostor(TestWorkflow()))
     }
     workflow.testRender(Unit)
-        .expectWorkflow<Nothing, Unit>(
+        .expectWorkflow(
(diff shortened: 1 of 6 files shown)


workflow-testing/src/test/java/com/squareup/workflow1/testing/RealRenderTesterTest.kt
```diff
@@ -715,7 +715,7 @@ class RealRenderTesterTest {
       renderChild(TestImpostor(TestWorkflow()))
     }
     workflow.testRender(Unit)
-        .expectWorkflow<Nothing, Unit>(
+        .expectWorkflow(
```
rjrjr, line 718: :)

PR #128, opened 2020-07-13
Pulled the deprecated Mutator and associated method into a sub-interface of WorkflowAction

zach-klippenstein:
This should be a find-and-replace migration, and allows `WorkflowAction` to require
implementors to override `apply()`. The new sub-interface is completely deprecated.

### DIFF
--- a/samples/containers/poetry/src/main/java/com/squareup/sample/poetry/PoemWorkflow.kt
+++ b/samples/containers/poetry/src/main/java/com/squareup/sample/poetry/PoemWorkflow.kt
@@ -28,12 +28,11 @@ import com.squareup.sample.poetry.StanzaWorkflow.Output.ShowNextStanza
 import com.squareup.sample.poetry.StanzaWorkflow.Output.ShowPreviousStanza
 import com.squareup.sample.poetry.StanzaWorkflow.Props
 import com.squareup.sample.poetry.model.Poem
-import com.squareup.workflow1.RenderContext
+import com.squareup.workflow1.MutatorWorkflowAction
+import com.squareup.workflow1.MutatorWorkflowAction.Mutator
 import com.squareup.workflow1.Snapshot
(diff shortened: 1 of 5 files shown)


rjrjr:
Oh, nice! I felt so filthy not doing this at the time.

👀

samples/containers/poetry/src/main/java/com/squareup/sample/poetry/PoemWorkflow.kt
```diff
@@ -107,7 +106,7 @@ object PoemWorkflow : StatefulWorkflow<Poem, Int, ClosePoem, OverviewDetailScree
     sink.writeInt(state)
   }
 
-  private sealed class Action : WorkflowAction<Poem, Int, ClosePoem> {
+  private sealed class Action : MutatorWorkflowAction<Poem, Int, ClosePoem> {
```
rjrjr, line 109: Well this is just embarrassing, didn't realize I still had stale samples. And such a core one, too.

workflow-core/src/main/java/com/squareup/workflow1/StatefulWorkflow.kt
```diff
@@ -287,11 +288,13 @@ fun <PropsT, StateT, OutputT, RenderingT>
  * in [toString].
  * @param update Function that defines the workflow update.
  */
+/* ktlint-disable parameter-list-wrapping */
```
rjrjr, line 291: sigh

samples/containers/poetry/src/main/java/com/squareup/sample/poetry/PoemWorkflow.kt
zach-klippenstein: There's a comment further down in the code, this sample was intentionally left using the old API.

PR #129, opened 2020-07-13
Convert Updater to an inner class.

zach-klippenstein:
Follow-up to #128 that does the `interface + class` to `abstract class + inner class` refactor again to get rid of the inner type params.

### DIFF
--- a/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/PlayerWorkflow.kt
+++ b/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/PlayerWorkflow.kt
@@ -35,16 +35,16 @@ class PlayerWorkflow(
   private val cellsPerSecond: Float = 15f
 ) : StatefulWorkflow<ActorProps, Movement, Nothing, Rendering>() {
 
-  sealed class Action : WorkflowAction<ActorProps, Movement, Nothing> {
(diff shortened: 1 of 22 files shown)


samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/PlayerWorkflow.kt
```diff
@@ -35,16 +35,16 @@ class PlayerWorkflow(
   private val cellsPerSecond: Float = 15f
 ) : StatefulWorkflow<ActorProps, Movement, Nothing, Rendering>() {
 
-  sealed class Action : WorkflowAction<ActorProps, Movement, Nothing> {
+  sealed class Action : WorkflowAction<ActorProps, Movement, Nothing>() {
 
     class StartMoving(private val direction: Direction) : Action() {
-      override fun Updater<ActorProps, Movement, Nothing>.apply() {
+      override fun Updater.apply() {
```
rjrjr, line 41: This is pretty nice. Makes this pattern that much more tolerable. And it does get more use than you might think, especially with teams that are better balanced btw. Android and iOS

rjrjr:
This PR swung me around. I don't see a downside. If it won't be a big upgrade burden, I'd go for it.

zach-klippenstein:
I can't think of any downsides of making this thing a class vs an interface. It can still participate in sealed class hierarchies, so i think it's fine.

PR #130, opened 2020-07-13
Make RenderTester throw a more realistic exception if the same workflow/side effect is ran twice.

### LINKED ISSUE #120: RenderTester allows duplicate workflows to be rendered.
Workflows with identical `WorkflowIdentifier`s will cause a real render to fail, but `RenderTester` allows them to succeed.

zach-klippenstein:
It now throws an exception that looks just like the real runtime. This doesn't _fail_ the test,
it _errors_ it, which is also better behavior since this case means the workflow-under-test is
written incorrectly, not that it doesn't match expected test behavior. Fixes #120.

This change also cleans up the way `WorkflowIdentifier`s are described in error and testing strings,
unifies the impostor and non-impostor calls, makes the string consistent between runtime and tests,
and removes a method from the public API.

### DIFF
--- a/workflow-core/api/workflow-core.api
+++ b/workflow-core/api/workflow-core.api
@@ -192,7 +192,6 @@ public final class com/squareup/workflow1/WorkflowAction$Updater {
 
 public final class com/squareup/workflow1/WorkflowIdentifier {
 	public static final field Companion Lcom/squareup/workflow1/WorkflowIdentifier$Companion;
-	public final fun describeRealIdentifier ()Ljava/lang/String;
 	public fun equals (Ljava/lang/Object;)Z
 	public final fun getRealIdentifierType ()Lkotlin/reflect/KAnnotatedElement;
 	public fun hashCode ()I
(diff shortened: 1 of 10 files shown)


workflow-core/api/workflow-core.api
```diff
@@ -192,7 +192,6 @@ public final class com/squareup/workflow1/WorkflowAction$Updater {
 
 public final class com/squareup/workflow1/WorkflowIdentifier {
 	public static final field Companion Lcom/squareup/workflow1/WorkflowIdentifier$Companion;
-	public final fun describeRealIdentifier ()Ljava/lang/String;
```
zach-klippenstein, line 195: Yay one less public method!

workflow-core/src/main/java/com/squareup/workflow1/WorkflowIdentifier.kt
```diff
@@ -113,20 +113,15 @@ class WorkflowIdentifier internal constructor(
...
-
-  /**
-   * Returns a description of this identifier including the name of its workflow type and any
-   * [ImpostorWorkflow.realIdentifier]s.
    */
   override fun toString(): String =
-    proxiedIdentifiers
-        .joinToString { it.typeName }
-        .let { "WorkflowIdentifier($it)" }
+    description?.invoke()
```
rjrjr, line 121: First time I've seen an unambiguously good reason to use `invoke()`

zach-klippenstein:
Both UI tests timed out. This has been happening more and more, we might need to increase the timeouts.

PR #136, opened 2020-07-17
Fix maven link in README

zach-klippenstein:


zach-klippenstein:
Assemble error is some weird infrastructure error, documentation shard is fine so merging.

Issue #109, opened 2020-07-09
Allow root workflows to emit multiple outputs to their activities/fragments


bnvinay92:
A multi-activity application may require the root fragment/activity to handle multiple navigation requests throughout the session of a workflow.

grandstaish:
I'm also looking at migrating one of my existing apps to use Workflow and am unsure of what to do about Activity navigation. It seems intuitive to me to use Workflow outputs as the trigger for a navigation event, e.g. [like this](https://github.com/grandstaish/workflow-legacy-app-migration-example/blob/master/app/src/main/java/nz/bradcampbell/legacyappwithworkflows/main/MainActivity.kt). But of course this code is broken because once a result is emitted, the Workflow is shutdown. 

Do you have any suggestions for how I could migrate my app differently?

PR #142, opened 2020-07-22
Allow multiple outputs from root workflow

bnvinay92:
Addresses https://github.com/square/workflow-kotlin/issues/109

Desired behavior:

- Root workflow can emit multiple outputs, each will be consumed in sequence at most once
- If an output x is emitted during config change, then succeeding outputs are paused until there's a new STARTED host that consumes x
- A worker's result is handled even if it was the worker was enqueued in a different host before a config change

I'm very unfamiliar with coroutines and after skimming some docs it looked like a rendezvous channel is what I'm looking for. I

### DIFF
--- a/workflow-ui/core-android/src/test/java/com/squareup/workflow1/ui/WorkflowRunnerViewModelTest.kt
+++ b/workflow-ui/core-android/src/test/java/com/squareup/workflow1/ui/WorkflowRunnerViewModelTest.kt
@@ -2,6 +2,8 @@ package com.squareup.workflow1.ui
 
 import androidx.savedstate.SavedStateRegistry.SavedStateProvider
 import com.google.common.truth.Truth.assertThat
+import com.nhaarman.mockito_kotlin.mock
+import com.nhaarman.mockito_kotlin.verify
 import com.squareup.workflow1.RenderingAndSnapshot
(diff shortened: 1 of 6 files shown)


workflow-ui/core-android/src/test/java/com/squareup/workflow1/ui/WorkflowRunnerViewModelTest.kt
```diff
@@ -62,24 +68,69 @@ class WorkflowRunnerViewModelTest {
...
-    val tester = testScope.async { runner.awaitResult() }
+    val tester = testScope.async { runner.receiveOutput() }
 
     assertThat(tester.isActive).isTrue()
     trigger.complete("fnord")
     assertThat(tester.isCompleted).isTrue()
     assertThat(tester.getCompleted()).isEqualTo("fnord")
   }
 
+  @Test fun `Factory propagates worker errors`() {
```
bnvinay92, line 112: @zach-klippenstein, is this what you were looking for?

zach-klippenstein:
We were planning to merge this, why was it closed?

zach-klippenstein:
Ah, looks like you split it into 2 prs?

PR #148, opened 2020-07-28
Make TreeSnapshot nullable throughout the runtime API and internally.

### LINKED ISSUE #134: renderWorkflowIn should take a nullable TreeSnapshot
The core API `StatefulWorkflow.createSnapshot` supports nullable snapshots now, allowing nulls for `TreeSnapshot` would make for a more ergonomic API.

zach-klippenstein:
Closes #134.


workflow-runtime/src/main/java/com/squareup/workflow1/TreeSnapshot.kt
```diff
@@ -20,6 +20,7 @@ import com.squareup.workflow1.TreeSnapshot.Companion.parse
 import com.squareup.workflow1.internal.WorkflowNodeId
 import okio.Buffer
 import okio.ByteString
+import kotlin.LazyThreadSafetyMode.NONE
```
zach-klippenstein, line 23: Wow, I have no idea how this even compiled before.

PR #154, opened 2020-07-31
Fix expectWorker to match on concrete types of generic workers.

### LINKED ISSUE #153: RenderTester.expectWorker doesn't match workers that are instances of generic worker types
E.g. `TypedWorker<T>`, which is the type of the workers returned by most `asWorker()` functions. If your workflow does `runningWorker(flowOf("hello").asWorker()) { … }`, and your test does `expectWorker(typeOf<Worker<String>>())`, it will not match, because `Worker<String>` is not a supertype of `TypedWorker<Any?>`, which is the erased type of the actual worker at runtime.

zach-klippenstein:
Fixes #153.

### DIFF
--- a/workflow-core/src/main/java/com/squareup/workflow1/RenderContext.kt
+++ b/workflow-core/src/main/java/com/squareup/workflow1/RenderContext.kt
@@ -21,7 +21,6 @@ package com.squareup.workflow1
 
 import com.squareup.workflow1.StatefulWorkflow.RenderContext
 import com.squareup.workflow1.WorkflowAction.Companion.noAction
-import com.squareup.workflow1.WorkflowAction.Updater
 import kotlin.reflect.KType
 import kotlin.reflect.typeOf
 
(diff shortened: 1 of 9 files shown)


workflow-core/src/main/java/com/squareup/workflow1/RenderContext.kt
```diff
@@ -173,10 +172,13 @@ fun <PropsT, StateT, OutputT, ChildRenderingT>
  *
  * @param key An optional string key that is used to distinguish between identical [Worker]s.
  */
-fun <PropsT, StateT, OutputT> BaseRenderContext<PropsT, StateT, OutputT>.runningWorker(
-  worker: Worker<Nothing>,
+/* ktlint-disable parameter-list-wrapping */
+inline fun <reified W : Worker<Nothing>, PropsT, StateT, OutputT>
+    BaseRenderContext<PropsT, StateT, OutputT>.runningWorker(
+  worker: W,
```
zach-klippenstein, line 178: Without this, matching a `Worker<Nothing>` will fail because the type captured here will be `Worker<java.lang.Void>`, which is not a subtype of `Worker<Nothing>`.

workflow-testing/src/test/java/com/squareup/workflow1/testing/WorkerRenderExpectationsTest.kt
```diff
@@ -385,4 +387,15 @@ class WorkerRenderExpectationsTest {
           )
         }
   }
+
+  @Test fun `expectWorker on TypedWorker`() {
```
zach-klippenstein, line 391: Verified manually that this test fails without the change.

PR #156, opened 2020-08-05
Fix RenderTester workflow class matching for mocked workflow instances.

### LINKED ISSUE #155: RenderTester expectWorkflow doesn't match workflows created with Mockito
Turns out that the KClasses of objects mockito returns for a given interface aren't considered subclasses of the KClass of that interface by kotlin reflection, which means that the logic to match a mock workflow to the expectation of its supertype fails.

zach-klippenstein:
Turns out that mocks made with at least some mocking libraries (tested
with Mockito and Mockk) aren't seen by Kotlin reflection as subclassing
their mocked types. That is, `Interface::class.isSuperclassOf(mock<Interface>()::class)`
returns false. This change works around that issue by falling back to java reflection
if `KClass.isSuperclassOf` returns false.

Fixes #155.

### DIFF
--- a/workflow-testing/src/main/java/com/squareup/workflow1/testing/RealRenderTester.kt
+++ b/workflow-testing/src/main/java/com/squareup/workflow1/testing/RealRenderTester.kt
@@ -334,7 +334,7 @@ internal fun WorkflowIdentifier.realTypeMatchesExpectation(
       expectedType.isSupertypeOf(actualType)
     }
     expectedType is KClass<*> && actualType is KClass<*> -> {
-      expectedType.isSuperclassOf(actualType)
+      expectedType.isSuperclassOf(actualType) || actualType.isJavaMockOf(expectedType)
     }
     else -> {
       error(
@@ -344,3 +344,15 @@ internal fun WorkflowIdentifier.realTypeMatchesExpectation(
     }
   }
 }
+
+/**
+ * Falls back to using Java reflection to determine subclass relationship.
+ *
+ * Kotlin's [isSuperclassOf] doesn't play nice with Mockito or Mockk:
+ * `Interface::class.isSuperclassOf(mock<Interface>()::class)` will return false.
+ *
+ * See https://github.com/square/workflow-kotlin/issues/155 and
+ * https://youtrack.jetbrains.com/issue/KT-40863.
+ */
+private fun KClass<*>.isJavaMockOf(type: KClass<*>): Boolean =
+  type.java.isAssignableFrom(this.java)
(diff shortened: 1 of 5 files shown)


workflow-testing/src/test/java/com/squareup/workflow1/testing/RealRenderTesterTest.kt
```diff
@@ -1201,4 +1234,10 @@ class RealRenderTesterTest {
 
   @Suppress("unused")
   private interface ContravariantGenericType<in T>
+
+  // For mocking tests. Interfaces can't be defined inside functions, and Mockito can't handle
+  // the class names of local classes defined in functions named like these test functioøns are.
```
steve-the-edwards, line 1239: is this l33t speak for function??

zach-klippenstein: 🤦

zach-klippenstein, line 1239:
```suggestion
  // the class names of local classes defined in functions named like these test functions are.
```

workflow-testing/src/main/java/com/squareup/workflow1/testing/RealRenderTester.kt
```diff
@@ -344,3 +344,14 @@ internal fun WorkflowIdentifier.realTypeMatchesExpectation(
...
+
+/**
+ * Falls back to using Java reflection to determine subclass relationship.
+ *
+ * Kotlin's [isSuperclassOf] doesn't play nice with Mockito or Mockk:
+ * `Interface::class.isSuperclassOf(mock<Interface>()::class)` will return false.
+ *
+ * See https://github.com/square/workflow-kotlin/issues/155.
+ */
+private fun KClass<*>.isJavaMockOf(type: KClass<*>): Boolean =
```
rjrjr, line 356: Any reason not to just use `isAssignableFrom` and drop `isSuperclassOf` entirely?

zach-klippenstein: Just keeping the intent clear and highlighting that this is an issue with java mocking libraries. I don't feel too strongly about keeping both checks.

zach-klippenstein: I added a link to the Kotlin reflection bug this is working around, would you be ok keeping both now?

zach-klippenstein:
Github seems to be having issues, I ran `./gradlew clean build connectedCheck` locally and everything passed, so I'm just going to merge without waiting.

PR #163, opened 2020-08-07
Demonstrate Non-Stacktrace Crashes

SalvatoreT:
@zach-klippenstein, this has a crash that  breaks the process but doesn't show why.

```
FATAL EXCEPTION: main
Process: com.squareup.sample.dungeon, PID: 4551
```


zach-klippenstein:
Verified that the exception is propagating all the correct ways, i.e. the default uncaught exception handler is getting this exception.

SalvatoreT:
Roger that. Thanks for looking into this!

PR #168, opened 2020-08-17
Allow multiple outputs to be emitted from a workflow

bnvinay92:


workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowRunnerViewModel.kt
```diff
@@ -74,26 +74,27 @@ internal class WorkflowRunnerViewModel<OutputT>(
       val config = configure()
       val props = config.props
       val scope = CoroutineScope(config.dispatcher)
-      val result = CompletableDeferred<OutputT>(parent = scope.coroutineContext[Job])
+      val outputChannel = Channel<OutputT>()
+      scope.coroutineContext[Job]!!.invokeOnCompletion {
+        outputChannel.close(it)
```
zach-klippenstein, line 79: I think we need to check if `it` is a `CancellationException` and pass null through if it is. I think you said there's test coverage for this though, so I guess it works?

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowRunner.kt
```diff
@@ -48,15 +49,11 @@ interface WorkflowRunner<out OutputT> {
...
+   * Returns the next [OutputT] value emitted by the workflow. Throws the cancellation
    * exception if the workflow was cancelled before emitting.
    *
-   * The output of the root workflow is treated as a result code, handy for use
-   * as a sign that the host Activity or Fragment should be finished. Thus, once
-   * a value is emitted the workflow is ended and its output value is reported through
-   * this field.
    */
-  suspend fun awaitResult(): OutputT
+  suspend fun receiveOutput(): OutputT
```
zach-klippenstein, line 56: Should we expose the channel directly? I don't think so, we don't want to expose another cancellation API. Too bad there isn't a subset of `ReceiveChannel` that only exposes the select hooks without cancellation. And if we really need it, we can add later.

zach-klippenstein:
LG but want @rjrjr to take a look too.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowRunnerViewModel.kt
bnvinay92: These are the test scenarios AFAICT:
1. Verify cancellation is propagated to running workers when the runner's scope is cleared. `Factory cancels runtime when cleared`
2. Verify that the coroutine that is suspend-waiting on the next output is canceled when the runner's scope is cleared. `Factory cancels result when cleared`
3. When the runner scope is canceled, verify that a CancellationException is invoked on completion of runnerScope.coroutineContext[Job], and that the cause-chain of this exception are only CancellationExceptions. `hostCancelledOnCleared`

Is there another scenario we should test?

>check if it is a CancellationException and pass null through if it is
So we want future usages of the channel to throw a ClosedSendChannelException/ClosedReceiveChannelException instead of a CancellationException? Or is there another reason to do this?

bnvinay92:
Closes #109

PR #169, opened 2020-08-17
Allow fragments to consume outputs & customise layoutParams

bnvinay92:
Allows fragments to consume outputs of root workflows similar to activities.

If you commit a WorkflowFragment to a FrameLayout, then the created WorkflowLayout has layout params MATCH_PARENT. If you commit the fragment to a CoordinatorLayout though, it has layout params WRAP_CONTENT (due to different generateDefaultLayoutParams(). This change allows subclasses to specify layout params.

Stacked on top of https://github.com/square/workflow-kotlin/pull/168


rjrjr:
Thanks!

Issue #171, opened 2020-08-17
Upgrade to Kotlin 1.4


zach-klippenstein:
Currently blocked by https://github.com/charleskorn/kaml/issues/26.

PR #173, opened 2020-08-18
Remove dependency on Kaml for Yaml parsing.

zach-klippenstein:
This dependency always slows us down upgrading to new Kotlin versions, since
serialization is still pre-release and often breaks, requiring us to wait for
Kaml to release a new version before upgrading.

Unblocks #171.


samples/dungeon/app/src/main/assets/boards/simple_board.txt
```diff
@@ -1,5 +1,5 @@
 ---
-name: Simple Board
+{name: "Simple Board"}
 ---
 🌳🌳🌳🌳🌳🌳🌳🌳🌳🌳🌳🌳🌳🌳🌳🌳
```
rjrjr, line 4: I will never stop being delighted that this is "text".

zach-klippenstein:
Kaml just got updated to support 1.4, but I still want to drop it.

PR #174, opened 2020-08-18
Make Sink and LayoutRunner fun interfaces.

zach-klippenstein:
This allows consumers to create anonymous instances of them with Kotlin's lambda syntax when using Kotlin 1.4.


zach-klippenstein:
ktlint's not happy. Fix is on the way, see https://github.com/pinterest/ktlint/issues/830

steve-the-edwards (approved):
What does this gain? Less boilerplate with the SAM style?

zach-klippenstein:
Just rebased now that we landed 1.4.

zach-klippenstein:
> What does this gain? Less boilerplate with the SAM style?

Yes, we can pass lambdas now instead of having to create anonymous objects.

zach-klippenstein:
Need to update ktlint to land this, but that then complains about a bunch of our formatting. Tried reformatting in ide as well as using ktlintFormat task, it just got more and more confused about indentation. Need to review the tricks @vRallev  used to make this work.

zach-klippenstein:
This should finally work, after upgrading out ktlint version in #277.

Issue #170, opened 2020-08-17
Incorrect layout when update() is not called on all WorkflowViewStubs in a LinearLayout


bnvinay92:
Reproduced here: https://github.com/square/workflow-kotlin/compare/main...careem:reproduce-workflow-viewstub-issue

Layout:

```
<LinearLayout orientation="vertical">
  <WorkflowViewStub
    background="red"
    height="0dp"
    weight="1" />

  <TextView
    background="orange"
    height="wrap_content" />

  <com.squareup.workflow1.ui.WorkflowViewStub
    background="green"
    height="wrap_content" />
</LinearLayout>

```
Rendered as:

![Screenshot_1597680505](https://user-images.githubusercontent.com/22435681/90418355-13e2d280-e0c6-11ea-8478-5991ce6d032c.png)

If the red viewStub is inflated but the green one is not, then the red viewstub has zero height and the green view stub height matches parent. Inflating both results in a correct rendering.
More here: https://workflow-community.slack.com/archives/CHTFPR277/p1597683241264200?thread_ts=1597672694.261300&cid=CHTFPR277

rjrjr:
Discussion in #177 distills down to: add a `updateSetsVisibility` val / xml attribute, default to `true`.

rjrjr:
@bnvinay92 @zach-klippenstein Seems like we'd have the same issue with `setBackground`, no?

rjrjr:
Actually I guess that's not necessary, since `background` is nullable and only clobbers the background if it is set to something. The lack of parallelism between the two attributes is kind of uncomfortable, but then the value types and the use cases are pretty distinct too. Having the child be in charge of its own background is a pretty different concern than having it be in charge of its own visibility, and I think expectations differ.

PR #177, opened 2020-08-26
Don't inherit visibility from WorkflowViewStub

bnvinay92:
Setting the default visibility to GONE mimics android's `ViewStub` and addresses the scenario specified in #170.
This then causes the next view to be also GONE though, since visibilities are inherited.

Not sure how/when Square is setting visibilities on WorkflowViewStubs, and whether this would break things or not.
Perhaps letting users override visibility when calling update() is another solution?

This change allows LayoutRunners to set their own visibility to GONE if required.

### DIFF
--- a/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowViewStub.kt
+++ b/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowViewStub.kt
@@ -94,6 +94,7 @@ class WorkflowViewStub @JvmOverloads constructor(
     inflatedId = attrs.getResourceId(R.styleable.WorkflowViewStub_inflatedId, NO_ID)
     attrs.recycle()
 
+    visibility = GONE
     setWillNotDraw(true)
   }
 
@@ -114,8 +115,8 @@ class WorkflowViewStub @JvmOverloads constructor(
   }
 
   /**
-   * Sets the visibility of this stub as usual, and also that of [actual].
-   * Any new views created by [update] will be assigned this visibility.
+   * Sets the visibility of this stub as usual, and also that of [actual] if it exists.
+   * Any new views created by [update] will NOT be assigned this visibility.
    */
   override fun setVisibility(visibility: Int) {
     super.setVisibility(visibility)
@@ -166,6 +167,7 @@ class WorkflowViewStub @JvmOverloads constructor(
     actual.takeIf { it.canShowRendering(rendering) }
         ?.let {
           it.showRendering(rendering, viewEnvironment)
+          super.setVisibility(it.visibility)
           return it
         }
 
@@ -177,10 +179,10 @@ class WorkflowViewStub @JvmOverloads constructor(
     return viewEnvironment[ViewRegistry].buildView(rendering, viewEnvironment, parent)
         .also { newView ->
           if (inflatedId != NO_ID) newView.id = inflatedId
-          newView.visibility = visibility
           background?.let { newView.background = it }
           replaceOldViewInParent(parent, newView)
           actual = newView
+          super.setVisibility(actual.visibility)
         }
   }
 }
--- a/samples/dungeon/app/src/main/java/com/squareup/sample/dungeon/BoardsListLayoutRunner.kt
+++ b/samples/dungeon/app/src/main/java/com/squareup/sample/dungeon/BoardsListLayoutRunner.kt
@@ -16,7 +16,6 @@
 package com.squareup.sample.dungeon
 
 import android.view.View
-import android.view.View.INVISIBLE
 import android.view.View.VISIBLE
 import android.view.ViewGroup
 import android.widget.TextView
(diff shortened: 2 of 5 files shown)


workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowViewStub.kt
```diff
@@ -94,6 +94,7 @@ class WorkflowViewStub @JvmOverloads constructor(
     inflatedId = attrs.getResourceId(R.styleable.WorkflowViewStub_inflatedId, NO_ID)
     attrs.recycle()
 
+    visibility = GONE
```
bnvinay92, line 97: Makes WorkflowViewStub behave similar to ViewStub in the scenario specified in https://github.com/square/workflow-kotlin/issues/170

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowViewStub.kt
```diff
@@ -114,8 +115,8 @@ class WorkflowViewStub @JvmOverloads constructor(
   }
 
   /**
-   * Sets the visibility of this stub as usual, and also that of [actual].
-   * Any new views created by [update] will be assigned this visibility.
+   * Sets the visibility of this stub as usual, and also that of [actual] if it exists.
+   * Any new views created by [update] will NOT be assigned this visibility.
```
rjrjr, line 119: Here's a thought: what if we always propagate the visibility of `actual`

rjrjr: Maybe combination of that and your `update` idea. Give `update` an optional `@IdRes visibility : Int?` argument that defaults to null. If it's not set, we copy from `actual` (remember that `actual` defaults to `this`).

bnvinay92: Thinking some more about it, setting the visibility through update might not be a great idea since the order of inflation/visibility setting is ambiguous to the reader. It's easy enough to explicitly set the visibility after update if it needs to be overriden.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowViewStub.kt
```diff
@@ -177,10 +179,10 @@ class WorkflowViewStub @JvmOverloads constructor(
     return viewEnvironment[ViewRegistry].buildView(rendering, viewEnvironment, parent)
         .also { newView ->
           if (inflatedId != NO_ID) newView.id = inflatedId
-          newView.visibility = visibility
           background?.let { newView.background = it }
           replaceOldViewInParent(parent, newView)
           actual = newView
+          super.setVisibility(actual.visibility)
```
bnvinay92, line 185: Is this what you meant by 
>what if we always propagate the visibility of actual

I'm guessing the usecase is to allow reads of `actual.visibility` using the view stub as a proxy?

rjrjr: No. I meant that when a new view is created, it is assigned the visibility of the previous view — the current `actual`.

rjrjr: Basically, change nothing in this class except for line 180, and that becomes:

```
newView.visibility = actual.visibility
```

bnvinay92: The most minimal change that would fix the issue I was seeing would be to remove the code that sets visibility on actual in `update()`.

The issue occurs when the parent does not want to have any opinion on what the visibility of an inflated workflowViewStub should be.

For example: I have a toast-like view which may or may not be visible after a certain expiration timestamp.

```
data class MapToolTipRendering(
    val text: String,
    val expirationTimeMs: Long?,
)
```

Because the parent always overrides the visibility of the child, it would need to either:
- Know the specifics of how to calculate visibility from a sequence of `MapToolRendering`, which defeats the purpose of `WorkflowViewStub`.
- or, Not override the visibility of `actual`.

```
newView.visibility = actual.visibility
```
The above would make the visibility of `actual` harder to reason about since it would depend on the order of `previous` renderings and their associated runners that are sharing and mutating visibility.

rjrjr:
I'm still concerned that this change bends too far over backwards in the direction of this use case. It'll be pretty surprising to a lot of existing uses — I suspect we'll be plagued with bugs where child views start re-appearing on update.

How does this sound: WVS defaults to being unopinionated about the child's visibility. Once `WVS.setVisiblity` is called, it moves to a mode where it behaves like I suggested above, propgating the visibility of the previous `actual`. To keep this from being a one-way change, we can provide a `val updateSetsVisibility: Boolean`.

WDYT?

rjrjr:
Or really, the side effect notion is nasty. How about we just expose that `updateSetsVisibility` property / xml attribute. I think it should default to false, though, as I really believe the 90% case is for the parent to own that concern.

If that works for you, I should be able to put up a PR on Monday.

bnvinay92:
`updateSetsVisibility` sounds good.

PR #181, opened 2020-09-02
Fix ungrammatical error message

jedp:


zach-klippenstein:
@jedp Could I interest you in writing a lint check for this? Would save you and @rjrjr both a lot of trouble, since I am sure I will never get this right.

jedp:
Fat-fingered the "close" button like the GitHub Pro that I am

PR #182, opened 2020-09-07
Enforce distinct id and inflatedId values.

### LINKED ISSUE #176: WorkflowViewStub crashes when restoring wrong saved view state
It probably needs to do the same `restoredChildState` stuff that `WorkflowLayout` does. Surely we can consolidate these things.

rjrjr:
Closes #176

Cargo culting had lead to them using the same value throughout our codebase,
but we tend not to find out until it breaks view persistence, if ever.
It's easy enough to fail fast in that situation:

```
Caused by: java.lang.IllegalArgumentException: inflatedId and id must be distinct
at com.squareup.workflow1.ui.WorkflowViewStub.setInflatedId(WorkflowViewStub.kt:86)
at com.squareup.workflow1.ui.WorkflowViewStub.<init>(WorkflowViewStub.kt:96)
at com.squareup.workflow1.ui.WorkflowViewStub.<init>(WorkflowViewStub.kt:71)
at com.squareup.workflow1.ui.WorkflowViewStub.<init>(Unknown Source:8)
... 224 more
```

### DIFF
--- a/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowViewStub.kt
+++ b/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowViewStub.kt
@@ -98,7 +98,13 @@ class WorkflowViewStub @JvmOverloads constructor(
    * The id to be assigned to new views created by [update]. If the inflated id is
    * [View.NO_ID] (its default value), new views keep their original ids.
    */
-  @IdRes var inflatedId: Int
+  @IdRes var inflatedId: Int = NO_ID
+    set(value) {
+      require(value == NO_ID || value != id) {
+        "inflatedId and id must be distinct"
+      }
+      field = value
+    }
 
   init {
     val attrs = context.obtainStyledAttributes(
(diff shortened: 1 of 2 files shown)


workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowViewStub.kt
```diff
@@ -98,7 +98,13 @@ class WorkflowViewStub @JvmOverloads constructor(
    * The id to be assigned to new views created by [update]. If the inflated id is
    * [View.NO_ID] (its default value), new views keep their original ids.
    */
-  @IdRes var inflatedId: Int
+  @IdRes var inflatedId: Int = NO_ID
+    set(value) {
+      require(value == NO_ID || value != id) {
+        "inflatedId and id must be distinct"
```
zach-klippenstein, line 104: Nit: include the names of the id resources in this (and the other) message?

zach-klippenstein (approved):

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowViewStub.kt
rjrjr: Oh, of course. Will follow up.

rjrjr: https://github.com/square/workflow-kotlin/pull/184

PR #184, opened 2020-09-08
Improves exception message for conflicting ids in stubs.

rjrjr:


workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowViewStub.kt
```diff
@@ -119,7 +119,7 @@ class WorkflowViewStub @JvmOverloads constructor(
 
   override fun setId(@IdRes id: Int) {
     require(id == NO_ID || id != inflatedId) {
-      "id and inflatedId must be distinct"
+      "id must be distinct from inflatedId: ${resources.getResourceName(id)}"
```
steve-the-edwards, line 122: ${resources.getResourceName(**inflatedId**)

rjrjr: `id` is fine, it binds to the method parameter.

steve-the-edwards: ahh of course! they are the same :).

steve-the-edwards (approved):

zach-klippenstein (approved):
Beautiful.

PR #188, opened 2020-09-14
Upgrade (not Kotlin) and other dependencies

SalvatoreT:
To get a list of dependencies that could be updated, I used the versions plugin.

    gradle dependencyUpdates -Drevision=release

There were a number of libraries that I couldn't get upgraded quickly (dokka and ktlint among them), but hopefully this helps move thing forward.


build.gradle.kts
```diff
@@ -105,7 +106,7 @@ subprojects {
 }
 
 apply(from = rootProject.file(".buildscript/binary-validation.gradle"))
-
+apply(plugin = "com.github.ben-manes.versions")
```
zach-klippenstein, line 109: Why was this added?

samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/board/Parser.kt
```diff
@@ -106,7 +103,7 @@ private fun BufferedSource.readHeader(): String? = buildString {
   while (true) {
     val line = readUtf8Line() ?: throw IllegalArgumentException("Expected --- but found EOF.")
     if (line == YAML_DELIMITER) return@buildString
-    appendln(line)
+    appendLine(line)
```
zach-klippenstein, line 106: We can't use the stdlib from kotlin 1.4 yet, since we can't use kotlin 1.4 internally.

versions.properties
```diff
@@ -3,46 +3,47 @@
...
 version.io.reactivex.rxjava2..rxandroid=2.1.1
 version.io.reactivex.rxjava2..rxjava=2.2.19
 version.junit.junit=4.13
-version.kotlin=1.3.72
-version.kotlinx.coroutines=1.3.7
-version.kotlinx.serialization=0.20.0
-version.me.champeau.gradle..jmh-gradle-plugin=0.5.0
-version.moshi=1.9.3
+version.kotlin=1.4.10
+version.kotlinx.coroutines=1.3.9
```
zach-klippenstein, line 37: Can't upgrade coroutines yet because we can't require the 1.4 stdlib.

versions.properties
```diff
@@ -3,46 +3,47 @@
...
 version.io.mockk..mockk=1.10.0
 version.io.reactivex.rxjava2..rxandroid=2.1.1
 version.io.reactivex.rxjava2..rxjava=2.2.19
 version.junit.junit=4.13
-version.kotlin=1.3.72
-version.kotlinx.coroutines=1.3.7
-version.kotlinx.serialization=0.20.0
-version.me.champeau.gradle..jmh-gradle-plugin=0.5.0
-version.moshi=1.9.3
+version.kotlin=1.4.10
```
zach-klippenstein, line 36: We can build with 1.4, but we need to restrict to the 1.3.x stdlib.

zach-klippenstein (changes requested):
Unfortunately we can't really use 1.4 yet.

build.gradle.kts
SalvatoreT: This allows us to see which libraries need updating with the following command:

```
gradle dependencyUpdates -Drevision=release
```

samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/board/Parser.kt
SalvatoreT: ☹️

versions.properties
SalvatoreT: ☹️

build.gradle.kts
zach-klippenstein: We're already using one plugin to manage dependencies, I'm very reluctant to add a second.

SalvatoreT:
updated to remove Kotlin 1.4 and libraries that require Kotlin 1.4

build.gradle.kts
```diff
@@ -105,7 +105,6 @@ subprojects {
 }
 
 apply(from = rootProject.file(".buildscript/binary-validation.gradle"))
-
```
zach-klippenstein, line 108: This blank line was intentional, since the below lines are unrelated.

zach-klippenstein (changes requested):
One last (accidental?) change that needs to be reverted, but then we can merge this.

SalvatoreT:
Updated!

zach-klippenstein:
LG, that instrumentation test failure was an infra issue but I want to see it pass just to make sure, then I'll merge.

PR #191, opened 2020-09-16
fixes #190: Add guarded setOutput to enterState ReplaceWith expression

### LINKED ISSUE #190: Update ReplaceWith expressions for WorkflowAction `enterState` deprecation to account for output
If we use auto-complete right now when emittingOutput is non-null, we can lose the output.

steve-the-edwards:
This is so we do not accidentally wipe out outputs when using suggested action.

fixes #190

### DIFF
--- a/workflow-core/src/main/java/com/squareup/workflow1/WorkflowAction.kt
+++ b/workflow-core/src/main/java/com/squareup/workflow1/WorkflowAction.kt
@@ -78,7 +78,10 @@ abstract class WorkflowAction<in PropsT, StateT, out OutputT> {
     @Deprecated(
         message = "Use action",
         replaceWith = ReplaceWith(
-            expression = "action { state = newState }",
+            expression = "action {\n" +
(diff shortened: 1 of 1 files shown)


workflow-core/src/main/java/com/squareup/workflow1/WorkflowAction.kt
```diff
@@ -78,7 +78,10 @@ abstract class WorkflowAction<in PropsT, StateT, out OutputT> {
     @Deprecated(
         message = "Use action",
         replaceWith = ReplaceWith(
-            expression = "action { state = newState }",
+            expression = "action {\n" +
+                "  state = newState\n" +
+                "  emittingOutput?.let { setOutput(it) }\n" +
```
zach-klippenstein, line 83: I think this might work?
```suggestion
                "  emittingOutput?.let(::setOutput)\n" +
```

zach-klippenstein (approved):

PR #196, opened 2020-09-29
First tutorial with code.

zach-klippenstein:


.github/workflows/tutorials.yml
```diff
@@ -0,0 +1,25 @@
...
+
+jobs:
+  build:
+    name: Build Tutorials
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      # These setup steps should be common across all jobs in this workflow.
+      - uses: actions/checkout@v2
+      - name: set up JDK 1.8
```
rjrjr, line 19: Not 1.11?

samples/tutorial/Tutorial1.md
```diff
@@ -0,0 +1,239 @@
...
+  /** The current name that has been entered. */
+  val name: String,
+  /** Callback when the name changes in the UI. */
+  val onNameChanged: (String) -> Unit,
+  /** Callback when the login button is tapped. */
+  val onLoginTapped: () -> Unit
+)
+```
+
+Then we need to create a `ViewFactory` that knows how to create an Android `View` to draw the actual screen. The easiest way to create a `ViewFactory` is to create a layout runner. A layout runner is a class that has a reference to the view and knows how to update the view given an instance of a screen. In a typical app, every screen will have a layout runner. Layout runners can also work with AndroidX `ViewBinding`s, which we'll use to define the `WelcomeLayoutRunner`. We have a pre-built `WelcomeViewBinding` that you can use, but if you would like to create and lay out the view yourself, feel free to do so!
```
rjrjr, line 46: Shouldn't the first _layout runner_ instead be `LayoutRunner`?

rjrjr (approved):

samples/tutorial/Tutorial1.md
zach-klippenstein: Isn't it? 
image

samples/tutorial/tutorial-1-complete/src/main/java/workflow/tutorial/WelcomeWorkflow.kt
```diff
@@ -0,0 +1,41 @@
...
+    snapshot: Snapshot?
+  ): State = State(name = "")
+
+  override fun render(
+    props: Unit,
+    state: State,
+    context: RenderContext
+  ): WelcomeScreen = WelcomeScreen(
+    name = state.name,
+    onNameChanged = { context.actionSink.send(onNameChanged(it)) },
```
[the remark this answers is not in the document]
zach-klippenstein: Feel free to file an issue, we can think about it more.

.github/workflows/tutorials.yml
zach-klippenstein: #239

samples/tutorial/Tutorial1.md
steve-the-edwards: I think he means the class reference, not the general noun.

samples/tutorial/README.md
```diff
@@ -0,0 +1,31 @@
...
+## Layout
+
+The project has both a starting point, as well as an example of the completed tutorial.
+
+To help with the setup, we have created a few helpers:
+
+- `tutorial-views`: A set of 3 views for the 3 screens we will be building, `Welcome`, `TodoList`,
+  and `TodoEdit`.
+- `tutorial-base`: This is the starting point to build out the tutorial. It contains view
+  controllers that host the views from `TutorialViews` to see how they display.
```
steve-the-edwards, line 17: "view controllers"? Was this from the iOS one?

```suggestion
- `tutorial-base`: This is the starting point to build out the tutorial.
```

samples/tutorial/README.md
```diff
@@ -0,0 +1,31 @@
...
+## Overview
+
+Oh hi! Looks like you want build some software with Workflows! It's a bit different from traditional
+Android development, so let's go through building a simple little TODO app to get the basics down.
+
+## Layout
+
+The project has both a starting point, as well as an example of the completed tutorial.
+
+To help with the setup, we have created a few helpers:
```
steve-the-edwards, line 12:
```suggestion
To help with the setup, we have created a few helper modules:
```

samples/tutorial/Tutorial1.md
```diff
@@ -0,0 +1,245 @@
...
+Start by creating a new workflow and screen by creating a new file with the [file templates](../../install-templates.sh), adding it to the `tutorial-base` module:
+
+![New Workflow](images/new-workflow.png)
+![Workflow Name](images/workflow-name.png)
+
+Follow the same steps using the `Layout Runner (ViewBinding)` template.
+
+### Screens, View Factories, and Layout Runners
+
+Let's start with what a "screen" is, and how it relates to the view controller.
```
steve-the-edwards, line 30:
```suggestion
Let's start with what a "screen" is, and how it relates to the Android `View`.
```

```suggestion
Let's start with what a "screen" is, and how it relates to the Workflow.
```

```suggestion
Let's start with what a "screen" is, and how it relates to a view model.
```

samples/tutorial/Tutorial1.md
```diff
@@ -0,0 +1,245 @@
...
+  ): State = State(name = "")
+
+  // …
+}
+
+```
+
+Now that we have the state modeled, we'll send it to the UI every time a render pass happens. The text field will overwrite its value with whatever was provided.
+
+If you run the app again, you'll see that it still behaves the same, letting your type into the name field. This is because we have only rendered the screen once.
```
steve-the-edwards, line 167:
```suggestion
If you run the app again, you'll see that it still behaves the same, letting you type into the name field. This is because we have only rendered the screen once.
```

samples/tutorial/Tutorial1.md
```diff
@@ -0,0 +1,245 @@
...
+
+You may have noticed that your workflow only has access to its `State` in a few functions, and even then in many cases it is read-only. This is intentional. The Workflow infrastructure manages the state for the various workflows that are running and triggers a re-render when appropriate. In order to update the workflow's internal state, we need to add an "Action",
```
steve-the-edwards, line 169:
```suggestion
You may have noticed that your workflow only has access to its `State` in a few functions, and even then in many cases it is read-only. This is intentional. Instead of allowing your Workflow to modify the state directly, the Workflow infrastructure manages the state for the various workflows that are running and triggers a re-render when appropriate. In order to update the workflow's internal state, we need to add an "Action",
```

samples/tutorial/Tutorial1.md
```diff
@@ -0,0 +1,245 @@
...
+```kotlin
+  private fun onNameChanged(name: String) =
+    object : WorkflowAction<Unit, State, Nothing>() {
+      override fun Updater.apply() {
+        this.state = this.state.copy(name = name)
+      }
+    }
+```
+
+We need to send this action back to the workflow any time the name changes. Update the `render` method to send it through a sink back to the workflow whenever the `onNameChanged` closure is called:
```
steve-the-edwards, line 194: While 'sync' is a general concept, it could use a quick sidebar/explanation here.

samples/tutorial/Tutorial1.md
```diff
@@ -0,0 +1,245 @@
...
+    - `render` is called on the workflow. A new screen is returned with the updated `name` from the internal state.
+5) The layout runner is provided the new screen with the call to `fun showRendering(rendering: WelcomeScreen, viewEnvironment: ViewEnvironment)`.
+    - This layout runner updates the text field with the received name value, and also updates the callbacks for when the name changes or login is pressed.
+6) The workflow waits for the next `WorkflowAction` to be received, and then the goes through the same update loop.
+
+## Summary
+
+In this tutorial, we covered creating a Screen, `LayoutRunner`, Workflow, and binding them together in an Activity with `ViewRegistry` and `setContentWorkflow`. We also covered the Workflow being responsible for the state of the UI instead of the `LayoutRunner` or `View` being responsible.
+
+Next, we will create a second screen and workflow, and the use composition to navigate between them.
```
steve-the-edwards, line 243:
```suggestion
Next, we will create a second screen and workflow, and then use composition to navigate between them.
```

samples/tutorial/Tutorial1.md
zach-klippenstein: @rjrjr Can you elaborate? I'm very confused.

samples/tutorial/README.md
zach-klippenstein: Good catch, found a few more as well.

samples/tutorial/Tutorial1.md
zach-klippenstein: Added a bit.

PR #198, opened 2020-09-30
Introduce helpers for working with EditTexts from LayoutRunners.

zach-klippenstein:
This is similar to what we have internally, but also lets you update selection by passing it in via the incoming `CharSequence`.

Related to https://github.com/square/workflow/issues/1213, although doesn't attempt to solve
any cross-platform issues yet.

### DIFF
--- a/workflow-ui/core-android/src/androidTest/java/com/squareup/workflow1/ui/EditTextsTest.kt
+++ b/workflow-ui/core-android/src/androidTest/java/com/squareup/workflow1/ui/EditTextsTest.kt
@@ -1,0 +1,108 @@
+package com.squareup.workflow1.ui
+
+import android.text.Selection
+import android.text.SpannableStringBuilder
+import android.widget.EditText
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import androidx.test.platform.app.InstrumentationRegistry
+import com.google.common.truth.Truth.assertThat
+import org.junit.Assert.fail
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@OptIn(WorkflowUiExperimentalApi::class)
+@RunWith(AndroidJUnit4::class)
+class EditTextsTest {
+
+  private val instrumentation = InstrumentationRegistry.getInstrumentation()
+  private val editText = EditText(instrumentation.context)
+
+  @Test fun updateText_setsTextValue() {
+    assertThat(editText.text.toString()).isEqualTo("")
+
+    editText.updateText("h")
+    assertThat(editText.text.toString()).isEqualTo("h")
+
+    editText.updateText("hello")
+    assertThat(editText.text.toString()).isEqualTo("hello")
+  }
+
+  @Test fun updateText_setsSelectionWhenSpecified() {
+    editText.setText("hello")
+    assertThat(editText.selectionStart).isEqualTo(0)
+    assertThat(editText.selectionEnd).isEqualTo(0)
+
+    val newText = SpannableStringBuilder("hello")
+        .also { Selection.setSelection(it, 1, 3) }
+    editText.updateText(newText)
+    assertThat(editText.text.toString()).isEqualTo("hello")
+    assertThat(editText.selectionStart).isEqualTo(1)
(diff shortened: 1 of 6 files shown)


workflow-ui/core-android/src/androidTest/java/com/squareup/workflow1/ui/EditTextsTest.kt
```diff
@@ -0,0 +1,108 @@
...
+import org.junit.Assert.fail
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@OptIn(WorkflowUiExperimentalApi::class)
+@RunWith(AndroidJUnit4::class)
+class EditTextsTest {
+
+  private val instrumentation = InstrumentationRegistry.getInstrumentation()
+  private val editText = EditText(instrumentation.context)
```
rjrjr, line 18: whoa

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/EditTexts.kt
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * Helper for setting the text value of an [EditText] without changing any metadata about the text,
+ * unless [text] includes a selection. If a selection is passed in, then the selection of [editText]
+ * is set to the incoming selection.
+ *
+ * Intended to be used by [LayoutRunner]s for updating [EditText]s from workflow renderings.
+ */
+@WorkflowUiExperimentalApi
+fun updateText(
```
zach-klippenstein, line 17: Intentionally didn't define these as extension functions, since they are specifically for use with workflows. But open to changing that if people think it makes sense.

workflow-ui/core-android/src/androidTest/java/com/squareup/workflow1/ui/EditTextsTest.kt
rjrjr: I had no idea that we could test isolated views his way.

Should be doing this to test `ViewRegistry` _much_ better.

rjrjr (approved):

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/EditTexts.kt
rjrjr: I'm pro-extension and pro-discoverability. Is there any reason to think they're dangerous for non-workflow uses?

workflow-ui/core-android/src/androidTest/java/com/squareup/workflow1/ui/EditTextsTest.kt
zach-klippenstein: Neither did I, really didn't expect this to work lol.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/EditTexts.kt
zach-klippenstein: I felt more strongly about this yesterday, but I don't remember why. I'll make them extensions.

zach-klippenstein: Oh, it might have been because, when the listener is an extension, you see these three options in the autocomplete:
```
fun addTextChangedListener
fun removeTextChangedListener
fun setTextChangedListener
```
and it's not obvious that the last has a completely different listener type than the others.

rjrjr: I dunno. If you see those it's because you opted in by using Workflow in the first place. And on top of that you'll get an experimental warning.

rjrjr: Maybe you should offer them in an androidx PR. :)

zach-klippenstein: :sweatingtowelguy:

PR #201, opened 2020-10-01
Fix some issues with the initial EditTexts implementations.

zach-klippenstein:
Remove the recyclerview sample since it doesn't really represent any real-world
use cases.


rjrjr (approved):

rjrjr:
Just kind of amazing that none of this is built in.

rjrjr:
This really would make a good addition to androidx. Copious spare time and all that.

zach-klippenstein:
> This really would make a good addition to androidx.

It is called Compose ;)

PR #202, opened 2020-10-02
wip: ViewRegistry overhaul, ViewRendering, ModalRendering

### LINKED ISSUE #195: Decouple modality from modal visual treatment
@bencochran @zach-klippenstein @AquaGeek @dhavalshreyas 

I had a chat last week with @helios175, and he schooled me on decoupling modal visual style from layering.

e.g., in Kotlin right now we have a number of samples where the top level rendering type is something like…

```
AlertContainerScreen<CardContainerScreen<BackstackScreen<Any>>
```

That is, you can show alerts over cards. Not so bad. But in our prodution app, where we have both a number of modal layers _and_ a number of modal styles, the combinatorics get kind of nuts, something like:

```
typealias Root =
    AlertContainerScreen<
        TutorialBarsContainerScreen<
            CardContainerScreen<
                SheetContainerScreen<
                    CardContainerScreen<Any, Any>,
                    Any
                    >,
                Any
                >
            >
        >
```

because we want
 * alerts to be modal over the tutorial bars
 * tutorial bars to be outside the area covered by cards and sheets
 * the ability to put sheets over cards visually, but then sometimes another card on top of that

Instead of this, the "big idea" (probably obvious to anyone who isn't me) is to have a single container type that is able to make things modal, wrapping others that own the visual style. This works because our modal screens already hold lists to allow visual stacking.

So the above would reduce to something like:

```
typealias Root =
    ModalContainerScreen<
        TutorialBarsContainerScreen<
            ModalContainerScreen<Any, Any>
            >,
        Any
        >
```

With `AlertModal<T>`, `CardModal<T>`, `SheetModal<T>`, `BottomSheetModal<T>`, etc., as possible contituents. If I want to show an alert over a card over a sheet, my root workflow assembles the equivalent of:

```
ModalContainerScreen(
  modals = listOf(AlertModal(...)),
  body = TutorialBarsContainerScreen(
    ModalContainerScreen(
      modals = listOf(
        SheetModal(...), CardModal(...)
      )
      body = ...
    )
  )
)
```

The downside here is that some layering rules become slightly more runtime. E.g. we don't want every child workflow that may want to show modals to have to know anything about tutorials, or these layering rules. An inner workflow should be able to render something like:

```
ModalContainerScreen(
  modals = listOf(CardModal(...), AlertModal("Are you sure?", ...)),
  body = ...
)
```

So some tutorial-aware outer workflow will need to filter the `modals` lists of its children, and move `AlertModal` instances to an outer layer. But that doesn't seem like a big deal.

Anyway, I'm giving myself permission to play with this today because I need to solve a problem where I'm getting double-darkness when showing cards over cards.

WDYT?

rjrjr:
Builds the simpler, more flexible scheme for modals proposed in #195. Only
AlertDialog takes advantage of it so far, and only the `hellobackbutton` and
`helloworkflow` samples are ported.

closes #195

 * We introduce the `ViewRendering` and `ModalRendering` interfaces,
   similar in sprit to `AnyScreen` in Swift.

 * `ViewRegistry` changes to be able to hold a collection of more arbitrary entry types, not just `ViewFactory`.
    * Its `buildView` method is deprecated.
    * Once we delete that method, we should be able to move `ViewRegistry` over to the
      non-Android, core UI module.

 * There are three implementations of `ViewRegistry.Entry`
    * Our old friend `ViewFactory<T: Any>`, now deprecated.
    * `ViewBuilder<T: ViewRendering>`, which replaces `ViewFactory`
    * `DialogBuilder<T: ModalRendering>`

 * Extension method `ViewRendering.buildView()` replaces `ViewRegistry.buildView()`. We also introduce  `ModalRendering.buildDialog()`.

    * These methods find the appropriate builder in `ViewRegistry`, based on the concrete type of the rendering, just like `ViewRegistry.buildView()` did.

    * The new ability of `ViewRegistry` to hold arbitrary entries means that we can add as many rendering:view-machinery pairings as we want to. Maybe cleans up Compose integration? (@zach-klippenstein )

 * The abstract `ModalContainer` and its subclasses (`AlertContainer`, `ModalViewContainer`, our internal ones) are all deprecated in favor of a single final class: `ModalContainerView`, which is able to display any kind of `ModalRendering` associated with a `DialogBuilder` implementation.

 * `LayoutRunner<T: Any>` is deprecated, replaced by `ViewRunner<T: ViewRendering>`. That's going to be a tedious migration for existing `LayoutRunner` implementations, but it avoids a huge breaking change —
 the new mechanism works in parallel with the deprecated one.

 * `WorkflowViewStub.display(rendering: ViewRendering)` replaces `WorkflowViewStub.update(rendering: Any)`, now deprecated. But the latter is able to delegate to the former so incremental migration is possible.


workflow-ui/core-android/build.gradle.kts
```diff
@@ -40,6 +40,7 @@ dependencies {
   api(Dependencies.Kotlin.Stdlib.jdk6)
 
   implementation(Dependencies.AndroidX.activity)
+  implementation(Dependencies.AndroidX.appcompat)
```
rjrjr, line 43: Uh-oh. Did we move this out of core to keep @Armaxis happy? Is that still a concern?

I think I can avoid it, but it'll be a bit tedious.

Armaxis: No longer a concern, with the amount of things we need these day, appcompat is imminent. 
Bring it back. Sorry for bothering you remove it in first place :/

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/BespokeViewBuilder.kt
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * This pattern allows us to assemble a [ViewRegistry] out of the
+ * custom classes themselves.
+ *
+ *    val TicTacToeViewBuilders = ViewRegistry(
+ *        MyView, GamePlayViewRunner, GameOverViewRunner
+ *    )
+ */
+@WorkflowUiExperimentalApi
+class BespokeViewBuilder<RenderingT : ViewRendering>(
```
rjrjr, line 39: Not even sure if this is worth providing. Should try just implementing the interface somewhere.

workflow-ui/core-android/build.gradle.kts
rjrjr: Yay! Glad I asked.

samples/containers/hello-back-button/src/main/java/com/squareup/sample/hellobackbutton/AreYouSureWorkflow.kt
```diff
@@ -75,7 +64,7 @@ object AreYouSureWorkflow : StatefulWorkflow<Unit, State, Finished, AlertContain
         )
       }
       Quitting -> {
-        val dialog = AlertScreen(
+        val dialog = AlertModalRendering(
```
rjrjr, line 67: This is the punchline. `AlertModalRendering` is bound in the `ViewRegistry` to `AlertDialogBuilder`. We didn't have to make a subclass of `ModalContainerView` in order to show this particular kind of modal.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewRegistry.kt
```diff
@@ -153,8 +104,66 @@ fun <RenderingT : Any> ViewRegistry.buildView(
...
 @WorkflowUiExperimentalApi
 operator fun ViewRegistry.plus(other: ViewRegistry): ViewRegistry = ViewRegistry(this, other)
+
+@WorkflowUiExperimentalApi
+fun <RenderingT : ViewRendering> RenderingT.buildView(
+  initialViewEnvironment: ViewEnvironment,
+  contextForNewView: Context,
+  container: ViewGroup? = null
+): View {
+  val builder = initialViewEnvironment[ViewRegistry].getEntryFor(this::class)
```
rjrjr, line 119: Very tempting to allow the `RenderingT` itself to implement `ViewBuilder`

```kotlin
  val builder = (this as? ViewBuilder<RenderingT>) 
      ?: initialViewEnvironment[ViewRegistry].getEntryFor(this::class)
```

rjrjr: Actually, better to do it in the other order so that `ViewRegistry` can override default implementations.


```kotlin
  val builder = initialViewEnvironment[ViewRegistry].getEntryFor(this::class)
      ?: (this as? ViewBuilder<RenderingT>)
```

zach-klippenstein: IIRC the way we talked ourselves out of that was because we didn’t want to encourage screen value types from being coupled to the UI layer. But I think in reality, screen types and view factories are often defined in public-facing, API, android-dependent modules already. And this isn’t requiring that, it merely allows it, which can be super useful for things like the `WorkflowEditText` wiring, the `ComposeRendering` PoC, etc. So, I'm on board.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/DisplayRendering.kt
```diff
@@ -0,0 +1,233 @@
...
+    else -> displaying as RenderingT
+  }
+}
+
+/**
+ * Returns the most recent rendering displayed by this [Dialog], or null if [Dialog.bindDisplayFunction]
+ * has never been called.
+ */
+@WorkflowUiExperimentalApi
+fun <RenderingT : ModalRendering> Dialog.getDisplaying(): RenderingT? {
```
steve-the-edwards, line 171: Is it worth an interface with `showRenderingTag` and `displaying` so we can abstract some of these duplicated functions into the same implementation?

steve-the-edwards: nvm I see these extension properties are defined below (which of course they are as this is Dialog and View). No clear way to pull out an abstraction as an extension on Dialog and View I guess. 🤔

steve-the-edwards: I guess you do almost already have an extension abstraction in ShowRenderingTag - why can't all these other extensions delegate into functions defined on that class itself. That way we wouldn't have the double maintenance. Granted this may not scale much beyond Dialog and View.

steve-the-edwards: Except for would it if we introduce Compose?

rjrjr:
To Do

- [ ] Add an optional window tear down hook so that we can nudge leak canary when the window should go away

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ModalContainerView.kt
```diff
@@ -0,0 +1,168 @@
...
+    newScreen: ModalContainerViewRendering<*, *>,
+    viewEnvironment: ViewEnvironment
+  ) {
+    baseView.show(newScreen.beneathModals, viewEnvironment)
+
+    val updateDialogs = mutableListOf<Dialog>()
+    for ((i, modalRendering) in newScreen.modals.withIndex()) {
+      updateDialogs +=
+        if (i < dialogs.size && compatible(dialogs[i].getDisplaying()!!, modalRendering)) {
+          dialogs[i].apply { display(modalRendering, viewEnvironment) }
```
bnvinay92, line 44: AFAIK alert dialogs can't be updated in place. Perhaps AlertModalRendering should implement Compatible?

zach-klippenstein:
This change seems to require all screen renderings to be completely opinionated about whether they should be a modal or not. Is that true? What happens if a rendering type wants to be usable as either a modal or a non-modal, and so implements both marker interfaces? Do we explicitly allow/forbid that, do we check for it anywhere?

As a straw-view, consider [`ComposeRendering`](https://github.com/square/workflow-kotlin-compose/blob/main/core-compose/src/main/java/com/squareup/workflow/ui/compose/ComposeRendering.kt): This is a generic rendering type that can render itself. It doesn't care if it's in a modal or not, it just defines _how_ to render. Could it implement both marker interfaces?

samples/containers/hello-back-button/src/main/java/com/squareup/sample/hellobackbutton/AreYouSureWorkflow.kt
```diff
@@ -25,23 +10,27 @@ import com.squareup.workflow1.Snapshot
...
-import com.squareup.workflow1.ui.modal.AlertScreen.Button.NEGATIVE
-import com.squareup.workflow1.ui.modal.AlertScreen.Button.POSITIVE
-import com.squareup.workflow1.ui.modal.AlertScreen.Event.ButtonClicked
-import com.squareup.workflow1.ui.modal.AlertScreen.Event.Canceled
 import com.squareup.workflow1.ui.toParcelable
 import com.squareup.workflow1.ui.toSnapshot
 import kotlinx.android.parcel.Parcelize
 
+@OptIn(WorkflowUiExperimentalApi::class)
+private typealias Rendering = ModalContainerViewRendering<*, AlertModalRendering>
```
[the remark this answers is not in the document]
rjrjr: Note that this one is private -- it was strictly a convenience to avoid some copy / paste, and arguably a bad choice for the reaons you mentioned. Our trend has been to use fewer and fewer type aliases, especially public ones.

Some of the lingering ones are leftovers from before we had the more concise `action {}` lambdas.

rjrjr:
> This change seems to require all screen renderings to be completely opinionated about whether they should be a modal or not. Is that true? What happens if a rendering type wants to be usable as either a modal or a non-modal, and so implements both marker interfaces? Do we explicitly allow/forbid that, do we check for it anywhere?
> 
> As a straw-view, consider [`ComposeRendering`](https://github.com/square/workflow-kotlin-compose/blob/main/core-compose/src/main/java/com/squareup/workflow/ui/compose/ComposeRendering.kt): This is a generic rendering type that can render itself. It doesn't care if it's in a modal or not, it just defines _how_ to render. Could it implement both marker interfaces?

My original thought was that `ComposeRendering` would be a completely different beast, the fourth implementation of `ViewRegistry.Entry`. Haven't dug in at all to see if that makes sense.

PR #203, opened 2020-10-06
Make core RenderTester.expectWorkflow method internal.

zach-klippenstein:
To do this, RenderTester has been turned into an abstract class, with all abstract methods.

This solves the issue of AS/IJ autocomplete giving priority to the least-used method, and then once you've made that mistake, not offering to import the extensions, because now everything is an extension.

cc @steve-the-edwards

### DIFF
--- a/workflow-testing/src/main/java/com/squareup/workflow1/testing/RenderTester.kt
+++ b/workflow-testing/src/main/java/com/squareup/workflow1/testing/RenderTester.kt
@@ -221,7 +221,7 @@ fun <PropsT, StateT, OutputT, RenderingT>
  *     rendering.onCancelClicked()
  * ```
  */
-interface RenderTester<PropsT, StateT, OutputT, RenderingT> {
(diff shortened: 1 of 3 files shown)


workflow-testing/src/main/java/com/squareup/workflow1/testing/RenderTester.kt
```diff
@@ -237,7 +237,7 @@ interface RenderTester<PropsT, StateT, OutputT, RenderingT> {
    * expectation by returning a [ChildWorkflowMatch]. If the expectation matches, the function
    * must include the rendering and optional output for the child workflow.
    */
-  fun expectWorkflow(
+  internal abstract fun expectWorkflow(
```
zach-klippenstein, line 240: We don't expect there to be any public callers of this method, so just hiding it completely. We can always make it public later.

rjrjr (approved):

steve-the-edwards (approved):

PR #212, opened 2020-10-14
Hello RenderContext.eventHandler

### LINKED ISSUE #207: Bring back RenderContext.onEvent()
Because this:

```kotlin
onBackPressed = context.onEvent {
  action { setOutput(BackSelected) }
}
```
is a lot nicer than this:
```kotlin
onSplitTenderPressed = {
  context.actionSink.send(
      action { setOutput(SplitTenderSelected) }
  )
}
```

### LINKED ISSUE #208: Can we do better than RenderContext.onEvent?
Here's the dream:

```kotlin
    onSplitTenderPressed = context.eventHandler { event -> 
      state = when(event.foo) { ... }
      setOutput(event.blah)
    }
```

I _think_ we can get that done by adding an `EventT` param to `WorkflowAction.Updater`, and `action {}`. That should be palatable if we can provide convenience overloads for when `EventT` is `Unit`, so that existing `action {}` calls don't break. If we can't pull that off, doesn't seem worthwhile.

rjrjr:
Introduces `BaseRenderContext.eventHandler` to replace the deprecated
`BaseRenderContext.onEvent`. Also deprecates
`BaseRenderContext.makeEventSink`.

Tried to make these extension methods rather than building them into the
interace, but that caused the IDE to complain that it couldn't choose the
correct overload.

closes #207, closes #208

### DIFF
--- a/samples/tictactoe/common/src/main/java/com/squareup/sample/gameworkflow/RunGameWorkflow.kt
+++ b/samples/tictactoe/common/src/main/java/com/squareup/sample/gameworkflow/RunGameWorkflow.kt
@@ -1,32 +1,7 @@
-/*
- * Copyright 2017 Square Inc.
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *     http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
 package com.squareup.sample.gameworkflow
 
 import com.squareup.sample.container.panel.PanelContainerScreen
 import com.squareup.sample.container.panel.firstInPanelOver
-import com.squareup.sample.gameworkflow.Action.CancelNewGame
-import com.squareup.sample.gameworkflow.Action.ConfirmQuit
-import com.squareup.sample.gameworkflow.Action.ConfirmQuitAgain
-import com.squareup.sample.gameworkflow.Action.ContinuePlaying
-import com.squareup.sample.gameworkflow.Action.Exit
-import com.squareup.sample.gameworkflow.Action.HandleLogGame
-import com.squareup.sample.gameworkflow.Action.PlayAgain
-import com.squareup.sample.gameworkflow.Action.StartGame
-import com.squareup.sample.gameworkflow.Action.StopPlaying
-import com.squareup.sample.gameworkflow.Action.TrySaveAgain
 import com.squareup.sample.gameworkflow.Ending.Quitted
 import com.squareup.sample.gameworkflow.GameLog.LogResult.LOGGED
 import com.squareup.sample.gameworkflow.GameLog.LogResult.TRY_LATER
@@ -43,7 +18,7 @@ import com.squareup.sample.gameworkflow.SyncState.SAVING
 import com.squareup.workflow1.Snapshot
 import com.squareup.workflow1.StatefulWorkflow
 import com.squareup.workflow1.Workflow
-import com.squareup.workflow1.WorkflowAction
+import com.squareup.workflow1.action
 import com.squareup.workflow1.runningWorker
 import com.squareup.workflow1.rx2.asWorker
 import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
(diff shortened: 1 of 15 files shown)


workflow-core/src/main/java/com/squareup/workflow1/StatefulWorkflow.kt
```diff
@@ -85,7 +85,13 @@ abstract class StatefulWorkflow<
...
-  inner class RenderContext internal constructor(
+  open inner class RenderContext internal constructor(
+    baseContext: BaseRenderContext<PropsT, StateT, OutputT>
+  ) : BaseRenderContext<@UnsafeVariance PropsT, StateT, @UnsafeVariance OutputT> by baseContext
+
+  inner class RenderContextDeluxe internal constructor(
+    val props: @UnsafeVariance PropsT,
+    val state: StateT,
     baseContext: BaseRenderContext<PropsT, StateT, OutputT>
   ) : BaseRenderContext<@UnsafeVariance PropsT, StateT, @UnsafeVariance OutputT> by baseContext
```
zach-klippenstein, line 96: Not sure why `RenderContext` needs to be open here, `RenderContextDeluxe` isn't subclassing it (and that's good – the less inheritance the better).

workflow-core/src/main/java/com/squareup/workflow1/StatefulWorkflow.kt
```diff
@@ -138,11 +144,17 @@ abstract class StatefulWorkflow<
...
     props: PropsT,
     state: StateT,
     context: RenderContext
-  ): RenderingT
+  ): RenderingT {
+    return RenderContextDeluxe(props, state, context).render()
+  }
+
+  open fun RenderContextDeluxe.render(): RenderingT {
+    error { "Must override one render method or the other" }
```
zach-klippenstein, line 156: Ugh, this pattern again.

workflow-core/src/main/java/com/squareup/workflow1/StatefulWorkflow.kt
```diff
@@ -85,7 +85,13 @@ abstract class StatefulWorkflow<
     out RenderingT
     > : Workflow<PropsT, OutputT, RenderingT> {
 
-  inner class RenderContext internal constructor(
+  open inner class RenderContext internal constructor(
+    baseContext: BaseRenderContext<PropsT, StateT, OutputT>
+  ) : BaseRenderContext<@UnsafeVariance PropsT, StateT, @UnsafeVariance OutputT> by baseContext
+
+  inner class RenderContextDeluxe internal constructor(
```
zach-klippenstein, line 92: In similar sketches I've done, since this change is largely to reduce boilerplate, I just call this type `Ctx` so it doesn't take up as much visual space, or as long to type if people are typing these by hand.

zach-klippenstein (commented):
I like this change in general, because it's the same thing I always end up with trying to reduce generics boilerplate, and lets us potentially do interesting things with delegate properties for props/state in the future if we want to be smarter. 

That said, I think this change is too big to land this late into the 1.0 process. Having two render methods means we can no longer ask the IDE to "implement superclass methods" and get all the methods we care about – now you have to know there's a render method, know which one to use, and then override it manually. Maybe the convenience of that isn't as important as I think, but having no type safety around which render method you implement still is a big loss.

workflow-core/src/main/java/com/squareup/workflow1/StatefulWorkflow.kt
rjrjr: I was just hacking and slashing to get something to work quickly.

rjrjr:
I was mainly trying to bang out an existence proof. I agree that losing the template abstract method is too big a price to pay. Racking my brain to come up with something unobtrusive before I completely give up on the idea of landing it.

rjrjr:
Tracking the _RenderContext as receiver_ notion separately, #213.

samples/tictactoe/common/src/main/java/com/squareup/sample/gameworkflow/RunGameWorkflow.kt
```diff
@@ -214,9 +102,13 @@ class RealRunGameWorkflow(
       alertScreen(
           base = GamePlayScreen(state.playerInfo, state.completedGame.lastTurn),
           alert = maybeQuitScreen(
-              confirmQuit = { context.actionSink.send(ConfirmQuit) },
-              continuePlaying = {
-                context.actionSink.send(ContinuePlaying(state.playerInfo, state.completedGame.lastTurn))
+              confirmQuit = context.eventHandler {
+                val oldState = this.state as MaybeQuitting
```
rjrjr, line 106: Should use `as?` here, set a consistent example.

workflow-core/src/main/java/com/squareup/workflow1/RenderContext.kt
```diff
@@ -239,6 +224,7 @@ internal fun <T, PropsT, StateT, OutputT>
  * Alternative to [RenderContext.actionSink] that allows externally defined
  * event types to be mapped to anonymous [WorkflowAction]s.
  */
+@Deprecated("Use eventHandler")
```
rjrjr, line 227:
```suggestion
@Deprecated("Use StatefulWorkflow.RenderContext.eventHandler or its counterpart in StatelessWorkflow")
```

workflow-core/src/main/java/com/squareup/workflow1/StatefulWorkflow.kt
```diff
@@ -166,6 +150,164 @@ abstract class StatefulWorkflow<
...
+    return { e1, e2, e3, e4, e5, e6, e7, e8, e9, e10 ->
+      val action = object : NamedAction<PropsT, StateT, OutputT>(name) {
+        override fun Updater.apply() = update(e1, e2, e3, e4, e5, e6, e7, e8, e9, e10)
+      }
+
+      actionSink.send(action)
+    }
+  }
+
+  private abstract class NamedAction<PropsT, StateT, Output>(
```
rjrjr, line 304: There's a dup of this in `StatelessWorkflow`. Should be able to share something.

rjrjr:
`TimeMachineWorkflowTest` is broken by the deprecation of `makeEventSink`, and I can't use `eventHandler` there b/c `Workflow.stateful` and `Workflow.stateless` are tied directly to `BaseRenderContext`. Gonna try to fix that.

rjrjr:
Yeah, this is real. The compiler can't do it.

Maybe I can make these methods on `RenderContext` itself.

rjrjr:
> Maybe I can make these methods on RenderContext itself.

That works. Does it mean that they can be methods w/default interfaces on the BaseRenderContext interface?

rjrjr:
Update makes the methods default implementations on `BaseRenderContext`. Eliminates duplication and makes them available for use by `Workflow.stateful()` and `Workflow.stateless()`. Also adds missing unit tests, and converts the last few `makeActionSink` calls.

Again, I'd much prefer them to be extensions rather than part of the interface, but the compiler refuses.

workflow-core/src/main/java/com/squareup/workflow1/StatefulWorkflow.kt
```diff
@@ -275,9 +259,9 @@ inline fun <StateT, OutputT, RenderingT> Workflow.Companion.stateful(
  */
 fun <PropsT, StateT, OutputT, RenderingT>
     StatefulWorkflow<PropsT, StateT, OutputT, RenderingT>.action(
-      name: String = "",
-      update: WorkflowAction<PropsT, StateT, OutputT>.Updater.() -> Unit
-    ) = action({ name }, update)
+  name: String = "",
```
rjrjr, line 262: sigh

workflow-core/api/workflow-core.api
```diff
@@ -1,10 +1,43 @@
 public abstract interface class com/squareup/workflow1/BaseRenderContext {
```
zach-klippenstein, line 1: Can we make this an abstract class? Would save us all the synthetic methods, and all the concrete impls in both concrete subclasses.

zach-klippenstein (commented):
LG so far. Re-request a review whenever you're ready.

rjrjr:
This is done, as far as I'm concerned. Just trying to get GitHub actions to finish.

rjrjr:
Fine with me. I'll try it out tomorrow.

rjrjr:
Update rebases to catch up with CI improvements. Time to merge this thing.

workflow-core/src/main/java/com/squareup/workflow1/RenderContext.kt
```diff
@@ -115,9 +100,159 @@ interface BaseRenderContext<out PropsT, StateT, in OutputT> {
...
+        override fun Updater.apply() = update(e1, e2, e3, e4, e5, e6, e7, e8, e9, e10)
+      }
+
+      actionSink.send(action)
+    }
+  }
 }
 
-@Deprecated("Use RenderContext.actionSink.")
+@Deprecated("Use RenderContext.eventHandler.")
```
rjrjr, line 255: Should say `BaseRenderContext`

rjrjr: But if that's the only problem, gonna merge this.

workflow-core/api/workflow-core.api
rjrjr: Tried, failed. The version with no event type confuses the compiler if they're extensions. Same in Kotlin 1.4. We suspect a compiler bug, will file.

workflow-core/src/main/java/com/squareup/workflow1/RenderContext.kt
```diff
@@ -115,9 +100,159 @@ interface BaseRenderContext<out PropsT, StateT, in OutputT> {
...
+   * @param name A string describing the update, included in the action's [toString]
+   * as a debugging aid
+   * @param update Function that defines the workflow update.
+   */
+  fun eventHandler(
+    name: String = "eventHandler",
+    update: WorkflowAction<PropsT, StateT, OutputT>.Updater.() -> Unit
+  ): () -> Unit {
+    return {
+      actionSink.send(action(name, update))
```
zach-klippenstein, line 118: Why is `action` good enough here, but we need `NamedAction` for the others? Why do we need `NamedAction` at all, since `action{}` takes a name?

zach-klippenstein (approved):

workflow-core/src/main/java/com/squareup/workflow1/RenderContext.kt
rjrjr: I…don't remember. I thought I had a reason that I couldn't use action, but I'm not sure I actually believe that.

rjrjr:
Update gets rid of `NamedAction`, changes the `name` params to `() -> String`, and restores accidentally clobbered test for `makeEventSink`.

workflow-core/api/workflow-core.api
rjrjr: https://youtrack.jetbrains.com/issue/KT-42741

rjrjr:
Forgot to update the `apiDump`.

PR #217, opened 2020-10-16
Consolidates CI check shards.

rjrjr:
They're all pretty fast, and we do see shards get queued sometimes. Fewer
shards, fewer lines to wait in. Relies on gradle's `--continue` flag to ensure
we see as many errors at once as possible.


zach-klippenstein (approved):
Nice!

PR #220, opened 2020-10-20
Fixes memory leak in ModalContainer

### LINKED ISSUE #206: Stop leaking dialog windows
Every time we close a dialog before the activity is torn down, we leak a reference to it via its lifecycle hook. Here's the leaky bit, courtesy @pyricau : 

```kotlin
          dialog.lifecycleOrNull()
              ?.addObserver(OnDestroy { dialog.dismiss() })
        }
```

rjrjr:
The code that prevents a nagging logcat warning about leaking dialog windows
on config change was in turn leaking references to them on the `Lifecycle`.

closes #206


elizrus (approved):
Makes sense and I also verified that this fixes the leak I was seeing- thanks!

zach-klippenstein (approved):
So if I'm reading this right, it basically scopes the lifecycle observation to only being active while the dialog's view is attached to the window, right?

rjrjr:
> So if I'm reading this right, it basically scopes the lifecycle observation to only being active while the dialog's view is attached to the window, right?

Yes.

Issue #187, opened 2020-09-11
Allow lazy construction of Workers


bnvinay92:
`render()` passes construct `Worker` instances every time even if they won't be run. Allow users to pass in a factory method that is invoked only when the runtime determines the need to enqueue some work.

PR #221, opened 2020-10-24
Introduce LazyWorker

bnvinay92:
A wrapper worker that delays construction of the underlying worker.

Addresses #187.

### DIFF
--- a/workflow-core/src/main/java/com/squareup/workflow1/BaseRenderContext.kt
+++ b/workflow-core/src/main/java/com/squareup/workflow1/BaseRenderContext.kt
@@ -311,6 +311,37 @@ inline fun <T, reified W : Worker<T>, PropsT, StateT, OutputT>
 }
 
 /**
+ * Ensures the worker produce by [workerProvider] is running. When the [Worker] emits an output,
+ * [handler] is called to determine the [WorkflowAction] to take. When the worker finishes, nothing
+ * happens (although another render pass may be triggered).
+ *
+ * Unlike the other [runningWorker] methods, this one requires a mandatory [key] which is used to
+ * determine if workers are to be started, kept alive, or destroyed across multiple render passes.
+ * This allows the runtime to delay construction of workers until necessary and avoid reflection.
+ *
+ * @param key A string key that is used to distinguish [Worker]s emitted by [workerProvider].
+ */
+/* ktlint-disable parameter-list-wrapping */
+inline fun <T, reified W : Worker<T>, PropsT, StateT, OutputT>
+    BaseRenderContext<PropsT, StateT, OutputT>.runningWorker(
+  noinline workerProvider: () -> W,
+  key: String,
+  noinline handler: (T) -> WorkflowAction<PropsT, StateT, OutputT>
+) {
+/* ktlint-enable parameter-list-wrapping */
+  runningWorker(LazyWorker(key, workerProvider), LazyWorkerType, key, handler)
+}
+
+/**
+ * Worker equality is first determined by their [KType], and if those match then by
+ * by their implementation of [Worker.doesSameWorkAs]. [LazyWorker]s do not construct
+ * their underlying [Worker]s until they are about to be run and therefore use this
+ * default [KType], relying entirely on their keys to be differentiated.
+ */
+@OptIn(ExperimentalStdlibApi::class)
+val LazyWorkerType: KType = typeOf<LazyWorker<*>>()
+
+/**
  * Ensures [worker] is running. When the [Worker] emits an output, [handler] is called
  * to determine the [WorkflowAction] to take. When the worker finishes, nothing happens (although
  * another render pass may be triggered).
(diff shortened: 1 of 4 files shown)


zach-klippenstein:
There are a couple issues with this PR, but I don't understand what problem this addresses. Aside from the changes to the API wrt keys, it just seems to be replacing a worker allocation with a lambda allocation. 

Like @rjrjr pointed out in Slack, if you have a worker that is expensive to construct it's better to just keep it in a val in the workflow class if possible.

bnvinay92:
So one issue is that workers take in arguments from props/state eliminating the worker caching strategy most of the time.
The other issue is that most of the time that goes into worker construction/running is due to two reflection invocations.

One is when there's an invocation of runningWorker. https://github.com/square/workflow-kotlin/blob/421d5114aca73e475becd8344d9758a970ae2414/workflow-core/src/main/java/com/squareup/workflow1/BaseRenderContext.kt#L310

The other one, which is more avoidable but specific to our migration, is when RxWorkers converts Rx primitives to a Flow.

AFAICT, workflow first compares types before using keys to differentiate work. If keys are sufficiently unique though, then it would be nice to ignore the types and avoid the associated reflection. 

Given the profiler bloats everything up, I guess the absolute value is not as meaningful but it's worth nothing that `typeOf` invocations form a big % chunk of the main thread:
Screen Shot 2020-10-27 at 7 00 01 AM

zach-klippenstein:
> AFAICT, workflow first compares types before using keys to differentiate work. If keys are sufficiently unique though, then it would be nice to ignore the types and avoid the associated reflection.

Skipping the reflection check if keys are distinct is an interesting optimization, but unfortunately we can't know if keys will be distinct until after the first worker is created, so we can't skip the reflection on the first call. And the implementation of RenderContext can't run at all to do this sort of logic until the type has been captured anyway (we probably wouldn't want to inline that much logic even if we could, e.g. by changing `RenderContext` to an abstract class).

One optimization you could do is just stop using workers and use the side effect API directly (`runningSideEffect`) - side effects have no notion of type, only use (and require) string keys, and don't use any reflection. If you need to reuse this code, you can wrap your implementation in a sharable workflow to make the public API nicer.

workflow-core/src/main/java/com/squareup/workflow1/BaseRenderContext.kt
```diff
@@ -310,6 +310,37 @@ inline fun <T, reified W : Worker<T>, PropsT, StateT, OutputT>
...
+ */
+/* ktlint-disable parameter-list-wrapping */
+inline fun <T, reified W : Worker<T>, PropsT, StateT, OutputT>
+    BaseRenderContext<PropsT, StateT, OutputT>.runningWorker(
+  noinline workerProvider: () -> W,
+  key: String,
+  noinline handler: (T) -> WorkflowAction<PropsT, StateT, OutputT>
+) {
+/* ktlint-enable parameter-list-wrapping */
+  runningWorker(LazyWorker(key, workerProvider), LazyWorkerType, key, handler)
```
bnvinay92, line 332: @zach-klippenstein the `typeOf` has been replaced with a singleton `LazyWorkerType` used across all invocations that pass in a key, assuming these keys will be unique.

Referring to:

>And the implementation of RenderContext can't run at all to do this sort of logic until the type has been captured anyway

Would `LazyWorkerType` break RenderContext implementations in some insidious way? If so, I'm guessing I would have to implement a `String` key-based `WorkflowIdentifier` for my usecase, right?

zach-klippenstein: It should work, and if you have a use case for it, this technique should be fine. 

However changing the use of type information in the standard worker builder functions at this point could break code that is written assuming they take both type and key into account.

It would be safer to create separate builder functions that are explicit about ignoring type (eg `asUntypedWorker()`). And for that matter, I think `UntypedWorker` would be a better name for this construct, since the most significant thing about it is its lack of type information, not its laziness (all workers are "lazy" in a sense).

I think I would be ok with landing this stuff if it didn't change any existing behavior. @rjrjr WDYT?

rjrjr: My kneejerk is always going to be to push back against additions to the library that users can implement themselves. Do you both think its of general interest?

zach-klippenstein: I agree, although technically Worker itself falls under this category now I believe.

I'd rather not merge this in the core library for now, keep the API surface smaller.

bnvinay92: Makes sense, I can add the builders to our project instead.

zach-klippenstein:
Sounds like we've reached a resolution for now, so I'm going to close this PR. Please feel free to continue to give us feedback about how this pattern works for you!

Issue #224, opened 2020-10-27
Exceptions thrown by RenderingAndSnapshot.equals are swallowed.


rjrjr:
The `fail()` call in the test below is swallowed.

To see this exact failure, you'll have to revert the fix for #223 if that has landed. Or else replace the String rendering type with something that will throw from `equals`.

```kotlin
    lateinit var sink: Sink<String>

    val workflow = Workflow.stateful<Unit, String, Nothing, String>(
        initialState = { _, _ -> "unchanging state" },
        snapshot = { Snapshot.of {
          fail("Snapshot must not be resolved.")
        } },
        render = { _, state ->
          sink = actionSink.contraMap { action { this.state = it } }
          state
        }
    )
    val props = MutableStateFlow(Unit)
    val renderings = renderWorkflowIn(workflow, scope, props) {}

    renderings.value.rendering.let { rendered ->
      assertEquals("unchanging state", rendered)
      sink.send("unchanging state")
    }
```

zach-klippenstein:
Yikes. Surprised we don't have tests around this either.

zach-klippenstein:
@rjrjr If I understand correctly, this is fixed by #225 because renderings are now never compared by equals? I'll still add unit tests around rendering and snapshot equals to make sure nothing else is eating them up.

PR #225, opened 2020-10-27
Makes snapshots lazy again.

### LINKED ISSUE #223: Snapshot laziness is broken by RenderingSnapshot.equals
I suspect that this call in `renderWorkflowIn` has broken it:

```kotlin
      renderingsAndSnapshots.value = runner.nextRendering()
```
b/c of a call to `equals` in `StateFlowImpl`

Screen Shot 2020-10-27 at 10 23 56 AM

Should be a simple fix, make `RenderingAndSnapshot.equals` and `hashCode` ignore the `Snapshot`.

https://workflow-community.slack.com/archives/CHTFPR277/p1603819369039000

rjrjr:
`renderWorkflowIn` uses a `MutableStateFlow` as the source of its
`RenderingAndSnapshot` stream, and that class does an equality check as an
optimization.

Fix is to change `RenderingAndSnapshot` to ignore its `Snapshot` in `equal`,
`hashCode`, and while we're at it `toString`. Danger here is that we'll drop
updates that change only the `Snapshot`, but that seems like a pretty weird
case anyway.

Alternative would be to make `RenderingAndSnapshot` always fail equality, but
that seems like it'd be an even bigger landmine.

closes #223

### DIFF
--- a/workflow-runtime/src/main/java/com/squareup/workflow1/RenderingAndSnapshot.kt
+++ b/workflow-runtime/src/main/java/com/squareup/workflow1/RenderingAndSnapshot.kt
@@ -1,24 +1,16 @@
-/*
- * Copyright 2019 Square Inc.
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *     http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
 package com.squareup.workflow1
 
 /**
  * Tuple of rendering and snapshot used by [renderWorkflowIn].
+ *
+ * Note that this class keeps the default identity equality
+ * implementation it inherits from `Any`, rather than comparing
+ * its [rendering] or [snapshot].
  */
-data class RenderingAndSnapshot<out RenderingT>(
+class RenderingAndSnapshot<out RenderingT>(
   val rendering: RenderingT,
   val snapshot: TreeSnapshot
-)
+) {
+  operator fun component1() = rendering
+  operator fun component2() = snapshot
+}
--- a/workflow-runtime/src/test/java/com/squareup/workflow1/RenderWorkflowInTest.kt
+++ b/workflow-runtime/src/test/java/com/squareup/workflow1/RenderWorkflowInTest.kt
@@ -32,11 +32,13 @@ import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.suspendCancellableCoroutine
 import kotlinx.coroutines.test.TestCoroutineScope
+import okio.ByteString
 import org.junit.Test
 import kotlin.test.assertEquals
 import kotlin.test.assertFailsWith
 import kotlin.test.assertFalse
 import kotlin.test.assertNotNull
+import kotlin.test.assertNotSame
 import kotlin.test.assertNull
 import kotlin.test.assertTrue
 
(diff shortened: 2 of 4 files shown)


zach-klippenstein:
> Danger here is that we'll drop updates that change only the Snapshot

Snapshots should be a pure function of state, and renderings are always re-created at the same time as snapshots, whenever state changes. So I think this is impossible for well-behaved workflows, and definitely something we should explicitly not support if it ever actually comes up (which i've never seen it do).

zach-klippenstein:
in general, with these sorts of changes, I'm worried about the implications of violating the assumption that data classes are equatable. However, this particular class is only used for multiple return values, and I'm pretty sure it's nearly always immediately split into components, so I don't expect it will actually throw anything off.

steve-the-edwards (approved):

workflow-runtime/src/main/java/com/squareup/workflow1/RenderingAndSnapshot.kt
```diff
@@ -21,4 +21,21 @@ package com.squareup.workflow1
 data class RenderingAndSnapshot<out RenderingT>(
   val rendering: RenderingT,
   val snapshot: TreeSnapshot
-)
+) {
+
+  override fun equals(other: Any?): Boolean {
+    return when {
+      this === other -> true
+      javaClass != other?.javaClass -> false
```
zach-klippenstein, line 29: `other !is RenderingAndSnapshot ->` reads better and is more idiomatic, imo.

workflow-runtime/src/main/java/com/squareup/workflow1/RenderingAndSnapshot.kt
```diff
@@ -21,4 +21,21 @@ package com.squareup.workflow1
...
+      else -> rendering == (other as RenderingAndSnapshot<*>).rendering
```
zach-klippenstein, line 30: We could still do an identity comparison for `snapshot`, but then we'd need to use `System.identityHashCode` for the hashCode implementation and I don't believe there's any multiplatform version of that function (although there is a [native one](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.native/identity-hash-code.html), apparently). Anyway, probably more work than it's worth.

Although I'm kind of thinking it would be better to just make this a simple identity comparison and not include even the rendering, since it seems overly complicated, and not useful, to have only partial equality.

workflow-runtime/src/test/java/com/squareup/workflow1/RenderWorkflowInTest.kt
```diff
@@ -151,6 +153,35 @@ class RenderWorkflowInTest {
...
+  // https://github.com/square/workflow-kotlin/issues/223
+  @Test fun `snapshots are lazy`() {
+    lateinit var sink: Sink<String>
+    var snapped = false
+
+    val workflow = Workflow.stateful<Unit, String, Nothing, String>(
+        initialState = { _, _ -> "unchanging state" },
+        snapshot = { Snapshot.of {
+          // We need to use snapped b/c the fail call is swallowed.
+          // https://github.com/square/workflow-kotlin/issues/223
```
zach-klippenstein, line 165: Nit: I think you linked the wrong issue here, did you mean #224?

rjrjr:
> in general, with these sorts of changes, I'm worried about the implications of violating the assumption that data classes are equatable. However, this particular class is only used for multiple return values, and I'm pretty sure it's nearly always immediately split into components, so I don't expect it will actually throw anything off.

At the very least I should punch add kdoc making this behavior explicit.

workflow-runtime/src/test/java/com/squareup/workflow1/RenderWorkflowInTest.kt
```diff
@@ -151,6 +153,35 @@ class RenderWorkflowInTest {
...
+    )
+    val props = MutableStateFlow(Unit)
+    val renderings = renderWorkflowIn(workflow, scope, props) {}
+
+    renderings.value.rendering.let { rendered ->
+      assertEquals("unchanging state", rendered)
+      sink.send("unchanging state")
+    }
+
+    assertFalse(snapped)
```
zach-klippenstein, line 182: I'd probably add one more check here to make sure the next snapshot is not equal, and that the first snapshot is still not snapped.

workflow-runtime/src/main/java/com/squareup/workflow1/RenderingAndSnapshot.kt
```diff
@@ -21,4 +21,21 @@ package com.squareup.workflow1
 data class RenderingAndSnapshot<out RenderingT>(
   val rendering: RenderingT,
   val snapshot: TreeSnapshot
-)
+) {
+
+  override fun equals(other: Any?): Boolean {
```
zach-klippenstein, line 26: We should write a small test for this equals/hashcode too, now that it's explicit.

zach-klippenstein (changes requested):

workflow-runtime/src/test/java/com/squareup/workflow1/RenderWorkflowInTest.kt
rjrjr: Yup, thanks.

workflow-runtime/src/main/java/com/squareup/workflow1/RenderingAndSnapshot.kt
rjrjr: > Although I'm kind of thinking it would be better to just make this a simple identity comparison and not include even the rendering, since it seems overly complicated, and not useful, to have only partial equality.

Yeah, I like that better, seems less magical. And that could be as simple as dropping `data` from the class declaration.

zach-klippenstein (approved):

PR #226, opened 2020-10-27
Add tests for exceptions thrown from equals/hashcode methods in renderings.

### LINKED ISSUE #224: Exceptions thrown by RenderingAndSnapshot.equals are swallowed.

zach-klippenstein:
These tests were written to catch situations like #224, which was actually fixed
in #225 by a PR to another issue. Nonetheless, these tests should help prevent
regressions.

Also makes other tests more strict by correctly calling `cleanupTestCoroutines()` on the default scope after the tests.

Closes #224 (doesn't really, it's obsolete, but close enough).

### DIFF
--- a/workflow-runtime/src/test/java/com/squareup/workflow1/RenderWorkflowInTest.kt
+++ b/workflow-runtime/src/test/java/com/squareup/workflow1/RenderWorkflowInTest.kt
@@ -20,6 +20,7 @@ import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers.Unconfined
 import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.FlowPreview
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.channels.Channel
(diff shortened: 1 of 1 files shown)


zach-klippenstein:
@rjrjr Feel free to merge on approval.

workflow-runtime/src/test/java/com/squareup/workflow1/RenderWorkflowInTest.kt
```diff
@@ -42,28 +46,43 @@ import kotlin.test.assertNotSame
...
+@OptIn(ExperimentalCoroutinesApi::class, FlowPreview::class)
 class RenderWorkflowInTest {
 
   // TestCoroutineScope doesn't actually create a Job, so isActive will always return true unless
   // explicitly give it a job.
-  private val scope = TestCoroutineScope(Job())
+
+  /**
+   * A [CoroutineScope] that will fail the test if it has uncaught exceptions after the test
+   * completes. Use this scope to test success cases.
```
rjrjr, line 57: Nice

rjrjr (approved):

workflow-runtime/src/test/java/com/squareup/workflow1/RenderWorkflowInTest.kt
```diff
@@ -485,5 +507,85 @@ class RenderWorkflowInTest {
     )
   }
 
+  // https://github.com/square/workflow-kotlin/issues/224
+  @Test fun `exceptions from Snapshots don't fail runtime`() {
```
steve-the-edwards, line 511: Can you add to these test names 'don't fail runtime' shadows the fact that you are also testing that they *do fail* in Snapshot builder or equals when they should (i.e. don't get swallowed).

zach-klippenstein: Fair enough, although I think the better fix here is to not test the thing this test does not actually care about (stop testing that `throw` works).

PR #232, opened 2020-11-16
A couple of Worker<Nothing> fixes.

rjrjr:
* `LifecycleWorker.doesSameWorkAs` now actually does compare concrete types,
   like its kdoc claims.
 * Fixes a problem where `RenderTester` thought `Worker<Nothing>` matched
   all types registered with `expectWorker`

### DIFF
--- a/workflow-core/src/main/java/com/squareup/workflow1/LifecycleWorker.kt
+++ b/workflow-core/src/main/java/com/squareup/workflow1/LifecycleWorker.kt
@@ -73,5 +73,7 @@ abstract class LifecycleWorker : Worker<Nothing> {
   /**
    * Equates [LifecycleWorker]s that have the same concrete class.
    */
-  override fun doesSameWorkAs(otherWorker: Worker<*>): Boolean = true
+  override fun doesSameWorkAs(otherWorker: Worker<*>): Boolean {
+    return otherWorker::class == this::class
+  }
 }
--- a/workflow-testing/src/main/java/com/squareup/workflow1/testing/RenderTesterWorkers.kt
+++ b/workflow-testing/src/main/java/com/squareup/workflow1/testing/RenderTesterWorkers.kt
@@ -159,7 +159,7 @@ inline fun <PropsT, StateT, OutputT, RenderingT, WorkerOutputT, WorkerT : Worker
  * @param description Optional string that will be used to describe this expectation in error
  * messages.
  */
-@OptIn(ExperimentalWorkflowApi::class)
+@OptIn(ExperimentalWorkflowApi::class, ExperimentalStdlibApi::class)
 /* ktlint-disable parameter-list-wrapping */
 fun <PropsT, StateT, OutputT, RenderingT>
     RenderTester<PropsT, StateT, OutputT, RenderingT>.expectWorker(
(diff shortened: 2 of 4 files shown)


workflow-testing/src/test/java/com/squareup/workflow1/testing/RealRenderTesterTest.kt
```diff
@@ -668,6 +670,38 @@ class RealRenderTesterTest {
     )
   }
 
+  @Test fun `typeOf LifecycleWorker does not match typeOf Flow asWorker`() {
```
rjrjr, line 673: Should prefix both of these method names with `runningWorker`

workflow-testing/src/test/java/com/squareup/workflow1/LifecycleWorkerTest.kt
```diff
@@ -62,4 +62,10 @@ class LifecycleWorkerTest {
       assertTrue(onCancelledCalled)
     }
   }
+
+  @Test fun `doesSameWorkAs compares concrete types`() {
```
rjrjr, line 66: Inadequate

workflow-testing/src/main/java/com/squareup/workflow1/testing/RenderTesterWorkers.kt
```diff
@@ -175,7 +175,16 @@ fun <PropsT, StateT, OutputT, RenderingT>
...
+    val ruleExpectsNothing = typeOf<Worker<Nothing>>().isSupertypeOf(workerType)
+    val actualExpectsNothing = typeOf<Worker<Nothing>>().isSupertypeOf(actualWorkerType)
+
+    // We have to take some care when the actual worker is `Worker<Nothing>`, b/c
+    // Worker<Something>.isSupertypeOf(Worker<Nothing>) is always true -- Nothing is
+    // the bottom type. So, we only make that check if the rule is `Worker<Nothing>`,
+    // or both the rule and the actual are `Worker<Something>`.
+
+    (key == actualKey &&
+        (ruleExpectsNothing || !actualExpectsNothing) && workerType.isSupertypeOf(actualWorkerType))
```
zach-klippenstein, line 187: I am glad you added that comment and tests for this, i am having trouble figuring out what this actually means just from the code.

workflow-core/src/main/java/com/squareup/workflow1/LifecycleWorker.kt
```diff
@@ -73,5 +73,7 @@ abstract class LifecycleWorker : Worker<Nothing> {
   /**
    * Equates [LifecycleWorker]s that have the same concrete class.
    */
-  override fun doesSameWorkAs(otherWorker: Worker<*>): Boolean = true
+  override fun doesSameWorkAs(otherWorker: Worker<*>): Boolean {
+    return otherWorker::class == this::class
```
zach-klippenstein, line 77: My only concern with this is that it's a different default than other workers get. But I don't know if that's actually a valid concern. The usual "when in doubt, use keys" rule still applies.

zach-klippenstein (approved):

workflow-core/src/main/java/com/squareup/workflow1/LifecycleWorker.kt
rjrjr: I think the difference makes sense here. I can't think of any other workers that we customize via subclassing this way.

PR #234, opened 2020-11-17
Add the second tutorial page.

zach-klippenstein:
This is based on #196.

PR #235, opened 2020-11-18
Add the third tutorial section.

zach-klippenstein:
This is based on #234.


rjrjr:
I think your rebasing must be off, this includes redundant commits despite being targeted at `zachklipp/tutorial2`. Or maybe that one is hosed?

samples/tutorial/Tutorial3.md
```diff
@@ -0,0 +1,415 @@
...
+
+Start from the implementation of `tutorial-2-complete` if you're skipping ahead.
+
+## Editing todo items
+
+Now that a user can "log in" to their todo list, we want to add the ability to edit the todo items listed.
+
+### State ownership
+
+In the workflow framework, data flows _down_ the tree as properties to child workflows, and actions come _up_ as output events (as in the traditional computer science sense that trees that grow downward).
```
rjrjr, line 17: "actions come _up_" seems a little confusing, since `WorkflowAction` and `action {}` actually mean something. 

Weird, it won't let me suggest changes. Perhaps:

In the workflow framework, data flows _down_ the tree as properties set by parents on their child workflows, and comes back _up_  from children as output events to their parents (as in the traditional computer science sense that trees that grow downward).

samples/tutorial/Tutorial3.md
```diff
@@ -0,0 +1,415 @@
...
+
+What this means is that state should be created as far down the tree as possible, to limit the scope of state to be as small as possible. Additionally, there should be only one "owner" of the state in the tree - if it's passed farther down the tree, it should be a copy or read-only version of it - so there is no shared mutable state in multiple workflows.
+
+When a child workflow has a copy of the state from its parent, it should change it by emitting an _output_ back to the parent, requesting that it be changed. The child will then receive an updated snapshot of the data from the parent - keeping ownership at a single level of the tree.
```
rjrjr, line 21: Avoid overloading "snapshot" here. Perhaps "receive an updated copy of the data"

samples/tutorial/Tutorial3.md
```diff
@@ -0,0 +1,415 @@
...
+```kotlin
+class TodoEditLayoutRunner(
+  private val binding: TodoEditViewBinding
+) : LayoutRunner<TodoEditScreen> {
+
+  override fun showRendering(
+    rendering: TodoEditScreen,
+    viewEnvironment: ViewEnvironment
+  ) {
+    binding.root.backPressedHandler = rendering.discardChanges
```
rjrjr, line 87: :D

samples/tutorial/Tutorial3.md
```diff
@@ -0,0 +1,415 @@
...
+  override fun onPropsChanged(
+    old: EditProps,
+    new: EditProps,
+    state: State
+  ): State {
+    // The `Todo` from our parent changed. Update our internal copy so we are starting from the same
+    // item. The "correct" behavior depends on the business logic - would we only want to update if
+    // the users hasn't changed the todo from the initial one? Or is it ok to delete whatever edits
+    // were in progress if the state from the parent changes?
+    if (old.initialTodo != new.initialTodo) {
```
rjrjr, line 137: Isn't this check redundant these days?

samples/tutorial/Tutorial3.md
```diff
@@ -0,0 +1,415 @@
...
+}
+```
+
+## Todo Editing in the full flow
+
+### Updating the current workflows to prepare to add the edit workflow
+
+We want the todo edit screen to be shown when a user taps on an item on the todo list screen. To do this, we will modify the todo list workflow to show the edit screen when we are editing.
+
+Because the `TodoEditWorkflow` returns a different screen type, we will first need to modify the todo list workflow to return a list of `Any`s as the rendering.
```
rjrjr, line 224: It'd be more idiomatic for it to emit `BackStackScreen<Any>`. They have appropriate `plus` operators. Maybe you make that change later?

The sentence is also a little strange -- how does returning a different screen type make us decide to render a `List<Any>` rather than just `Any`?

rjrjr (approved):

samples/tutorial/Tutorial3.md
zach-klippenstein: Good point, this is a bit odd. This is directly copied from the swift tutorial, so might want to fix it there too (cc @AquaGeek )

samples/tutorial/Tutorial3.md
zach-klippenstein: Yes, although this doesn't hurt since if other properties were added to the props, we'd probably have to add this back in.

samples/tutorial/Tutorial3.md
zach-klippenstein: I actually changed it back intentionally. Seems better to me to use more simple, more general types for these children workflows that don't actually need to use the special view binding.

zach-klippenstein: > The sentence is also a little strange -- how does returning a different screen type make us decide to render a List<Any> rather than just Any?

That's a good point though. How about: "Because the parent workflow needs to know about the fact that this workflow renders a list, so it can append its screens to the backstack, we will first need to modify the todo list workflow to return a _list_ of `Any`s as the rendering."?

samples/tutorial/Tutorial3.md
zach-klippenstein: Done.

samples/tutorial/Tutorial3.md
rjrjr: Def. better. But still leaves implicit _why_ we're returning a list, I think. Did you say anything like _we're going to return one or two screens that the parent can use to build a back stack_.

zach-klippenstein: > The `TodoListWorkflow` will now occasionally need to render two screens instead of just the one. Its parent workflow will add the one or two screens to the backstack it constructs. Since `TodoListWorkflow`'s rendering type is `Any`, it can just return a list of screens. However, the parent workflow needs to know about the list, so it can pull the screens out and add them to its backstack. We'll change the rendering type from `Any` to `List<Any>`:

How's this?

samples/tutorial/Tutorial3.md
```diff
@@ -0,0 +1,415 @@
...
+```kotlin
+object TodoListWorkflow : StatefulWorkflow<ListProps, State, Back, List<Any>>() {
+
+  data class ListProps(val name: String)
+
+  data class State(
+    val todos: List<TodoModel>,
+    val step: Step
+  ) {
+    sealed class Step {
```
steve-the-edwards, line 312: Any good resources we can point to for forming the shape of State for common use cases? (I mean for even React or any uni-directional data flow state machine). It seems like this is one of the 'art' areas of using Workflow where extra resources will help.

PR #236, opened 2020-11-18
Add the fourth tutorial section.

zach-klippenstein:
This is based on #235.


samples/tutorial/Tutorial4.md
```diff
@@ -0,0 +1,343 @@
...
+  // …
+
+}
+```
+
+#### Moving Edit Output handling to the TodoWorkflow
+
+The `TodoWorkflow` now can handle the outputs from the `TodoListWorkflow`. Next, let's add handling for the `TodoEditWorkflow` output events.
+
+Since the types of output and actions are pretty different from their origin, make a *second* action type on the `TodoWorkflow`:
```
rjrjr, line 259: We don't acually have action types in this example, do we?

samples/tutorial/Tutorial4.md
```diff
@@ -0,0 +1,343 @@
...
+        return listOf(todoListScreen, todoEditScreen)
+      }
+    }
+  }
+
+  // …
+}
+```
+
+That's it! There is now a workflow for both of our current steps of the Todo flow. We also used the ability to define multiple actions for a single workflow to keep the logic contained to the expected state we would receive the actions from.
```
rjrjr, line 335: > We also used the ability to define multiple actions for a single workflow to keep the logic contained to the expected state we would receive the actions from.

Did we though?

samples/tutorial/Tutorial4.md
```diff
@@ -0,0 +1,343 @@
...
+```
+
+#### Moving logic from the TodoList to the TodoWorkflow
+
+Move the `ListState` state, input, and outputs from the `TodoListWorkflow` up to the `TodoWorkflow`. It will be owner the list of todo items, and the `TodoListWorkflow` will simply show whatever is passed into its input:
+
+```kotlin
+object TodoWorkflow : StatefulWorkflow<TodoProps, State, Back, List<Any>>() {
+
+  data class TodoProps(val name: String)
```
rjrjr, line 40: Cross cutting nit: thinking that `userName` would be clearer for this and the like state fields.

rjrjr (approved):

samples/tutorial/Tutorial4.md
```diff
@@ -0,0 +1,343 @@
...
+  override fun initialState(
+    props: ListProps,
+    snapshot: Snapshot?
+  ) = Unit
+
+  // …
+}
+```
+
+TODO add section about converting to StatelessWorkflow.
```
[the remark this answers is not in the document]
zach-klippenstein: I'm gonna do this in a follow up, since i need to make this change in the fifth section as well.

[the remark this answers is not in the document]
zach-klippenstein: #245

samples/tutorial/Tutorial4.md
zach-klippenstein: Nope. Changing this to "*second* set of actions"

samples/tutorial/Tutorial4.md
zach-klippenstein: We did not. Removed this sentence.

samples/tutorial/Tutorial4.md
zach-klippenstein: #246, gonna do in a follow up so i can change everywhere at once.

samples/tutorial/Tutorial4.md
```diff
@@ -0,0 +1,343 @@
...
+    // When a todo item is selected, edit it.
+    state = state.copy(step = Step.Edit(index))
+  }
+
+  private fun newTodo() = action {
+    // Append a new todo model to the end of the list.
+    state = state.copy(
+        todos = state.todos + TodoModel(
+            title = "New Todo",
+            note = ""
```
steve-the-edwards, line 199: Nit: May want to add a note to say that we have no separate screen for adding a todo, we just fill out a default one and drop it into the list (i.e. I was expecting us to change to Step.Edit) here, but we didn't.

PR #238, opened 2020-11-18
Add note about updating tutorial workflow version to the release docs.

zach-klippenstein:
Depends on #196.


rjrjr:
Doesn't seem like you really want this to target `main`.

rjrjr (approved):
Second SHA LGTM

zach-klippenstein:
Fixed the second commit.

PR #247, opened 2020-11-19
Upgrade Kotlin to 1.4.10, coroutines 1.4.1.

zach-klippenstein:


versions.properties
```diff
@@ -32,16 +32,16 @@ version.io.mockk..mockk=1.10.0
...
+version.kotlinx.coroutines=1.4.1
+version.kotlinx.serialization=1.0.0
 version.me.champeau.gradle..jmh-gradle-plugin=0.5.1
 version.moshi=1.9.3
 version.okio=2.7.0
 version.org.hamcrest..hamcrest-core=2.2
 version.org.jetbrains..annotations=19.0.0
 version.org.jetbrains.dokka..dokka-gradle-plugin=0.10.1
-version.org.jetbrains.kotlin..kotlin-reflect=1.3.72
+version.org.jetbrains.kotlin..kotlin-reflect=1.4.0
```
0legg, line 44: Should it be 1.4.10 as well?

zach-klippenstein: Yep, fixed.

versions.properties
```diff
@@ -32,16 +32,16 @@ version.io.mockk..mockk=1.10.0
 version.io.reactivex.rxjava2..rxandroid=2.1.1
 version.io.reactivex.rxjava2..rxjava=2.2.19
 version.junit.junit=4.13
-version.kotlin=1.3.72
-version.kotlinx.coroutines=1.3.7
-version.kotlinx.serialization=0.20.0
+version.kotlin=1.4.10
+version.kotlinx.coroutines=1.4.1
+version.kotlinx.serialization=1.0.0
```
0legg, line 37: 1.0.1 was released 3 weeks ago with few patches

workflow-testing/src/main/java/com/squareup/workflow1/testing/WorkflowTestRuntime.kt
```diff
@@ -435,9 +435,9 @@ private fun <T> unwrapCancellationCause(block: () -> T): T {
   } catch (e: CancellationException) {
     throw generateSequence(e as Throwable) { e.cause }
         // Stop the sequence if an exception's cause is itself.
-        .scanReduce { error, cause ->
+        .runningReduce { error, cause ->
```
0legg, line 438: Nit: for the same behavior you can utilize `windowed` function, and it'll be more straightforward for a reader. I won't  guarantee it'll fit well with what you're trying to achieve in the next call, but anyway this part could benefit from some rethink of conditions and tinkering with stdlib functions.

At the same time, no objections in merging it as is.

0legg (approved):

workflow-testing/src/main/java/com/squareup/workflow1/testing/WorkflowTestRuntime.kt
zach-klippenstein: I don't see `windowed` in the docs – what am I missing?

zach-klippenstein: Yea, we'd need another operator to reduce the emitted lists I believe.

versions.properties
zach-klippenstein: This is only used for samples, but better to be up-to-date.

workflow-testing/src/main/java/com/squareup/workflow1/testing/WorkflowTestRuntime.kt
0legg: Either https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/zip-with-next.html or https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/windowed.html could be a good candidate, but I don't know how will it help with the methods down the chain.

zach-klippenstein: Yea, both those would require another operator to flatten the lists they return. `runningReduce` does it in one. I'm not sure which is more readable.

rjrjr (approved):

PR #248, opened 2020-11-19
Introduces AndroidViewRendering, de-emphasizes ViewRegistry

rjrjr:
`AndroidViewRendering` is implemented by renderings that can provide their own
`ViewFactory`. `ViewRegistry.buildView` looks to see if a given rendering
implements that method if it fails to find an entry for it -- note that we
look for a registration first, so `ViewRegistry` is able override the default.

Given this, it's no longer crucial for developers to define and provide a
`ViewRegistry`, so we get rid of all the nannying that forced them to provide
one.

Updates the samples to use the new interface and drop their `ViewRegistry`
instances where possible.


rjrjr:
This is still a draft, but I want to get a sanity check before I do the rest of the work to make it shippable.

rjrjr:
Not sure that `ViewRendering` is actually accomplishing anything.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewRegistry.kt
```diff
@@ -105,7 +103,7 @@ fun ViewRegistry(): ViewRegistry = TypedViewRegistry()
  * It is usually more convenient to use [WorkflowViewStub] than to call this method directly.
  *
  * Creates a [View] to display [initialRendering], which can be updated via calls
- * to [View.showRendering].
+ * to [View.showRendering]. Uses
```
rjrjr, line 106: Uses…

rjrjr:
For that matter, `AndroidContainerView` might be a wart. Original thought was to let something like `HelloWorkflow.Rendering` implement `ViewFactory` itself, but doing that w/o boilerplate was tricky. Maybe I should try harder.

Need to take care not to paint ourselves into any corners that will make it harder to get to the simpler modals sketched in #202.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/AndroidViewContainer.kt
```diff
@@ -0,0 +1,29 @@
+package com.squareup.workflow1.ui
+
+import android.content.Context
+import android.view.View
+import android.view.ViewGroup
+
+@OptIn(WorkflowUiExperimentalApi::class)
+class AndroidViewContainer<V : ViewRendering>(
```
zach-klippenstein, line 8: Needs kdoc.

zach-klippenstein, line 8: I always imagined this as an interface, but I think I understand why that doesn't work and this is better. Do I understand correctly:

The infra will use the vals on this type to create the initial view to show the rendering. On subsequent updates, they will use the usual mechanism to try to update the old view with the new rendering. If this fails, then they'll ask the viewFactory in this class to create a new view. This means that if a workflow returns an `AndroidViewContainer<String>` in one render pass, and an `AndroidViewContainer<Int>` in the next render pass, the view will be appropriately recreated with the `Int` view factory the second time around.

I think this _could_ be an interface if we had self types. Then we could do something like:
```kotlin
interface AndroidViewRendering {
  val viewFactory: ViewFactory<Self>
}
```

The way this would look with fake self-types is:
```kotlin
interface AndroidViewRendering<R : AndroidViewRendering<R>> {
  val viewFactory: ViewFactory<R>
}
```

So for example:
```kotlin
data class Rendering(
  val message: String,
  val onClick: () -> Unit
) : AndroidViewRendering<Rendering> {
  // This is kind of awkward, maybe it should just be a fun.
  override val viewFactory get() = LayoutRunner.bind(HelloGoodbyeLayoutBinding::inflate) { r, _ ->
    helloMessage.text = r.message
    helloMessage.setOnClickListener { r.onClick() }
  }
}
```

That seems less boilerplaty to me than having to create a function just to return the correct `AndroidViewContainer` type. WDYT?

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/ViewRendering.kt
```diff
@@ -0,0 +1,4 @@
+package com.squareup.workflow1.ui
+
+@WorkflowUiExperimentalApi
+interface ViewRendering
```
zach-klippenstein, line 4: I still don't see the point of this class. Why shouldn't I be able to create an `AndroidViewContainer<String>` if I want to?

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/AndroidViewContainer.kt
rjrjr: Ass written, the second AVC would never be shown. To fix that would have to make AVC implement Compatibile. I think that's another reason to avoid the wrapper and go instead with a base class or something.

rjrjr: As I've written it, the second AVC would never be shown. To fix that would have to make AVC implement Compatibile. I think that's another reason to avoid the wrapper and go instead with a base class or something.

I hadn't thought about making a new interface, was trying to just make ViewFactory work, and I think it was the unrecognized need for the self typing that derailed me.

I'll try that and the base class, see how they feel.

zach-klippenstein: > I hadn't thought about making a new interface

Well, you already made `ViewRendering`, so what's one more? 😜 

Joking aside, I'm curious to see what you come up with. Renderings implementing an interface seems like the most straightforward API to me. I'm not sure what advantages you see from an abstract class over an interface.

rjrjr: Yeah, sleeping on it (more or less) and looking again at your sketch, the interface seems like a clear win.

rjrjr: Also, I don't see any reason for that `val viewFactory` to use `get()`.

rjrjr: OMG SO GOOD

```kotlin
data class HelloView(
  val message: String,
  val onClick: () -> Unit
) : AndroidViewRendering<HelloView> {
  override val viewFactory: ViewFactory<HelloView> =
    LayoutRunner.bind(HelloGoodbyeLayoutBinding::inflate) { r, _ ->
      helloMessage.text = r.message
      helloMessage.setOnClickListener { r.onClick() }
    }
}
```

```kotlin
  override fun render(
    props: Unit,
    state: State,
    context: RenderContext
  ): HelloView {
    return HelloView(
      message = state.name,
      onClick = { context.actionSink.send(helloAction) }
    )
  }
```

rjrjr: I wish I could delete the type declaration on the `val viewFactory` implementation, cant understand why it breaks inference. Oh well.

samples/containers/app-poetry/src/main/java/com/squareup/sample/poetryapp/PoemListView.kt
```diff
@@ -25,29 +10,41 @@ import androidx.recyclerview.widget.RecyclerView
...
+import com.squareup.workflow1.ui.AndroidViewRendering
 import com.squareup.workflow1.ui.LayoutRunner
-import com.squareup.workflow1.ui.LayoutRunner.Companion.bind
-import com.squareup.workflow1.ui.ViewFactory
 import com.squareup.workflow1.ui.ViewEnvironment
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
 
 @OptIn(WorkflowUiExperimentalApi::class)
-class PoemListLayoutRunner(view: View) : LayoutRunner<PoemListRendering> {
+data class PoemListView(
```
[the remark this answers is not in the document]
zach-klippenstein: > The rendering class for FooWorkflow is named FooView

Instead of `FooRendering`? I think it's simpler to just stick with "Rendering".

> If it has a LayoutRunner implementation, that's private and declared in the same file.

Definitely, I assumed these would always be private implementations.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/AndroidViewRendering.kt
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Interface implemented by a rendering class to allow it to drive an Android UI
+ * via an appropriate [ViewFactory] implementation.
+ *
+ * You will rarely, if ever, write a [ViewFactory] yourself.  Instead
+ * use [LayoutRunner.bind] to work with XML layout resources, or
+ * [BuilderViewFactory] to create views from code.  See [LayoutRunner] for more
+ * details.
+
```
zach-klippenstein, line 11: Nit?: Missing `*` (not sure if this affects parsing)

zach-klippenstein (approved):

rjrjr:
Update renames *View back to *Rendering.

rjrjr:
Rebasing, rebasing, I can't keep up with Zach!

steve-the-edwards (commented):
Cool; will this be the first in a category of "Platform"ViewRendering interfaces do you think>

zach-klippenstein:
The idea was always for workflow to be able to support different ui frameworks. That doesn't change with this. Although we don't have any immediate plans to implement support for other ones at this time.

PR #264, opened 2020-12-14
Fixes error message when no view binding is found.

rjrjr:
And adds explicit visibility to some existing tests to quiet IDE warnings.

### DIFF
--- a/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewRegistry.kt
+++ b/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewRegistry.kt
@@ -123,7 +123,7 @@ public fun <RenderingT : Any> ViewRegistry.buildView(
 ): View {
   @Suppress("UNCHECKED_CAST")
   val factory: ViewFactory<RenderingT> = getFactoryFor(initialRendering::class)
-    ?: (initialRendering as? AndroidViewRendering<*>)?.viewFactory as ViewFactory<RenderingT>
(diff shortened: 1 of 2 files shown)


workflow-ui/core-android/src/test/java/com/squareup/workflow1/ui/CompositeViewRegistryTest.kt
```diff
@@ -6,9 +6,9 @@ import kotlin.reflect.KClass
 import kotlin.test.assertFailsWith
 
 @OptIn(WorkflowUiExperimentalApi::class)
-class CompositeViewRegistryTest {
+public class CompositeViewRegistryTest {
```
zach-klippenstein, line 9: We don't need public visibility modifiers in tests. They are explicitly excluded from the gradle plugin, Android Studio is just clueless.

zach-klippenstein (approved):
Looks good, but the public modifiers for tests is unnecessary.

workflow-ui/core-android/src/test/java/com/squareup/workflow1/ui/CompositeViewRegistryTest.kt
rjrjr: Yeah, but it's also drowning out all the real warnings. Not sure what to do instead.

rjrjr: Out of scope for this PR so I'll back them out.

steve-the-edwards (approved):

PR #267, opened 2020-12-17
Make runningSideEffect's function accept a CoroutineScope receiver.

### LINKED ISSUE #263: runningSideEffect should have a CoroutineScope receiver
This function is effectively a coroutine builder, and it is idiomatic for such functions to take a scope receiver. All the standard ones do, as well as Compose's `LaunchedEffect`. We recently saw a use case where someone was doing gymnastics to create a scope in a side effect, which is further evidence we should just expose it.

zach-klippenstein:
This is more idiomatic for coroutine-builder-style functions like this,
and makes it easy to write fire-and-forget side effects.

Closes #263.

### DIFF
--- a/workflow-runtime/src/main/java/com/squareup/workflow1/internal/WorkflowNode.kt
+++ b/workflow-runtime/src/main/java/com/squareup/workflow1/internal/WorkflowNode.kt
@@ -123,7 +123,7 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
 
   override fun runningSideEffect(
     key: String,
-    sideEffect: suspend () -> Unit
+    sideEffect: suspend CoroutineScope.() -> Unit
   ) {
     // Prevent duplicate side effects with the same key.
     sideEffects.forEachStaging {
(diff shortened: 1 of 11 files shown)


rjrjr:
I give up — where do we actually invoke the method against the scope?

workflow-runtime/src/main/java/com/squareup/workflow1/internal/WorkflowNode.kt
```diff
@@ -225,10 +225,10 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
 
   private fun createSideEffectNode(
     key: String,
-    sideEffect: suspend () -> Unit
+    sideEffect: suspend CoroutineScope.() -> Unit
   ): SideEffectNode {
     val scope = this + CoroutineName("sideEffect[$key] for $id")
-    val job = scope.launch(start = LAZY) { sideEffect() }
+    val job = scope.launch(start = LAZY, block = sideEffect)
```
rjrjr, line 231: Wow, is this the spot?

rjrjr (approved):

workflow-runtime/src/main/java/com/squareup/workflow1/internal/WorkflowNode.kt
zach-klippenstein: Yep. This change wasn't actually necessary, since _this_ block already has the receiver, but this way we're allocating one fewer lambda.

PR #275, opened 2021-01-20
Tests for ViewStateCache and BackStackContainer.

zach-klippenstein:
This lays the groundwork for starting to TDD support for AndroidX view tree owners.

cc @laurakelly

### DIFF
--- a/versions.properties
+++ b/versions.properties
@@ -13,6 +13,7 @@ version.androidx.gridlayout=1.0.0
 version.androidx.lifecycle=2.2.0
 version.androidx.recyclerview=1.1.0
 version.androidx.savedstate=1.0.0
+version.androidx.test.core=1.3.0
 version.androidx.test.espresso=3.3.0
 version.androidx.test.ext.junit=1.1.2
 version.androidx.test.uiautomator=2.2.0
--- a/workflow-ui/backstack-android/build.gradle.kts
+++ b/workflow-ui/backstack-android/build.gradle.kts
@@ -10,11 +10,16 @@ java {
 }
 
 apply(from = rootProject.file(".buildscript/configure-maven-publish.gradle"))
-
 apply(from = rootProject.file(".buildscript/configure-android-defaults.gradle"))
+apply(from = rootProject.file(".buildscript/android-ui-tests.gradle"))
 
-// See https://github.com/Kotlin/kotlinx.coroutines/issues/1064#issuecomment-479412940
-android.packagingOptions.exclude("**/*.kotlin_*")
+android {
+  // See https://github.com/Kotlin/kotlinx.coroutines/issues/1064#issuecomment-479412940
+  packagingOptions.exclude("**/*.kotlin_*")
+
+  // Disable transition animations.
+  testOptions.animationsDisabled = true
+}
 
 dependencies {
   api(project(":workflow-core"))
@@ -31,9 +36,6 @@ dependencies {
   implementation(Dependencies.Kotlin.Coroutines.android)
   implementation(Dependencies.Kotlin.Coroutines.core)
 
-  testImplementation(Dependencies.Test.junit)
-  testImplementation(Dependencies.Test.truth)
-  testImplementation(Dependencies.Kotlin.Coroutines.test)
-  testImplementation(Dependencies.Kotlin.Test.jdk)
-  testImplementation(Dependencies.Kotlin.Test.mockito)
+  androidTestImplementation(Dependencies.Test.AndroidX.core)
+  androidTestImplementation(Dependencies.Test.AndroidX.truthExt)
 }
--- a/workflow-ui/backstack-android/src/androidTest/java/com/squareup/workflow1/ui/backstack/test/BackStackContainerTest.kt
+++ b/workflow-ui/backstack-android/src/androidTest/java/com/squareup/workflow1/ui/backstack/test/BackStackContainerTest.kt
@@ -1,0 +1,171 @@
+package com.squareup.workflow1.ui.backstack.test
+
+import android.view.View
+import androidx.test.espresso.Espresso.onView
+import androidx.test.espresso.assertion.ViewAssertions.matches
+import androidx.test.espresso.matcher.ViewMatchers.isCompletelyDisplayed
+import androidx.test.espresso.matcher.ViewMatchers.withTagValue
+import androidx.test.ext.junit.rules.ActivityScenarioRule
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import com.google.common.truth.Truth.assertThat
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+import com.squareup.workflow1.ui.backstack.test.fixtures.BackStackTestActivity
+import com.squareup.workflow1.ui.backstack.test.fixtures.BackStackTestActivity.TestRendering
+import com.squareup.workflow1.ui.backstack.test.fixtures.ViewStateTestView
+import com.squareup.workflow1.ui.backstack.test.fixtures.ViewStateTestView.ViewHooks
+import com.squareup.workflow1.ui.backstack.toBackStackScreen
+import org.hamcrest.Matchers.equalTo
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@OptIn(WorkflowUiExperimentalApi::class)
+@RunWith(AndroidJUnit4::class)
+class BackStackContainerTest {
+
+  @Rule @JvmField val scenarioRule = ActivityScenarioRule(BackStackTestActivity::class.java)
+  private val scenario get() = scenarioRule.scenario
+
+  @Test fun restores_view_on_pop_without_config_change() {
+    lateinit var firstScreen: TestRendering
+    lateinit var firstView: View
+
+    scenario.onActivity { activity ->
+      // Set some view state to be saved and restored.
+      activity.currentTestView.viewState = "hello world"
+      firstView = activity.currentTestView
+      firstScreen = activity.backstack!!.top
+    }
+
+    // Navigate to another screen.
+    setBackstack(firstScreen, TestRendering("new screen"))
+    assertThat(viewForScreen("new screen").viewState).isEqualTo("")
+
+    // Navigate back.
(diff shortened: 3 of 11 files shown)


workflow-ui/backstack-android/src/main/java/com/squareup/workflow1/ui/backstack/ViewStateCache.kt
```diff
@@ -76,7 +76,9 @@ public class ViewStateCache private constructor(
     if (oldViewMaybe != null) {
       oldViewMaybe.namedKey.takeIf { hiddenKeys.contains(it) }
           ?.let { savedKey ->
-            val saved = SparseArray<Parcelable>().apply { oldViewMaybe.saveHierarchyState(this) }
+            val saved = SparseArray<Parcelable>().apply {
+              oldViewMaybe.saveHierarchyState(this)
+            }
```
zach-klippenstein, line 81: Intentional, easier to debug.

workflow-ui/backstack-android/src/androidTest/java/com/squareup/workflow1/ui/backstack/test/BackStackContainerTest.kt
```diff
@@ -0,0 +1,175 @@
...
+    )
+
+    // Navigate back.
+    setBackstack(firstRendering)
+    waitForScreen(firstRendering.name)
+    assertThat(consumeEvents()).containsExactly(
+      "first onViewCreated viewState=",
+      "first onShowRendering viewState=",
+      "first onShowRendering viewState=",
+      "first onRestore viewState=hello",
```
zach-klippenstein, line 149: Note that the first `showRendering` happens before state is restored. @rjrjr is this intentional? Is it required? I'm guessing yes, since Activity will only emit onRestoreInstanceState after onStart, at which point all the views should have received their initial values, so we want the same ordering here. Makes sense that you'd only want to restore scroll position after the scrollable content had been initialized too.

rjrjr:
Very intentional, for the reasons you said. Otherwise the rendering always
stomps the restore. Think about config change in particular. The rendering
is usually something that has already been shown in the previous config,
and me have been typed over since then.

workflow-ui/backstack-android/src/androidTest/java/com/squareup/workflow1/ui/backstack/test/BackStackContainerTest.kt
rjrjr: Very intentional, for the reasons you said. Otherwise the rendering always
stomps the restore. Think about config change in particular. The rendering
is usually something that has already been shown in the previous config,
and me have been typed over since then.

rjrjr: And be assured tests will fail if you change it. There's pretty good coverage here.

versions.properties
```diff
@@ -46,3 +46,6 @@ version.org.jetbrains.kotlinx..binary-compatibility-validator=0.2.3
 version.org.jlleitschuh.gradle..ktlint-gradle=9.2.1
 version.org.openjdk.jmh..jmh-core=1.25.2
 version.org.openjdk.jmh..jmh-generator-annprocess=1.25.2
+
+version.androidx.test.core=1.3.0
+##             # available=1.3.1-alpha03
```
rjrjr, line 51: ?

workflow-ui/backstack-android/build.gradle.kts
```diff
@@ -10,11 +10,16 @@ java {
...
+apply(from = rootProject.file(".buildscript/android-ui-tests.gradle"))
+
+android {
+  // See https://github.com/Kotlin/kotlinx.coroutines/issues/1064#issuecomment-479412940
+  packagingOptions.exclude("**/*.kotlin_*")
 
-// See https://github.com/Kotlin/kotlinx.coroutines/issues/1064#issuecomment-479412940
-android.packagingOptions.exclude("**/*.kotlin_*")
+  // Disable transition animations.
+  testOptions.animationsDisabled = true
```
rjrjr, line 21: Oooh, thanks!

workflow-ui/backstack-android/src/androidTest/java/com/squareup/workflow1/ui/backstack/test/BackStackContainerTest.kt
```diff
@@ -0,0 +1,175 @@
...
+import com.squareup.workflow1.ui.backstack.test.fixtures.ViewStateTestView.ViewHooks
+import com.squareup.workflow1.ui.backstack.toBackStackScreen
+import org.hamcrest.Matchers.equalTo
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@OptIn(WorkflowUiExperimentalApi::class)
+@RunWith(AndroidJUnit4::class)
+class BackStackContainerTest {
```
rjrjr, line 24: Thanks for this. Just to be sure you know, there was already some coverage in the tests of the Tic Tac Toe app, including for config change — in particular, for restoring state in and under backstacks in modals as well.

workflow-ui/backstack-android/src/androidTest/java/com/squareup/workflow1/ui/backstack/test/BackStackContainerTest.kt
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  @Test fun restores_view_after_config_change() {
+    scenario.onActivity { activity ->
+      // Set some view state to be saved and restored.
+      activity.currentTestView.viewState = "hello world"
+    }
+
+    // Destroy and recreate the activity.
+    scenario.recreate()
```
rjrjr, line 63: OMG.

O.
M.
G.

I didn't know this was possible. We need a lot of this in house.

workflow-ui/backstack-android/src/androidTest/java/com/squareup/workflow1/ui/backstack/test/fixtures/BackStackTestActivity.kt
```diff
@@ -0,0 +1,101 @@
...
+import com.squareup.workflow1.ui.backstack.BackStackScreen
+import com.squareup.workflow1.ui.backstack.test.fixtures.BackStackTestActivity.TestRendering
+import com.squareup.workflow1.ui.backstack.test.fixtures.ViewStateTestView.ViewHooks
+import com.squareup.workflow1.ui.bindShowRendering
+import com.squareup.workflow1.ui.showRendering
+
+/**
+ * Basic activity that only contains a [BackStackContainer] for [BackStackContainerTest].
+ * You can "navigate" by setting the [backstack] property. The backstack consists of named
+ * [TestRendering] objects. The backstack value will be preserved across configuration changes via
```
rjrjr, line 24: Clever, nice.

rjrjr (commented):
This is great!

zach-klippenstein:
@rjrjr can you think of other significant test cases I missed? I covered the basic cases that I wanted to make sure we handle correctly for the view tree owners, but I probably missed some.

rjrjr:
> @rjrjr can you think of other significant test cases I missed? I covered the basic cases that I wanted to make sure we handle correctly for the view tree owners, but I probably missed some.

Nothing jumps out at me. Integration tests are out of scope here, I think, and already covered by Tic Tac Toe tests. Configuration is the most important bit and that seems covered.

versions.properties
zach-klippenstein: My response exactly. Will get rid of this.

workflow-ui/backstack-android/src/androidTest/java/com/squareup/workflow1/ui/backstack/test/BackStackContainerTest.kt
zach-klippenstein: That's reassuring, but when I add view tree owner support I need to test even more complex interactions, and writing tests for that with the TTT app would be difficult to do and make the sample a lot messier.

zach-klippenstein:
I feel like setting the orientation to test activity lifecycle interaction is kind of a hack, and `recreate()` is a much more direct way to test behavior when an activity is destroyed and recreated. Rotate might be better for integration tests though since it represents a more realistic interaction, and also tests potentially loading different resources, etc.

workflow-ui/backstack-android/src/androidTest/java/com/squareup/workflow1/ui/backstack/test/BackStackContainerTest.kt
rjrjr: Oh hell yeah, this is great.

rjrjr (approved):

zach-klippenstein:
I'm going to add some tests for `DecorativeViewFactory`, there's a lot going on there and I don't see any tests aside from whatever tests the backstack stuff.

workflow-ui/backstack-android/src/androidTest/java/com/squareup/workflow1/ui/backstack/test/BackStackContainerTest.kt
```diff
@@ -0,0 +1,175 @@
...
+    waitForScreen(secondRendering.name)
+    assertThat(consumeEvents()).containsExactly(
+      "second onViewCreated viewState=",
+      "second onShowRendering viewState=",
+      "second onShowRendering viewState=",
+      "first onSave viewState=hello",
+      "first onDetach viewState=hello",
+      "second onAttach viewState=",
+      "first onAttach viewState=hello",
+      "first onDetach viewState=hello"
```
zach-klippenstein, line 139: I don't know why the exiting view gets attached and detached at the end of the transition here.

zach-klippenstein:
Tests are failing on API 24, and when I run them locally they also fail on API 21. According to the error message, the entering view is still completely off-screen when we're expecting to see it. I have no idea why this is happening on older API levels. Disabling transitions (by subclassing `BackStackContainer` and overriding `performTransition`) fixes the tests, and also gets weird of that weird attach/detach that seems to be happening at the end of the transition.

PR #276, opened 2021-01-20
Fix the double showRendering call when initializing DecorativeViewFactory views.

zach-klippenstein:
Introduces `ViewInitializer` which gets executed by `bindShowRendering` between setting
the tag and actually invoking `showRendering`, which gives the `DecorativeViewFactory`
logic the chance to cancel the original `showRendering` call since it wraps it.

This initializer concept will also be used in a follow-up PR to allow `ViewStateCache`
to correctly initialize AndroidX view tree owners.

### DIFF
--- a/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewInitializer.kt
+++ b/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewInitializer.kt
@@ -1,0 +1,67 @@
+package com.squareup.workflow1.ui
+
+import android.view.View
+
+/**
+ * A single-method interface that can be passed to [buildView] to run logic after the [ViewFactory]
+ * creates the view, but before it's [ViewShowRendering] is invoked for the first time.
+ *
+ * See the documentation on [onViewCreated] for more information about how to implement this
+ * interface.
+ */
+@WorkflowUiExperimentalApi
+public fun interface ViewInitializer {
+  /**
+   * Called by [bindShowRendering] the first time it's called for a particular [View], inside a
+   * [ViewFactory.buildView] method. This function will run before the [ViewShowRendering], which
+   * gives you the chance to initialize any properties of the view that need to be set before the
+   * first [ViewShowRendering] call (e.g. [LayoutRunner.showRendering]).
+   *
+   * The [ShowRenderingTag] tag is set _before_ calling this function, so this function can use
+   * [showRenderingTag], [getShowRendering], [getRendering], etc.
+   *
+   * This function must invoke the ViewFactory's initial [`showRendering`][ViewShowRendering] logic,
+   * and it can do so in a number of ways:
+   *
+   *  - Call [getShowRendering] to get the [ViewShowRendering] and pass it the initial rendering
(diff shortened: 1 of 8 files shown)


workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewInitializer.kt
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * TODO kdoc
+ */
+@WorkflowUiExperimentalApi
+public fun interface ViewInitializer {
+  /**
+   * TODO kdoc
+   *
+   * Implementations of this function _must_ call [showRendering].
+   */
```
zach-klippenstein, line 14: TODO

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewInitializer.kt
```diff
@@ -0,0 +1,28 @@
...
+  public fun onViewCreated(view: View)
```
zach-klippenstein, line 15: Note to self: If it's important that `ViewStateCache` only call `restoreSavedInstanceState` _after_ the view's initial `showRendering`, then this will either have to be changed to allow initializers to run code both before and after the `showRendering` call, or we'll need two `update` functions that `BackStackContainer`s will be required to call. 

I'm leaning towards the latter – it keeps this API much simpler and more focused, and while it makes the `ViewStateCache` more complex, the extra verbosity might actually make the code more clear.

rjrjr (approved):

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewInitializer.kt
zach-klippenstein: initView is called in the same point in the initialization process, so they are related. 
But it's different than initView in that it's not tied to a particular view factory, but to a particular invocation of buildView.

doShowRendering is unrelated though, that gets called on every update.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewInitializer.kt
```diff
@@ -0,0 +1,67 @@
...
+   *      innerShowRendering(rendering, environment)
+   *    }
+   *    ```
+   *
+   * If none of these are done, or if the initial [ViewShowRendering] is invoked more than once,
+   * an [IllegalStateException] will be thrown.
+   *
+   * @param view The [View] that was just created by the [ViewFactory].
+   */
+  public fun onViewCreated(view: View)
```
zach-klippenstein, line 45: cc @laurakelly i removed that parameter like we discussed.

zach-klippenstein:
So I thought this would be required to implement AndroidX ViewTreeOwners support, but I sketched out an idea in #280 that doesn't actually require this at all. I haven't tested that yet though so it might need this after all. But if it doesn't, then I think this is probably too over-engineered just to solve this one case, and we should rethink how to fix the double-render.

rjrjr:
This was made obsolete by #397, I believe.

PR #277, opened 2021-01-21
Upgrade ktlint and make our .editorconfig match Square code styles.

zach-klippenstein:
The editorconfig file was taken from Anvil, with some extra rules disabled (indent + parameter-list-wrapping) because ktlint doesn't handle them right (see comments). I removed individual disablings of the parameter-list-wrapping rule, and had to add some for max line length for some test names.

This new ktlint version should let us use kotlin 1.4 features like fun interfaces and trailing commas.


zach-klippenstein:
Crap I forgot to actualy update the ktlint version 🤦

rjrjr (approved):

FrostRocket (approved):

zach-klippenstein:
This hasn't gone green yet because UI tests keep flaking, however another PR, #275, on which this is based, has passed, and that PR just adds more tests, so by the transitive property, this PR must be green. I'm going to go ahead and merge.

PR #280, opened 2021-01-23
WIP Playing with view tree owner support

zach-klippenstein:
I haven't actually tried running this code, but I think this rough sketch is a good start. 

# Lifecycle 

Any time you call `ViewRegistry.buildView`, we set up the `ViewTreeLifecycleOwner` on the view for you. The instance we install is fancy: 
1. It waits until the view is attached.
2. Then it tries to find the nearest parent lifecycle. It starts by looking up the view tree starting with its parent view, then looks up the context wrapper chain. This means the parent lifecycle will be wired up correctly even in our weird internal case where we have an `AppCompatActivity` but we've disabled the default view tree owners.
3. If a parent lifecycle was found, it mirrors the parent lifecycle. If no parent was found, it sets itself to RESUMED and stays there.
4. If the parent lifecycle becomes DESTROYED, our lifecycle becomes DESTROYED and will never change – this is a terminal state.
5. There is also one extra public method we expose on our special lifecycle owner type: `destroyOnDetach()`. When this method is called, the next time the view is detached, it will move to the DESTROYED terminal state.

This means that implementors of container views only need to worry about one extra thing: the lifecycle is automatically initialized for you, you just need to call `WorkflowLifecycleOwner.destroyOnDetach()` when you are about to replace the view with a different view from the factory. I wired this up in `WorkflowViewStub`, `ViewStateCache` (for backstacks), and `ModalContainer`.

This technique composes well. If you have a tree of `WorkflowViewStub`s, each stub will have its own lifecycle that will be DESTROYED if the stub gets an incompatible rendering. However, if the stub gets detached and reattached with the same rendering, the lifecycle will not be destroyed (I think this is required for RecyclerView support). In other words, the lifecycle won't be destroyed until it's actually being navigated away from. Each lifecycle will also be aware of the lifecycle of the stub above it in the hierarchy. If an entire tree is removed, all the lifecycles in the tree will be destroyed. 

# SavedState

Saved state management only needs to happen in two places: the backstack, and modals.

For the backstack, most of the changes are in `ViewStateCache` and `ViewStateFrame`. `ViewStateFrame` does more of the work now. It is responsible for both the management of a single frame's classic `save/restoreHierarchyState` calls, as well as the androidx `SavedStateRegistry`. The `ViewStateFrame` gets the `WorkflowLifecycleOwner` from the view (installed as described above), and then wraps it with a special `SavedStateRegistryOwner` implementation. The frame contains both the `SparseArray` for legacy state, and a `Bundle` for the androidx state. 

`ViewStateCache` works a little differently now. Instead of only storing frames for hidden views, it now eagerly creates a frame the first time it sees a new view. It restores the view from the frame if necessary. Outgoing views are handled similar to before, but most of the code has moved to the frame class so it's a bit cleaner. The `ViewStateCache` also tracks the current frame, and when it's asked to save its state by the `BackStackContainer`, it first asks the current frame to save its androidx saved state (the classic view state doesn't need to be handled explicitly, it should be traversed automatically). The biggest unknown here is the ordering of the restore calls from `BackStackContainer` – i'm not sure they will come in soon enough to actually restore the saved state registry.

The other place that manages saved state is `ModalContainer`, which keeps some saved state for each dialog. To make this easier to do, I moved the `ViewStateFrame` class into `core-android` and used it from `DialogRef` to manage the androidx registry and classic saved state for free.

# ViewModel

Ugh, gross. Actually this was the _least_ gross thing to do! I decided to not bake this in to the core infra like the other two because you really shouldn't be using view models if you're using workflows. However, some annoying libraries, not to name names, still require you to provide a `VIewTreeViewModelStoreOwner`, just in case some code somewhere ever decides to use one. The contract of this owner is harder to satisfy in pure view world than the other two, because the actual object itself must be retained across configuration instances, and there's no pluggable way to access `onRetainConfigurationInstance` (that's kind of the whole point of ViewModels). However, workflows are a sibling alternative solution to this problem. So to make this integration work, I'm just providing a simple workflow to hold a `ViewModelStore` in the workflow state and expose it as a rendering (yes it's mutable, but the state is private and the rendering type is opaque). This could also be just a simple data holder type that we ask people to put in their own states, it would use the same idea. The rendering type also wraps an arbitrary rendering value, and makes one simple promise: the `View` created for the wrapped rendering will be able to use `ViewTreeViewModelStoreOwner` to access a `ViewModelStore` that is scoped to the wrapper workflow. The `ViewFactory` for the wrapper rendering type uses `DecorativeViewFactory` to hook into the view initialization and set the owner tag on the view. The actual `ViewModelStore` is never exposed publicly, so it's a nice limited API surface.

There are some other approaches we could take here if this doesn't work. One thing I started playing with a little but didn't merge is to wrap the entire `ViewRegistry` so that every factory-created `View` gets its own `ViewModelStore`. This wrapper could hook into the `WorkflowLifecycleOwner` provided above to automatically dispose view models when views disappear, as per the VMS contract. The `ViewRegistry` wrapper would probably be owned by somewhere high enough in the object graph to also be retained across configuration changes. I think the only change to the lifecycle stuff we'd need to make this work is to expose the `destroyOnDetach` flag to determine whether a view model should be cleared or retained.

cc @laurakelly @rjrjr


workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewRegistry.kt
```diff
@@ -139,10 +139,12 @@ public fun <RenderingT : Any> ViewRegistry.buildView(
...
   )
-    .apply {
-      check(this.getRendering<Any>() != null) {
-        "View.bindShowRendering should have been called for $this, typically by the " +
+    .also { newView ->
+      check(newView.getRendering<Any>() != null) {
+        "View.bindShowRendering should have been called for $newView, typically by the " +
           "${ViewFactory::class.java.name} that created it."
+
+        WorkflowLifecycleOwner.installOn(newView)
```
zach-klippenstein, line 147: I ended up actually not using the view initializer thing for this. I haven't tested this, so i'm not sure if it actually works, but I suspect that it's reasonable to require anyone trying to use the view tree stuff to wait until the view is attached (which should happen long after both the initial showRendering and this call). The compose infra does exactly that – it doesn't validate the view tree owners until it's attached.

rjrjr:
Do you want to revert the view initializer then?

zach-klippenstein:
I never merged it, wanted to wait to validate it with this use case. There's still a chance it might be needed, but if not, I will probably just decline that PR. I'm not super worried about the double showRendering thing at this point.

zach-klippenstein:
TODO Write up something in the description about the tradeoffs of hardcoding the support in like this vs building it on top of some more generic mechanism.

zach-klippenstein:
Closing this in favor of #283 and #284.

PR #283, opened 2021-01-28
Introduce support for ViewTreeLifecycleOwner in Android container views.

zach-klippenstein:
This is the first step to supporting Compose in our navigation primitives. The next step will be introducing support for `ViewTreeSavedStateRegistryOwner` in #284.

The PR looks pretty big, but a lot of that is fairly verbose tests.

The main goal of this PR is simply to ensure that, if you call `VIewTreeLifecycleOwner` on any view that is a descendent of a view created by the workflow machinery, then the `LifecycleOwner` that gets returned will be correctly destroyed when the workflow machinery decides to not show that view anymore. This requires support from container views, but that support is as simple as calling `WorkflowLifecycleOwner.get(oldView).destroyOnDetach()` before replacing a child view with a different one from `ViewRegistry.buildView`.

cc @laurakelly


zach-klippenstein:
This is now tested, documented, and ready for review.

samples/tictactoe/app/src/androidTest/java/com/squareup/sample/TicTacToeEspressoTest.kt
```diff
@@ -58,6 +58,7 @@ class TicTacToeEspressoTest {
       IdlingRegistry.getInstance()
           .unregister(activity.idlingResource)
     }
+    scenario.close()
```
zach-klippenstein, line 61: Not related, but this was missing before and could be causing test flakes.

versions.properties
```diff
@@ -8,11 +8,11 @@ version.androidx.activity=1.1.0
...
 version.androidx.databinding=4.0.0
-version.androidx.fragment=1.2.5
+version.androidx.fragment=1.3.0-rc02
 version.androidx.gridlayout=1.0.0
-version.androidx.lifecycle=2.2.0
+version.androidx.lifecycle=2.3.0-rc01
 version.androidx.recyclerview=1.1.0
-version.androidx.savedstate=1.0.0
+version.androidx.savedstate=1.1.0-rc01
 version.androidx.test.core=1.3.0
```
zach-klippenstein, line 16: The whole ViewTree*Owner set of classes is only available in these pre-release versions. I've discussed using these versions with @rjrjr in the past, internally, and we're fine with it. This being a _library_, I'm not sure it's something we want to do – but we don't really have a choice unless we want to wait to be able to start supporting Compose until all these are released as stable.

workflow-ui/internal-testing-android/api/internal-testing-android.api
```diff
@@ -0,0 +1,16 @@
...
+	protected fun onCreate (Landroid/os/Bundle;)V
+	protected fun onDestroy ()V
+	protected fun onPause ()V
+	protected fun onResume ()V
+	protected fun onStart ()V
+	protected fun onStop ()V
+	public final fun recreateRenderingOnNextUpdate ()V
+	protected final fun update (Ljava/lang/Object;)Landroid/view/View;
+}
+
```
zach-klippenstein, line 16: This isn't actually public API – we don't even publish this module to maven, but this is simpler than manually excluding it and then forgetting to unexclude it if we ever decide to ship.

workflow-ui/modal-android/src/main/java/com/squareup/workflow1/ui/modal/ModalContainer.kt
```diff
@@ -155,6 +179,14 @@ public abstract class ModalContainer<ModalRenderingT : Any> @JvmOverloads constr
       }
     }
 
+    internal fun dismiss() {
```
zach-klippenstein, line 182:
```suggestion
    /**
     * Call this instead of calling `dialog.dismiss()` directly – this method ensures that the modal's
     * [WorkflowLifecycleOwner] is destroyed correctly.
     */
    internal fun dismiss() {
```

zach-klippenstein:
Maybe the `destroyOnDetach` thing is too fancy, and the lifecycle should always be destroyed on detach and then recreated if reattached. 

Pros:
- Doesn't require any container support. 
- Which makes it impossible to introduce memory leaks by containers forgetting to behave correctly. 

Cons:
- This would effectively disable a compose feature that lets you keep a composition alive while a view is detached and reattached frequently, the canonical use case for that being recycler view children. Probably not a feature we really care about anyway?
- I don't think any existing implementations of this view tree owner return different `Lifecycle` instances for the same owner over the lifetime of a single view, and so I'm not sure if it would break anything.

zach-klippenstein:
> Maybe the destroyOnDetach thing is too fancy, and the lifecycle should always be destroyed on detach and then recreated if reattached.

I tried doing this, and I don't think it will work. With this approach, I think we need to "reset" the `Lifecycle` returned by the `LifecycleOwner` as soon as the lifecycle is destroyed on detach, so that any code that tries to read/subscribe to the lifecycle before the view is re-attached sees it in the `INITIALIZED` state (subscribing to a `DESTROYED` lifecycle is like subscribing to a completed observable – it's a terminal state). However, if we create a new `LifecycleRegistry` as soon as the view is detached, and it's never re-attached, then we'll be holding on to references to any observers that have subscribed to the new lifecycle indefinitely, and might leak them, since the new lifecycle will never get destroyed. For that reason, I think we need to keep `destroyOnDetach`.

You might have noticed that this is actually a problem with the current solution too – if a `WorkflowLifecycleOwner` is set on a new view that never gets attached, then any lifecycle observers could be leaked. However, I think it's reasonable to asssume/require that any views onto which a WLO is installed _must_ be attached at least once. We rely on this assumption a lot in our internal `doOnAttach`-style functions as well, and since the `installOn` API is a more advanced API only meant for container view authors, I think it's sufficiently hard to get cut on this sharp edge. I will add a note about this to the kdoc.

zach-klippenstein:
I'm going to try to consolidate the two backstack testing activities so we can also use that activity for adding state registry tests later.

zach-klippenstein:
These UI test failures seem legit.

samples/tictactoe/app/src/androidTest/java/com/squareup/sample/TicTacToeEspressoTest.kt
zach-klippenstein: I ended up changing this to a rule instead, even cleaner.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLifecycleOwner.kt
```diff
@@ -0,0 +1,180 @@
...
+/**
+ * An extension of [LifecycleOwner] that is always owned by a [View], is logically a child lifecycle
+ * of the next-nearest [ViewTreeLifecycleOwner] above it (it mirrors its parent's lifecycle until
+ * it's destroyed), and can be [asked to destroy][destroyOnDetach] itself early.
+ *
+ * This type is meant to help integrate with [ViewTreeLifecycleOwner] by allowing the creation of a
+ * tree of [LifecycleOwner]s that mirrors the view tree.
+ *
+ * Custom container views that use [ViewRegistry.buildView] to create their children _must_ ensure
+ * they call [destroyOnDetach] on the outgoing view before they replace children with new views.
```
steve-the-edwards, line 26: And it doesn't make sense for us to wrap this swap (enforcing `destroyOnDetach`) in a base container type?

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLifecycleOwner.kt
```diff
@@ -0,0 +1,180 @@
...
+  public companion object {
+    /**
+     * Creates a new [WorkflowLifecycleOwner] and sets it on [view]'s tags so it can be later
+     * retrieved with [get].
+     *
+     * It's very important that, once this function is called with a given view, that view gets
+     * attached at least once to ensure that the lifecycle eventually gets destroyed (either by
+     * being destroyed explicitly via [destroyOnDetach] or because its parent is destroyed).
+     * If this is not done, any observers registered with the [Lifecycle] may be leaked as they will
+     * never see the destroy event.
```
steve-the-edwards, line 50: I don't understand why this is necessary since we are by default hanging off the parent lifecycle anyway? if this view never got attached couldn't we make sure any LIfecycle observers got our destroyed event ?

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLifecycleOwner.kt
```diff
@@ -0,0 +1,180 @@
...
+      parentState != null -> parentState
+      // We have no parent but we just attached, so just assume we're resumed.
+      isAttached -> RESUMED
+      // We have no parent and we're detached, so just stay where we are. We don't want to destroy
+      // here since any view in the parent can detach and re-attach us at any time, and that doesn't
+      // mean that we're going away.
+      else -> localState
+    }.also { newState ->
+      if (newState == DESTROYED) {
+        // We're now in a terminal DESTROY state. Be a good citizen and make sure to detach from
```
steve-the-edwards, line 173: Are we not able to 'push' the destroy to our observers in the case that the view hasn't been attached? I don't see why that is a necessary step.

workflow-ui/internal-testing-android/src/main/java/com/squareup/workflow1/ui/internal/test/AbstractLifecycleTestActivity.kt
```diff
@@ -0,0 +1,154 @@
...
+ * Create an `ActivityScenarioRule` in your test that launches your subclass of this activity, and
+ * then have your subclass expose a method that calls [update] with whatever rendering type your
+ * test wants to use. Then call [consumeLifecycleEvents] to get a list of strings back that describe
+ * what lifecycle-related events occurred since the last call.
+ *
+ * Subclasses must override [viewRegistry] to specify the [ViewFactory]s they require. All views
+ * will be hosted inside a [WorkflowViewStub].
+ */
+@WorkflowUiExperimentalApi
+public abstract class AbstractLifecycleTestActivity : AppCompatActivity() {
```
steve-the-edwards, line 35: Neat!

workflow-ui/modal-android/src/main/java/com/squareup/workflow1/ui/modal/ModalContainer.kt
```diff
@@ -41,25 +43,46 @@ public abstract class ModalContainer<ModalRenderingT : Any> @JvmOverloads constr
...
     newScreen: HasModals<*, ModalRenderingT>,
     viewEnvironment: ViewEnvironment
   ) {
     baseViewStub.update(newScreen.beneathModals, viewEnvironment)
 
+    // // This view should only ever be created by a ViewFactory, which means that well-behaved
+    // // containers should have wired it up with a lifecycle.
+    // val myLifecycle = checkNotNull(WorkflowLifecycleOwner.get(this)) {
+    //   "Expected ModalContainer's parent to set a WorkflowLifecycleOwner on it."
+    // }
```
steve-the-edwards, line 58: Meant to remove this check completely? If we want to ensure parentLifecyleOwner why make it nullable?

steve-the-edwards:
A few questions mostly to wrap my head around the goals/limitations of what is happening here.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLifecycleOwner.kt
zach-klippenstein: How would we do that? The containers all have different logic around this code. E.g. compare `BackStackContainer` and `ModalContainer`.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLifecycleOwner.kt
zach-klippenstein: We don't get the parent lifecycle until the view is attached though. Although this is slightly incorrect, as it would also be ok to never attach the view but just call `destroyOnDetach()` directly – if the view is detached when this is called, the lifecycle will be destroyed immediately (TODO need docs here and on `destroyOnDetach`).

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLifecycleOwner.kt
zach-klippenstein: That will happen, but it's not related to this code so I'm not sure I understand what you mean. If the lifecycle has listeners when this code is executed, they will be notified of the destroyed event when we set the value on the `LifecycleRegistry`.

This is just ensuring that we clear our own listener of our parent when we no longer care to listen.

workflow-ui/modal-android/src/main/java/com/squareup/workflow1/ui/modal/ModalContainer.kt
zach-klippenstein: Hm, probably.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLifecycleOwner.kt
```diff
@@ -0,0 +1,180 @@
...
+      return
+    }
+
+    localLifecycle.currentState = when {
+      // We've been queued for destruction.
+      // Stay attached to the parent's lifecycle until we re-attach, since the parent could be
+      // destroyed while we're detached.
+      destroyOnDetach && !isAttached -> DESTROYED
+      // We may or may not be attached, but we have a parent lifecycle so we just blindly follow it.
+      parentState != null -> parentState
```
zach-klippenstein, line 164: I think this is too permissive. We should crash if there's no parent lifecycle, since the root of the lifecycle tree should always be a `Fragment` or `ComponentActivity`. If it's not, we might leak all over the place.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewRegistry.kt
```diff
@@ -112,6 +112,12 @@ public fun ViewRegistry(): ViewRegistry = TypedViewRegistry()
  * If that returns null, falls back to the factory provided by the rendering's
  * implementation of [AndroidViewRendering.viewFactory], if there is one.
  *
+ * The returned view will have a [WorkflowLifecycleOwner] set on it, the caller _must_ ensure the
```
rjrjr, line 115: What would you think about adding a new optional param to `buildView` to handle the bookending:

```kotlin
public fun <RenderingT : Any> ViewRegistry.buildView(
  initialRendering: RenderingT,
  contextForNewView: Context,
  container: ViewGroup? = null,
  replaces: View? = null
): View
```

It's a shame the way it kind of collides with `container`, don't like it.

Could also just add a new extension, and derive most of the noise from `previous`:

```kotlin
public fun <RenderingT : Any> ViewRegistry.replaceView(
  initialRendering: RenderingT,
  previous: View? = null
): View
```

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewRegistry.kt
```diff
@@ -153,6 +161,12 @@ public fun <RenderingT : Any> ViewRegistry.buildView(
  * Creates a [View] to display [initialRendering], which can be updated via calls
  * to [View.showRendering].
  *
+ * The returned view will have a [WorkflowLifecycleOwner] set on it, the caller _must_ ensure the
```
rjrjr, line 164: I'm concerned about the must-attach-once part of the contract. e.g. I've just learned that our app is littered with cases where a window shown and dismissed so quickly fast that its decorView is never attached.

rjrjr: Ah, you said below the attach actually isn't required.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLifecycleOwner.kt
```diff
@@ -0,0 +1,180 @@
...
+        view.addOnAttachStateChangeListener(it)
+      }
+    }
+
+    /**
+     * Looks for the nearest [ViewTreeLifecycleOwner] on [view] and returns it if it's an instance
+     * of [WorkflowLifecycleOwner]. Convenience function for retrieving the owner set by
+     * [installOn].
+     */
+    public fun get(view: View): WorkflowLifecycleOwner? =
```
rjrjr, line 76: nit: I usually make this kind of thing an extension, `fun View.workflowLifecycleOwnerOrNull()`

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLifecycleOwner.kt
```diff
@@ -0,0 +1,180 @@
...
+    private fun findViewTreeLifecycle(view: View): Lifecycle? {
+      // Start at our view's parent – if we look on our own view, we'll just get this back.
+      return (view.parent as? View)?.let { ViewTreeLifecycleOwner.get(it) }?.lifecycle
+        ?: view.context.lifecycleOrNull()
+    }
+  }
+}
+
+/**
+ * @param enforceMainThread Allows disabling the main thread check for testing.
```
rjrjr, line 93: Thanks for the doc

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLifecycleOwner.kt
rjrjr: When exactly should we crash? You don't want to require getting attached, so parent lifecycle really can be null forever, right?

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLifecycleOwner.kt
zach-klippenstein: It probably makes sense to do this automatically in Compose, since I expect container implementations will all look a lot more uniform and Compose actually has a stronger concept of lifetime than classic android views do.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLifecycleOwner.kt
zach-klippenstein: The following invariants should hold:
- If we have ever been attached, and we have not been destroyed, then there must be a parent lifecycle.
- If there is no parent lifecycle, then either we must be in the INITIALIZED state (and never been attached), OR we must be in the DESTROYED state.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewRegistry.kt
zach-klippenstein: I don't think either of those would work with `ModalContainer`, which both needs to be able to independently destroy any of it's modal children (not replace them), and delegate the `buildView` call to an abstract method.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewRegistry.kt
zach-klippenstein: As long as the container actually destroys its lifecycle. Another mitigating factor is that, I believe, in practice nothing will ask for a `ViewTreeLifecycleOwner`, let alone actually start observing it, until the view is attached (because the assumption is that in 99% of cases the view itself won't have its own owner, and the view tree must be traversed to find one).

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLifecycleOwner.kt
rjrjr: That'll be a good comment.

rjrjr (approved):

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLifecycleOwner.kt
zach-klippenstein: It'll be an even better set of unit tests.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLifecycleOwner.kt
steve-the-edwards: Sorry ya my comment was confusing - It was related to your previous kdoc note of making sure we *attach* the view. I was commenting here because I thought this was the place we handled destorying whether ot not the view was ever attached so I didn't see a reason why we needed to do what you cautioned we must do (up on line 50)

steve-the-edwards (approved):

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLifecycleOwner.kt
zach-klippenstein: Done.

workflow-ui/modal-android/src/main/java/com/squareup/workflow1/ui/modal/ModalContainer.kt
zach-klippenstein: Removed this commented block. This is now part of the contract of `WorkflowLifecycleOwner` and enforced at runtime by it.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLifecycleOwner.kt
zach-klippenstein: Docs added.

steve-the-edwards (approved):

PR #284, opened 2021-01-28
Introduce support for ViewTreeSavedStateRegistryOwner in Android container views.

zach-klippenstein:
Based on #283.

cc @laurakelly


zach-klippenstein:
TODO:

- [ ] Write tests
- [ ] Split the ViewStateCache/ViewStateFrame API cleanup into a prefix PR

zach-klippenstein:
This is now ready for review.

workflow-ui/backstack-android/api/backstack-android.api
```diff
@@ -31,36 +29,11 @@ public final class com/squareup/workflow1/ui/backstack/BackStackContainer$Compan
 	public fun getType ()Lkotlin/reflect/KClass;
 }
 
-public final class com/squareup/workflow1/ui/backstack/ViewStateCache : android/os/Parcelable {
-	public static final field CREATOR Lcom/squareup/workflow1/ui/backstack/ViewStateCache$CREATOR;
+public final class com/squareup/workflow1/ui/backstack/ViewStateCache {
 	public fun <init> ()V
-	public fun describeContents ()I
+	public final fun installOnContainer (Landroid/view/View;)V
```
zach-klippenstein, line 34: Really happy that I managed to whittle this public API down to all of 3 methods.

workflow-ui/modal-android/src/main/java/com/squareup/workflow1/ui/modal/ViewTrees.kt
```diff
@@ -0,0 +1,61 @@
...
+ * Forwarding is best-effort – if no owner is set, none will be forwarded.
+ *
+ * @param anchorView A [View] that is attached and exists in the "main" view tree. E.g. a modal
+ * container view. View tree owners will be read from this view.
+ * @param rootViews A [Sequence] of [View]s that will be shown, but will not be part of the same
+ * view tree as [anchorView]. E.g. the root views for some dialogs that are managed by [anchorView].
+ * View tree owners from [anchorView] will be set on all these views.
+ */
+@OptIn(WorkflowUiExperimentalApi::class)
+internal fun linkViewTreeOwners(
```
zach-klippenstein, line 30: Part of the reason for pulling this out into a separate file is that I think we will need to add a public, general hook into this mechanism to allow external code to wire up arbitrary connections between view trees. Namely, for Compose specifically, we'll need to make sure to forward the `CompositionReference` that's stored in the view tree to the dialog roots as well. I'm planning to put a hook for that in the `ViewEnvironment` and read it here, at which point we'll probably want to move this tree wiring code to the `core-android` module.

workflow-ui/backstack-android/src/main/java/com/squareup/workflow1/ui/backstack/ViewStateFrame.kt
```diff
@@ -1,19 +1,179 @@
...
+ * This class manages both the Android view state traditionally saved and restored through the
+ * [View.onSaveInstanceState]/[View.onRestoreInstanceState], as well as the modern
+ * [SavedStateRegistry] for the view exposed via [ViewTreeSavedStateRegistryOwner].
+ *
+ * A frame can either be [Detached] if it's simply holding the saved data for a hidden view, or
+ * [Attached] when it's managing a visible view.
  */
-internal data class ViewStateFrame(
+@OptIn(WorkflowUiExperimentalApi::class)
+internal class ViewStateFrame(
```
zach-klippenstein, line 39: This class shares even more similarities now with things like `Fragment` and [`NavBackStackEntry`](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:navigation/navigation-runtime/src/main/java/androidx/navigation/NavBackStackEntry.java;l=60?q=backstackentry&sq=).

workflow-ui/backstack-android/src/main/java/com/squareup/workflow1/ui/backstack/ViewStateFrame.kt
```diff
@@ -1,19 +1,179 @@
...
+        event: Event
+      ) {
+        if (event == ON_DESTROY) {
+          changeToDetachedState()
+        }
+      }
+    })
+
+    ViewTreeSavedStateRegistryOwner.set(view, this)
+    ViewTreeLifecycleOwner.set(view, this)
```
zach-klippenstein, line 106: Notice the parallels with [this](https://android-review.googlesource.com/c/platform/frameworks/support/+/1577946).

workflow-ui/backstack-android/src/main/java/com/squareup/workflow1/ui/backstack/BackStackContainer.kt
```diff
@@ -38,7 +37,9 @@ public open class BackStackContainer @JvmOverloads constructor(
   defStyleRes: Int = 0
 ) : FrameLayout(context, attributeSet, defStyle, defStyleRes) {
 
-  private val viewStateCache = ViewStateCache()
+  private val viewStateCache = ViewStateCache().also {
+    it.installOnContainer(this)
+  }
```
zach-klippenstein, line 42: Merging this into our internal fork should be a cinch.

zach-klippenstein:
Started adding a complex integration test mixing backstacks, modals, and compose leaf views, but ran into an issue where having multiple backstacks (`beneachModals` and in the modals) means they all try using the same key for the SavedStateRegistry, which is illegal. This is a new issue with using the SavedStateRegistry for all state saving, instance state didn't have this issue because it didn't use global keys.

workflow-ui/compose/src/androidTest/java/com/squareup/workflow1/ui/compose/ComposeViewTreeIntegrationTest.kt
```diff
@@ -156,10 +177,245 @@ internal class ComposeViewTreeIntegrationTest {
...
+
+    composeRule.runOnIdle {
+      assertThat(state!!.value).isEqualTo("saved")
+    }
+  }
+
+  @Test fun complex_modal_backstack_with_text_entry() {
+    fun buildScreen(name: String): ComposeRendering = ComposeRendering(compatibilityKey = name) {
+      val view = AmbientView.current
+      // TODO remove this once BackStackContainer uses compatibility key
```
steve-the-edwards, line 292: Issue for this?

workflow-ui/internal-testing-android/src/main/java/com/squareup/workflow1/ui/internal/test/AbstractLifecycleTestActivity.kt
```diff
@@ -57,6 +70,11 @@ public abstract class AbstractLifecycleTestActivity : WorkflowUiTestActivity() {
     // changes. This is intentional, since our ViewRegistry probably contains a leafBinding which
     // captures the events list.
     viewEnvironment = ViewEnvironment(mapOf(ViewRegistry to viewRegistry + NamedViewFactory))
+
+    // customNonConfigurationData["onAttachListener"]?.let
+    // TODO
+    // // Must set attached listener first so that it will be invoked when update sets the view.
+    // onViewAttachStateChangedListener = it.attachedStateListener
```
steve-the-edwards, line 77: wut?

workflow-ui/modal-android/src/main/java/com/squareup/workflow1/ui/modal/ViewTrees.kt
steve-the-edwards: I get `anchorView` if we are generalizing for this, but I don't "get" `rootViews`, wouldn't it be more like `subTreeViews`?

steve-the-edwards (approved):
I'm not sure I follow everything that is happening, but I tried.

zach-klippenstein:
Sorry @steve-the-edwards, I moved this back to a draft because it's still not working, and forgot to remove reviewers.

zach-klippenstein:
Just had a brainstorming session with @rjrjr offline, and we figured some things out. We came up with the following principles for containers:

1. Custom containers must have their own view ID set, and either a) ensure that the root of every child's subtree must have a unique view ID set, or b) manually manage the state for its children by using ViewStateFrame to host a SavedStateRegistry and wire up the correct save/restore calls for both the registry and legacy state mechanism and saving its composite state in its parent SavedStateRegistry itself.
2. If a custom container is managing its childrens state itself, then the key it uses to register in its parent SavedStateRegistry must incorporate the view IDs of all the views between itself and the nearest ancestor ViewStateFrame. This ensures that every container that shares a given parent ViewStateFrame will use a key that is unique within the namespace of that shared registry, as long as the views are all assigned unique IDs as is required by the legacy "hierarchy state" mechanism.

BackStackContainer and ModalContainer fall under the second subcase, 1b. BackStackContainer (i.e. ViewStateCache) uses Named renderings to generate unique string keys for each of its frames, which it uses to namespace each frames' state. ModalContainer has an unnamed list of modals, and effectively uses the indices of the modals in that list as keys to namespace its childrens' state. I think that ModalContainer may also need to explicitly manage a ViewStateFrame for the base, non-modal rendering, for namespacing reasons.

WorkflowViewStub falls under 1a. doesn't manage any state itself (it's effectively just a passthrough), which means that the View framework will automatically dispatch save/restore hierarchy state calls to its children, and that when searching for a view tree SavedStateRegistry, its children will locate the registry of the nearest ancestor ViewStateFrame above the WVS. Both of these cases require the WVS to have its own unique View ID to work correctly when there are multiple sibling WVSs at the same level.

zach-klippenstein:
Also we should add more tests around these cases, this late-breaking bug would have slipped through if I hadn't tried adding that fancy compose test just for kicks.

zach-klippenstein:
Ugh, I woke up today thinking a much simpler way to do this would be simply to wrap the `SavedStateRegistry` set by the `ModalContainer` with one that prefixed all the keys. But `SavedStateRegistry`, unlike `Lifecycle`, is a final class, not an interface. f%#@$*

workflow-ui/backstack-android/src/androidTest/java/com/squareup/workflow1/ui/backstack/test/BackstackContainerTest.kt
```diff
@@ -159,7 +166,179 @@ internal class BackstackContainerTest {
...
+
+    // Navigate forward to save it.
+    scenario.onActivity {
+      it.update(firstScreen, LeafRendering("new screen"))
+    }
+
+    // Navigate back.
+    scenario.onActivity {
+      it.update()
+    }
```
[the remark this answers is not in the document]
zach-klippenstein: It navigates back two screens, but it's one action. And technically navigating back 400 screens is still "Navigating back"

workflow-ui/backstack-android/src/main/java/com/squareup/workflow1/ui/backstack/ViewStateCache.kt
```diff
@@ -171,6 +275,23 @@ private val View.namedKey: String
...
     return checkNotNull(rendering?.compatibilityKey) {
       "Expected $this to be showing a ${Named::class.java.simpleName}<*> rendering, " +
-          "found $rendering"
+        "found $rendering"
     }
   }
+
+/**
+ * Computes a string key for using with a `SavedStateRegistry` that will be globally unique.
+ * Requires View IDs to be set for any views that are siblings and have the same rendering type.
```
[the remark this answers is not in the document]
zach-klippenstein: It's a widely known requirement for view state saving in general.

workflow-ui/modal-android/src/main/java/com/squareup/workflow1/ui/modal/ViewTrees.kt
zach-klippenstein: They are effectively the roots of their view trees though. Sure there's a DecorView and whatever on top, but for all of our intents and purposes, these are the root views.

zach-klippenstein:
Ok I think this should be ready for review again now. All the UI tests are passing locally.

workflow-ui/internal-testing-android/src/main/java/com/squareup/workflow1/ui/internal/test/AbstractLifecycleTestActivity.kt
zach-klippenstein, line 77:
```suggestion
```

workflow-ui/modal-android/src/androidTest/java/com/squareup/workflow1/ui/modal/test/ModalViewContainerLifecycleActivity.kt
```diff
@@ -68,5 +70,29 @@ internal class ModalViewContainerLifecycleActivity : AbstractLifecycleTestActivi
...
 
+  val currentModalView: ViewStateTestView<LeafRendering>?
+    get() {
+      val modalContainer = rootRenderedView as ModalContainer<*>
+      val firstModalDecor = modalContainer.dialogDecorViews.firstOrNull() ?: return null
+      @Suppress("UNCHECKED_CAST")
+      return (((firstModalDecor as ViewGroup)
+        .getChildAt(0) as ViewGroup)
+        .getChildAt(0) as ViewGroup)
+        .getChildAt(0) as ViewStateTestView<LeafRendering>
```
zach-klippenstein, line 81: I think this assumption of structure is failing ui tests on certain Android versions:

```

com.squareup.workflow1.ui.modal.test.ModalViewContainerLifecycleTest > modal_hierarchy_state_restored_after_config_change[test(AVD) - 5.0.2] FAILED 
1099
	java.lang.RuntimeException: java.lang.ClassCastException: androidx.appcompat.widget.AppCompatTextView cannot be cast to android.view.ViewGroup
1100
	at androidx.test.runner.MonitoringInstrumentation.runOnMainSync(MonitoringInstrumentation.java:450)
1101

```

workflow-ui/modal-android/src/main/java/com/squareup/workflow1/ui/modal/ViewTrees.kt
```diff
@@ -0,0 +1,51 @@
...
+ * in case any of the owners will resolve to different values.
+ *
+ * Forwarding is best-effort – if no owner is set, none will be forwarded.
+ *
+ * @param anchorView A [View] that is attached and exists in the "main" view tree. E.g. a modal
+ * container view. View tree owners will be read from this view.
+ * @param rootViews A [Sequence] of [View]s that will be shown, but will not be part of the same
+ * view tree as [anchorView]. E.g. the root views for some dialogs that are managed by [anchorView].
+ * View tree owners from [anchorView] will be set on all these views.
+ */
```
rjrjr, line 25:
```suggestion
 * @param anchorView A [View] that is attached and exists in the "main" view tree, 
 * typically a descedent of an Activity contentView -- e.g. a modal container view.
 * View tree owners will be read from this view.
 * @param rootViews A [Sequence] of [View]s that will be shown, but will not be part of the same
 * view tree as [anchorView]. E.g. the root views for some dialogs that are managed by [anchorView].
 * View tree owners from [anchorView] will be set on all these views.
 */
```

rjrjr (approved):

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewStateFrame.kt
```diff
@@ -0,0 +1,220 @@
...
+import androidx.lifecycle.ViewTreeLifecycleOwner
+import androidx.savedstate.SavedStateRegistry
+import androidx.savedstate.SavedStateRegistryController
+import androidx.savedstate.SavedStateRegistryOwner
+import androidx.savedstate.ViewTreeSavedStateRegistryOwner
+import com.squareup.workflow1.ui.ViewStateFrame.AttachState.Attached
+import com.squareup.workflow1.ui.ViewStateFrame.AttachState.Detached
+
+/**
+ * Used by [ViewStateCache] to record the [viewState] data for the view identified
```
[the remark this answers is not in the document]
zach-klippenstein: Nope, i'll add more info about other consumers

workflow-ui/internal-testing-android/src/main/java/com/squareup/workflow1/ui/internal/test/AbstractLifecycleTestActivity.kt
```diff
@@ -57,6 +70,11 @@ public abstract class AbstractLifecycleTestActivity : WorkflowUiTestActivity() {
     // changes. This is intentional, since our ViewRegistry probably contains a leafBinding which
     // captures the events list.
     viewEnvironment = ViewEnvironment(mapOf(ViewRegistry to viewRegistry + NamedViewFactory))
+
+    // customNonConfigurationData["onAttachListener"]?.let
```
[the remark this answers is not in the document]
zach-klippenstein: I just removed this while you were reviewing.

workflow-ui/modal-android/src/androidTest/java/com/squareup/workflow1/ui/modal/test/ModalViewContainerLifecycleActivity.kt
zach-klippenstein: Fixed

workflow-ui/modal-android/src/main/java/com/squareup/workflow1/ui/modal/ModalContainer.kt
```diff
@@ -37,14 +42,68 @@ public abstract class ModalContainer<ModalRenderingT : Any> @JvmOverloads constr
   defStyleRes: Int = 0
 ) : FrameLayout(context, attributeSet, defStyle, defStyleRes) {
 
+  /**
+   * Wrap the [baseViewStub] in another view, since [WorkflowViewStub] swaps itself out for its
+   * child view, but we need a stable reference to its view subtree to set up ViewTreeOwners.
+   * Wrapping with a container view is simpler than trying to make always make sure we keep
+   * [baseViewStub]'s [actual][WorkflowViewStub.actual] view updated.
+   *
+   * Note that this isn't a general problem that needs to be solved for WVS, it's only because we're
```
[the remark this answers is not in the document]
zach-klippenstein: Probably. #347

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowAndroidXSupport.kt
```diff
@@ -15,13 +27,162 @@ public object WorkflowAndroidXSupport {
...
+   * By explicitly looking at the context if the view tree is missing an owner, we can automatically
+   * handle cases where a custom Activity, for example, has overridden setContentView and thus
+   * disabled the ViewTreeOwner integration.
    */
   @WorkflowUiExperimentalApi
   public fun lifecycleOwnerFromViewTreeOrContext(view: View): LifecycleOwner? =
     ViewTreeLifecycleOwner.get(view) ?: view.context.lifecycleOwnerOrNull()
 
+  /**
+   * Tries to get the parent [SavedStateRegistryOwner] from the current view's view via
```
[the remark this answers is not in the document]
zach-klippenstein: done

workflow-ui/backstack-android/src/androidTest/java/com/squareup/workflow1/ui/backstack/test/BackstackContainerTest.kt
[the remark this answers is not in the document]
steve-the-edwards: I'm saying that it's not clear that using update() without params will reach the root. Just saying some help would be nice!

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowAndroidXSupport.kt
```diff
@@ -15,13 +27,162 @@ public object WorkflowAndroidXSupport {
...
+   * The key is generated by combining the rendering [compatibility keys][Named.keyFor] of each
+   * rendering starting with the current one and going up the view tree to the root. Each rendering
+   * key is also associated with the ID of the [View] it's tagged on.
+   */
+  @WorkflowUiExperimentalApi
+  public fun compositeViewIdKey(view: View): String {
+    return generateSequence(view) { it.parent as? ViewGroup }
+      // This will generate the list of ancestor IDs from child -> parent, so it will be in the
+      // reverse order we want (child/parent/root).
+      .map { it.id.toString() }
```
steve-the-edwards, line 64: Faster to do this after list conversion? I guess doesn't matter unless its parallelized for the list.

workflow-ui/modal-android/src/main/java/com/squareup/workflow1/ui/modal/ViewTrees.kt
steve-the-edwards: yes. Understood. I still like `subTreeRootViews` but fine either way.

steve-the-edwards (approved):

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowAndroidXSupport.kt
zach-klippenstein: if we did it after the list conversion, that would be two list traversals instead of one. Of course, it probably doesn't matter anyway since this list will unlikely ever be longer than 20-ish elements.

workflow-ui/modal-android/src/main/java/com/squareup/workflow1/ui/modal/ViewTrees.kt
zach-klippenstein: They're _not_ subtrees though, that's the whole reason any of this logic needs to exist.

zach-klippenstein:
Emulator being garbage again, and a timeout, so i'm going to bump the time limit for the shards since we have so many more UI tests now.
```
	com.android.builder.testing.api.DeviceException: com.android.ddmlib.InstallException: Unknown failure: cmd: Can't find service: package
		at com.android.build.gradle.internal.testing.ConnectedDevice.installPackage(ConnectedDevice.java:132)
```

zach-klippenstein:
Latest failure on that 29 emulator is a crash:
```
com.squareup.workflow1.ui.tests.ViewTreeIntegrationTest > complex_modal_backstack_with_text_entry[test(AVD) - 10] FAILED 
Test failed to run to completion. Reason: 'Test run failed to complete. Expected 1 tests, received 0'. Check device logcat for details
Tests on test(AVD) - 10 failed: Test run failed to complete. Expected 1 tests, received 0. onError: commandError=false message=INSTRUMENTATION_ABORTED: System has crashed.
```

Can't debug this more without seeing logcat, but I've ran these tests on a level 29 emulator locally and everything passes, so I'm confident this is purely an infrastructure issue and the tests are actually green. @rjrjr is ready to cut the release, so I'm going to merge.

PR #287, opened 2021-02-04
Fix UI test flakiness by tweaking our RootMatchers.

zach-klippenstein:
These tests have been flaky for months, and in the last couple weeks have felt worse than ever.

This PR fixes them by:
- Introducing `onWorkflowView`, which is like `Espresso.onView` but looks for views in _all_ roots, instead of using the default root matcher. This introduces a risk of ambiguous matches in new tests, but that should be obvious if we ever hit it, and we can reduce the root scope then.
- Introducing `workflowPressBack` which is like `Espresso.pressBack` but looks only in the focused window, to make sure it doesn't end up finding a hidden window and sending the back event to the wrong view, which ends up looking like a freeze in the tests.
    
Also disables animations in all UI tests by default, instead of individually.

### DIFF
--- a/workflow-ui/internal-testing-android/src/main/java/com/squareup/workflow1/ui/internal/test/Espresso.kt
+++ b/workflow-ui/internal-testing-android/src/main/java/com/squareup/workflow1/ui/internal/test/Espresso.kt
@@ -1,0 +1,44 @@
+package com.squareup.workflow1.ui.internal.test
+
+import android.view.View
+import androidx.test.espresso.Espresso
+import androidx.test.espresso.Root
+import androidx.test.espresso.ViewInteraction
+import androidx.test.espresso.action.ViewActions
+import androidx.test.espresso.matcher.ViewMatchers.isDisplayed
+import androidx.test.espresso.matcher.ViewMatchers.isRoot
+import org.hamcrest.Description
+import org.hamcrest.Matcher
+import org.hamcrest.Matchers.allOf
+import org.hamcrest.Matchers.any
+import org.hamcrest.TypeSafeMatcher
+
+/**
+ * Fork of [Espresso.onView] that looks in all [Root]s, not just the one matched by the default
+ * root matcher. The default root matcher will sometimes fail to find views in our dialogs.
+ */
+public fun onWorkflowView(viewMatcher: Matcher<View>): ViewInteraction {
+  return Espresso.onView(viewMatcher).inRoot(any(Root::class.java))
+}
+
+/**
+ * Fork of [Espresso.pressBack] that finds the root view of the focused window only, instead of
+ * using the default root matcher. This is necessary because when we are showing dialogs,
+ * the default matcher will sometimes match the wrong window, and the back press won't do
+ * anything.
+ */
+public fun workflowPressBack() {
+  val rootHasFocusMatcher = object : TypeSafeMatcher<Root>() {
+    override fun describeTo(description: Description) {
+      description.appendText("has window focus")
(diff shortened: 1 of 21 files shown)


steve-the-edwards (approved):
+100

zach-klippenstein:
And of course now I hit our other flake, the dungeon test:

```

1342
com.squareup.sample.dungeon.DungeonAppTest > loadsBoardsList[test(AVD) - 10] FAILED 
1343
	androidx.test.espresso.NoMatchingViewException: No views in hierarchy found matching: with string from resource id: <2131558429>[boards_list_label] value: Select a board:
1344
	

```

zach-klippenstein:
This one is new:
```

com.squareup.sample.hellobackbutton.HelloBackButtonEspressoTest > wrappedTakesPrecedence[test(AVD) - 10] FAILED 
1162
	androidx.test.espresso.PerformException: Error performing 'single click - At Coordinates: 159, 331 and precision: 16, 16' on view 'with id is <com.squareup.sample.hellobackbutton:id/hello_message>'.
1163
	at androidx.test.espresso.PerformException$Builder.build(PerformException.java:5)
1164

```

zach-klippenstein:
I moved the new helpers into our shared test util module and changed ALL the UI tests to use the new `onWorkflowView` and `workflowPressBack` instead of the Espresso ones. Hopefully this fixes all the flakiness…

zach-klippenstein:
I'm going to merge now since I've already discussed the overall strategy with @rjrjr and @steve-the-edwards approved it. Happy to address any further feedback on new function names in a follow-up.

zach-klippenstein:
Welp, I guess this didn't fix all the flakes. This just happened in  #282:

```
com.squareup.sample.hellobackbutton.HelloBackButtonEspressoTest > wrappedTakesPrecedence[test(AVD) - 10] FAILED 
	androidx.test.espresso.PerformException: Error performing 'single click - At Coordinates: 159, 331 and precision: 16, 16' on view 'with id is <com.squareup.sample.hellobackbutton:id/hello_message>'.
	at androidx.test.espresso.PerformException$Builder.build(PerformException.java:5)
```

workflow-ui/internal-testing-android/src/main/java/com/squareup/workflow1/ui/internal/test/Espresso.kt
```diff
@@ -0,0 +1,44 @@
...
+import org.hamcrest.Matcher
+import org.hamcrest.Matchers.allOf
+import org.hamcrest.Matchers.any
+import org.hamcrest.TypeSafeMatcher
+
+/**
+ * Fork of [Espresso.onView] that looks in all [Root]s, not just the one matched by the default
+ * root matcher. The default root matcher will sometimes fail to find views in our dialogs.
+ */
+public fun onWorkflowView(viewMatcher: Matcher<View>): ViewInteraction {
```
rjrjr, line 20: Drive by bikeshed: there's really nothing workflow specific about either of these methods. I'd suggest `inAnyRoot()` and `pressBackThatActuallyFuckingWorks()`.

zach-klippenstein: I like `inAnyRoot` much better, how about `actuallyPressBack` for the other?

zach-klippenstein: #308

PR #288, opened 2021-02-04
Start running UI tests on API 21.

zach-klippenstein:


zach-klippenstein:
Some of the lifecycle tests are failing on 21, so that's a real issue.

rjrjr (approved):

zach-klippenstein:
Turns out the activity lifecycle seems to be a little different on API level 21: `scenario.moveToState(STARTED)` will stop the activity, not just pause it. To avoid having branching logic in the tests, I just changed the tests to move between `RESUMED` and `STOPPED` instead, which is consistent across all levels from what I can tell.

zach-klippenstein:
This is a weird error, haven't seen it locally at all: 

```
com.squareup.sample.mainactivity.TodoAppTest > navigatesToListAndBack_portrait[test(AVD) - 5.0.2] FAILED 
1472
	java.lang.NullPointerException
1473
	at com.squareup.sample.mainactivity.ToDoActivity.onCreate(ToDoActivity.kt:31)
1474
Tests on test(AVD) - 5.0.2 failed: Instrumentation run failed due to '
```

rjrjr (approved):

PR #301, opened 2021-02-04
Import workflow-kotlin-compose repo into the compose/ directory.

### LINKED ISSUE #289: Import the workflow-kotlin-compose repo into this one.
I would like to bring the compose repo into the main one. I probably won’t merge the Gradle projects yet, since compose is currently using a weird kotlin version, but there’s no reason we can’t have the separate project in the same repo. There never was, I don’t know why I ever made a separate repo for it. If there’s no objections, I’ll bring it in tomorrow and archive the workflow-kotlin-compose repo.

zach-klippenstein:
- As far as gradle are concerned, this is entirely separate from the main workflow gradle project. We might merge them later, but this PR does not attempt to do so.
- Includes all commits from the compose repo (which includes some duplicates from the early workflow days – this is a little weird, but i don't think it's critical, and fixing that would be very difficult). 
- Merges the CODEOWNERs files. 
- The compose github actions configurations are already in their own file, so no merge conflict. 
- I have already transferred open issues over
- Will push branches from compose to this repo so we can pick those up from where we left them.

Fixes #289.


zach-klippenstein:
Since there are a lot of commits here, note the last one in particular: 35455c3. This commit excludes the compose directory from some of our build tooling.

rjrjr (approved):

PR #302, opened 2021-02-05
Remove explicit buildToolsVersion.

zach-klippenstein:
This has not been required since AGP 3.1.0: https://developer.android.com/studio/releases/gradle-plugin#3-1-0

### DIFF
--- a/.buildscript/configure-android-defaults.gradle
+++ b/.buildscript/configure-android-defaults.gradle
@@ -1,6 +1,5 @@
 android {
   compileSdkVersion Versions.targetSdk
-  buildToolsVersion '29.0.2'
 
   compileOptions {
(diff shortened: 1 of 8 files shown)


zach-klippenstein:
Oops i broke the CODEOWNERs file, fixing in #303.

steve-the-edwards (approved):

steve-the-edwards:
Is that a thing now where it doesn't have to be specified?

zach-klippenstein:
@steve-the-edwards Ugh, thanks google for giving broken links 🤦 . I fixed the description:

> This has not been required since AGP 3.1.0: https://developer.android.com/studio/releases/gradle-plugin#3-1-0

PR #304, opened 2021-02-05
Introduce basic integration tests for Compose integration.

zach-klippenstein:
Currently there's no explicit compose support, but we need to test to make sure that
Compose views behave correctly given our custom implementations of the general view
contracts that it expects.

This integration coverage is mostly for #283. I verified that removing the `WorkflowLifecycleOwner.installOn` in `ViewStateCache` causes the `DisposeOnLifecycleDestroyed` strategy test to fail.


samples/tutorial/build.gradle
```diff
@@ -7,7 +7,7 @@ buildscript {
     workflow_version = "1.0.0-alpha.7"
 
     deps = [
-        agp: "com.android.tools.build:gradle:4.1.1",
+        agp: "com.android.tools.build:gradle:4.2.0-beta04",
```
zach-klippenstein, line 10: Compose requires AGP 4.2.0 at least.

versions.properties
```diff
@@ -3,7 +3,7 @@
 ## Dependencies and Plugin versions with their available updates
 ## Generated by $ ./gradlew refreshVersions
 ## Please, don't put extra comments in that file yet, keeping them is not supported yet.
-plugin.android=4.1.1
+plugin.android=4.2.0-beta04
```
zach-klippenstein, line 6: Compose requires AGP 4.2.0 at least.

samples/tutorial/build.gradle
zach-klippenstein, line 10: Sample AGP version must match main project's AGP version.

workflow-ui/internal-testing-android/src/main/java/com/squareup/workflow1/ui/internal/test/AbstractLifecycleTestActivity.kt
```diff
@@ -206,7 +207,7 @@ public abstract class AbstractLifecycleTestActivity : AppCompatActivity() {
 
   public open class LeafView<R : Any>(
     context: Context
-  ) : View(context) {
+  ) : FrameLayout(context) {
```
zach-klippenstein, line 210: Needed to be able to add child views.

workflow-ui/compose/src/androidTest/java/com/squareup/workflow1/ui/compose/NoTransitionBackStackContainer.kt
```diff
@@ -0,0 +1,39 @@
...
+import com.squareup.workflow1.ui.backstack.R
+import com.squareup.workflow1.ui.bindShowRendering
+
+/**
+ * A subclass of [BackStackContainer] that disables transitions to make it simpler to test the
+ * actual backstack logic. Views are just swapped instantly.
+ */
+// TODO ship this?
+@OptIn(WorkflowUiExperimentalApi::class)
+internal class NoTransitionBackStackContainer(context: Context) : BackStackContainer(context) {
```
zach-klippenstein, line 20: @rjrjr I really think this should be the default behavior, and maybe we ship a subclass that uses the transition manager. Or maybe we don't. Either way, I am not convinced the default base class behavior should be to animate.

versions.properties
rjrjr: This kind of thing is why you kept a separate repo earlier. So glad that's done.

samples/tutorial/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.7-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.8.1-all.zip
```
zach-klippenstein, line 3: Had to do this to update AGP, but it's in a prefix commit.

workflow-ui/compose/src/androidTest/java/com/squareup/workflow1/ui/compose/NoTransitionBackStackContainer.kt
rjrjr: Completely agree.

I also think that any configuration of transition animation should be decoupled from backstacks, it's more fundamental than that. In non-compose, I want to tie it to `WorkflowViewStub` instead. I might have a notion of how to use `ViewRegistry` for that, if I can ever land the work to generalize that.

rjrjr (approved):

versions.properties
zach-klippenstein: True, but that was only _required_ because it was using a fork of kotlin itself. We've used pre-release AGP versions before in this project, and I've been using the 4.2.0 canaries forever so I'm really not worried.

workflow-ui/compose/src/androidTest/java/com/squareup/workflow1/ui/compose/NoTransitionBackStackContainer.kt
zach-klippenstein: That sounds nice! Ok, I'll leave this copy in here for now, and will follow up to clean these up (#306).

zach-klippenstein:
Failure looks like another test flake:
```
com.squareup.sample.hellobackbutton.HelloBackButtonEspressoTest > wrappedTakesPrecedence[test(AVD) - 10] FAILED 
	androidx.test.espresso.PerformException: Error performing 'single click - At Coordinates: 159, 331 and precision: 16, 16' on view 'with id is <com.squareup.sample.hellobackbutton:id/hello_message>'.
	at androidx.test.espresso.PerformException$Builder.build(PerformException.java:5)
```

PR #311, opened 2021-02-08
Fix a small typo in samples/tutorial/tutorial-1-complete

afollestad:


rjrjr:
Thanks!

rjrjr (approved):
LG, but we're sorting out why that ownership check is failing.

zach-klippenstein:
Ownership failure is due to #316, but it's not a blocker for this PR since the other checks are green. Thanks!

PR #312, opened 2021-02-08
Update README.md to account for compose folder

EzraLopez:
The compose repo content was moved to the compose folder.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,8 +86,8 @@ higher-level integrations with UI frameworks.
 [Jetpack Compose](https://developer.android.com/jetpack/compose) is the new (under-development,
 pre-release) UI toolkit for Android. It is comparable to SwiftUI for iOS. The main UI artifacts in
 this repository support standard Android Views, but various types of Compose integrations are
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -86,8 +86,8 @@ higher-level integrations with UI frameworks.
 [Jetpack Compose](https://developer.android.com/jetpack/compose) is the new (under-development,
 pre-release) UI toolkit for Android. It is comparable to SwiftUI for iOS. The main UI artifacts in
 this repository support standard Android Views, but various types of Compose integrations are
-provided in the sidecar repository **[square/workflow-kotlin-compose](https://github.com/square/workflow-kotlin-compose)**.
-See that repo for usage info and documentation.
+provided under the **[compose](https://github.com/square/workflow-kotlin/tree/main/compose)** folder.
```
zach-klippenstein, line 89: I think this can just be a relative link:
```suggestion
provided under the **[compose](/compose)** folder. 
```

zach-klippenstein:
Ownership failure is due to #316. It's not a blocker for this PR, but I put up a fix so if you rebase it should be green.

zach-klippenstein (changes requested):
Would prefer to use relative URLs if possible. Thanks for fixing this though!

Issue #314, opened 2021-02-08
ModalContainer doesn't tear down dialogs when destroyed


zach-klippenstein:
It does this when its context's lifecycle is destroyed, but not its local lifecycle is being torn down. Another possible approach would be to tear down on detach. Not sure which is more correct yet.

zach-klippenstein:
Discussion from slack:

Zach:
> @Ray Ryan Any reason by ModalContainer doesn’t just tear the dialogs down when the view is detached? It does so when its context’s lifecycle is destroyed, but container views’ primary lifecycle is their attach state, and this container just seems to assume it will be attached forever.
> I think the container should tear them down when detached, and also make sure to re-show them when attached

Ray:
> Totally agree, nice catch
> Not sure the attach side is necessary?
> Well, I guess we can’t count on update being called after attach

Zach:
> if we tear down when we detach, then we need to re-setup when attached, otherwise if some parent just detaches and reattaches for some reason then the dialogs won’t show.
> that’s unlikely, but it should work if done 

Ray:
> Yeah

Zach:
> also the only reason i even noticed this was because i’m writing a test for some of the androidx integration that assumes that this is all wired correctly

Ray:
> It would be great to get rid of the lifecycle thing. Just have to hope that Android’s nag comes after the activity content view is detached. Surely…

Zach: 
> actually the real bug right now is that the dialog teardown listens to the context’s lifecycle, not the local one that is now being provided from the parent container
> that would be a less drastic change too, only a few lines and leave the attach behavior the same. At least then dialogs would go away correctly if the container will never be re-attached

Ray:
> It’s such a wart. The code is entirely there to prevent noise in logcat, there’s no actually bad behavior.
> I’d love to do your first proposal and light that block on fire
> It’s already caused one or two memory leaks

rjrjr:
See also #204

zach-klippenstein:
Yea, I think if we make `ModalContainer` do the right thing, #204 will get fixed as well.

rjrjr:
I think we're going to discover that waiting for `onDetach` is too late. I'm debugging something right now where that comes after Android's "dude you leaked a dialog" warning in logcat.

rjrjr:
Fixed by #595

PR #315, opened 2021-02-08
Make ModalContainer look for the parent lifecycle using ViewTreeLifecycleOwner.

zach-klippenstein:
This stops the leak of #314, but doesn't implement the full fix agreed upon there.

This change also changes the API a bit so it doesn't pollute the entire auto-complete
namespace for `Context` (or `View`) for consumers, which is important because I'm about
to add some more functions there and Workflow isn't an "AndroidX support support library".


workflow-ui/modal-android/src/main/java/com/squareup/workflow1/ui/modal/ModalContainer.kt
```diff
@@ -76,7 +76,9 @@ public abstract class ModalContainer<ModalRenderingT : Any> @JvmOverloads constr
                 override fun onViewAttachedToWindow(v: View) {
                   // Note this is a different lifecycle than the WorkflowLifecycleOwner – it will
                   // probably be the owning AppCompatActivity.
-                  lifecycle = ref.dialog.decorView?.context?.lifecycleOrNull()
+                  lifecycle = ref.dialog.decorView
+                    ?.let(::lifecycleOwnerFromViewTreeOrContext)
```
zach-klippenstein, line 80: TODO unit tests that this actually fixes #314

rjrjr (approved):

zach-klippenstein:
@rjrjr I know you already approved, but checkout the tests this broke: https://github.com/square/workflow-kotlin/compare/43ed69687a44022eb09a9826cccbb99dae2f5cc8..cdca31cecf62accc50c1594f45e3fa622d47a328

We could have been leaking a lot.

rjrjr:
Yikes.

zach-klippenstein:
The failure on API 21 looks like something that could be legit:
```
	---
```

However I can't repro it locally on 21 or 29.

The failure on 29 looks like a flake.

PR #318, opened 2021-02-10
Upgrade compose to alpha12 and Kotlin to 1.4.30

zach-klippenstein:
3 commits:

1. Extract Compose dependencies to Dependencies.kt.
1. Upgrade Compose to 1.0.0-alpha12.
1. Upgrade to kotlin 1.4.30.
    This was blocked by upgrading Compose to alpha12, and while the Compose module seems to compile fine with just the compose options specifying the new kotlin version that makes me nervous, so I'm going to do the kotlin update in the same branch/PR.


rjrjr (approved):

zach-klippenstein:
Two test failures, I think both are flakes:

21:
```
	---
```
(I've seen that flake once, although not sure how, lifecycle should be deterministic I would think?)

29:
```
com.squareup.sample.dungeon.DungeonAppTest > loadsBoardsList[test(AVD) - 10] FAILED 
	androidx.test.espresso.NoMatchingViewException: No views in hierarchy found matching: with string from resource id: <2131558429>[boards_list_label] value: Select a board:
```
This flakes a lot, should probably fix next.

laurakelly (approved):

zach-klippenstein:
Dungeon failed again, maybe it's real…

```
com.squareup.sample.dungeon.DungeonAppTest > loadsBoardsList[test(AVD) - 10] FAILED 
	androidx.test.espresso.NoMatchingViewException: No views in hierarchy found matching: with string from resource id: <2131558429>[boards_list_label] value: Select a board:

```

zach-klippenstein:
That test is literally the simplest test, it just checks for a basic text view on the first screen in the app. I don't know how it could be flaky.

PR #320, opened 2021-02-10
Stop leaking the viewbinding dependency to consumers.

zach-klippenstein:
We specify it as a `compileOnly` dependency, but we also had `viewBinding = true` by default for all Android modules.
This was causing the viewbinding artifact to be added as an `api` dependency for `:workflow-ui:core-android`, which meant it was leaking to consumers as a transitive dependency.

cc @Armaxis @christiandeange

### DIFF
--- a/workflow-ui/backstack-android/api/backstack-android.api
+++ b/workflow-ui/backstack-android/api/backstack-android.api
@@ -63,12 +63,3 @@ public final class com/squareup/workflow1/ui/backstack/ViewStateCache$SavedState
 	public synthetic fun newArray (I)[Ljava/lang/Object;
 }
 
-public final class com/squareup/workflow1/ui/backstack/databinding/ViewStackLayoutBinding : androidx/viewbinding/ViewBinding {
-	public final field viewStack Lcom/squareup/workflow1/ui/backstack/BackStackContainer;
-	public static fun bind (Landroid/view/View;)Lcom/squareup/workflow1/ui/backstack/databinding/ViewStackLayoutBinding;
-	public synthetic fun getRoot ()Landroid/view/View;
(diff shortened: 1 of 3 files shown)


workflow-ui/backstack-android/api/backstack-android.api
```diff
@@ -63,12 +63,3 @@ public final class com/squareup/workflow1/ui/backstack/ViewStateCache$SavedState
...
 
-public final class com/squareup/workflow1/ui/backstack/databinding/ViewStackLayoutBinding : androidx/viewbinding/ViewBinding {
-	public final field viewStack Lcom/squareup/workflow1/ui/backstack/BackStackContainer;
-	public static fun bind (Landroid/view/View;)Lcom/squareup/workflow1/ui/backstack/databinding/ViewStackLayoutBinding;
-	public synthetic fun getRoot ()Landroid/view/View;
-	public fun getRoot ()Lcom/squareup/workflow1/ui/backstack/BackStackContainer;
-	public static fun inflate (Landroid/view/LayoutInflater;)Lcom/squareup/workflow1/ui/backstack/databinding/ViewStackLayoutBinding;
-	public static fun inflate (Landroid/view/LayoutInflater;Landroid/view/ViewGroup;Z)Lcom/squareup/workflow1/ui/backstack/databinding/ViewStackLayoutBinding;
-}
-
```
zach-klippenstein, line 74: Oops, we were leaking this too.

Armaxis (approved):

steve-the-edwards (approved):

zach-klippenstein:
Oops I broke the samples.

christiandeange (approved):
Thanks for the quick turnaround 🙏

zach-klippenstein:
This is flaking almost 50% on API 21. I'm going to dig into what's going on here soon, just retrying for now.
```
```

Issue #307, opened 2021-02-05
Rename state, props params of render() method to renderedState and renderedProps


rjrjr:
Simple, low-impact change that will should help people avoid capturing stale state in inline lambdas. Because this is undisoverable:

```kotlin
return when (state) {
  is SelectingAge -> 
    Rendering(
      onClick = eventHandler { newAge ->
        this.state = this.state.copy(age = newAge)
      }
    )
  //...
```

And this is at least a bit less so:

```kotlin
return when (renderedState) {
  is SelectingAge -> 
    Rendering(
      onClick = eventHandler { newAge ->
       state = state.copy(age = newAge)
      }
    )
  //...
```

PR #322, opened 2021-02-11
Prefix Props & Context with "Render"

forrestbice:
Prefixes `Workflow#render` `props` and `context` with "render" to better
remind users they should avoid using using the props and state of the
rendering pass as there is the possibility they are stale.

https://github.com/square/workflow-kotlin/issues/307.

### DIFF
--- a/fileTemplates/Stateful Workflow.kt
+++ b/fileTemplates/Stateful Workflow.kt
@@ -22,8 +22,8 @@ class ${WorkflowName} : StatefulWorkflow<Props, State, Output, Rendering>() {
   ): State = TODO("Initialize state")
 
   override fun render(
-    props: Props,
-    state: State,
+    renderProps: Props,
+    renderState: State,
     context: RenderContext
   ): Rendering {
     TODO("Render")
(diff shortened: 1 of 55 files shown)


steve-the-edwards (approved):
Behold, a person of action @forrestbice . Grateful!

rjrjr:
@forrestbice Is there a reason this is still marked as a draft?

forrestbice:
Was just waiting for the build to go green. Looks like warnings are treated as errors so I had to revert the changes in `compose/`. We can change those parameter names once a new version is released and we bump it for the Compose subdirectory.

zach-klippenstein (approved):

rjrjr:
@zach-klippenstein Forrest is struggling to get green UI shards. Do these failures look like the same flakes you've been seeing?

rjrjr (approved):

rjrjr:
@forrestbice if you have it in you, a pass replacing `this.state` with `state` would be a good addition, since that was a big part of the point of this change.

forrestbice:
Yeah, I'll do that in a separate commit later today.

rjrjr:
> @zach-klippenstein Forrest is struggling to get green UI shards. Do these failures look like the same flakes you've been seeing? I'm inclined to merge.

Only API 29 shard is failing, the others are fine. I'm merging.

PR #324, opened 2021-02-12
Cleanup `this.state`

forrestbice:
Now that `props` and `state` of `Workflow#render` have been [renamed](https://github.com/square/workflow-kotlin/pull/322) to
`renderProps` and `renderState`, `this` prefixing of `this.state` in
`WorkflowAction.Updater` lambdas is extraneous.


forrestbice:
@rjrjr looks like this time around both the API 21 and 24 UI tests were cancelled due to what I assume is a timeout at 20 minutes? The other test shards passed.

rjrjr (approved):

rjrjr:
Thanks for this, nice clean up. Less than I expected.

I'll bang on the retry button a couple of more times I guess. :/

forrestbice:
Yeah, less than I expected as well. Worth spot checking me; happy to followup if I missed some.

PR #332, opened 2021-02-18
Add a helper class for intercepting BaseRenderContext calls.

### LINKED ISSUE #330: WorkflowInterceptor (or related) needs hooks for actions being sent into sinks
This hook should fire as early as possible, e.g. on this line: https://github.com/square/workflow-kotlin/blob/79337dac0bb91128e4f091d49cbcd31917ccd4fc/workflow-runtime/src/main/java/com/squareup/workflow1/internal/RealRenderContext.kt#L51

The use case is logging stack traces that indicate what/who is sending events.

### LINKED ISSUE #331: Consider giving WorkflowInterceptor a more direct hook into side effects
If you want to hook into every time a side effect is started, you can do so currently by creating a `WorkflowInterceptor` whose `render` method wraps the `BaseRenderContext` with a custom implementation of `runningSideEffect` that runs your hook, and then passing that wrapper to the proceed function.

This works, but it's not very discoverable and requiring the wrapping of a somewhat complicated type like `BaseRenderContext` feels awkward. Should we provide a more direct hook, maybe just a `fun onSideEffectStarted()`-type function on `WorkflowInterceptor`?

zach-klippenstein:
Instead of passing `BaseRenderContext` back to its `proceed` function, 
`WorkflowInterceptor.onRender` now can provide an implementation of  a new
interface, `RenderContextInterceptor`. This is similar to
`WorkflowInterceptor`, with a bunch of methods that mirror
`BaseRenderContext`, each with its own `proceed` function. This makes it much
simpler to intercept calls to `actionSink.send` and to intercept side effects.

Also uses the new class in `SimpleLoggingWorkflowInterceptor` to add additional logging.

Fixes #330 and fixes #331.

Co-authored-by: Ray Ryan <ray@squareup.com>

### DIFF
--- a/workflow-runtime/api/workflow-runtime.api
+++ b/workflow-runtime/api/workflow-runtime.api
@@ -23,8 +23,8 @@ public final class com/squareup/workflow1/RenderingAndSnapshot {
 public class com/squareup/workflow1/SimpleLoggingWorkflowInterceptor : com/squareup/workflow1/WorkflowInterceptor {
 	public fun <init> ()V
 	protected fun log (Ljava/lang/String;)V
-	protected fun logAfterMethod (Ljava/lang/String;Lcom/squareup/workflow1/WorkflowInterceptor$WorkflowSession;)V
-	protected fun logBeforeMethod (Ljava/lang/String;Lcom/squareup/workflow1/WorkflowInterceptor$WorkflowSession;)V
+	protected fun logAfterMethod (Ljava/lang/String;Lcom/squareup/workflow1/WorkflowInterceptor$WorkflowSession;[Lkotlin/Pair;)V
+	protected fun logBeforeMethod (Ljava/lang/String;Lcom/squareup/workflow1/WorkflowInterceptor$WorkflowSession;[Lkotlin/Pair;)V
 	protected fun logError (Ljava/lang/String;)V
 	public fun onInitialState (Ljava/lang/Object;Lcom/squareup/workflow1/Snapshot;Lkotlin/jvm/functions/Function2;Lcom/squareup/workflow1/WorkflowInterceptor$WorkflowSession;)Ljava/lang/Object;
 	public fun onPropsChanged (Ljava/lang/Object;Ljava/lang/Object;Ljava/lang/Object;Lkotlin/jvm/functions/Function3;Lcom/squareup/workflow1/WorkflowInterceptor$WorkflowSession;)Ljava/lang/Object;
@@ -59,6 +59,18 @@ public final class com/squareup/workflow1/WorkflowInterceptor$DefaultImpls {
 	public static fun onRender (Lcom/squareup/workflow1/WorkflowInterceptor;Ljava/lang/Object;Ljava/lang/Object;Lcom/squareup/workflow1/BaseRenderContext;Lkotlin/jvm/functions/Function3;Lcom/squareup/workflow1/WorkflowInterceptor$WorkflowSession;)Ljava/lang/Object;
 	public static fun onSessionStarted (Lcom/squareup/workflow1/WorkflowInterceptor;Lkotlinx/coroutines/CoroutineScope;Lcom/squareup/workflow1/WorkflowInterceptor$WorkflowSession;)V
 	public static fun onSnapshotState (Lcom/squareup/workflow1/WorkflowInterceptor;Ljava/lang/Object;Lkotlin/jvm/functions/Function1;Lcom/squareup/workflow1/WorkflowInterceptor$WorkflowSession;)Lcom/squareup/workflow1/Snapshot;
+}
+
+public abstract interface class com/squareup/workflow1/WorkflowInterceptor$RenderContextInterceptor {
+	public abstract fun onActionSent (Lcom/squareup/workflow1/WorkflowAction;Lkotlin/jvm/functions/Function1;)V
+	public abstract fun onRenderChild (Lcom/squareup/workflow1/Workflow;Ljava/lang/Object;Ljava/lang/String;Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function4;)Ljava/lang/Object;
+	public abstract fun onRunningSideEffect (Ljava/lang/String;Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function2;)V
+}
+
+public final class com/squareup/workflow1/WorkflowInterceptor$RenderContextInterceptor$DefaultImpls {
+	public static fun onActionSent (Lcom/squareup/workflow1/WorkflowInterceptor$RenderContextInterceptor;Lcom/squareup/workflow1/WorkflowAction;Lkotlin/jvm/functions/Function1;)V
+	public static fun onRenderChild (Lcom/squareup/workflow1/WorkflowInterceptor$RenderContextInterceptor;Lcom/squareup/workflow1/Workflow;Ljava/lang/Object;Ljava/lang/String;Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function4;)Ljava/lang/Object;
+	public static fun onRunningSideEffect (Lcom/squareup/workflow1/WorkflowInterceptor$RenderContextInterceptor;Ljava/lang/String;Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function2;)V
 }
 
 public abstract interface class com/squareup/workflow1/WorkflowInterceptor$WorkflowSession {
--- a/workflow-runtime/src/main/java/com/squareup/workflow1/SimpleLoggingWorkflowInterceptor.kt
+++ b/workflow-runtime/src/main/java/com/squareup/workflow1/SimpleLoggingWorkflowInterceptor.kt
@@ -1,5 +1,6 @@
 package com.squareup.workflow1
 
+import com.squareup.workflow1.WorkflowInterceptor.RenderContextInterceptor
 import com.squareup.workflow1.WorkflowInterceptor.WorkflowSession
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Job
@@ -42,10 +43,10 @@ public open class SimpleLoggingWorkflowInterceptor : WorkflowInterceptor {
     renderProps: P,
     renderState: S,
     context: BaseRenderContext<P, S, O>,
-    proceed: (P, S, BaseRenderContext<P, S, O>) -> R,
+    proceed: (P, S, RenderContextInterceptor<P, S, O>?) -> R,
     session: WorkflowSession
   ): R = logMethod("onRender", session) {
-    proceed(renderProps, renderState, context)
+    proceed(renderProps, renderState, SimpleLoggingContextInterceptor(session))
   }
 
   override fun <S> onSnapshotState(
@@ -59,11 +60,12 @@ public open class SimpleLoggingWorkflowInterceptor : WorkflowInterceptor {
   private inline fun <T> logMethod(
     name: String,
     session: WorkflowSession,
+    vararg extras: Pair<String, Any?>,
     block: () -> T
   ): T {
-    invokeSafely("logBeforeMethod") { logBeforeMethod(name, session) }
+    invokeSafely("logBeforeMethod") { logBeforeMethod(name, session, *extras) }
     return block().also {
-      invokeSafely("logAfterMethod") { logAfterMethod(name, session) }
+      invokeSafely("logAfterMethod") { logAfterMethod(name, session, *extras) }
     }
   }
 
@@ -88,9 +90,10 @@ public open class SimpleLoggingWorkflowInterceptor : WorkflowInterceptor {
    */
   protected open fun logBeforeMethod(
     name: String,
-    session: WorkflowSession
+    session: WorkflowSession,
+    vararg extras: Pair<String, Any?>
   ) {
-    log("START| $name($session)")
+    log("START| ${formatLogMessage(name, session, extras)}")
   }
 
   /**
@@ -98,9 +101,10 @@ public open class SimpleLoggingWorkflowInterceptor : WorkflowInterceptor {
    */
   protected open fun logAfterMethod(
     name: String,
-    session: WorkflowSession
+    session: WorkflowSession,
+    vararg extras: Pair<String, Any?>
   ) {
-    log("  END| $name($session)")
+    log("  END| ${formatLogMessage(name, session, extras)}")
   }
 
   /**
@@ -111,4 +115,40 @@ public open class SimpleLoggingWorkflowInterceptor : WorkflowInterceptor {
   }
 
   protected open fun logError(text: String): Unit = System.err.println(text)
+
+  private fun formatLogMessage(
+    name: String,
+    session: WorkflowSession,
+    extras: Array<out Pair<String, Any?>>
+  ): String = if (extras.isEmpty()) {
+    "$name($session)"
+  } else {
+    "$name($session, ${extras.toMap()})"
+  }
+
+  private inner class SimpleLoggingContextInterceptor<P, S, O>(
+    private val session: WorkflowSession
+  ) : RenderContextInterceptor<P, S, O> {
+
+    override fun onActionSent(
+      action: WorkflowAction<P, S, O>,
+      proceed: (WorkflowAction<P, S, O>) -> Unit
+    ) {
+      logMethod("onActionSent", session, "action" to action) {
+        proceed(action)
+      }
+    }
+
+    override fun onRunningSideEffect(
+      key: String,
+      sideEffect: suspend () -> Unit,
+      proceed: (key: String, sideEffect: suspend () -> Unit) -> Unit
+    ) {
+        proceed(key) {
+          logMethod("onSideEffectRunning", session, "key" to key) {
+            sideEffect()
+          }
+        }
+    }
+  }
 }
--- a/workflow-runtime/src/main/java/com/squareup/workflow1/WorkflowInterceptor.kt
+++ b/workflow-runtime/src/main/java/com/squareup/workflow1/WorkflowInterceptor.kt
@@ -1,7 +1,11 @@
 package com.squareup.workflow1
 
+import com.squareup.workflow1.WorkflowInterceptor.RenderContextInterceptor
 import com.squareup.workflow1.WorkflowInterceptor.WorkflowSession
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Job
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.coroutineContext
 
 /**
  * Provides hooks into the workflow runtime that can be used to instrument or modify the behavior
@@ -25,7 +29,7 @@ import kotlinx.coroutines.CoroutineScope
  *
  * A single workflow may be rendered by different parents at the same time, or the same parent at
  * different, disjoint times. Each continuous sequence of renderings of a particular workflow type,
- * with the same key passed to [RenderContext.renderChild], is called an "session" of that
+ * with the same key passed to [BaseRenderContext.renderChild], is called an "session" of that
  * workflow. The workflow's [StatefulWorkflow.initialState] method will be called at the start of
  * the session, and its state will be maintained by the runtime until the session is finished.
  * Each session is identified by the [WorkflowSession] object passed into the corresponding method
@@ -82,9 +86,9 @@ public interface WorkflowInterceptor {
     renderProps: P,
     renderState: S,
     context: BaseRenderContext<P, S, O>,
-    proceed: (P, S, BaseRenderContext<P, S, O>) -> R,
+    proceed: (P, S, RenderContextInterceptor<P, S, O>?) -> R,
     session: WorkflowSession
-  ): R = proceed(renderProps, renderState, context)
+  ): R = proceed(renderProps, renderState, null)
 
   /**
    * Intercepts calls to [StatefulWorkflow.snapshotState].
@@ -105,7 +109,7 @@ public interface WorkflowInterceptor {
     public val identifier: WorkflowIdentifier
 
     /**
-     * The string key argument that was passed to [RenderContext.renderChild] to render this
+     * The string key argument that was passed to [BaseRenderContext.renderChild] to render this
      * workflow.
      */
     public val renderKey: String
@@ -119,6 +123,102 @@ public interface WorkflowInterceptor {
 
     /** The parent [WorkflowSession] of this workflow, or null if this is the root workflow. */
     public val parent: WorkflowSession?
+  }
+
+  /**
+   * Provides hooks for intercepting calls to a [BaseRenderContext], to be used from [onRender].
+   *
+   * For use by [onRender] methods that want to hook into action and
+   * side effect events. See documentation on methods for more information about the individual
+   * hooks:
+   *  - [RenderContextInterceptor.onActionSent]
+   *  - [RenderContextInterceptor.onRunningSideEffect]
+   *
+   * E.g.:
+   * ```
+   * override fun <P, S, O, R> onRender(
+   *   renderProps: P,
+   *   renderState: S,
+   *   proceed: (P, S, RenderContextInterceptor<P, S, O>) -> R,
+   *   session: WorkflowSession
+   * ): R = proceed(renderProps, renderState, object : RenderContextInterceptor<P, S, O> {
+   *   override fun onActionSent(
+   *     action: WorkflowAction<P, S, O>,
+   *     proceed: (WorkflowAction<P, S, O>) -> Unit
+   *   ) {
+   *     log("Action sent: $action")
+   *     proceed(action)
+   *   }
+   *
+   *   override fun onRunningSideEffect(
+   *     key: String,
+   *     sideEffect: suspend () -> Unit,
+   *     proceed: (key: String, sideEffect: suspend () -> Unit) -> Unit
+   *   ) {
+   *     proceed(key) {
+   *       log("Side effect started: $key")
+   *       sideEffect()
+   *       log("Side effect ended: $key")
+   *     }
+   *   }
+   * })
+   * ```
+   */
+  public interface RenderContextInterceptor<P, S, O> {
+
+    /**
+     * Intercepts calls to [send][Sink.send] on the [BaseRenderContext.actionSink].
+     *
+     * This method will be called from inside the actual [Sink.send] stack frame, so any stack
+     * traces captured from it will include the code that is actually making the send call.
+     */
+    public fun onActionSent(
+      action: WorkflowAction<P, S, O>,
+      proceed: (WorkflowAction<P, S, O>) -> Unit
+    ) {
+      proceed(action)
+    }
+
+    /**
+     * Intercepts calls to [BaseRenderContext.runningSideEffect], allowing the
+     * interceptor to wrap or replace the [sideEffect] and its [key]. This could
+     * be used to prevent a side effect from running, or to augment it with
+     * further effects.
+     *
+     * The [sideEffect] function will perform the actual suspending side effect, and only
+     * return when the side effect is complete – this may be far in the future. This means
+     * the interceptor can be notified when the side effect _ends_ by simply running code
+     * after [sideEffect] returns or throws.
+     *
+     * The interceptor may run [sideEffect] in a different [CoroutineContext], e.g to change
+     * its dispatcher or name, but should take care to use the original [Job], or otherwise
+     * ensure that the structured concurrency contract is not broken.
+     */
+    public fun onRunningSideEffect(
+      key: String,
+      sideEffect: suspend () -> Unit,
+      proceed: (key: String, sideEffect: suspend () -> Unit) -> Unit
+    ) {
+      proceed(key, sideEffect)
+    }
+
+    /**
+     * Intercepts calls to [BaseRenderContext.renderChild], allowing the
+     * interceptor to wrap or replace the [child] Workflow, its [childProps],
+     * [key], and the [handler] function to be applied to the child's output.
+     */
+    public fun <CP, CO, CR> onRenderChild(
+      child: Workflow<CP, CO, CR>,
+      childProps: CP,
+      key: String,
+      handler: (CO) -> WorkflowAction<P, S, O>,
+      proceed: (
+        child: Workflow<CP, CO, CR>,
+        childProps: CP,
+        key: String,
+        handler: (CO) -> WorkflowAction<P, S, O>
+      ) -> CR
+    ): CR = proceed(child, childProps, key, handler)
   }
 }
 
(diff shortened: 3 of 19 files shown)


workflow-runtime/src/main/java/com/squareup/workflow1/WorkflowInterceptor.kt
```diff
@@ -120,6 +120,103 @@ public interface WorkflowInterceptor {
...
+   *         } finally {
+   *           log("Side effect ended: $key")
+   *         }
+   *       }
+   *     }
+   *   }
+   * )
+   * ```
+   */
+  public abstract class RenderContextInterceptor<P, S, O>(
```
zach-klippenstein, line 167: Some problems with this approach:

1. Inheritance. Should this be an interface instead, or even just a function that takes some lambdas? Would eliminate the "don't call super" problem.
2. Requires allocating twice the number of context objects on each render pass than would otherwise be required. If we made these hooks more first-class (built them into `WorkflowInterceptor` directly, called them from `RealRenderContext`) then we could avoid that. But we're also allocating all these lambdas anyway, so what's one more object?

workflow-runtime/src/main/java/com/squareup/workflow1/WorkflowInterceptor.kt
```diff
@@ -120,6 +120,103 @@ public interface WorkflowInterceptor {
...
+    private val baseRenderContext: BaseRenderContext<P, S, O>
+  ) : BaseRenderContext<P, S, O>, Sink<WorkflowAction<P, S, O>> {
+
+    /**
+     * TODO kdoc
```
zach-klippenstein, line 172: TODO

workflow-runtime/src/main/java/com/squareup/workflow1/WorkflowInterceptor.kt
```diff
@@ -120,6 +120,103 @@ public interface WorkflowInterceptor {
...
+     */
+    protected open fun onActionSent(
+      action: WorkflowAction<P, S, O>,
+      proceed: (WorkflowAction<P, S, O>) -> Unit
+    ) {
+      proceed(action)
+    }
+
+    /**
+     * TODO kdoc
```
zach-klippenstein, line 184: TODO

workflow-runtime/src/main/java/com/squareup/workflow1/WorkflowInterceptor.kt
```diff
@@ -120,6 +120,103 @@ public interface WorkflowInterceptor {
     /** The parent [WorkflowSession] of this workflow, or null if this is the root workflow. */
     public val parent: WorkflowSession?
   }
+
+  /**
+   * A helper function intended for use by [onRender] methods that want to hook into action and
```
rjrjr, line 125: Not a function (yet?).

workflow-runtime/src/main/java/com/squareup/workflow1/WorkflowInterceptor.kt
rjrjr: Don't care even a tiny little bit about the lambdas.

I agree that an abstract class is awkward. Hard to choose between an interface or a factory function. Maybe an extension function on `BaseRenderContext`? Or even a couple / three, for tidier syntax if you only want to use one method or the other.

workflow-runtime/src/main/java/com/squareup/workflow1/SimpleLoggingWorkflowInterceptor.kt
```diff
@@ -45,7 +46,28 @@ public open class SimpleLoggingWorkflowInterceptor : WorkflowInterceptor {
...
+            sideEffect()
+          }
+        }
+      }
+
+      override fun onActionSent(
+        action: WorkflowAction<P, S, O>,
+        proceed: (WorkflowAction<P, S, O>) -> Unit
+      ) {
+        logMethod("onActionSent", session, "action" to action) {
```
rjrjr, line 66: Very excited about this, should be so illuminating.

workflow-runtime/src/main/java/com/squareup/workflow1/WorkflowInterceptor.kt
zach-klippenstein: I don't like the extension function syntax since it will pollute the autocomplete namespace for all workflows' `render` methods that have even transient dependencies on the runtime. If we had multiple receivers, I could do an extension on the union of (`WorkflowInterceptor & BaseRenderContext`).

zach-klippenstein: I don't like putting these functions on the main `Workflowinterceptor` interface because then they don't have access to any of the state/props from the render pass that the context was created for, which could be useful to associate.

workflow-runtime/src/main/java/com/squareup/workflow1/WorkflowInterceptor.kt
zach-klippenstein: It is now.

workflow-runtime/src/main/java/com/squareup/workflow1/WorkflowInterceptor.kt
```diff
@@ -120,6 +124,146 @@ public interface WorkflowInterceptor {
...
+            contextInterceptor.onSideEffectRunning(key) {
+              // We don't want to use the CoroutineScope receiver from the outer runningSideEffect
+              // here since the interceptor could be calling this function from a different context
+              // (e.g. wrap it with a different dispatcher or name).
+              coroutineScope {
+                val innerScope = this
+                // If the interceptor is wrapping the coroutine context, it can't change the job
+                // because that could break structured concurrency.
+                check(outerScope.coroutineContext.job === innerScope.coroutineContext.job) {
+                  // TODO unit tests
```
zach-klippenstein, line 257: TODO

workflow-runtime/src/main/java/com/squareup/workflow1/WorkflowInterceptor.kt
```diff
@@ -120,6 +124,146 @@ public interface WorkflowInterceptor {
...
+     * be notified when the side effect _ends_ by simply running code after [proceed] returns or
+     * throws.
+     *
+     * The interceptor may run [proceed] in a different [CoroutineContext], but the context _must_
+     * have the same [Job] or an exception will be thrown. This is to ensure that structured
+     * concurrency is not broken. The context can change things like dispatcher, name, etc.
+     */
+    public suspend fun onSideEffectRunning(
+      key: String,
+      proceed: suspend () -> Unit,
```
zach-klippenstein, line 165: These all being suspend functions makes a really simple API that is really powerful. I'm pretty excited about the idea of "suspending interceptors".

zach-klippenstein:
I think the API is in a state I'm happy with. Just need to update and write more tests.

rjrjr:
> This makes it much simpler to intercept calls to actionSink.send 

Oh hell yeah. We can wrap the actions and make them log updated state and emitted output.

rjrjr:
Should we change `onSessionStarted` to use the new suspending proceed idiom? Feels inconsistent right now, and it's always been a bit weird that the `CoroutineScope` is exposed directly.

```kotlin
  public fun onSessionStarted(
    session: WorkflowSession,
    proceed: suspend (WorkflowSession) -> Unit
  ): Unit = Unit
```

workflow-runtime/src/main/java/com/squareup/workflow1/WorkflowInterceptor.kt
```diff
@@ -165,3 +264,55 @@ internal fun <P, S, O, R> WorkflowInterceptor.intercept(
...
+      // it on subsequent consecutive calls with the same key.
+      interceptor.onSideEffectRunning(key) {
+        // We don't want to use the CoroutineScope receiver from the outer runningSideEffect
+        // here since the interceptor could be calling this function from a different context
+        // (e.g. wrap it with a different dispatcher or name).
+        coroutineScope {
+          val innerScope = this
+          // If the interceptor is wrapping the coroutine context, it can't change the job
+          // because that could break structured concurrency.
+          check(outerScope.coroutineContext.job === innerScope.coroutineContext.job) {
```
rjrjr, line 309: This can't succeed, because `coroutineScope` introduces a new `Job`:

```
 * The provided scope inherits its [coroutineContext][CoroutineScope.coroutineContext] from the outer scope, but overrides
 * the context's [Job].
```

Still figuring out how to fix it.

rjrjr: I'm stuck. Tried this:

```kotlin
    baseRenderContext.runningSideEffect(key) {
      val outerScope = this
      println("job 1: $coroutineContext, ${coroutineContext.job}")

      // Only invoke the hook when the side effect is actually starting – i.e. don't invoke
      // it on subsequent consecutive calls with the same key.
      interceptor.onSideEffectRunning(key) {
        println("job 4: $coroutineContext, ${coroutineContext.job}")
        // If the interceptor is wrapping the coroutine context, it can't change the job
        // because that could break structured concurrency.
        check(outerScope.coroutineContext.job === coroutineContext.job) {
          "Expected onSideEffectRunning not to call proceed with a different Job."
        }

        // We don't want to use the CoroutineScope receiver from the outer runningSideEffect
        // here since the interceptor could be calling this function from a different context
        // (e.g. wrap it with a different dispatcher or name).
        coroutineScope {
          sideEffect()
        }
      }
    }
```

With a `RenderContextInterceptor` that looks like this:

```kotlin
          override suspend fun onSideEffectRunning(
            key: String,
            proceed: suspend () -> Unit
          ) {
            println("job 2: $coroutineContext, ${coroutineContext.job}")
            // Creates a new Job, so sneaky.
            coroutineScope {
              println("job 3: $coroutineContext, ${coroutineContext.job}")
              proceed()
            }
          }
```

Prints this, revealing that `check()` call is just going to capture the
context that `runningSideEffect` was called in, it's blind to what the
interceptor method does behind its back. I don't see how I can detect it.

```
job 1: [CoroutineId(1), "coroutine#1":BlockingCoroutine{Active}@79efed2d, BlockingEventLoop@2928854b], "coroutine#1":BlockingCoroutine{Active}@79efed2d
job 2: [CoroutineId(1), "coroutine#1":BlockingCoroutine{Active}@79efed2d, BlockingEventLoop@2928854b], "coroutine#1":BlockingCoroutine{Active}@79efed2d
job 3: [CoroutineId(1), "coroutine#1":ScopeCoroutine{Active}@2bbf180e, BlockingEventLoop@2928854b], "coroutine#1":ScopeCoroutine{Active}@2bbf180e
job 4: [CoroutineId(1), "coroutine#1":BlockingCoroutine{Active}@79efed2d, BlockingEventLoop@2928854b], "coroutine#1":BlockingCoroutine{Active}@79efed2d
```

rjrjr: Ooh, maybe I can check the job's super job? Is that a thing?

rjrjr: Hot doggy!

```
          check(innerScope.coroutineContext.job in outerScope.coroutineContext.job.children) {
```

rjrjr:
I've pushed an update (no rebasing) that fixes the unit tests, fixes the job-enforcement thing I called out above, and adds three of the four missing unit tests. Have to call time now, but I'll pick this back up on Monday.

rjrjr:
I'm stumped on how to write the final unit test. I can see that the interceptor's context is being used, but I can't figure out how to make that fact visible anywhere in the test code, because of the `coroutineScope {}` call in `InterceptedRenderContext#runningSideEffect` hiding everything.

@zach-klippenstein , got time to pair on Tuesday?

workflow-runtime/src/main/java/com/squareup/workflow1/WorkflowInterceptor.kt
zach-klippenstein: Neat! I am a little scared though, this feels brittle (why just children, and not grandchildren? Etc.)

rjrjr: Yeah, I had the same thought. (Just children should be fine, thought, since we are in complete control of creating the new job, even if we're delegating to `CoroutineScope` to do so.)

I'm still stuck, though, on how to write the last unit test. By using either `CoroutineScope` builder here, I'm obscuring any visibility of the `CoroutineContext` created by the interceptor, since properties like `CoroutineName` aren't inherited. Do you have any tricks to teach me there?

zach-klippenstein: If you use `CoroutineScope(coroutineContext)`, that scope will point to the exact same context instance as the side effect gets. Not sure I understand why this wouldn't work? I guess I should read that test…

workflow-runtime/src/test/java/com/squareup/workflow1/WorkflowInterceptorTest.kt
```diff
@@ -73,7 +82,173 @@ class WorkflowInterceptorTest {
...
+        renderState: S,
+        proceed: (P, S, RenderContextInterceptor<P, S, O>?) -> R,
+        session: WorkflowSession
+      ): R {
+        return proceed(renderProps, renderState, object : RenderContextInterceptor<P, S, O> {
+          override suspend fun onSideEffectRunning(
+            key: String,
+            proceed: suspend () -> Unit
+          ) {
+            val context = coroutineContext + coroutineContext.job + CoroutineName("intercepted!")
```
zach-klippenstein, line 172: ` + coroutineContext.job` is redundant

workflow-runtime/src/test/java/com/squareup/workflow1/WorkflowInterceptorTest.kt
```diff
@@ -73,7 +82,173 @@ class WorkflowInterceptorTest {
...
+            CoroutineScope(context).run {
+              proceed()
+            }
```
zach-klippenstein, line 175: This is a no-op, since `proceed` doesn't take the `CoroutineScope` as a receiver. You want this:
```suggestion
            withContext(CoroutineName("intercepted!")) {
              proceed()
            }
```

zach-klippenstein: One might ask, why not just make `proceed` take the scope as a receiver? Because it still needs to be a suspend function, and that would make it possible for the interceptor to provide a different coroutine context implicitly vs via the scope receiver, which is invalid. By forcing the implicit context as the only way to change the context to `proceed`, we eliminate that potential ambiguity.

zach-klippenstein:
@rjrjr don't forget to add `Co-authored by: Ray Ryan <email@domain.tld>` at the end of your eventual squashed commit so you get commit credit.

workflow-runtime/src/main/java/com/squareup/workflow1/WorkflowInterceptor.kt
rjrjr: Back to this, and I've forgotten how to replace 

```kotlin
coroutineScope { sideEffect() }
```

with 

```kotlin
val scope = CoroutineScope(coroutineContext)
```

How exactly do I use that scope to call `sideEffect`?

workflow-runtime/src/main/java/com/squareup/workflow1/WorkflowInterceptor.kt
```diff
@@ -165,3 +264,55 @@ internal fun <P, S, O, R> WorkflowInterceptor.intercept(
...
+            // TODO unit tests
+            "Expected onSideEffectStarting not to call proceed with a different Job."
+          }
+          sideEffect()
+        }
```
zach-klippenstein:
```suggestion
      interceptor.onSideEffectRunning(key) {
        // We don't want to use the CoroutineScope receiver from the outer runningSideEffect
        // here since the interceptor could be calling this function from a different context
        // (e.g. wrap it with a different dispatcher or name).
        // TODO(ray): Double-check that `coroutineContext` here is the `suspend val` property, and not the property from the outer scope receiver.
        val innerScope = CoroutineScope(coroutineContext)
        // If the interceptor is wrapping the coroutine context, it can't change the job
        // because that could break structured concurrency.
        check(outerScope.coroutineContext.job === innerScope.coroutineContext.job) {
          // TODO unit tests
          "Expected onSideEffectStarting not to call proceed with a different Job."
        }
        innerScope.sideEffect()
      }
```

rjrjr: Remember when we believed that `withContext` just passes the `Job` through? Those were simpler times.

```kotlin
            println("1. ${coroutineContext.job}")
            withContext(TestElement) {
              println("2. ${coroutineContext.job}")
              proceed()
            }
```

```
1. "coroutine#3":BlockingCoroutine{Active}@77847718
2. "coroutine#3":UndispatchedCoroutine{Active}@9597028
```

rjrjr: This does the trick:

```kotlin
            println("1. ${activeCoroutineContext().job}")
            CoroutineScope(TestElement).run {
              println("2. ${activeCoroutineContext().job}")
              proceed
            }
```

rjrjr: Also, this doesn't do what you think it does. The scope receiver has no effect on the `CoroutineContext` that `sideEffect` runs in.

```kotlin
        val innerScope = CoroutineScope(coroutineContext)
        innerScope.sideEffect()
```

But this does:

```kotlin
        val innerScope = CoroutineScope(coroutineContext)
        innerScope.launch { sideEffect() }
```

samples/tutorial/build.gradle
```diff
@@ -7,7 +7,7 @@ buildscript {
     workflow_version = "1.0.0-alpha.7"
 
     deps = [
-        agp: "com.android.tools.build:gradle:4.2.0-beta04",
+        agp: "com.android.tools.build:gradle:4.2.0-beta05",
```
rjrjr, line 10: oops

trace-encoder/build.gradle.kts
```diff
@@ -17,7 +17,6 @@ dependencies {
   api(Dependencies.Kotlin.Stdlib.jdk6)
   api(Dependencies.Kotlin.Coroutines.core)
 
-  implementation(Dependencies.Kotlin.reflect)
```
rjrjr, line 20: Wrong PR

workflow-runtime/src/main/java/com/squareup/workflow1/WorkflowInterceptor.kt
```diff
@@ -120,6 +125,97 @@ public interface WorkflowInterceptor {
...
+   * ): R = proceed(renderProps, renderState, object : RenderContextInterceptor<P, S, O> {
+   *   override fun onActionSent(
+   *     action: WorkflowAction<P, S, O>,
+   *     proceed: (WorkflowAction<P, S, O>) -> Unit
+   *   ) {
+   *     log("Action sent: $action")
+   *     proceed(action)
+   *   }
+   *
+   *   override suspend fun onSideEffectStarting(
```
rjrjr, line 154: Should be `onSideEffectRunning`

workflow-runtime/src/main/java/com/squareup/workflow1/WorkflowInterceptor.kt
```diff
@@ -120,6 +123,102 @@ public interface WorkflowInterceptor {
...
+    public fun onRunningSideEffect(
+      key: String,
+      sideEffect: suspend () -> Unit,
+      proceed: (key: String, sideEffect: suspend () -> Unit) -> Unit
+    ) {
+      proceed(key, sideEffect)
+    }
+
+    /**
+     * Intercepts calls to [BaseRenderContext.renderChild], allowing the
```
rjrjr, line 205: @zach-klippenstein PTAL, new kdoc and new param names.

workflow-runtime/src/main/java/com/squareup/workflow1/WorkflowInterceptor.kt
```diff
@@ -165,3 +268,54 @@ internal fun <P, S, O, R> WorkflowInterceptor.intercept(
...
+  ): ChildRenderingT =
+    interceptor.onRenderChild(child, props, key, handler) { iChild, iProps, iKey, iHandler ->
+      baseRenderContext.renderChild(iChild, iProps, iKey, iHandler)
+    }
+
+  override fun runningSideEffect(
+    key: String,
+    sideEffect: suspend CoroutineScope.() -> Unit
+  ) {
+    // We don't want to invite the interceptor to shoot itself in the foot
```
rjrjr, line 299: @zach-klippenstein PTAL, from our pairing.

workflow-runtime/src/test/java/com/squareup/workflow1/internal/ChainedWorkflowInterceptorTest.kt
```diff
@@ -173,44 +171,102 @@ class ChainedWorkflowInterceptorTest {
...
 
     assertEquals(
-        "r1: r2: (props2: props1: props|state2: state1: state|context2: context1: context)",
-        finalRendering
+      "r1: r2: (props2: props1: props|state2: state1: state)",
+      finalRendering
+    )
+  }
+
+  @Test fun `chains calls with RenderContextInterceptor in left-to-right order`() {
```
rjrjr, line 199: @zach-klippenstein PTAL, new test

rjrjr (approved):

workflow-runtime/src/main/java/com/squareup/workflow1/SimpleLoggingWorkflowInterceptor.kt
```diff
@@ -111,4 +114,41 @@ public open class SimpleLoggingWorkflowInterceptor : WorkflowInterceptor {
...
+    }
+
+    override fun onRunningSideEffect(
+      key: String,
+      sideEffect: suspend () -> Unit,
+      proceed: (key: String, sideEffect: suspend () -> Unit) -> Unit
+    ) {
+      // Note that because logMethod is inline, it automatically acts like a suspend function here.
+      // The START will be logged immediately, and the END will be logged when the side effect
+      // completes, asynchronously.
```
zach-klippenstein, line 148: This comment is now out of date. And this whole method implementation is probably wrong. I think it still makes sense for it to log the lifetime of the actual side effect, as it was originally written.

workflow-runtime/src/main/java/com/squareup/workflow1/SimpleLoggingWorkflowInterceptor.kt
```diff
@@ -111,4 +114,41 @@ public open class SimpleLoggingWorkflowInterceptor : WorkflowInterceptor {
...
+      logMethod("onSideEffectRunning", session, "key" to key) {
+        proceed(key, sideEffect)
+      }
```
zach-klippenstein, line 151:
```suggestion
      proceed(key) {
        logMethod("onSideEffectRunning", session, "key" to key) {
          sideEffect()
        }
      }
```

workflow-runtime/src/main/java/com/squareup/workflow1/WorkflowInterceptor.kt
```diff
@@ -120,6 +123,102 @@ public interface WorkflowInterceptor {
...
+      key: String,
+      sideEffect: suspend () -> Unit,
+      proceed: (key: String, sideEffect: suspend () -> Unit) -> Unit
+    ) {
+      proceed(key, sideEffect)
+    }
+
+    /**
+     * Intercepts calls to [BaseRenderContext.renderChild], allowing the
+     * interceptor to wrap or replace the [child] Workflow, its [childProps],
```
zach-klippenstein, line 206: Also the key

workflow-runtime/src/main/java/com/squareup/workflow1/WorkflowInterceptor.kt
```diff
@@ -165,3 +268,54 @@ internal fun <P, S, O, R> WorkflowInterceptor.intercept(
...
+      }
+    }
+  }
+
+  /**
+   * In a block with a CoroutineScope receiver, calls to `coroutineContext` bind
+   * to `CoroutineScope.coroutineContext` instead of `suspend val coroutineContext`.
+   * Call this and always get the latter.
+   */
+  private suspend fun activeCoroutineContext(): CoroutineContext {
```
zach-klippenstein, line 318: Make this inline? We just need to force the compiler to call the right method, no need to generate a whole new suspend state machine.

workflow-runtime/src/main/java/com/squareup/workflow1/WorkflowInterceptor.kt
```diff
@@ -120,6 +123,102 @@ public interface WorkflowInterceptor {
...
+   * ): R = proceed(renderProps, renderState, object : RenderContextInterceptor<P, S, O> {
+   *   override fun onActionSent(
+   *     action: WorkflowAction<P, S, O>,
+   *     proceed: (WorkflowAction<P, S, O>) -> Unit
+   *   ) {
+   *     log("Action sent: $action")
+   *     proceed(action)
+   *   }
+   *
+   *   override suspend fun onSideEffectRunning(
```
zach-klippenstein, line 152: This example is wrong now

workflow-runtime/src/test/java/com/squareup/workflow1/RecordingWorkflowInterceptor.kt
```diff
@@ -8,25 +8,27 @@ import com.squareup.workflow1.WorkflowInterceptor.WorkflowSession
...
     name: String,
-    session: WorkflowSession
+    session: WorkflowSession,
+    vararg extras: Pair<String, Any?>
   ) {
     events += "END|$name"
   }
 
-  fun consumeEvents(): List<String> = events
+  private fun consumeEvents(): List<String> = events
```
zach-klippenstein, line 31: This was a typo? Maybe was used for tests at some point

workflow-runtime/src/test/java/com/squareup/workflow1/internal/ChainedWorkflowInterceptorTest.kt
zach-klippenstein: Lg, although I probably would have made that transcript a list of strings instead. This works though, thanks for writing!

workflow-runtime/src/main/java/com/squareup/workflow1/WorkflowInterceptor.kt
```diff
@@ -81,10 +85,9 @@ public interface WorkflowInterceptor {
   public fun <P, S, O, R> onRender(
     renderProps: P,
     renderState: S,
-    context: BaseRenderContext<P, S, O>,
```
zach-klippenstein, line 84: I wonder if this removal was too aggressive. When the interceptor gets the actual context here, it can render its own workflow children or run side effects itself. Now it can't. Should we add it back in? I don't think it can hurt, and it's actually a less drastic change.

workflow-runtime/src/main/java/com/squareup/workflow1/WorkflowInterceptor.kt
rjrjr:
```kotlin
   *   override fun onRunningSideEffect(
   *     key: String,
   *     sideEffect: suspend () -> Unit,
   *     proceed: (key: String, sideEffect: suspend () -> Unit) -> Unit
   *   ) {
   *     proceed(key) {
   *       log("Side effect started: $key")
   *       sideEffect()
   *       log("Side effect ended: $key")
   *     }
   *   }
```

workflow-runtime/src/test/java/com/squareup/workflow1/RecordingWorkflowInterceptor.kt
rjrjr: Not a typo, now only used here.

workflow-runtime/src/test/java/com/squareup/workflow1/internal/ChainedWorkflowInterceptorTest.kt
rjrjr: It is…

```kotlin
    val transcript = mutableListOf<String>()
```

workflow-runtime/src/main/java/com/squareup/workflow1/WorkflowInterceptor.kt
rjrjr: Would we pass it back through `proceed()` as well? That seems redundant with the interceptor.

workflow-runtime/api/workflow-runtime.api
```diff
@@ -61,6 +61,18 @@ public final class com/squareup/workflow1/WorkflowInterceptor$DefaultImpls {
 	public static fun onSnapshotState (Lcom/squareup/workflow1/WorkflowInterceptor;Ljava/lang/Object;Lkotlin/jvm/functions/Function1;Lcom/squareup/workflow1/WorkflowInterceptor$WorkflowSession;)Lcom/squareup/workflow1/Snapshot;
 }
 
+public abstract interface class com/squareup/workflow1/WorkflowInterceptor$RenderContextInterceptor {
+	public abstract fun onActionSent (Lcom/squareup/workflow1/WorkflowAction;Lkotlin/jvm/functions/Function1;)V
+	public abstract fun onRenderChild (Lcom/squareup/workflow1/Workflow;Ljava/lang/Object;Ljava/lang/String;Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function4;)Ljava/lang/Object;
+	public abstract fun onRunningSideEffect (Ljava/lang/String;Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function2;)V
+}
+
+public final class com/squareup/workflow1/WorkflowInterceptor$RenderContextInterceptor$DefaultImpls {
```
zach-klippenstein, line 70: This reminds me we should just probably target 1.8 bytecode and turn on `@JvmDefaults`.

workflow-runtime/src/main/java/com/squareup/workflow1/WorkflowInterceptor.kt
zach-klippenstein: Nope. The utility of passing the context here is simply to allow the interceptor to do workflow-y things with it.

workflow-runtime/src/main/java/com/squareup/workflow1/WorkflowInterceptor.kt
```diff
@@ -25,7 +29,7 @@ import kotlinx.coroutines.CoroutineScope
  *
  * A single workflow may be rendered by different parents at the same time, or the same parent at
  * different, disjoint times. Each continuous sequence of renderings of a particular workflow type,
- * with the same key passed to [RenderContext.renderChild], is called an "session" of that
+ * with the same key passed to [BaseRenderContext.renderChild], is called an "session" of that
```
zach-klippenstein, line 32: Oops, thanks.

workflow-runtime/src/main/java/com/squareup/workflow1/SimpleLoggingWorkflowInterceptor.kt
```diff
@@ -111,4 +115,40 @@ public open class SimpleLoggingWorkflowInterceptor : WorkflowInterceptor {
...
+    }
+
+    override fun onRunningSideEffect(
+      key: String,
+      sideEffect: suspend () -> Unit,
+      proceed: (key: String, sideEffect: suspend () -> Unit) -> Unit
+    ) {
+        proceed(key) {
+          logMethod("onSideEffectRunning", session, "key" to key) {
```
zach-klippenstein, line 148: ❤️

zach-klippenstein (commented):
I can't approve my own PR, but I approve of this.

PR #341, opened 2021-03-01
Update to Jetpack Compose UI beta

### LINKED ISSUE #297: renderAsState potentially leaks coroutines
Some Compose devs recently [clarified](https://kotlinlang.slack.com/archives/CJLTWPH7S/p1591660760450000?thread_ts=1591653158.447300&cid=CJLTWPH7S) some details about the lifecycle of compositions, namely how `onCommit` and `onDispose` are related.

Workflows can perform side effects in workers. Workers ran during the initial render pass will be launched before the render pass completes. This means that once the first render pass completes, the root `WorkflowNode` _must_ be cancelled if the runner coroutine never runs, otherwise those worker coroutines will be leaked. 

The current implementation of `renderAsState` starts the runtime outside of an `onCommit` which means that if the composition fails, the runtime will never be cancelled, and any workers started by the runtime in the initial render pass will be leaked.

Some solutions to this:
1. Hide `renderAsState` as an internal-only API. Bring back the version of `WorkflowContainer` that accepts a content composable, so we can ensure the lambda is only composed once the first rendering is available.
2. Keep `renderAsState`, give it a required `initialRendering` parameter, and maybe an overload for non-null rendering types that defaults the initial rendering to null. Asking code outside the workflow to produce a rendering is certainly a bit odd, but would be hidden from consumers of the `WorkflowContainer` API, and probably a non-issue since most root renderings are non-nullable or container types anyway, often both, and would be covered either by the non-nullable overload or could use an empty container as an initial rendering. (This option looks a lot like the built-in `collectAsState` APIs.)
3. Make `renderAsState`'s return type some sort of explicit Optional type. Gross.

### LINKED ISSUE #300: Use the new View.compositionReference for nested view factories instead of ViewEnv
This new extension property is introduced in this PR: https://android-review.googlesource.com/c/platform/frameworks/support/+/1497121

It seems specifically designed to allow arbitrary views in a hierarchy to also nest their compositions without directly knowing about their parents. I think this matches our use case directly. Regardless, we should definitely take advantage of it for the root. 

This also might let us drop the awkward root composition wrapper machinery we've hacked together.

SalvatoreT:
- use the latest Compose UI API
- update the Workflow library
- cross my fingers a bunch

This page saved my life. It just says Ambient was renamed to
CompositionLocal. https://foso.github.io/Jetpack-Compose-Playground/general/ambient/

Fixes #297
Fixes #300

cc @zach-klippenstein @rjrjr @steve-the-edwards


compose/buildSrc/src/main/java/Dependencies.kt
```diff
@@ -30,20 +30,20 @@ object Versions {
 
 @Suppress("unused")
 object Dependencies {
-  const val android_gradle_plugin = "com.android.tools.build:gradle:4.2.0-alpha15"
+  const val android_gradle_plugin = "com.android.tools.build:gradle:4.2.0-beta05"
```
zach-klippenstein, line 33:
```suggestion
  const val android_gradle_plugin = "com.android.tools.build:gradle:7.0.0-alpha08"
```

compose/core-compose/src/main/java/com/squareup/workflow/ui/compose/CompositionRoot.kt
```diff
@@ -86,8 +86,8 @@ fun ViewRegistry.withCompositionRoot(root: CompositionRoot): ViewRegistry =
     // If the ambient is false, this is the first time this function has appeared in the composition
     // so far. We provide a true value for the ambient for everything below us, so any recursive
     // calls to this function will hit the if case above and not re-apply the wrapper.
-    Providers(HasViewFactoryRootBeenApplied provides true) {
-      val safeRoot: CompositionRoot = remember(root) { safeCompositionRoot(root) }
+    CompositionLocalProvider(HasViewFactoryRootBeenApplied provides true) {
+      val safeRoot: CompositionRoot = remember(root as Any?) { safeCompositionRoot(root) }
```
zach-klippenstein, line 90: Why the cast?

compose/core-compose/src/main/java/com/squareup/workflow/ui/compose/RenderAsState.kt
```diff
@@ -220,11 +220,13 @@ private class WorkflowState<PropsT, OutputT : Any, RenderingT>(
...
 
-  override fun onEnter() {}
+  override fun onRemembered() {}
 
-  override fun onLeave() {
+  override fun onForgotten() {
     workflowScope.cancel()
   }
+
+  override fun onAbandoned() {}
```
zach-klippenstein, line 229: Should also cancel the scope here, since this means the composition was unsuccessful. This was actually a latent bug (#297), so I'm really excited that this new API allows us to fix it. 
```suggestion
  override fun onAbandoned() {
    workflowScope.cancel()
  }
```

Then add "Fixes #297" to the commit message please.

compose/core-compose/src/main/java/com/squareup/workflow/ui/compose/ComposeViewFactory.kt
```diff
@@ -126,11 +125,11 @@ internal class ComposeViewFactory<RenderingT : Any>(
...
       // This lambda will be executed synchronously before bindShowRendering returns.
       composeContainer.bindShowRendering(
-        initialRendering,
-        initialViewEnvironment
+          initialRendering,
+          initialViewEnvironment
       ) { rendering, environment ->
         // Entry point to the world of Compose.
-        composeContainer.setContent(parentComposition) {
+        composeContainer.setContent {
```
zach-klippenstein, line 132: Could you add "Fixes #300" to the commit message too?

compose/core-compose/src/main/java/com/squareup/workflow/ui/compose/internal/ViewFactories.kt
```diff
@@ -92,7 +91,7 @@ import kotlin.properties.Delegates.observable
   // Plumb the current composition through the ViewEnvironment so any nested composable factories
   // get access to any ambients currently in effect.
   val parentComposition = remember { ParentComposition() }
-  parentComposition.reference = compositionReference()
+  parentComposition.reference = rememberCompositionContext()
```
zach-klippenstein, line 94: This line, and the `reference` property, and all the manual wiring up of parent compositions, can go away because of #300. You already deleted the consuming side above, and this can be deleted as well.

compose/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.7-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.7.1-all.zip
```
zach-klippenstein, line 3: While you're here, might as well bump this to 6.8.2 (i think the latest version of AGP requires that anyway).

zach-klippenstein (approved):
Wow, I can't believe this was so small… I expected there to be a lot more changes. Thanks for doing this!

compose/core-compose/src/main/java/com/squareup/workflow/ui/compose/CompositionRoot.kt
SalvatoreT: Screen Shot 2021-03-13 at 12 58 29 PM

SalvatoreT:
@zach-klippenstein, I have everything compiling! I need a help on a block or two, but I _think_ we're pretty close there. If you start the example, however, you'll get the following error.

```
E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.squareup.sample, PID: 28382
    java.lang.NullPointerException: Parameter specified as non-null is null: method kotlin.jvm.internal.Intrinsics.checkNotNullParameter, parameter rendering
        at com.squareup.workflow1.ui.compose.ViewEnvironmentsKt.WorkflowRendering(Unknown Source:5)
        at com.squareup.workflow1.ui.compose.WorkflowContainerKt.WorkflowContainer(WorkflowContainer.kt:67)
        at com.squareup.sample.hellocompose.ComposableSingletons$AppKt$lambda-1$1.invoke(App.kt:58)
        at com.squareup.sample.hellocompose.ComposableSingletons$AppKt$lambda-1$1.invoke(App.kt:38)
        at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(ComposableLambda.kt:121)
        at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(ComposableLambda.kt:46)
        at androidx.compose.runtime.CompositionLocalKt.CompositionLocalProvider(CompositionLocal.kt:193)
        at androidx.compose.material.TextKt.ProvideTextStyle(Text.kt:246)
        ...
```

compose/core-compose/src/main/java/com/squareup/workflow/ui/compose/CompositionRoot.kt
zach-klippenstein: Ah yes, this is such a weird bug.

compose/core-compose/src/main/java/com/squareup/workflow/ui/compose/RenderAsState.kt
```diff
@@ -184,45 +182,65 @@ private class WorkflowState<PropsT, OutputT : Any, RenderingT>(
...
+  override fun onRemembered() {}
+
+  override fun onForgotten() {
+    workflowScope.cancel()
+  }
 
-  override fun onLeave() {
+  override fun onAbandoned() {
     workflowScope.cancel()
   }
```
zach-klippenstein, line 245: Minor nit: I like to keep `onAbandoned` above `onRemembered` and `onForgotten` to hint that it happens earlier in the lifecycle.

compose/core-compose/src/main/java/com/squareup/workflow/ui/compose/RenderAsState.kt
```diff
@@ -184,45 +182,65 @@ private class WorkflowState<PropsT, OutputT : Any, RenderingT>(
...
+        initialSnapshot = snapshotState.value?.bytes?.let { TreeSnapshot.parse(it) },
+        onOutput = { exit(it) }
+    )
+        .map { it.rendering }
+        .produceIn(workflowScope)
+
+    workflowScope.launch {
+      propsFlow.collect {
+        // TODO ask Zach
+      }
```
zach-klippenstein, line 219: Why do you think you need to collect the props flow here? It's already been passed to the runtime above.

compose/core-compose/src/main/java/com/squareup/workflow/ui/compose/RenderAsState.kt
```diff
@@ -184,45 +182,65 @@ private class WorkflowState<PropsT, OutputT : Any, RenderingT>(
...
+    }
+
+    val renderings = renderWorkflowIn(
+        workflow = workflow,
+        scope = workflowScope,
+        props = propsFlow,
+        initialSnapshot = snapshotState.value?.bytes?.let { TreeSnapshot.parse(it) },
+        onOutput = { exit(it) }
+    )
+        .map { it.rendering }
```
zach-klippenstein, line 213: Need to save the snapshots so we can restore too.

compose/core-compose/src/main/java/com/squareup/workflow/ui/compose/internal/ViewFactories.kt
```diff
@@ -84,17 +85,14 @@ import kotlin.properties.Delegates.observable
...
   viewEnvironment: ViewEnvironment
 ) {
-  // Plumb the current composition through the ViewEnvironment so any nested composable factories
-  // get access to any ambients currently in effect.
-  val parentComposition = remember { ParentComposition() }
-  parentComposition.reference = compositionReference()
   val wrappedEnvironment = remember(viewEnvironment) {
-    viewEnvironment + (ParentComposition to parentComposition)
+    viewEnvironment
   }
```
zach-klippenstein, line 96: This is a noop – just a remember that remembers itself. Delete it.

compose/core-compose/src/main/java/com/squareup/workflow/ui/compose/RenderAsState.kt
```diff
@@ -184,45 +182,65 @@ private class WorkflowState<PropsT, OutputT : Any, RenderingT>(
...
-        .launchIn(this)
+    val exitOutput = CompletableDeferred<OutputT>()
+    fun exit(code: OutputT) {
+      // If we don't cancel the workflow runtime explicitly, coroutineScope will hang waiting for it to
+      // finish.
+      workflowScope.coroutineContext.cancelChildren(
+          CancellationException("WorkflowState completed with exit code $exitOutput")
+      )
+      exitOutput.complete(code)
+    }
```
zach-klippenstein, line 204: The workflow runtime doesn't have the concept of "exiting". It runs until it is cancelled.

compose/core-compose/src/main/java/com/squareup/workflow/ui/compose/RenderAsState.kt
```diff
@@ -184,45 +182,65 @@ private class WorkflowState<PropsT, OutputT : Any, RenderingT>(
...
+
+    val renderings = renderWorkflowIn(
+        workflow = workflow,
+        scope = workflowScope,
+        props = propsFlow,
+        initialSnapshot = snapshotState.value?.bytes?.let { TreeSnapshot.parse(it) },
+        onOutput = { exit(it) }
+    )
+        .map { it.rendering }
+        .produceIn(workflowScope)
```
zach-klippenstein, line 214: This should just be collected to the `renderingState` object – e.g. `.onEach { renderingState.value = it }.launchIn(workflowScope)`. As far as I can see, nothing is ever setting `renderingState.value`, which is probably why you're getting an NPE.

zach-klippenstein (changes requested):
Thanks for addressing my previous comments, but it looks like there are now a lot more invasive changes to `RenderAsState` that significantly change behavior and break things.

SalvatoreT:
@zach-klippenstein, all of the examples work except for the Nested Renderings and five of fifty tests.

I feel like we're _so close_!

Screenshot_1615871490

compose/core-compose/src/main/java/com/squareup/workflow/ui/compose/RenderAsState.kt
```diff
@@ -68,9 +67,8 @@ import okio.ByteString
 @Composable
 fun <PropsT, OutputT : Any, RenderingT> Workflow<PropsT, OutputT, RenderingT>.renderAsState(
   props: PropsT,
-  onOutput: (OutputT) -> Unit,
-  diagnosticListener: WorkflowDiagnosticListener? = null
```
zach-klippenstein, line 72: This should take a list of interceptors instead, but I can follow up.

compose/core-compose/src/main/java/com/squareup/workflow/ui/compose/RenderAsState.kt
```diff
@@ -155,22 +150,25 @@ inline fun <RenderingT> Workflow<Unit, Nothing, RenderingT>.renderAsState(
...
   // TODO Pass Dispatchers.Main.immediate and merge two scope vals when this bug is fixed:
   //  https://issuetracker.google.com/issues/165674304
   val baseScope = rememberCoroutineScope()
   val workflowScope = remember { baseScope + Dispatchers.Main.immediate }
-  val snapshotState = savedInstanceState(key = snapshotKey, saver = SnapshotSaver) { null }
+  val snapshotState =
+    rememberSaveable(
+        key = snapshotKey,
+        stateSaver = SnapshotSaver
+    ) { mutableStateOf(SnapshotSaver.restore(byteArrayOf())) }
```
zach-klippenstein, line 163: We shouldn't need to manually call any of the saver APIs.

compose/core-compose/src/main/java/com/squareup/workflow/ui/compose/RenderAsState.kt
```diff
@@ -183,46 +181,48 @@ private class WorkflowState<PropsT, OutputT : Any, RenderingT>(
...
+        workflow = workflow,
+        scope = workflowScope,
+        props = propsFlow,
+        initialSnapshot = TreeSnapshot.forRootOnly(snapshotState.value),
+        onOutput = { outputRef.value!!.invoke(it) }
+    )
+
+    renderings
+        .map { it.snapshot }
+        .onEach { snapshotState.value = SnapshotSaver.restore(it.toByteString().toByteArray()) }
```
zach-klippenstein, line 206: We should just be storing the snapshot itself here.

compose/core-compose/src/main/java/com/squareup/workflow/ui/compose/ViewEnvironments.kt
```diff
@@ -60,7 +62,7 @@ import com.squareup.workflow.ui.compose.internal.WorkflowRendering
   val viewRegistry = remember(viewEnvironment) { viewEnvironment[ViewRegistry] }
   val renderingType = rendering::class
   val viewFactory = remember(viewRegistry, renderingType) {
-    viewRegistry.getFactoryFor(renderingType)
+    viewRegistry.getFactoryFor(renderingType)!!
```
zach-klippenstein, line 65: We should throw a more helpful error message if no factory is found.

compose/core-compose/src/main/java/com/squareup/workflow/ui/compose/internal/ParentComposition.kt
```diff
@@ -32,8 +33,9 @@ import com.squareup.workflow.ui.ViewEnvironmentKey
  * link its composition to the outer one.
  */
 internal class ParentComposition(
```
zach-klippenstein, line 34: This class can be deleted.

compose/core-compose/src/main/java/com/squareup/workflow/ui/compose/internal/ViewRegistries.kt
```diff
@@ -31,7 +33,7 @@ internal fun ViewRegistry.mapFactories(
   override fun <RenderingT : Any> getFactoryFor(
     renderingType: KClass<out RenderingT>
   ): ViewFactory<RenderingT> {
-    val transformedFactory = transform(this@mapFactories.getFactoryFor(renderingType))
+    val transformedFactory = transform(this@mapFactories.getFactoryFor(renderingType)!!)
```
zach-klippenstein, line 36: Better message needed

compose/samples/src/main/java/com/squareup/sample/nestedrenderings/NestedRenderingsActivity.kt
```diff
@@ -17,31 +17,37 @@ package com.squareup.sample.nestedrenderings
...
 )
 
-private val viewEnvironment = ViewEnvironment(viewRegistry).withCompositionRoot { content ->
-  Providers(BackgroundColorAmbient provides Color.Green, children = content)
-}
+@OptIn(WorkflowUiExperimentalApi::class)
+private val viewEnvironment =
+  ViewEnvironment(mapOf(ViewRegistry to viewRegistry)).withCompositionRoot { content ->
+    CompositionLocalProvider(BackgroundColorAmbient provides Color.Green) {
+      // TODO talk to Zach about the right way to do this
```
zach-klippenstein, line 39: Just call the function I think

compose/samples/src/main/java/com/squareup/sample/nestedrenderings/RecursiveViewFactory.kt
```diff
@@ -82,9 +81,10 @@ val RecursiveViewFactory = composedViewFactory<Rendering> { rendering, viewEnvir
   }
 }
 
+@OptIn(WorkflowUiExperimentalApi::class)
 @Preview
 @Composable fun RecursiveViewFactoryPreview() {
-  Providers(BackgroundColorAmbient provides Color.Green) {
+  CompositionLocalProvider(BackgroundColorAmbient provides Color.Green) {
```
zach-klippenstein, line 87: Should rename this ambient

zach-klippenstein (changes requested):

zach-klippenstein (approved):

zach-klippenstein:
Looks like we forgot to suppress some deprecations.

PR #343, opened 2021-03-03
Makes the names logged by TracingWorkflowInterceptor human readable.

rjrjr:
Also:

 * Eliminates accidental dependency on kotlin-reflect

 * Drops use of SimpleLoggingWorkflowInterceptor in the samples,
   we don't want it cargo culted

Screen Shot 2021-03-02 at 4 54 09 PM


rjrjr:
It would be wise to ignore whitespace changes while reviewing this.

zach-klippenstein:
Unfortunately that is not something the mobile app seems to know how to do 😒

workflow-tracing/src/main/java/com/squareup/workflow1/diagnostic/tracing/TracingWorkflowInterceptor.kt
```diff
@@ -481,15 +484,46 @@ public class TracingWorkflowInterceptor internal constructor(
...
+
+  val params = arguments.map { projection ->
+    when (val type = projection.type) {
+      is KType -> type.toLoggingName()
+      else -> "*"
+    }
+  }
+
+  return if (params.isEmpty()) classifierName else "$classifierName<${params.joinToString(", ")}>"
+}
```
zach-klippenstein, line 522: Who needs the reflection library when you can write it yourself? 😅

zach-klippenstein (approved):

workflow-tracing/src/main/java/com/squareup/workflow1/diagnostic/tracing/TracingWorkflowInterceptor.kt
rjrjr: I love extension methods so much. Nothing but syntactic sugar, but they make such a huge difference.

PR #352, opened 2021-03-04
Fix links in KDoc for BaseRenderContext.renderChild()

Armaxis:
I noticed that links in the documentation were pointing to old method name and didn't work. Fixed.﻿

### DIFF
--- a/workflow-core/src/main/java/com/squareup/workflow1/BaseRenderContext.kt
+++ b/workflow-core/src/main/java/com/squareup/workflow1/BaseRenderContext.kt
@@ -57,9 +57,9 @@ public interface BaseRenderContext<out PropsT, StateT, in OutputT> {
    * through this context method.**
    *
    * 1. If the child _wasn't_ already running, it will be started either from
-   *    [initialState][Workflow.initialState] or its snapshot.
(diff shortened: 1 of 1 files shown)


zach-klippenstein (approved):
Thanks for fixing this! Although I think the docs actually need a bigger change. These references are specific to `StatefulWorkflow`, but `BaseRenderContext` is used by `StatelessWorkflow` as well. So its docs should be more general. This is still better though, so I've filed #353 to follow up and fix that.

Armaxis:
Signed the CLA :)

zach-klippenstein:
```
com.squareup.workflow1.ui.tests.ViewTreeIntegrationTest > complex_modal_backstack_with_text_entry[test(AVD) - 10] FAILED 
Test failed to run to completion. Reason: 'Test run failed to complete. Expected 1 tests, received 0'. Check device logcat for details
Tests on test(AVD) - 10 failed: Test run failed to complete. Expected 1 tests, received 0. onError: commandError=false message=INSTRUMENTATION_ABORTED: System has crashed.
```

This failure is happening on 29 almost 100% of the time now. I just filed #355 to investigate. I don't think restarting it's gonna help. I will just merge.

PR #354, opened 2021-03-05
Revert "Introduce support for ViewTreeSavedStateRegistryOwner in Android container views."

zach-klippenstein:
This reverts commit 31f758d45d015bd312c361ff734e2d057869f51f.
We noticed a bunch of tests failing internally, so clearly this repo is missing some tests.


rjrjr (approved):

zach-klippenstein:
Lol, this also reverts the CI timeout and now the shards are timing out again. I am surprised – with the tests reverted, I don't know why they would take so long still. Plus, the emulator action recently started using 2 cores instead of 1 for the emulators, which should reduce times not increase them. Anyway, I'm pretty confident this doesn't actually break anything, we just might need to re-configure CI again after landing this. I'm going to go ahead and merge without babysitting CI for another day.

PR #364, opened 2021-03-08
Reverts introduction of Compose, ViewTreeLifecycleOwner.

rjrjr:
Drastic, but doing so requiress us to depend on some androidx RCs, and we just
can't do that. As collateral damage, we lose a lot of good work that was done
to add much improved testing of BackStackContainer et al. It will be back.

To accomplish this, I did something like the following:

```
git co -b ray/decompose origin/main
git co v1.0.0-alpha.9
git diff ray/decompose --binary > ~/diff.patch
git co ray/decompose
cat ~/diff.patch | git apply
```

And then:

* `git cherry-pick` each keeper SHA
* squash rebase on origin/main


zach-klippenstein:
Oh crap, this reverts the Compose repo import too - i bet that's gross, because I imported the whole history there. Mind if I give it a try just reverting individual commits?

rjrjr:
> Oh crap, this reverts the Compose repo import too - i bet that's gross, because I imported the whole history there. Mind if I give it a try just reverting individual commits?

That's fine with me, and thanks.

But I don't think what we have here is so bad in terms of the history, really. Your branch will start w/a quick revert of this one, and we won't lose any historical info once yours returns to `main`.

zach-klippenstein:
I give up. Shippit.

zach-klippenstein (approved):

PR #376, opened 2021-03-17
Upgrade Compose project's AGP to latest alpha.

zach-klippenstein:
This requires using the latest canary of Arctic Fox, which has all the latest compose tooling features.


zach-klippenstein:
Only CI failure was a timeout in a UI test shard for the main project, not compose. I'll raise the timeouts, but that isn't a blocking failure for this PR.

rjrjr (approved):

PR #377, opened 2021-03-17
Fix some warnings and comment issues in compose project (multiple commits)

zach-klippenstein:


zach-klippenstein:
Timeouts on non-compose shards, this PR only affects the compose project. Timeouts fixed in #378. Merge away.

rjrjr (approved):

PR #382, opened 2021-03-19
Deletes ViewRegistry(vararg registries: ViewRegistry).

rjrjr:
By having two vararg factory functions, we made it impossible to use the
spread operator with either, because these are ambiguous:

```
val registries = listOf(ViewRegistry(), ViewRegistry())
val reg = ViewRegistry(*registries)
```

```
val factories: List<ViewFactory<*>> = ...
val reg = ViewRegistry(*factories)
```

And note that there are no uses of the deleted variant across all the samples.


0xMatthewGroves (approved):
💯

rjrjr:
Oooh, this works:

```kotlin
ViewRegistry(*it.toTypedArray())
```

I'm still inclined to delete, though. It's just redundant. @zach-klippenstein , WDYT?

zach-klippenstein:
Gross. Fewer functions is better.

PR #383, opened 2021-03-22
Update docs on BuilderViewFactory to reflect name change

seanfreiburg:
Update docs on BuilderViewFactory to reflect name change. It was incorrectly `BuilderBinding`


rjrjr (commented):
Much obliged.

PR #385, opened 2021-03-23
Start publishing a Gradle platform POM.

### LINKED ISSUE #228: Publish BOM/Gradle platform artifact
https://docs.gradle.org/current/userguide/java_platform_plugin.html

This would allow consumers to not have to specify the version for every single workflow artifact by just depending on the platform artifact.

zach-klippenstein:
See https://docs.gradle.org/current/userguide/platforms.html.

Closes #228.

### DIFF
--- a/platform/README.md
+++ b/platform/README.md
@@ -1,0 +1,7 @@
+# platform
+
+This module does not contain any source code – it simply defines a "platform" that consumers can
+depend upon to avoid having to specify version numbers for every individual workflow library they
+use.
+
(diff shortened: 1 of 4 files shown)


steve-the-edwards (approved):
I'm presuming we will include core dependencies such as coroutines etc?

rjrjr (approved):

zach-klippenstein:
I'm not sure that's something we should do. The Spring example in the gradle docs suggests it might be appropriate, but I'm really not interested in solving that use case. I just want to make it easier for consumers who have to depend on our 42 different artifacts. I could probably be convinced though.

PR #386, opened 2021-03-23
Drop the final dot in alpha versions.

rjrjr:
Gradle and Android studio are stupid about semver, and keep insisting
that, say, `1.0.0-alpha.12` < `1.0.0-alpha.9`. To cope with that we drop
the last dot, so our next release will be `1.0.0-alpha13`.


zach-klippenstein:
We should file a Gradle bug about this, but if google couldn't get it fixed, I doubt we will.

zach-klippenstein (approved):

PR #387, opened 2021-03-23
WIP Migrate main project to use Gradle's new Version Catalogs.

zach-klippenstein:
See https://docs.gradle.org/nightly/userguide/platforms.html\#sub:central-declaration-of-dependencies.


zach-klippenstein:
This seems to work great except in the root build.gradle.kts, where it doesn't seem to work at all. I would be completely unsurprised, but very disappointed, if this was not a supported use case.

zach-klippenstein:
Soft-blocked by https://github.com/gradle/gradle/issues/16958

RBusarow:
superseded by #648

PR #389, opened 2021-03-24
Fix doc to be correct arg on expectWorker

seanfreiburg:
There are several ways to do it, but it just didn't match with changes made to the functions, updated doc to be more accurate


steve-the-edwards (approved):
Thanks for the catch!

PR #390, opened 2021-03-26
Add updates and clarification to Tutorial 4

tir38:


samples/tutorial/Tutorial4.md
```diff
@@ -8,9 +8,6 @@ To follow this tutorial, launch Android Studio and open this folder (`samples/tu
 
 Start from the implementation of `tutorial-3-complete` if you're skipping ahead.
 
-## Adding new todo items
-
-A gap in the usability of the todo app is that it does not let the user create new todo items. We will add an "add" button on the right side of the navigation bar for this.
```
tir38, line 13: We never actually added UI elements or wired up "new TODO" through the Screen / LayoutRunner.  Instead of adding those steps and further increasing this Tutorial I just removed them.

samples/tutorial/Tutorial4.md
```diff
@@ -75,10 +72,10 @@ object TodoWorkflow : StatefulWorkflow<TodoProps, State, Back, List<Any>>() {
 }
 ```
 
-Define the output events from the `TodoListWorkflow` to describe the `new` item action and selecting a todo item, as well as removing the todo list from the `State`:
+Define the output events from the `TodoListWorkflow` to include back and a new `SelectTodo` output. Also, We no longer need to maintain the todo list in the `State` and we can remove it:
 
 ```kotlin
-object TodoListWorkflow : StatefulWorkflow<ListProps, Unit, Output, TodoListScreen>() {
+object TodoListWorkflow : StatefulWorkflow<ListProps, State, Output, TodoListScreen>() {
```
tir38, line 78: We convert this State -> Unit later

samples/tutorial/Tutorial4.md
```diff
@@ -112,12 +108,10 @@ object TodoListWorkflow : StatefulWorkflow<ListProps, Unit, Output, TodoListScre
 }
 ```
 
-Because the `State` has no properties anymore, it can't be a data class, so we need to change it to an object. Since the only reason to have a custom type for state is to define the data we want to store, we don't need a custom type anymore so we can just use `Unit`.
```
tir38, line 115: Moved down a step

samples/tutorial/Tutorial4.md
```diff
@@ -130,14 +124,57 @@ object TodoListWorkflow : StatefulWorkflow<ListProps, Unit, Output, TodoListScre
     // Tell our parent that a todo item was selected.
     setOutput(SelectTodo(index))
   }
+}
+```
+
+Move the editing actions from the `TodoListWorkflow` to the `TodoWorkflow`. We won't be able to call these methods until we can respond to output from the `TodoEditWorkflow`, but doing this now helps clean up the `TodoListWorkflow`:
```
tir38, line 130: We need to move these now or app won't compile

samples/tutorial/Tutorial4.md
```diff
@@ -224,48 +260,23 @@ object TodoWorkflow : StatefulWorkflow<TodoProps, State, Back, List<Any>>() {
...
-      login(output.username)
-    }
-    backstackScreens += welcomeScreen
-
-    when (state) {
-      // When the state is Welcome, defer to the WelcomeWorkflow.
-      is Welcome -> {
-        // We always add the welcome screen to the backstack, so this is a no op.
-      }
+      // …
```
tir38, line 279: I replaced this unchanged code with `//...` to draw more attention to the code that is updated.

samples/tutorial/Tutorial4.md
```diff
@@ -277,45 +288,17 @@ object RootWorkflow : StatefulWorkflow<Unit, State, Nothing, BackStackScreen<*>>
...
   // …
-
 }
 ```
 
 #### Moving Edit Output handling to the TodoWorkflow
 
-The `TodoWorkflow` now can handle the outputs from the `TodoListWorkflow`. Next, let's add handling for the `TodoEditWorkflow` output events.
-
-Since the types of output and actions are pretty different from their origin, make a *second* set of actions on the `TodoWorkflow`:
```
tir38, line 293: (see above) we need to move these to sooner in the tutorial so app compiles.

steve-the-edwards (approved):

PR #362, opened 2021-03-07
Update templates

tir38:
IMO these templates would be more helpful if they let users set types in the creation dialog, not with a later refactor.

We don't always want state/props/rendering/output to have the same name as the workflow. Nor
do we always want to create those states/props/renderings/output. Sometimes those classes already
exist (e.g. Unit, Nothing, Any).

Stateful and Stateless Workflow templates now take in *optional* props/state/rendering/output. If these are left blank they will be auto created

Test
----

1. In Android Studio -> File -> New -> Edit File Templates....
2. create a new template or update existing and copy/paste code from these .kt files
3. File -> New -> select a template
4. see generated file

### optionals left blank

Screen Shot 2021-03-11 at 7 07 34 PM

generates:

```kotlin
package workflow.tutorial

import com.squareup.workflow1.Snapshot
import com.squareup.workflow1.StatefulWorkflow

import workflow.tutorial.MyFancyWorkflow.MyFancyProps
import workflow.tutorial.MyFancyWorkflow.MyFancyState
import workflow.tutorial.MyFancyWorkflow.MyFancyOutput
import workflow.tutorial.MyFancyWorkflow.MyFancyRendering

object MyFancyWorkflow : StatefulWorkflow<MyFancyProps, MyFancyState, MyFancyOutput, MyFancyRendering>() {

  data class MyFancyProps(
    // TODO add args
  )

  data class MyFancyState(
    // TODO add args
  )

  data class MyFancyOutput(
    // TODO add args
  )

  data class MyFancyRendering(
    // TODO add args
  )

  override fun initialState(
    props: MyFancyProps,
    snapshot: Snapshot?
  ): MyFancyState = TODO("Initialize state")

  override fun render(
    props: MyFancyProps,
    state: MyFancyState,
    context: RenderContext
  ): MyFancyRendering {
    TODO("Render")
  }

  override fun snapshotState(state: MyFancyState): Snapshot? = Snapshot.write {
    TODO("Save state")
  }
}
```

### optionals supplied

Screen Shot 2021-03-11 at 7 09 03 PM

generates:

```kotlin
package workflow.tutorial

import com.squareup.workflow1.Snapshot
import com.squareup.workflow1.StatefulWorkflow

object MyFancyWorkflow : StatefulWorkflow<ExistingProps, ExistingState, Unit, ExistingScreen>() {

  override fun initialState(
    props: ExistingProps,
    snapshot: Snapshot?
  ): ExistingState = TODO("Initialize state")

  override fun render(
[...]

PR #393, opened 2021-03-27
Update Tutorial1 after changes to templates

tir38:
My initial changes to Tutorial 1 didn't include any changes to the templates  in #362 . Now that #362 is merged I need to go back and update the Tutorial to reference this new template.


steve-the-edwards (approved):

tir38:
@steve-the-edwards i think you have to merge this. I don't have write acess

PR #403, opened 2021-04-07
Tweaks to Compatible API, DecorativeViewFactory kdoc.

### LINKED ISSUE #388: Make the DecorativeViewFactory include Compatible
When a rendering wraps another one, it usually needs to implement `Compatible` and delegate to thing the it wraps. None of the kdoc examples show this.

rjrjr:
Moves Named.compatibilityKey to Compatible.compatibilityKey, which should be a
bit less confusing.

Updates DecorativeViewFactory kdoc to call out its use -- can't make a useful
wrapper rendering without it.

Closes #388

### DIFF
--- a/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/DecorativeViewFactory.kt
+++ b/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/DecorativeViewFactory.kt
@@ -11,13 +11,19 @@ import kotlin.reflect.KClass
  * instances of [InnerT] to add information or behavior, without requiring wasteful wrapping
  * in the view system.
  *
+ * One general note: when creating a wrapper rendering, you're very likely to want it
+ * to implement [Compatible], to ensure that checks made to update or replace a view
+ * are based on the wrapped item. Each example below illustrates this.
+ *
  * ## Examples
  *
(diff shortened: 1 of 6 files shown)


workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/DecorativeViewFactory.kt
```diff
@@ -11,13 +11,19 @@ import kotlin.reflect.KClass
  * instances of [InnerT] to add information or behavior, without requiring wasteful wrapping
  * in the view system.
  *
+ * One general note: when creating a wrapper rendering, you're very likely to want it
+ * to implement [Compatible], to ensure that checks made to update or replace a view
```
steve-the-edwards, line 15:
```suggestion
 * to implement [Compatible] to ensure that checks made to update or replace a view
```

steve-the-edwards (approved):

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/DecorativeViewFactory.kt
rjrjr: Disagree, courteously.

PR #408, opened 2021-04-16
bindShowRendering no longer calls showRendering.

### LINKED ISSUE #397: Redundant calls to View.showRendering
`DecorativeViewFactory` makes redundant calls to `showRendering`, and these can compound. I think the proper fix is to factor a `rebindShowRendering` call out of `bindShowRendering` that replaces the function but doesn't call it.

rjrjr:
That was always a weird coupling, and it caused the nasty compounding double
update behavior in `DecorativeViewFactory`.

To prevent `ViewFactory` implementations from having to remember to call
`showRendering`, we make `ViewRegistry.buildView` do so. And to allow more
sophisticiated `ViewFactory` implementations to prevent that call from
happening prematurely, we build it into the default value for a new
`initializeView` method on `ViewRegistry.buildView`.

`DecorativeViewFactory` takes advantage by passing a no-op function in for
`initializeView` when the view is built, and then making its own explicit call
to `showRendering` after it's done playing wrapping games.

Closes #397

### DIFF
--- a/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewRegistry.kt
+++ b/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewRegistry.kt
@@ -98,12 +98,43 @@ public fun ViewRegistry(vararg bindings: ViewFactory<*>): ViewRegistry =
 public fun ViewRegistry(): ViewRegistry = TypedViewRegistry()
 
 /**
- * It is usually more convenient to use [WorkflowViewStub] than to call this method directly.
+ * It is usually more convenient to use [WorkflowViewStub] or [DecorativeViewFactory]
+ * than to call this method directly.
  *
- * Creates a [View] to display [initialRendering], which can be updated via calls
- * to [View.showRendering]. Prefers entries found via [ViewRegistry.getFactoryFor].
- * If that returns null, falls back to the factory provided by the rendering's
- * implementation of [AndroidViewRendering.viewFactory], if there is one.
+ * Returns the [ViewFactory] that builds [View] instances suitable to display the given [rendering],
+ * via subsequent calls to [View.showRendering].
+ *
+ * Prefers factories found via [ViewRegistry.getFactoryFor]. If that returns null, falls
+ * back to the factory provided by the rendering's implementation of
+ * [AndroidViewRendering.viewFactory], if there is one. Note that this means that a
+ * compile time [AndroidViewRendering.viewFactory] binding can be overridden at runtime.
+ *
+ * @throws IllegalArgumentException if no factory can be find for type [RenderingT]
+ */
+@WorkflowUiExperimentalApi
+public fun <RenderingT : Any>
+  ViewRegistry.getFactoryForRendering(rendering: RenderingT): ViewFactory<RenderingT> {
+  @Suppress("UNCHECKED_CAST")
+  return getFactoryFor(rendering::class)
+    ?: (rendering as? AndroidViewRendering<*>)?.viewFactory as? ViewFactory<RenderingT>
+    ?: throw IllegalArgumentException(
+      "A ${ViewFactory::class.qualifiedName} should have been registered to display " +
+        "${rendering::class.qualifiedName} instances, or that class should implement " +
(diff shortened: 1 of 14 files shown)


workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewRegistry.kt
```diff
@@ -155,8 +143,23 @@ public fun <RenderingT : Any> ViewRegistry.buildView(
 public fun <RenderingT : Any> ViewRegistry.buildView(
   initialRendering: RenderingT,
   initialViewEnvironment: ViewEnvironment,
-  container: ViewGroup
-): View = buildView(initialRendering, initialViewEnvironment, container.context, container)
+  contextForNewView: Context,
+  container: ViewGroup? = null,
+  initializeView: ViewShowRendering<RenderingT> = { rendering, env ->
+    showRendering(rendering, env)
+  }
```
rjrjr, line 150:
```suggestion
  initializeView: View.() -> Unit = {
    val r = getRendering<RenderingT>()
    val e = getEnvironment()!!
    showRendering(r, e)
  }
```

zach-klippenstein:
> Made it that much easier for ViewRegistry.buildView
to invoke it

Minor implementation inconvenience isn't a great reason to make API changes.

> I suspect it will come in handy in other places too.

Maybe, but I'd rather keep it locked down until we prove that, so we don't have to deprecated it later.

samples/containers/android/src/main/java/com/squareup/sample/container/panel/PanelContainer.kt
```diff
@@ -59,13 +59,15 @@ class PanelContainer @JvmOverloads constructor(
...
-        PanelContainer(contextForNewView).apply {
-          id = R.id.panel_container
-          layoutParams = ViewGroup.LayoutParams(MATCH_PARENT, MATCH_PARENT)
-          bindShowRendering(initialRendering, initialEnv, ::update)
+    type = PanelContainerScreen::class,
+    viewConstructor = { initialRendering, initialEnv, contextForNewView, _ ->
+      PanelContainer(contextForNewView).apply {
+        id = R.id.panel_container
+        layoutParams = ViewGroup.LayoutParams(MATCH_PARENT, MATCH_PARENT)
+        bindShowRendering(initialRendering, initialEnv) { rendering, env ->
```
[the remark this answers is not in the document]
rjrjr: Guess I'll be restoring them :D

workflow-ui/modal-android/src/main/java/com/squareup/workflow1/ui/modal/ModalViewContainer.kt
```diff
@@ -62,7 +62,12 @@ public open class ModalViewContainer @JvmOverloads constructor(
     initialViewEnvironment: ViewEnvironment
   ): DialogRef<Any> {
     val view = initialViewEnvironment[ViewRegistry]
-        .buildView(initialModalRendering, initialViewEnvironment, this)
+        .buildView(
```
rjrjr, line 65: This change and a couple of others like it are because I deleted a convenience overload of `buildView` that caused nothing but trouble, and which in practice no one but the library will ever use.

rjrjr:
> Haven't started reading the actual PR yet, but just some comments from the description:
> 
> > Made it that much easier for ViewRegistry.buildView
> > to invoke it
> 
> Minor implementation inconvenience isn't a great reason to make API changes.
> 
> > I suspect it will come in handy in other places too.
> 
> Maybe, but I'd rather keep it locked down until we prove that, so we don't have to deprecated it later. I don't know if we will, either, because `showRendering` typically needs to have either a more concretely-typed view, or is using cached `findViewById` results from one.

Yeah, it was a bad call. Turns out I didn't even use it. PR is smaller now.

workflow-ui/modal-android/src/main/java/com/squareup/workflow1/ui/modal/ModalViewContainer.kt
zach-klippenstein: Yes! Figured that's what was going on, and I've wanted to do that in the past too. Thanks!

zach-klippenstein (approved):

PR #409, opened 2021-04-16
Cleans up view initialization in DecorativeViewFactory.

### LINKED ISSUE #396: DecorativeViewFactory signature is weird
You're forced to provide the `map` function, but if you provide your own `doShowRendering` it's not used.

rjrjr:
```kotlin
initView: (OuterRenderingT, ViewEnvironment) -> Unit
```

is now

```kotlin
initializeView: View.() -> Unit
```

just like the param on `ViewRegistry.buildView`. We also eliminate
the hard coded call to `showRendering()`, giving authors complete control
over when / if it's called. Again, this is in parallel with `ViewRegistry`'s
behavior.

`View.showFirstRendering()` is introduced, to simplify the 99% case.

Closes #396

### DIFF
--- a/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/DecorativeViewFactory.kt
+++ b/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/DecorativeViewFactory.kt
@@ -32,7 +32,7 @@ import kotlin.reflect.KClass
  *
  * To make a decorator type that adds information to the [ViewEnvironment]:
  *
- *    class NeutronFlowPolarity(val reversed) {
+ *    class NeutronFlowPolarity(val reversed: Boolean) {
  *      companion object : ViewEnvironmentKey<NeutronFlowPolarity>(NeutronFlowPolarity::class) {
  *        override val default: NeutronFlowPolarity = NeutronFlowPolarity(reversed = false)
  *      }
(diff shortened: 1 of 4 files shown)


rjrjr:
Flurry of updates tidied kdoc and inline comments. I'm done now.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/DecorativeViewFactory.kt
```diff
@@ -99,18 +102,20 @@ import kotlin.reflect.KClass
...
  * [InnerT], allowing pre- and post-processing. Default implementation simply
- * applies [map] and makes the function call.
+ * uses [map] to extract the [InnerT] instance from [OuterT] and makes the function call.
  */
 @WorkflowUiExperimentalApi
 public class DecorativeViewFactory<OuterT : Any, InnerT : Any>(
   override val type: KClass<OuterT>,
   private val map: (OuterT, ViewEnvironment) -> Pair<InnerT, ViewEnvironment>,
-  private val initView: (OuterT, View) -> Unit = { _, _ -> },
+  private val initializeView: View.() -> Unit = { showFirstRendering<OuterT>() },
```
steve-the-edwards, line 118: Extension function type! 🤯  so cool.

steve-the-edwards (approved):

zach-klippenstein (approved):
Nifty!

0xMatthewGroves (approved):

PR #417, opened 2021-05-05
Kotlin 1.5.0

vRallev:
You might want to wait until we're ready internally, though.


trace-encoder/api/trace-encoder.api
```diff
@@ -1,9 +1,79 @@
+public final class com/squareup/tracing/ChromeTraceEvent {
```
zach-klippenstein, line 1: I'm a bit confused as to why these wouldn't have been here before… seems like they should have been?

zach-klippenstein (commented):
These abi changes are baffling to me, but LG.

zach-klippenstein (approved):

trace-encoder/api/trace-encoder.api
vRallev: I'd think so, too 🤷🏻

zach-klippenstein: Ok this wasn't here before because it is internal. Which means the fact that it's showing up now seems like a big.

workflow-core/api/workflow-core.api
```diff
@@ -192,8 +196,11 @@ public abstract interface class com/squareup/workflow1/Worker {
 }
 
 public final class com/squareup/workflow1/Worker$Companion {
+	public final synthetic fun create (Lkotlin/jvm/functions/Function2;)Lcom/squareup/workflow1/Worker;
 	public final fun createSideEffect (Lkotlin/jvm/functions/Function1;)Lcom/squareup/workflow1/Worker;
 	public final fun finished ()Lcom/squareup/workflow1/Worker;
+	public final synthetic fun from (Lkotlin/jvm/functions/Function1;)Lcom/squareup/workflow1/Worker;
+	public final synthetic fun fromNullable (Lkotlin/jvm/functions/Function1;)Lcom/squareup/workflow1/Worker;
```
zach-klippenstein, line 203: These three functions are inline. I wouldn't expect they'd show up here either, although that seems less incorrect than internal functions.

workflow-core/api/workflow-core.api
```diff
@@ -202,6 +209,21 @@ public final class com/squareup/workflow1/Worker$DefaultImpls {
 	public static fun doesSameWorkAs (Lcom/squareup/workflow1/Worker;Lcom/squareup/workflow1/Worker;)Z
 }
 
+public final class com/squareup/workflow1/WorkerWorkflow : com/squareup/workflow1/StatefulWorkflow, com/squareup/workflow1/ImpostorWorkflow {
```
zach-klippenstein, line 212: This is internal too 😒

zach-klippenstein (changes requested):

trace-encoder/api/trace-encoder.api
zach-klippenstein: I just noticed we're a few versions behind on the validator plugin. Updating it in #419, testing locally to see if that fixes it…

zach-klippenstein: Yep that fixes everything!

zach-klippenstein (approved):
Approving this again - please rebase onto #419 to get rid of all the incorrect api diffs.

zach-klippenstein:
I've rebased both the branch and the PR onto #419.

zach-klippenstein:
Anyone wanna try this out as a snapshot internally?

vRallev:
Internal build is good.

vRallev:
> I think we're good to merge this then.

Up to you, but I would wait until we rolled out 1.5 internally.

zach-klippenstein:
How come? We've verified that even if we merge this, cut a release, and bring the release in internally, it doesn't break ci so is very likely completely safe. Doesn't seem like holding off brings us much.

vRallev:
> We've verified that even if we merge this, cut a release, and bring the release in internally, it doesn't break ci so is very likely completely safe. Doesn't seem like holding off brings us much.
> 
> I'll let @rjrjr break the tie.

I did not test this release with Kotlin 1.4 internally, only with 1.5 that we plan to roll out next week.

PR #418, opened 2021-05-05
Upgrade the coroutines library

vRallev:


workflow-runtime/src/test/java/com/squareup/workflow1/internal/WorkflowRunnerTest.kt
```diff
@@ -230,7 +233,7 @@ internal class WorkflowRunnerTest {
     scope.cancel("foo")
 
     dispatcher.advanceUntilIdle()
-    assertTrue(output.isCancelled)
+    assertFalse(output.isActive)
```
vRallev, line 236: This PR is only a draft because of this change and the one above. There seems to be behavior change in the new coroutines library which seems quite scary. It seems wrong that these two tests started to fail.

workflow-runtime/src/test/java/com/squareup/workflow1/internal/WorkflowRunnerTest.kt
```diff
@@ -202,7 +205,7 @@ internal class WorkflowRunnerTest {
     dispatcher.advanceUntilIdle()
     assertTrue(output.isActive)
 
-    scope.cancel("foo")
+    output.cancel("foo")
```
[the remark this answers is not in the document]
vRallev: Yep, `output.isCancelled` isn't true anymore. Not sure if this is a bug in the coroutines library (that would be a fundamental problem and they should have caught that) or it's an issue with the test that made assumptions. I unfortunately don't know.

workflow-runtime/src/test/java/com/squareup/workflow1/internal/WorkflowRunnerTest.kt
vRallev: `output.isCanceled` is false even though the test calls `scope.cancel()`. Same as above, it's unclear to me why this changed.

samples/hello-terminal/terminal-workflow/src/main/java/com/squareup/sample/helloterminal/terminalworkflow/KeyStroke.kt
```diff
@@ -40,6 +41,7 @@ data class KeyStroke(
   }
 }
 
+@ObsoleteCoroutinesApi
```
steve-the-edwards, line 44: Haha `Obsolete` co-existing with `Experimental`. Life comes at you fast!

workflow-runtime/src/test/java/com/squareup/workflow1/internal/WorkflowRunnerTest.kt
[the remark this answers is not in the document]
steve-the-edwards: We need to stick with cancelling the scope here. I think the problem is that the `dispatcher.advanceUntilIdle()` will resume before the job is registered as cancelled. As it stands now this test is not really testing any of our code, just confirming the coroutines library function.

My guess is there is a change of behavior for when the jobs within a scope are *marked* as cancelled in the new coroutines library and how this interacts with when the underlying dispatcher is idle.

steve-the-edwards:
Realized my comments offer no help. Will try to find a time to pull this PR and play around with what needs to be done wrt the scope cancellation timing.

steve-the-edwards:
After spending the morning debugging through this I believe this is actually a bug with 1.5-rc01 - https://github.com/Kotlin/kotlinx.coroutines/issues/2698 

CC @zach-klippenstein  @vRallev

vRallev:
> After spending the morning debugging through this I believe this is actually a bug with 1.5-rc01 - [Kotlin/kotlinx.coroutines#2698](https://github.com/Kotlin/kotlinx.coroutines/issues/2698)
> 
> CC @zach-klippenstein @vRallev

Thanks for filing the ticket and digging into it!

steve-the-edwards:
> > After spending the morning debugging through this I believe this is actually a bug with 1.5-rc01 - [Kotlin/kotlinx.coroutines#2698](https://github.com/Kotlin/kotlinx.coroutines/issues/2698)
> > CC @zach-klippenstein @vRallev
> 
> Thanks for filing the ticket and digging into it!

The plot thickens. My previous reproducer had a bug and is not the difference between 1.4.2 and 1.5. Currently suspecting it has to do with the WorkflowRunner::nextOutput change to `onReceiveCatching` in this PR.

steve-the-edwards:
@zach-klippenstein @vRallev The 'bug' was a false alarm and actually expected behavior for that particular use of the testDispatcher when there is only a single suspend resumption in the coroutine under test.

The actual problem was not checking for a closed channel and throwing the exception in `WorkflowRunner::nextOutput`. All tests pass now and @zach-klippenstein https://github.com/square/workflow/issues/512 can be closed because `onReceiveCatching` is designed to enable this now.

steve-the-edwards:
Fixes #512

vRallev:
@steve-the-edwards So is this PR good to merge?

vRallev:
I tested this release internally and all is good 👍🏻

steve-the-edwards (approved):

steve-the-edwards:
Yes IMHO this is good to merge, but I was wanting @zach-klippenstein to do a quick review since he wrote the initial test and I didn't want to review my own change :)

vRallev:
I'll update the branch with the stable coroutines library on Monday.

samples/hello-terminal/terminal-workflow/src/main/java/com/squareup/sample/helloterminal/terminalworkflow/KeyStroke.kt
zach-klippenstein: Nit: this should be opting into this annotation, not using it directly.

workflow-testing/src/test/java/com/squareup/workflow1/WorkerStressTest.kt
```diff
@@ -66,6 +67,7 @@ class WorkerStressTest {
     }
   }
 
+  @ObsoleteCoroutinesApi
```
zach-klippenstein, line 70: Nit: Tests are never obsolete or experimental themselves, they should always opt in.

zach-klippenstein (approved):
Great detective work @steve-the-edwards!

PR #420, opened 2021-05-10
Upgrade gradle to 7.0.

zach-klippenstein:


zach-klippenstein:
Seems like we finally need to get off our super-old Dokka version to unblock this.

zach-klippenstein:
Dokka upgrade in #421 should unblock this.

zach-klippenstein:
Looks like we need to upgrade the ktlint plugin as well, latest version is 10.0.0.

zach-klippenstein:
And detekt, 1.16.0.

rjrjr:
Completely. I don't care even a little about detekt.

zach-klippenstein:
Removed detekt in #426, rebasing this on top of that…

patrickyin (approved):

zach-klippenstein:
Updated jmh (benchmarking) and ktlint gradle plugins to their latest versions, required for Gradle 7.

zach-klippenstein:
Added a separate commit that updates gradle and AGP for the compose project. I did both there because AGP was so far behind.

compose/build.gradle.kts
```diff
@@ -94,9 +94,20 @@ subprojects {
...
+        // https://github.com/pinterest/ktlint/issues/963
+        // https://github.com/pinterest/ktlint/issues/964
+        // https://github.com/pinterest/ktlint/issues/965
+        //
+        // We can't revert the KtLint version, because they only work with Kotlin 1.3 and would
+        // block Kotlin 1.4. We rather have a newer Kotlin version than a proper indent. The
+        // indent rule needs to be disabled globally due to another bug:
+        // https://github.com/pinterest/ktlint/issues/967
+        "indent",
+        "parameter-list-wrapping"
```
zach-klippenstein, line 110: I just copied this in from the main project's build.gradle.

compose/build.gradle.kts
```diff
@@ -94,9 +94,20 @@ subprojects {
         // IntelliJ refuses to sort imports correctly.
         // This is a known issue: https://github.com/pinterest/ktlint/issues/527
         "import-ordering",
```
maybeYouShouldBeTheLawyer, line 96: A heads-up just in case you're interested in a follow-on PR: it looks like the import-ordering issue [has been (mostly) fixed](https://youtrack.jetbrains.com/issue/KTIJ-4813#focus=Comments-27-4205728.0-0) thanks to the introduction of a [layout table for Kotlin imports](https://github.com/JetBrains/kotlin/pull/3336). (But note the misbehavior cited at the end of that conversation, including a comment by the tried-and-true Valeriy Ovechkin.)

compose/build.gradle.kts
maybeYouShouldBeTheLawyer: Seems we _should_ be able to re-enable `indent` and `parameter-list-wrapping` (assuming we're on a recent-enough version of ktlint and there aren't any new Kotlin-compatibility issues) 🤞. 

963 was allegedly fixed [here](https://github.com/pinterest/ktlint/pull/961); 964 [here](https://github.com/pinterest/ktlint/pull/1022) (transitively via 918); 965 [here](https://github.com/pinterest/ktlint/pull/972).

maybeYouShouldBeTheLawyer (approved):
A couple comments about the disabled ktlint rules. Otherwise lgtm.

compose/build.gradle.kts
zach-klippenstein: I'm not interesting in diverging these effectively duplicate configs. We can follow up and address in the main project once these have merged.

PR #422, opened 2021-05-11
Upgrade AGP to 4.2.0 stable.

zach-klippenstein:


zach-klippenstein:
Merging on green since this is a straightforward tooling update.

PR #425, opened 2021-05-14
424: Add Explicit Worker Tag for RenderTester

### LINKED ISSUE #424: Add option to explicitly require all Worker Expectations in RenderTester
This line - https://github.com/square/workflow-kotlin/blob/main/workflow-testing/src/main/java/com/squareup/workflow1/testing/RealRenderTester.kt#L123 - can be wrapped with a flag that could be toggled in RenderTester setup.

steve-the-edwards:
Closes #424 

Requested feature to be able to turn on explicit `Worker` expectation in Workflow1. This allows us to test whether a worker is *not run* for example.

### DIFF
--- a/workflow-testing/src/main/java/com/squareup/workflow1/testing/RenderTester.kt
+++ b/workflow-testing/src/main/java/com/squareup/workflow1/testing/RenderTester.kt
@@ -14,8 +14,8 @@ import kotlin.reflect.KType
 import kotlin.reflect.KTypeProjection
 
 @Deprecated(
-    "Renamed to testRender",
-    ReplaceWith("testRender(props)", "com.squareup.workflow1.testing.testRender")
+  "Renamed to testRender",
+  ReplaceWith("testRender(props)", "com.squareup.workflow1.testing.testRender")
 )
 @Suppress("NOTHING_TO_INLINE")
 public inline fun <PropsT, OutputT, RenderingT> Workflow<PropsT, OutputT, RenderingT>.renderTester(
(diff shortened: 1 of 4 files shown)


workflow-testing/src/main/java/com/squareup/workflow1/testing/RenderTester.kt
```diff
@@ -203,6 +206,8 @@ RealRenderTester(this, props, initialState)
  */
 public abstract class RenderTester<PropsT, StateT, OutputT, RenderingT> {
 
+  protected var explicitWorkerExpectationsRequired: Boolean = false
```
zach-klippenstein, line 209: There's no reason for this to exist in the base class, the only reason this isn't an interface is because we needed to add some inline methods or something. Please keep as much implementation and logic out of this class as possible.

workflow-testing/src/main/java/com/squareup/workflow1/testing/RenderTester.kt
```diff
@@ -262,6 +267,11 @@ public abstract class RenderTester<PropsT, StateT, OutputT, RenderingT> {
     block: (rendering: RenderingT) -> Unit = {}
   ): RenderTestResult<PropsT, StateT, OutputT>
 
+  public open fun requireExplicitWorkerExpectations():
```
zach-klippenstein, line 270: This should be abstract, like most of the methods in this class.

zach-klippenstein (approved):
LG once that property and method impl is moved to RealRenderTester.

workflow-testing/src/main/java/com/squareup/workflow1/testing/RenderTester.kt
steve-the-edwards: I was wondering about that. I thought since this *was* an abstract class I might as well share the implementation of this since it was simple and should be consistent.

workflow-testing/src/main/java/com/squareup/workflow1/testing/RenderTester.kt
steve-the-edwards: Like above, wasn't clear on intention.

PR #427, opened 2021-05-19
Upgrade compose to latest beta, 07 in compose project.

zach-klippenstein:
Also upgrades kotlin version in compose project to 1.4.32.

Based on #420.

### DIFF
--- a/compose/README.md
+++ b/compose/README.md
@@ -48,7 +48,6 @@ android {
     compose true
   }
   composeOptions {
-    kotlinCompilerVersion "1.4.0-dev-withExperimentalGoogleExtensions-20200720"
     kotlinCompilerExtensionVersion "${compose_version}"
   }
 }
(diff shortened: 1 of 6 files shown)


zach-klippenstein:
This seems to work great except in the root build.gradle.kts, where it doesn't seem to work at all. I would be completely unsurprised, but very disappointed, if this was not a supported use case.

compose/build.gradle.kts
```diff
@@ -73,7 +73,6 @@ subprojects {
       // https://developer.android.com/jetpack/androidx/releases/compose-runtime
       freeCompilerArgs += listOf(
         "-Xopt-in=kotlin.RequiresOptIn",
-        "-Xallow-jvm-ir-dependencies",
         "-Xskip-prerelease-check"
```
zach-klippenstein, line 76: Nit: this can go too

compose/README.md
```diff
@@ -48,7 +48,6 @@ android {
     compose true
   }
   composeOptions {
-    kotlinCompilerVersion "1.4.0-dev-withExperimentalGoogleExtensions-20200720"
```
SalvatoreT, line 51: 🎉

maybeYouShouldBeTheLawyer (approved):

PR #430, opened 2021-06-07
Replace experimental time API in TraceEncoder with custom implementation of TimeMark

JoelBryceAnderson:
* Replace experimental time API in TraceEncoder with custom implementation of TimeMark
* The default implementation, TraceEncoderTimeMark, holds a starting value from `System.nanoTime()` which it compares to future calls to `System.nanoTime()` to calculate elapsed time

### DIFF
--- a/trace-encoder/src/main/java/com/squareup/tracing/TimeMark.kt
+++ b/trace-encoder/src/main/java/com/squareup/tracing/TimeMark.kt
@@ -1,0 +1,15 @@
+package com.squareup.tracing
+
+/**
+ * Interface that represents a time point. Remains bound to the time source it was taken from and
+ * allows querying for the duration of time elapsed from that point (see the val [elapsedNow]).
+ */
+public interface TimeMark {
+  /**
+   * Returns the amount of time passed from this mark measured with the time source from which this
+   * mark was taken.
+   *
+   * Note that the content of this val can change on subsequent invocations.
+   */
+  public val elapsedNow: Long
+}
--- a/trace-encoder/src/main/java/com/squareup/tracing/TraceEncoder.kt
+++ b/trace-encoder/src/main/java/com/squareup/tracing/TraceEncoder.kt
@@ -12,8 +12,6 @@ import kotlinx.coroutines.channels.consumeEach
 import okio.BufferedSink
 import java.io.Closeable
 import java.util.concurrent.atomic.AtomicInteger
-import kotlin.time.ExperimentalTime
-import kotlin.time.TimeMark
 import kotlin.time.TimeSource
 
 /**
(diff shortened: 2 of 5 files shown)


trace-encoder/src/main/java/com/squareup/tracing/TimeMark.kt
```diff
@@ -0,0 +1,5 @@
+package com.squareup.tracing
+
+public interface TimeMark {
```
rjrjr, line 3: Needs kdoc.

trace-encoder/src/main/java/com/squareup/tracing/TraceEncoder.kt
```diff
@@ -12,27 +12,24 @@ import kotlinx.coroutines.channels.consumeEach
...
 import kotlin.time.TimeSource
 
 /**
  * Encodes and writes [trace events][TraceEvent] to an Okio [BufferedSink].
  *
  * @param scope The [CoroutineScope] that defines the lifetime for the encoder. When the scope is
  * cancelled or fails, the sink returned from [sinkProvider] will be closed.
- * @param start The [TimeMark] to consider the beginning timestamp of the trace. All trace events'
- * timestamps are relative to this mark.
+ * @param start The [TraceEncoderTimeMark] to consider the beginning timestamp of the trace. All
```
rjrjr, line 22: It's still a `TimeMark`, no?

```suggestion
 * @param start The [TimeMark] to consider the beginning timestamp of the trace. All
```

rjrjr:
Run `./gradlew trace-encoder:apiCheck` to fix the API check.

trace-encoder/src/main/java/com/squareup/tracing/TraceEncoder.kt
JoelBryceAnderson: Tried to revert this change and I think I ended up dropping the commit. Will fix.

zach-klippenstein:
I'm not sure I understand what this is actually fixing. I just ran the dungeon sample on `main` and everything worked fine. Could you give a little more context?

JoelBryceAnderson:
> I'm not sure I understand what this is actually fixing. I just ran the dungeon sample on `main` and everything worked fine. Could you give a little more context?

To be honest, I'm not entirely sure of the root cause, but there seems to be an issue invoking the method `TimeMark#elapsedNow` inside `TraceEncoder` from applications using kotlin 1.5.0. I'm curious if it's some kind of version miss-match? Essentially the experimental time API is't behaving properly and this PR works around that.

rjrjr:
The experimental warning on the Time classes particularly stress that it's a bad idea for libraries to use those methods, as binary compatibility is not assured.

I think what @JoelBryceAnderson found out is that our 1.5-based app is puking when it causes that call to be made against a 1.4-based release of the library.

rjrjr (approved):
LGTM, but let's see if Zach buys my library / app kotlin version mismatch theory before merging.

zach-klippenstein:
Ah, I forgot POS was still on such an old release. That makes lots of sense, and yea probably shouldn't use experimental stuff.

zach-klippenstein (approved):

PR #432, opened 2021-06-08
Adds unit tests for view state casting fix

rjrjr:
Introduces our first two Robolectric tests. They could cover more, but at least we have a toe hold.

### DIFF
--- a/workflow-ui/core-android/src/test/java/com/squareup/workflow1/ui/WorkflowLayoutTest.kt
+++ b/workflow-ui/core-android/src/test/java/com/squareup/workflow1/ui/WorkflowLayoutTest.kt
@@ -1,0 +1,41 @@
+package com.squareup.workflow1.ui
+
+import android.content.Context
+import android.os.Bundle
+import android.os.Parcelable
+import android.util.SparseArray
+import android.view.View
+import androidx.test.core.app.ApplicationProvider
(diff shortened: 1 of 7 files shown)


workflow-ui/core-android/src/test/java/com/squareup/workflow1/ui/WorkflowLayoutTest.kt
```diff
@@ -0,0 +1,40 @@
...
+import android.view.View
+import androidx.test.core.app.ApplicationProvider
+import com.google.common.truth.Truth.assertThat
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+
+@RunWith(RobolectricTestRunner::class)
+@Config(manifest = Config.NONE, sdk = [28])
```
Blisse, line 16: worth a comment about the error you were seeing (and why it needs to be specific here)

Blisse (approved):

workflow-ui/core-android/src/test/java/com/squareup/workflow1/ui/WorkflowLayoutTest.kt
rjrjr: Done.

PR #438, opened 2021-06-17
Stop tracking .idea/misc.xml in compose project.

zach-klippenstein:
cc @maybeYouShouldBeTheLawyer


rjrjr (approved):

zach-klippenstein:
Only CI failure was an instrumentation test, but this is just an IDE change so merging anyway.

PR #439, opened 2021-06-17
Make compose project's .editorconfig a symlink to main editorconfig.

zach-klippenstein:


rjrjr (approved):
I wonder what that does on windows boxes? :)

PR #447, opened 2021-07-08
Tweak kotlin module names to avoid META-INF conflicts.

rjrjr:
Make sure our module names don't conflict with those from pre-workflow1
releases, so that old and new META-INF/ entries don't stomp each other.
(This is only an issue for apps that are still migrating from workflow to
workflow1, and so need to import two versions of the library.)
https://blog.jetbrains.com/kotlin/2015/09/kotlin-m13-is-out/


steve-the-edwards (approved):
Thanks @rjrjr !

PR #449, opened 2021-07-29
Apply snapshots on the first render pass only.

rjrjr:


syadu1988 (approved):

steve-the-edwards (approved):

workflow-runtime/src/main/java/com/squareup/workflow1/internal/SubtreeManager.kt
```diff
@@ -110,8 +110,10 @@ internal class SubtreeManager<PropsT, StateT, OutputT>(
     // and must be torn down.
     children.commitStaging { child ->
       child.workflowNode.cancel()
-      snapshotCache -= child.id
```
zach-klippenstein, line 113: I did some blame digging, and this line has been here for a few commits. Doesn't make sense to me, so yea this change definitely seems like a big improvement.

workflow-runtime/src/test/java/com/squareup/workflow1/internal/SubtreeManagerTest.kt
```diff
@@ -79,6 +82,7 @@ internal class SubtreeManagerTest {
     override fun snapshotState(state: Unit): Snapshot {
       snapshots++
       return Snapshot.write {
+        it.writeByte(0)
```
zach-klippenstein, line 85: What's this for?

zach-klippenstein (approved):

workflow-runtime/src/test/java/com/squareup/workflow1/internal/SubtreeManagerTest.kt
rjrjr: w/o that initialState() gets called with a null Snapshot.

PR #452, opened 2021-07-30
Upgrade gradle to 7.1.1.

zach-klippenstein:
Trying to do the first part of https://github.com/square/workflow-kotlin/pull/451 separately, just to make sure it works on its own.


zach-klippenstein:
Merged @SalvatoreT's PR instead.

PR #453, opened 2021-07-30
Turn the dungeon lights back on (i.e. Upgrade ConstraintLayout library to 2.1.0-rc01)

zach-klippenstein:
It seems that version 2.0.4 had a bug that causes the BoardView in the dungeon sample's gameplay screen to be measured with zero width and height. It's fixed in this version.

ConstraintLayout is only used in samples, so we are safe to consume a pre-release version.

Thanks to @SalvatoreT for catching this sample bug!

### DIFF
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -21,7 +21,7 @@ object Dependencies {
       const val ui = "androidx.compose.ui:ui:1.0.0-alpha12"
     }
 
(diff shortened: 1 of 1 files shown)


SalvatoreT:
👍

SalvatoreT:
A+ title

SalvatoreT:
Works for me!
---
# Before
missing-dungeon-map

# After
dungeon-map-is-back

rjrjr (approved):

PR #454, opened 2021-07-30
Clean up color-related deprecation warning

SalvatoreT:
- use ContextCompat for loading color resource
- use the Material color resource instead of the deprecated system one
https://developer.android.com/reference/android/R.color#primary_text_dark
https://github.com/material-components/material-components-android/blob/master/lib/java/com/google/android/material/color/res/values/colors.xml#L23


rjrjr (approved):
Thanks Sal.

Hey, this just occured to me: you have to marry into a Dura family, so you can change your name to Salvatore Testa-Dura.

SalvatoreT:
If I marry a Verde, I could become related to a journeyman NFL quarterback.

PR #455, opened 2021-07-30
[compose] Upgrade Gradle to 7.1.1, AGP to 7.0.0 stable, Compose to 1.0.0 stable, and Workflow to latest release.

zach-klippenstein:
There seems to be a Compose bug where returning early from a `Box` composable lambda crashes the runtime.
Update: That's a known issue from last year already: https://issuetracker.google.com/issues/173651155


zach-klippenstein:
Also looks like nesting Android views is broken:
image

rjrjr (approved):

PR #462, opened 2021-08-06
Remove useless type parameter from showFirstRendering().

zach-klippenstein:


zach-klippenstein:
Network timeout caused one of three test runs to fail – I think it's fine to merge, this isn't a behavior change just compile-time change.

steve-the-edwards (approved):

rjrjr:
giphy

PR #467, opened 2021-08-06
Changelog for v1.0.0-alpha18.

zach-klippenstein:


gradle.properties
```diff
@@ -8,7 +8,7 @@ android.useAndroidX=true
 systemProp.org.gradle.internal.publish.checksums.insecure=true
 
 GROUP=com.squareup.workflow1
-VERSION_NAME=1.0.0-alpha18-SNAPSHOT
+VERSION_NAME=1.0.0-alpha18
```
rjrjr, line 11: Your don't want this to be part of the changelog PR

rjrjr (approved):

gradle.properties
zach-klippenstein: Why not? I thought we used to do that all the time.

zach-klippenstein: Reverted.

Issue #458, opened 2021-07-30
[Design] Overall design for officially supporting Compose from Workflow


zach-klippenstein:
_This issue was adapted from an [internal design doc](https://docs.google.com/document/d/1Q1INk3n9L7LKpChjsmNzhaPGZXkNZdKSDj00YKhSaNY) and a [blog](https://developer.squareup.com/blog/jetpack-compose-support-in-workflow/)._

## Timeline and Process

Compose entered beta in the first half of 2020. Since we were all locked in our homes with no social lives, it was the perfect time to start exploring what integration between Compose and Workflows would look like. This was very experimental work — Compose APIs were changing drastically every two weeks. To say the least, it was not “ready for production.” However, it was important to suss out what sort of integration points were available to us, what API shapes felt natural, and where the rough edges were. In addition to figuring out our own adoption story, we have also been able to contribute a lot of feedback to Google (see our [case study](https://developer.android.com/stories/apps/square-compose)), and some of the features we initially wrote specifically for workflow integration ended up making it into the library (e.g. automatic subcomposition linking in child `View`s).

## Goals and Non-Goals

At Square, when we start a project, we like to enumerate and distinguish goals and explicit non-goals to delineate the scope of the project.

### Goals

- Allow Workflow screens to be written in Compose.
- Allow interop in both directions using the usual idioms:
    - Non-Compose Workflow screens to nest Compose-based Workflow screens.
    - Compose-based Workflow screens to nest both Compose-based and non-Compose Workflow screens.
- Provide a way for Compose-based apps to host a Workflow runtime and display Compose apps, e.g. a parallel entry point to how non-Compose apps can use `WorkflowLayout`.
- Ensure that data flow through `CompositionLocal`s is propagated to child Compose-based screens, regardless of whether there are non-Workflow screens sitting in between them.

### Non-Goals

- Convert existing screens in our apps to Compose.
- Provide design system components in Compose. (This is planned, but as a separate project that depends on this one.)
- Anything with our own internal declarative UI toolkit, Mosaic (sunsetting it, integrating with it, or otherwise).

## Major Components

There are a few major areas this project needs to focus on to support Compose from Workflows: navigation, `ViewFactory` support, and hosting.

### Navigation support

[...]

PR #473, opened 2021-08-17
Introduce ComposeViewFactory as a convenient way to define ViewFactories as composable functions.

zach-klippenstein:
First part of #458.

This code was mostly copy/pasted from the compose sub-project, with some additional polish. Some of the docs refer to things that don't exist yet, but will be coming in immediate follow-ups.

I'll bring in the samples once more pieces are merged, so I don't have to edit them to work with a more limited set of primitives.

Archeology: The first version of this was introduced in https://github.com/square/workflow/pull/703.

### DIFF
--- a/workflow-ui/compose/api/compose.api
+++ b/workflow-ui/compose/api/compose.api
@@ -1,0 +1,11 @@
+public abstract class com/squareup/workflow1/ui/compose/ComposeViewFactory : com/squareup/workflow1/ui/ViewFactory {
+	public static final field $stable I
+	public fun <init> ()V
+	public abstract fun Content (Ljava/lang/Object;Lcom/squareup/workflow1/ui/ViewEnvironment;Landroidx/compose/runtime/Composer;I)V
+	public final fun buildView (Ljava/lang/Object;Lcom/squareup/workflow1/ui/ViewEnvironment;Landroid/content/Context;Landroid/view/ViewGroup;)Landroid/view/View;
+}
+
+public final class com/squareup/workflow1/ui/compose/ComposeViewFactoryKt {
+	public static final fun composeViewFactory (Lkotlin/reflect/KClass;Lkotlin/jvm/functions/Function4;)Lcom/squareup/workflow1/ui/ViewFactory;
+}
+
--- a/workflow-ui/compose/build.gradle.kts
+++ b/workflow-ui/compose/build.gradle.kts
@@ -38,11 +38,10 @@ dependencies {
   api(project(":workflow-ui:backstack-android"))
   api(project(":workflow-ui:core-android"))
 
-  api(Dependencies.Kotlin.Stdlib.jdk8)
+  api(Dependencies.AndroidX.Compose.foundation)
 
   androidTestImplementation(project(":workflow-runtime"))
   androidTestImplementation(Dependencies.AndroidX.activity)
-  androidTestImplementation(Dependencies.AndroidX.Compose.foundation)
   androidTestImplementation(Dependencies.AndroidX.Compose.ui)
   androidTestImplementation(Dependencies.Test.AndroidX.core)
   androidTestImplementation(Dependencies.Test.AndroidX.truthExt)
--- a/workflow-ui/compose/src/androidTest/AndroidManifest.xml
+++ b/workflow-ui/compose/src/androidTest/AndroidManifest.xml
@@ -1,1 +1,7 @@
-<manifest package="com.squareup.workflow1.ui.compose.test"/>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="com.squareup.workflow1.ui.compose.test">
+
+  <application>
+    <activity android:name="androidx.activity.ComponentActivity"/>
+  </application>
+</manifest>
--- a/workflow-ui/compose/src/androidTest/java/com/squareup/workflow1/ui/compose/ComposeViewFactoryTest.kt
+++ b/workflow-ui/compose/src/androidTest/java/com/squareup/workflow1/ui/compose/ComposeViewFactoryTest.kt
@@ -1,0 +1,139 @@
+package com.squareup.workflow1.ui.compose
+
+import android.content.Context
+import android.widget.FrameLayout
+import androidx.compose.foundation.text.BasicText
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.test.assertIsDisplayed
+import androidx.compose.ui.test.assertTextEquals
+import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.compose.ui.test.onNodeWithTag
+import androidx.compose.ui.test.onNodeWithText
+import androidx.compose.ui.viewinterop.AndroidView
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import com.squareup.workflow1.ui.ViewEnvironment
+import com.squareup.workflow1.ui.ViewEnvironmentKey
+import com.squareup.workflow1.ui.ViewRegistry
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+import com.squareup.workflow1.ui.WorkflowViewStub
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@OptIn(WorkflowUiExperimentalApi::class)
+@RunWith(AndroidJUnit4::class)
+class ComposeViewFactoryTest {
+
+  @get:Rule val composeRule = createComposeRule()
+
+  @Test fun showsComposeContent() {
+    val viewFactory = composeViewFactory<Unit> { _, _ ->
+      BasicText("Hello, world!")
+    }
+    val viewRegistry = ViewRegistry(viewFactory)
+    val viewEnvironment = ViewEnvironment(mapOf(ViewRegistry to viewRegistry))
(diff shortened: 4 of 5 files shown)


rjrjr (approved):

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposedViewFactory.kt
```diff
@@ -0,0 +1,141 @@
...
+import com.squareup.workflow1.ui.bindShowRendering
+import kotlin.reflect.KClass
+
+/**
+ * Creates a [ViewFactory] that uses a [Composable] function to display the rendering.
+ *
+ * Simple usage:
+ *
+ * ```
+ * val FooViewFactory = composedViewFactory { rendering, _ ->
```
maybeYouShouldBeTheLawyer, line 24: Just double-checking that the capitalization of `FooViewFactory` was deliberate... (same below).

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposedViewFactory.kt
```diff
@@ -0,0 +1,141 @@
...
+ * val viewRegistry = ViewRegistry(FooViewFactory, …)
+ * ```
+ *
+ * If you need to write a class instead of a function, for example to support dependency injection,
+ * see [ComposedViewFactory].
+ *
+ * For more details about how to write composable view factories, see [ComposedViewFactory].
+ */
+@WorkflowUiExperimentalApi
+public inline fun <reified RenderingT : Any> composedViewFactory(
```
maybeYouShouldBeTheLawyer, line 39: What's the thinking behind `composedViewFactory` rather than `composeViewFactory`? 

(I get that the latter could be a hang-up in that the uninitiated might read 'compose' as a verb. But – being initiated on the Compose side of things – I found myself wondering whether the past participle wasn't mean to convey some subtlety of its own. Maybe there's no winning. A data point for your consideration, at any rate.)

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposedViewFactory.kt
```diff
@@ -0,0 +1,141 @@
...
+  @Composable override fun Content(
+    rendering: RenderingT,
+    viewEnvironment: ViewEnvironment
+  ) {
+    content(rendering, viewEnvironment)
+  }
+}
+
+/**
+ * A [ViewFactory] that uses a [Composable] function to display the rendering. It is the composable
```
maybeYouShouldBeTheLawyer, line 65: nit to disambiguate: 'It is the composable version' -> 'It is the Compose-based analogue' (or whatever)

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposedViewFactory.kt
```diff
@@ -0,0 +1,141 @@
...
+ *     rendering: Foo,
+ *     viewEnvironment: ViewEnvironment
+ *   ) {
+ *     Text(rendering.message)
+ *   }
+ * }
+ *
+ * …
+ *
+ * val viewRegistry = ViewRegistry(FooBinding, …)
```
maybeYouShouldBeTheLawyer, line 84: FooBinding -> FooViewFactory?

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposedViewFactory.kt
```diff
@@ -0,0 +1,141 @@
...
+ * ```
+ *
+ * ## Nesting child renderings
+ *
+ * Workflows can render other workflows, and renderings from one workflow can contain renderings
+ * from other workflows. These renderings may all be bound to their own [ViewFactory]s. Regular
+ * [ViewFactory]s and `LayoutRunner`s use
```
maybeYouShouldBeTheLawyer, line 91: nit: \`LayoutRunner\` ->  [LayoutRunner]? (We use the brackets with LayoutRunner above, at any rate.)

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposedViewFactory.kt
```diff
@@ -0,0 +1,141 @@
...
+ * [WorkflowViewStub][com.squareup.workflow1.ui.WorkflowViewStub] to recursively show nested
+ * renderings using the [ViewRegistry][com.squareup.workflow1.ui.ViewRegistry].
+ *
+ * View factories defined using this function may also show nested renderings. Doing so is as simple
+ * as calling [WorkflowRendering] and passing in the nested rendering. See the kdoc on that function
+ * for an example.
+ *
+ * Nested renderings will have access to any ambients defined in outer composable, even if there are
```
maybeYouShouldBeTheLawyer, line 99: nit: Do Compose docs still use the term 'ambients' or has that been replaced by composition locals? (Disregard if you want to cast a wider net than composition locals, of course.)

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposedViewFactory.kt
```diff
@@ -0,0 +1,141 @@
...
+ *     Text(rendering.message)
+ *   }
+ * }
+ *
+ * …
+ *
+ * val viewRegistry = ViewRegistry(FooBinding, …)
+ * ```
+ *
+ * ## Nesting child renderings
```
maybeYouShouldBeTheLawyer, line 87: Great docs 👍

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposedViewFactory.kt
```diff
@@ -0,0 +1,141 @@
...
+ * context, call [withCompositionRoot] on your top-level [ViewEnvironment]. The first time a
+ * [composedViewFactory] is used to show a rendering, its [Content] function will be wrapped
+ * with the [CompositionRoot]. See the documentation on [CompositionRoot] for more information.
+ */
+@WorkflowUiExperimentalApi
+public abstract class ComposedViewFactory<RenderingT : Any> : ViewFactory<RenderingT> {
+
+  /**
+   * The composable content of this [ViewFactory]. This method will be called any time [rendering]
+   * or [viewEnvironment] change. It is the composable version of [LayoutRunner.showRendering].
```
maybeYouShouldBeTheLawyer, line 116: nit as above: composable version -> Compose-based analogue (etc.)

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposedViewFactory.kt
```diff
@@ -0,0 +1,141 @@
...
+    contextForNewView: Context,
+    container: ViewGroup?
+  ): View = ComposeView(contextForNewView).also { composeView ->
+    // Update the state whenever a new rendering is emitted.
+    // This lambda will be executed synchronously before bindShowRendering returns.
+    composeView.bindShowRendering(
+      initialRendering,
+      initialViewEnvironment
+    ) { rendering, environment ->
+      // Entry point to the world of Compose.
```
maybeYouShouldBeTheLawyer, line 135: 💥 !

maybeYouShouldBeTheLawyer (approved):
Great to see this 💯

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposedViewFactory.kt
zach-klippenstein: yep

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposedViewFactory.kt
zach-klippenstein: i don't care what we call it

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposedViewFactory.kt
zach-klippenstein: crap, thought i got these

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposedViewFactory.kt
maybeYouShouldBeTheLawyer: No big opinions here either, but my mind wanted `composeViewFactory`, seemingly. I think Oleg may have asked the same question about the name in your retro design doc.

zach-klippenstein:
This depends on the `vtssr-support2` branch only for correct state save/restore behavior, so I'm rebasing onto main so I am not blocked by getting that sticky thing finished.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposedViewFactory.kt
zach-klippenstein: I did a quick poll on both [workflow](https://workflow-community.slack.com/archives/C013JGW58N7/p1629240387012500) and [internal](https://square.slack.com/archives/CE470RL5T/p1629240405170000) slack, and `composeViewFactory` won in both. I'll change it.

PR #486, opened 2021-08-18
Be even more careful about holding onto View references.

zach-klippenstein:


rjrjr (approved):

rjrjr:
Looks like `:samples:todo-android:app:connectedDebugAndroidTest` is failing consistently

zach-klippenstein:
Nulling out the view by hand was too aggressive, switched to using a weak reference instead, and the tests are passing locally for me.

rjrjr:
> Nulling out the view by hand was too aggressive, switched to using a weak reference instead, and the tests are passing locally for me.

Holy crap. LGTM.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLifecycleOwner.kt
```diff
@@ -111,9 +111,9 @@ internal class RealWorkflowLifecycleOwner(
   LifecycleEventObserver {
 
   /**
-   * This gets nulled out when we are destroyed.
+   * Weak reference ensures that we don't leak the view.
    */
-  private var view: View? = view
+  private var view: WeakReference<View> = WeakReference(null)
```
pyricau, line 116: 👀 would love to learn more about what "be even more careful" means. How comes we couldn't clear out the view ref when we needed to?

https://square.github.io/leakcanary/fundamentals-fixing-a-memory-leak/
image

rjrjr: For more context: this PR is part of the not-quite-finished effort to make Workflow get along with Compose. 

* Compose relies on SavedStateRegistry instead of classic View persistence (saveHierarchyState et al)
* SavedStateRegistry reliies on LifecycleOwner
* We need to set up LifecyleOwner + SavedStateRegistry per backstack frame and pre dialog
* The timing for all of that is hard to get right

Full details are in #458, which I'm going to have to internalize better.

rjrjr: #560 gets rid of the `WeakReference` with no apparent ill effects.

PR #488, opened 2021-08-19
Introduce WorkflowRendering to show child renderings inside a composedViewFactory.

### LINKED ISSUE #374: WorkflowRendering composable should support AndroidViewRendering

### LINKED ISSUE #375: WorkflowRendering composable should support Compatible

### LINKED ISSUE #478: WorkflowRendering should support WorkflowLifecycleOwner in the ComposeViewFactory path
In the android view path, this is done by `WorkflowViewStub`. But we don't use the stub if the rendering is a `ComposeViewFactory`, so `WorkflowRendering` needs to handle that itself.

zach-klippenstein:
`WorkflowRendering` is the compose analogue of `WorkflowViewStub`.

Part of #458.

Also introduces a sample of nesting legacy and compose renderings.

Archeology: Originally introduced in https://github.com/square/workflow-kotlin-compose/pull/7.

- Closes #374.
- Closes #375.
- Closes #478.

Kapture 2021-08-18 at 19 34 07

### DIFF
--- a/samples/compose-samples/src/androidTest/java/com/squareup/sample/compose/nestedrenderings/NestedRenderingsTest.kt
+++ b/samples/compose-samples/src/androidTest/java/com/squareup/sample/compose/nestedrenderings/NestedRenderingsTest.kt
@@ -1,0 +1,69 @@
+package com.squareup.sample.compose.nestedrenderings
+
+import androidx.compose.ui.test.SemanticsNodeInteraction
+import androidx.compose.ui.test.SemanticsNodeInteractionCollection
+import androidx.compose.ui.test.assertCountEquals
+import androidx.compose.ui.test.assertIsDisplayed
+import androidx.compose.ui.test.junit4.createAndroidComposeRule
+import androidx.compose.ui.test.onAllNodesWithText
+import androidx.compose.ui.test.onNodeWithText
+import androidx.compose.ui.test.performClick
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+
+private const val ADD_BUTTON_TEXT = "Add Child"
+
+@RunWith(AndroidJUnit4::class)
+class NestedRenderingsTest {
+
+  @OptIn(WorkflowUiExperimentalApi::class)
+  @get:Rule val composeRule = createAndroidComposeRule<NestedRenderingsActivity>()
+
+  @Test fun childrenAreAddedAndRemoved() {
+    composeRule.onNodeWithText(ADD_BUTTON_TEXT)
+      .assertIsDisplayed()
+      .performClick()
+
+    composeRule.onAllNodesWithText(ADD_BUTTON_TEXT)
+      .assertCountEquals(2)
+      .forEach { it.performClick() }
+
+    composeRule.onAllNodesWithText(ADD_BUTTON_TEXT)
+      .assertCountEquals(4)
+
+    resetAll()
+    composeRule.onAllNodesWithText(ADD_BUTTON_TEXT).assertCountEquals(1)
+  }
+
+  /**
+   * We can't rely on the order of nodes returned by [onAllNodesWithText], and the contents of the
+   * collection will change as we remove nodes, so we have to double-loop over all reset buttons and
+   * click them all until there is only one left.
+   */
+  private fun resetAll() {
+    var foundNodes = Int.MAX_VALUE
+    while (foundNodes > 1) {
+      foundNodes = 0
+      composeRule.onAllNodesWithText("Reset").forEach {
+        try {
+          it.assertExists()
+        } catch (e: AssertionError) {
+          // No more reset buttons, we're done.
+          return@forEach
+        }
+        foundNodes++
+        it.performClick()
+      }
+    }
+  }
+
+  private fun SemanticsNodeInteractionCollection.forEach(
+    block: (SemanticsNodeInteraction) -> Unit
+  ) {
+    val count = fetchSemanticsNodes().size
+    for (i in 0 until count) block(get(i))
+  }
+}
--- a/samples/compose-samples/src/main/java/com/squareup/sample/compose/nestedrenderings/LegacyRunner.kt
+++ b/samples/compose-samples/src/main/java/com/squareup/sample/compose/nestedrenderings/LegacyRunner.kt
@@ -1,0 +1,39 @@
+package com.squareup.sample.compose.nestedrenderings
+
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.tooling.preview.Preview
+import com.squareup.sample.compose.databinding.LegacyViewBinding
+import com.squareup.sample.compose.nestedrenderings.RecursiveWorkflow.LegacyRendering
+import com.squareup.workflow1.ui.LayoutRunner
+import com.squareup.workflow1.ui.LayoutRunner.Companion.bind
+import com.squareup.workflow1.ui.ViewEnvironment
+import com.squareup.workflow1.ui.ViewFactory
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+
+/**
+ * A [LayoutRunner] that renders [LegacyRendering]s using the legacy view framework.
+ */
+@OptIn(WorkflowUiExperimentalApi::class)
+class LegacyRunner(private val binding: LegacyViewBinding) : LayoutRunner<LegacyRendering> {
+
+  override fun showRendering(
+    rendering: LegacyRendering,
+    viewEnvironment: ViewEnvironment
+  ) {
+    binding.stub.update(rendering.rendering, viewEnvironment)
+  }
+
+  companion object : ViewFactory<LegacyRendering> by bind(
+    LegacyViewBinding::inflate, ::LegacyRunner
+  )
+}
+
+@OptIn(WorkflowUiExperimentalApi::class)
+@Preview(widthDp = 200, heightDp = 150, showBackground = true)
+@Composable private fun LegacyRunnerPreview() {
+  // TODO(#458) Uncomment once preview support is imported.
+  // LegacyRunner.preview(
+  //   rendering = LegacyRendering("child"),
+  //   placeholderModifier = Modifier.fillMaxSize()
+  // )
+}
--- a/samples/compose-samples/src/main/java/com/squareup/sample/compose/nestedrenderings/NestedRenderingsActivity.kt
+++ b/samples/compose-samples/src/main/java/com/squareup/sample/compose/nestedrenderings/NestedRenderingsActivity.kt
@@ -1,0 +1,59 @@
+package com.squareup.sample.compose.nestedrenderings
+
+import android.os.Bundle
+import androidx.activity.viewModels
+import androidx.appcompat.app.AppCompatActivity
+import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.ui.graphics.Color
+import androidx.lifecycle.SavedStateHandle
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.viewModelScope
+import com.squareup.workflow1.ui.ViewEnvironment
+import com.squareup.workflow1.ui.ViewRegistry
+import com.squareup.workflow1.ui.WorkflowLayout
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+import com.squareup.workflow1.ui.compose.withCompositionRoot
+import com.squareup.workflow1.ui.renderWorkflowIn
+import kotlinx.coroutines.flow.StateFlow
+
+@OptIn(WorkflowUiExperimentalApi::class)
+private val viewRegistry = ViewRegistry(
+  RecursiveViewFactory,
+  LegacyRunner
+)
+
+@OptIn(WorkflowUiExperimentalApi::class)
+private val viewEnvironment =
+  ViewEnvironment(mapOf(ViewRegistry to viewRegistry)).withCompositionRoot { content ->
+    CompositionLocalProvider(LocalBackgroundColor provides Color.Green) {
+      content()
+    }
+  }
+
+@WorkflowUiExperimentalApi
+class NestedRenderingsActivity : AppCompatActivity() {
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+
+    val model: NestedRenderingsModel by viewModels()
+    setContentView(
+      WorkflowLayout(this).apply {
+        start(
+          renderings = model.renderings,
+          environment = viewEnvironment
+        )
+      }
+    )
+  }
+
+  class NestedRenderingsModel(savedState: SavedStateHandle) : ViewModel() {
+    @OptIn(WorkflowUiExperimentalApi::class)
+    val renderings: StateFlow<Any> by lazy {
+      renderWorkflowIn(
+        workflow = RecursiveWorkflow,
+        scope = viewModelScope,
+        savedStateHandle = savedState
+      )
+    }
+  }
+}
--- a/samples/compose-samples/src/main/java/com/squareup/sample/compose/nestedrenderings/RecursiveViewFactory.kt
+++ b/samples/compose-samples/src/main/java/com/squareup/sample/compose/nestedrenderings/RecursiveViewFactory.kt
@@ -1,0 +1,128 @@
+@file:Suppress("RemoveEmptyParenthesesFromAnnotationEntry")
+
+package com.squareup.sample.compose.nestedrenderings
+
+import androidx.compose.foundation.layout.Arrangement.SpaceEvenly
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
+import androidx.compose.material.Button
+import androidx.compose.material.Card
+import androidx.compose.material.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.runtime.compositionLocalOf
+import androidx.compose.runtime.remember
+import androidx.compose.ui.Alignment.Companion.CenterHorizontally
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.compositeOver
+import androidx.compose.ui.res.dimensionResource
+import androidx.compose.ui.tooling.preview.Preview
+import com.squareup.sample.compose.R
+import com.squareup.sample.compose.nestedrenderings.RecursiveWorkflow.Rendering
+import com.squareup.workflow1.ui.ViewEnvironment
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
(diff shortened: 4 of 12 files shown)


maybeYouShouldBeTheLawyer (approved):
Nice – and thanks for the superb documentation and tests 💯.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/WorkflowRendering.kt
```diff
@@ -0,0 +1,201 @@
...
+  // If the view factory created an Android view, this will also remove the old one from the view
+  // hierarchy before replacing it with the new one.
+  key(viewFactory) {
+    // Just like WorkflowViewStub, we need to manage a Lifecycle for the child view. We just provide
+    // a local here – ViewFactoryAndroidView will handle setting the appropriate view tree owners
+    // on the child view when necessary. Because this function is surrounded by a key() call, when
+    // the rendering is incompatible, the lifecycle for the old view will be destroyed.
+    val lifecycleOwner = rememberChildLifecycleOwner()
+    CompositionLocalProvider(LocalLifecycleOwner provides lifecycleOwner) {
+      if (viewFactory is ComposeViewFactory) {
```
zach-klippenstein, line 94: Instead of having this if/else here, we could make a function `ViewFactory.asComposeViewFactory` that returns this or creates a wrapper factory that has the Android view integration logic, then this function wouldn't have to special case anything. I think that might be a bit cleaner? Or is it just over-engineered?

zach-klippenstein: I think I've convinced myself this is a good idea, since it makes extra, crystal clear the separation between what is general logic for managing the child, and what is specialized logic for handling Android view integration.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/WorkflowRendering.kt
```diff
@@ -0,0 +1,201 @@
...
+  // This will fetch a new view factory any time the new rendering is incompatible with the previous
+  // one, as determined by Compatible. This corresponds to WorkflowViewStub's canShowRendering
+  // check.
+  // We also use the ViewRegistry as a key, in case a new one gets passed down and we need to ask
+  // it for a more current factory.
+  // The view registry may return a new factory instance for a rendering every time we ask it, for
+  // example if an AndroidViewRendering doesn't share its factory between rendering instances. We
+  // intentionally don't ask it for a new instance every time to match the behavior of
+  // WorkflowViewStub and other containers, which only ask for a new factory when the rendering is
+  // incompatible.
```
steve-the-edwards, line 76: Had to read this 3 times to understand. Maybe start with the last paragraph as the overview and describe the intent? Something like:

> We always want an up-to-date viewFactory. The instance of the viewFactory may change with every rendering instance though (such as for [AndroidViewRendering]), so we don't want to key the viewFactory updates on the rendering itself. Instead there are 2 cases where we need a new viewFactory:
> 1. The viewRegistry has changed and may have an updated viewFactory.
> 2. The new Rendering is no longer [Compatible] with the previous.
> As a result, we key the update of the viewFactory on these two conditions.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/WorkflowRendering.kt
```diff
@@ -0,0 +1,201 @@
...
+  val renderingKey = Compatible.keyFor(rendering)
```
steve-the-edwards, line 77: nit: 
```suggestion
  val renderingCompatibilityKey = Compatible.keyFor(rendering)
```

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/WorkflowRendering.kt
```diff
@@ -0,0 +1,201 @@
...
+  val viewFactory = remember(renderingKey, viewRegistry) {
+    viewRegistry.getFactoryForRendering(rendering)
+  }
+
+  // By surrounding the rendering composable with this key function, any time the new rendering is
+  // not compatible with the previous rendering we'll tear down the previous subtree of the
```
steve-the-edwards, line 83:
```suggestion
  // not compatible with the previous rendering or the viewRegistry is updated we'll tear down the previous subtree of the
```

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/WorkflowRendering.kt
steve-the-edwards: Sorry: convinced yourself having the case in `WorkflowRendering` here is good? or having the extension on the `ViewFactory` is good?

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/WorkflowRendering.kt
```diff
@@ -0,0 +1,201 @@
...
+ *    of replacing itself in its parent doesn't play nicely with Compose.
+ *  - Allows us to pass the correct parent view for inflation (the root of the composition).
+ *  - Avoids `WorkflowViewStub` having to do its own lookup to find the correct [ViewFactory], since
+ *    we already have the correct one.
+ *  - Propagate the current [LifecycleOwner] from [LocalLifecycleOwner] by setting it as the
+ *    [ViewTreeLifecycleOwner] on the view.
+ *
+ * Like `WorkflowViewStub`, this function uses the [viewFactory] to create and memoize a [View] to
+ * display the [rendering], keeps it updated with the latest [rendering] and [viewEnvironment], and
+ * adds it to the composition.
```
steve-the-edwards, line 159: I've been reflecting about whether or not we don't want to start all this kdoc with "like `WorkflowViewStub`", "like `AndroidViewRendering`" etc. as we may want to include that as an addendum after describing this on its own terms. This may be the new normal after all. Not something needed for now, but we need to consider it.

steve-the-edwards (approved):

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/WorkflowRendering.kt
```diff
@@ -0,0 +1,201 @@
...
+  val lifecycleOwner = remember {
+    object : LifecycleOwner {
+      val registry = LifecycleRegistry(this)
+      override fun getLifecycle(): Lifecycle = registry
+    }
+  }
+  val parentLifecycle = LocalLifecycleOwner.current.lifecycle
+
+  DisposableEffect(parentLifecycle) {
+    val observer = LifecycleEventObserver { _, event ->
```
rjrjr, line 123: Hyper-nit: rename this `parentObserver`?

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/WorkflowRendering.kt
```diff
@@ -0,0 +1,201 @@
...
+              "${ViewFactory::class.java.name} that created it."
+          }
+
+          // Unfortunately AndroidView doesn't propagate this itself.
+          ViewTreeLifecycleOwner.set(view, lifecycleOwner)
+          // We don't propagate the (non-compose) SavedStateRegistryOwner, or the (compose)
+          // SaveableStateRegistry, because currently all our navigation is implemented as Android
+          // views, which only sets the state registry owner on a view that is a parent of this
+          // composition.
+          // If we want to support compose-based navigation using SaveableStateRegistry, we'd need
```
rjrjr, line 190: Should we track this?

rjrjr (approved):
This is magnificent.

samples/compose-samples/src/main/java/com/squareup/sample/compose/nestedrenderings/RecursiveViewFactory.kt
```diff
@@ -0,0 +1,129 @@
...
+/**
+ * Ambient of [Color] to use as the background color for a [RecursiveViewFactory].
+ */
+val LocalBackgroundColor = compositionLocalOf<Color> { error("No background color specified") }
+
+/**
+ * A `ViewFactory` that renders [RecursiveWorkflow.Rendering]s.
+ */
+@OptIn(WorkflowUiExperimentalApi::class)
+val RecursiveViewFactory = composeViewFactory<Rendering> { rendering, viewEnvironment ->
```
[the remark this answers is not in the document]
rjrjr: I think the answer is no, b/c `ComposeRendering` always creates a legacy `View` instance.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/WorkflowRendering.kt
zach-klippenstein: Sorry, pre-coffee comment. I've convinced myself I like the `asComposeFactory`. And just to be extra clear:

<table>
<tr><th>
With `asComposeViewFactory()`
</th><th>
With `if/else`
</th></tr>
<tr>
<td>

```kotlin
val factory = remember {
  registry.getFactoryForRendering(rendering)
   .asComposeViewFactory()
}

key(…) {
  CompositionLocalProvider(…) {
    Box(…) {
      factory.Content(rendering, viewEnvironment)
    }
  }
}
```

</td><td>

```kotlin
val factory = remember {
  registry.getFactoryForRendering(rendering)
}

key(…) {
  CompositionLocalProvider(…) {
    if (factory is ComposeViewFactory) {
      Box(…) {
        factory.Content(rendering, viewEnvironment)
      }
    } else {
      AndroidViewThingy(…)
    }
  }
}
```

</td>
</tr>
</table>

And `asComposeViewFactory()` would look like this:
```kotlin
private fun <R : Any> ViewFactory<R>.asComposeViewFactory() = 
  (this as? ComposeViewFactory) ?: object : ComposeViewFactory<R> {
    @Composable override fun Content(
      rendering: R,
    ) {
      AndroidViewThingy(rendering, viewEnvironment)
    }
  }
```

samples/compose-samples/src/main/java/com/squareup/sample/compose/nestedrenderings/LegacyRunner.kt
```diff
@@ -0,0 +1,39 @@
...
+}
+
+@OptIn(WorkflowUiExperimentalApi::class)
+@Preview(widthDp = 200, heightDp = 150, showBackground = true)
+@Composable private fun LegacyRunnerPreview() {
+  // TODO(#458) Uncomment once preview support is imported.
+  // LegacyRunner.preview(
+  //   rendering = LegacyRendering("child"),
+  //   placeholderModifier = Modifier.fillMaxSize()
+  // )
```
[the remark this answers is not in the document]
zach-klippenstein: It's in a separate PR that hasn't landed yet.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/WorkflowRendering.kt
steve-the-edwards: +1 I like `.asComposeViewFactory()` but I think it deserves some of your good documentation so folks know what is going on.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/WorkflowRendering.kt
zach-klippenstein: I am shocked that the small essay i wrote at midnight didn't turn out to be coherent 😅

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/WorkflowRendering.kt
zach-klippenstein: Yea, that thought definitely crossed my mind too. I'd cross-reference back as well, but the core-android module doesn't know about this one.

samples/compose-samples/src/main/java/com/squareup/sample/compose/nestedrenderings/LegacyRunner.kt
```diff
@@ -0,0 +1,39 @@
...
+class LegacyRunner(private val binding: LegacyViewBinding) : LayoutRunner<LegacyRendering> {
+
+  override fun showRendering(
+    rendering: LegacyRendering,
+    viewEnvironment: ViewEnvironment
+  ) {
+    binding.stub.update(rendering.rendering, viewEnvironment)
+  }
+
+  companion object : ViewFactory<LegacyRendering> by bind(
```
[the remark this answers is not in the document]
zach-klippenstein: Oh yea, definitely. I just copy-pasted this sample from a year ago.

samples/compose-samples/src/main/java/com/squareup/sample/compose/nestedrenderings/RecursiveViewFactory.kt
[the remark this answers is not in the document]
zach-klippenstein: Probably. I am planning to do a grooming pass once all the separate PRs have merged to make sure everything is using the right things.

samples/compose-samples/src/main/java/com/squareup/sample/compose/nestedrenderings/NestedRenderingsActivity.kt
```diff
@@ -0,0 +1,59 @@
...
+@OptIn(WorkflowUiExperimentalApi::class)
+private val viewRegistry = ViewRegistry(
+  RecursiveViewFactory,
+  LegacyRunner
+)
+
+@OptIn(WorkflowUiExperimentalApi::class)
+private val viewEnvironment =
+  ViewEnvironment(mapOf(ViewRegistry to viewRegistry)).withCompositionRoot { content ->
+    CompositionLocalProvider(LocalBackgroundColor provides Color.Green) {
```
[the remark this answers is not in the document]
zach-klippenstein: [`ProvidableCompositionLocal`](https://developer.android.com/reference/kotlin/androidx/compose/runtime/ProvidableCompositionLocal?authuser=1#provides(kotlin.Any)), aka what gets returned from `compositionLocalOf()`, aka what all the `Local*` vals are typed.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/WorkflowRendering.kt
rjrjr: Oh hell yeah. I really like `.asComposeViewFactory()`, sold.

samples/compose-samples/src/main/java/com/squareup/sample/compose/nestedrenderings/NestedRenderingsActivity.kt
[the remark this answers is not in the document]
steve-the-edwards: I think - https://developer.android.com/reference/kotlin/androidx/compose/runtime/ProvidableCompositionLocal#provides(kotlin.Any) 

@zach-klippenstein can correct me if I'm wrong.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/WorkflowRendering.kt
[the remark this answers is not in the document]
zach-klippenstein: WVS seems to think the contract of `ViewRegistry` is that it will always return the same factory for the total set of all renderings that are compatible with each other, and only asks for a new factory if that changes. That is:
```kotlin
if (view.canShowRendering(newRendering)) {
  view.showRendering(newRendering)
  return
}

// Else, ask for a new ViewFactory and initialize it.
```

I think changing this semantic could be significant, and I'd like to not do that here and instead stay consistent. One of the nice properties of this is that view factories' identify doesn't matter. E.g. the proposed `asComposeViewFactory` would return a different instance every time it's called – with the current semantic, we'd only call it once, and we would only ever care about or use the one returned from that call. With a more aggressive check, we'd be creating a new wrapper on every update, and then that wrapper would have to implement equality itself.

I think the current semantic is simpler and easier to reason about.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/WorkflowRendering.kt
zach-klippenstein: I don't think so. If we want to re-implement the containers in compose, that'll be a huge project and this will be one tiny little detail. Should I take the comment out?

rjrjr: Yeah.

samples/compose-samples/src/main/java/com/squareup/sample/compose/nestedrenderings/NestedRenderingsActivity.kt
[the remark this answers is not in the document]
zach-klippenstein: Owe me a coke.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/WorkflowRendering.kt
zach-klippenstein: I'll break out the top-shelf docs for this one

samples/compose-samples/src/androidTest/java/com/squareup/sample/compose/nestedrenderings/NestedRenderingsTest.kt
```diff
@@ -0,0 +1,69 @@
...
+          // No more reset buttons, we're done.
+          return@forEach
+        }
+        foundNodes++
+        it.performClick()
+      }
+    }
+  }
+
+  private fun SemanticsNodeInteractionCollection.forEach(
```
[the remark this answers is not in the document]
zach-klippenstein: 🤔

samples/compose-samples/src/main/java/com/squareup/sample/compose/nestedrenderings/LegacyRunner.kt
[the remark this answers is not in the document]
zach-klippenstein: Trying this, I think it might be too distracting. Gonna leave as-is.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/WorkflowRendering.kt
```diff
@@ -0,0 +1,201 @@
...
+ * This function fulfills a similar role as [WorkflowViewStub], but is much more convenient to use
+ * from Composable functions.
+ *
+ * ## Example
+ *
+ * ```
+ * data class FramedRendering(
+ *   val borderColor: Color,
+ *   val child: Any
+ * )
```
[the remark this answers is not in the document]
zach-klippenstein: Even better:
```
 * ```
 * data class FramedRendering<R : Any>(
 *   val borderColor: Color,
 *   val child: R
 * ) : ComposeRendering {
 * 
 *   @Composable override fun Content(viewEnvironment: ViewEnvironment) {
 *     Surface(border = Border(rendering.borderColor, 8.dp)) {
 *       WorkflowRendering(rendering.child, viewEnvironment)
 *     }
 *   }
 * }
 * ```
```

zach-klippenstein:
Comments addressed, merging on green.

PR #491, opened 2021-08-19
Introduce ComposeRendering, the compose version of AndroidViewRendering.

zach-klippenstein:
Part of #458.

Example:
```kotlin
data class HelloView(
  val message: String,
  val onClick: () -> Unit
) : ComposeRendering<HelloView> {

  @Composable override fun Content(viewEnvironment: ViewEnvironment) {
    Button(onClick) {
      Text(message)
    }
  }
}
```

### DIFF
--- a/workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposeRendering.kt
+++ b/workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposeRendering.kt
@@ -1,0 +1,68 @@
+package com.squareup.workflow1.ui.compose
+
+import androidx.compose.runtime.Composable
+import com.squareup.workflow1.ui.AndroidViewRendering
+import com.squareup.workflow1.ui.ViewEnvironment
+import com.squareup.workflow1.ui.ViewFactory
+import com.squareup.workflow1.ui.ViewRegistry
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+import kotlin.reflect.KClass
+
+/**
+ * Interface implemented by a rendering class to allow it to drive a composable UI via an
+ * appropriate [ComposeViewFactory] implementation, by simply overriding the [Content] method.
+ * This is the compose analog to [AndroidViewRendering].
+ *
+ * Note that unlike most workflow view functions, [Content] does not take the rendering as a
+ * parameter. Instead, the rendering is the receiver, i.e. the current value of `this`.
+ *
+ * Example:
+ *
+ * ```
+ * @OptIn(WorkflowUiExperimentalApi::class)
+ * data class HelloView(
+ *   val message: String,
+ *   val onClick: () -> Unit
+ * ) : ComposeRendering {
+ *
+ *   @Composable override fun Content(viewEnvironment: ViewEnvironment) {
+ *     Button(onClick) {
+ *       Text(message)
+ *     }
+ *   }
+ * }
+ * ```
+ *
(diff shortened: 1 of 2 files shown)


workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposeRendering.kt
```diff
@@ -0,0 +1,63 @@
...
+ * This is the simplest way to bridge the gap between your workflows and the UI, but using it
+ * requires your workflows code to reside in Android modules, instead of pure Kotlin. If this is a
+ * problem, or you need more flexibility for any other reason, you can use [ViewRegistry] to bind
+ * your renderings to [ComposeViewFactory] implementations at runtime.
+ */
+@WorkflowUiExperimentalApi
+public interface ComposeRendering<V : ComposeRendering<V>> : AndroidViewRendering<V> {
+
+  /** Don't override this, override [Content] instead. */
+  override val viewFactory: ViewFactory<V> get() = Companion
```
zach-klippenstein, line 40: I'm torn between making this a class so I can make this property `final override`, or keeping it an interface to keep it consistent with `AndroidViewRendering`. WDYT?

steve-the-edwards: `abstract class` makes sense to me if you want to enforce this.

steve-the-edwards (approved):
Though I like your suggestion to use a class rather than an interface to prevent overriding the viewFactory.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposeRendering.kt
```diff
@@ -0,0 +1,68 @@
...
+ * This is the simplest way to bridge the gap between your workflows and the UI, but using it
+ * requires your workflows code to reside in Android modules, instead of pure Kotlin. If this is a
+ * problem, or you need more flexibility for any other reason, you can use [ViewRegistry] to bind
+ * your renderings to [ComposeViewFactory] implementations at runtime.
+ */
+@WorkflowUiExperimentalApi
+public interface ComposeRendering : AndroidViewRendering<Nothing> {
+
+  /** Don't override this, override [Content] instead. */
+  override val viewFactory: ViewFactory<Nothing> get() = Companion
```
rjrjr, line 45: The ViewRegistry overhaul I've been trying to get back to since November(!) allows us to provide bindings for factories of arbitrary types, e.g. DialogFactory in addition to ViewFactory. When I finally get back to that (this quarter or bust), we can also make it vend a ComposableFactory and get rid of this `ViewFactory<Nothing>` wart.

rjrjr (approved):
🤩

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposeRendering.kt
zach-klippenstein: @steve-the-edwards votes for a class.
@rjrjr votes for… both? 😛

zach-klippenstein: FTR, the only really material argument I could think of for keeping it an interface was that renderings might be defined as sealed classes, and if only some cases would want to implement this, they'd need it to be an interface. But I think we actually discourage sealed classes for renderings, and even if someone wants to do that we have sealed interfaces now so it's not a problem.

PR #493, opened 2021-08-19
Introduce renderAsState, a handy way to run the workflow runtime from composables.

zach-klippenstein:
Part of #458.

`renderAsState` is a wrapper around the main workflow runtime entry point, `renderWorkflowIn`, that manages saving and restoring snapshot states as well as updating props and getting renderings and outputs from the workflow.

It's designed to follow the compose pattern for subscribing to reactive types: `Flow.collectAsState()`, `LiveData.observeAsState()`, `Observable.subscribeAsState()`.

Example:
```kotlin
@Composable fun App(workflow: Workflow<…>, props: Props) {
  val scaffoldState = …

  // Run the workflow in the current composition's coroutine scope.
  val rendering by workflow.renderAsState(props, onOutput = { output ->
    // Note that onOutput is a suspend function, so you can run animations
    // and call other suspend functions.
    scaffoldState.snackbarHostState
      .showSnackbar(output.toString())
  })
  val viewEnvironment = remember {
    ViewEnvironment(mapOf(ViewRegistry to appViewRegistry))
  }

  Scaffold(…) { padding ->
    // Display the root rendering using the view environment's ViewRegistry.
    WorkflowRendering(rendering, viewEnvironment, Modifier.padding(padding))
  }
}
```

### DIFF
--- a/workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/RenderAsState.kt
+++ b/workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/RenderAsState.kt
@@ -1,0 +1,215 @@
+@file:OptIn(ExperimentalWorkflowApi::class)
+
+package com.squareup.workflow1.ui.compose
+
+import androidx.annotation.VisibleForTesting
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.MutableState
+import androidx.compose.runtime.RememberObserver
+import androidx.compose.runtime.SideEffect
+import androidx.compose.runtime.State
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.rememberCoroutineScope
+import androidx.compose.runtime.rememberUpdatedState
+import androidx.compose.runtime.saveable.LocalSaveableStateRegistry
+import androidx.compose.runtime.saveable.Saver
+import androidx.compose.runtime.saveable.SaverScope
+import androidx.compose.runtime.saveable.rememberSaveable
+import com.squareup.workflow1.ExperimentalWorkflowApi
+import com.squareup.workflow1.Snapshot
+import com.squareup.workflow1.TreeSnapshot
+import com.squareup.workflow1.Workflow
+import com.squareup.workflow1.WorkflowInterceptor
+import com.squareup.workflow1.renderWorkflowIn
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.launchIn
+import kotlinx.coroutines.flow.onEach
+import okio.ByteString
+
+/**
+ * Runs this [Workflow] as long as this composable is part of the composition, and returns a
+ * [State] object that will be updated whenever the runtime emits a new [RenderingT]. Note that
+ * here, and in the rest of the documentation for this class, the "`State`" type refers to Compose's
+ * snapshot [State] type, _not_ the concept of the `StateT` type in a particular workflow.
+ *
+ * The workflow runtime will be started when this function is first added to the composition, and
+ * cancelled when it is removed or if the composition fails. The first rendering will be available
+ * immediately as soon as this function returns as [State.value]. Composables that read this value
+ * will automatically recompose whenever the runtime emits a new rendering. If you are driving UI
+ * from the Workflow tree managed by [renderAsState] then you will probably want to pass the
+ * returned [State]'s value (which is the Workflow rendering) to the [WorkflowRendering] composable.
+ *
+ * Note that the initial render pass will occur on whatever thread this function is called from.
+ * That may be a background thread, as Compose supports performing composition on background
+ * threads. Well-behaved workflows should have pure `initialState` and `render` functions, so this
+ * should not be a problem. Any side effects performed by workflows using the `runningSideEffect`
+ * method or Workers will be executed in [scope] as usual.
+ *
+ * [Snapshot]s from the runtime will automatically be saved and restored using Compose's
+ * [rememberSaveable].
+ *
+ * ## Example
+ *
+ * ```
+ * private val appViewRegistry = ViewRegistry(…)
+ *
+ * @Composable fun App(workflow: Workflow<…>, props: Props) {
+ *   val scaffoldState = …
+ *
+ *   // Run the workflow in the current composition's coroutine scope.
+ *   val rendering by workflow.renderAsState(props, onOutput = { output ->
+ *     // Note that onOutput is a suspend function, so you can run animations
+ *     // and call other suspend functions.
+ *     scaffoldState.snackbarHostState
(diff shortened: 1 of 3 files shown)


workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/RenderAsState.kt
```diff
@@ -0,0 +1,206 @@
...
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.launchIn
+import kotlinx.coroutines.flow.onEach
+import okio.ByteString
+
+/**
+ * Runs this [Workflow] as long as this composable is part of the composition, and returns a
+ * [State] object that will be updated whenever the runtime emits a new [RenderingT].
```
steve-the-edwards, line 35: Might want to be explicit about the overloading of State here. "returns a [State] object not in the Workflow sense but in the Compose sense"

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/RenderAsState.kt
```diff
@@ -0,0 +1,206 @@
...
+ *
+ * The workflow runtime will be started when this function is first added to the composition, and
+ * cancelled when it is removed or if the composition fails. The first rendering will be available
+ * immediately as soon as this function returns as [State.value]. Composables that read this value
+ * will automatically recompose whenever the runtime emits a new rendering. In most cases you
+ * probably want to pass the rendering value to the [WorkflowRendering] composable.
```
steve-the-edwards, line 41:
```suggestion
 * will automatically recompose whenever the runtime emits a new rendering. If you are driving UI form the Workflow tree invoked by [renderAsState] then you will
 * probably want to pass the State's value (the Workflow rendering) to the [WorkflowRendering] composable.
```

steve-the-edwards: There is just so much overloading going on between Compose/Workflow terms that I don't think you could be over explicit.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/RenderAsState.kt
```diff
@@ -0,0 +1,206 @@
...
+ * @param interceptors
+ * An optional list of [WorkflowInterceptor]s that will wrap every workflow rendered by the runtime.
+ * Interceptors will be invoked in 0-to-`length` order: the interceptor at index 0 will process the
+ * workflow first, then the interceptor at index 1, etc.
+ * @param scope
+ * The [CoroutineScope] in which to launch the workflow runtime. If not specified, the value of
+ * [rememberCoroutineScope] will be used. Any exceptions thrown in any workflows, after the initial
+ * render pass, will be handled by this scope, and cancelling this scope will cancel the workflow
+ * runtime and any running workers. Note that any dispatcher in this scope will _not_ be used to
+ * execute the very first render pass.
```
steve-the-edwards, line 85: which will instead be executed on the ... main dispatcher (?) ...?

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/RenderAsState.kt
```diff
@@ -0,0 +1,206 @@
...
+      onOutput = { updatedOnOutput(it) }
+    ).apply {
+      start(workflow, interceptors)
+    }
+  }
+
+  // Use a side effect to update props so that it waits for the composition to commit.
+  SideEffect {
+    state.setProps(props)
+  }
```
steve-the-edwards, line 138: So the rendering returned synchronously here is not based on the `initialProps`. Isn't that a problem?

steve-the-edwards: Oh wait, no they are passed to the State object. So why do we have to update the flow again after the composition committing?

steve-the-edwards (approved):
A couple questions.

rjrjr:
I _really_ want to use this in prod. 🤤

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/RenderAsState.kt
```diff
@@ -0,0 +1,206 @@
...
+  // committed, which will be after this function returns, and we need to run this immediately so we
+  // get the rendering synchronously. The thread running this composition might also not be the
+  // main thread or whatever thread the workflow context is configured to run on, but that should
+  // be fine as long as the workflows are correctly performing side effects in effects and not their
+  // render or related methods.
+  // The WorkflowState object remembered here is a RememberObserver – it will automatically cancel
+  // the workflow runtime when it leaves the composition or if the composition doesn't commit.
+  // The remember is keyed on any values that we can't update the runtime with dynamically, and
+  // therefore require completely restarting the runtime to take effect.
+  val state = remember(workflow, scope, interceptors) {
```
rjrjr, line 124: Everything that was hard about bootstrapping, Compose is making seamless. Thank God you came up with making the first rendering synchronous.

rjrjr (approved):

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/RenderAsState.kt
rjrjr: I think it'll be executed by whatever is running the caller, no?

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/RenderAsState.kt
zach-klippenstein: TBH I'm still not sure this won't blow up on us. I would be honestly quite surprised if, with all the workflows we've written internally at this point, at least one isn't doing something in its `initialState` method that requires being on the main thread.

rjrjr: Wouldn't want anyone to get bored.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/RenderAsState.kt
zach-klippenstein: The caller can pass in new props at any time.

zach-klippenstein:
Should we put this in a dedicated `compose-runtime` module? As-is, this requires the workflow runtime to be a transitive dependency of any consumers that want to just define some `ComposeViewFactory`s.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/RenderAsState.kt
zach-klippenstein: Yep. I'll make that explicit.

zach-klippenstein:
Comments addressed, merging on green.

PR #494, opened 2021-08-19
Move the rest of the samples over from the compose project.

zach-klippenstein:
Part of #458.

### DIFF
--- a/samples/compose-samples/src/androidTest/java/com/squareup/sample/compose/hellocompose/HelloComposeTest.kt
+++ b/samples/compose-samples/src/androidTest/java/com/squareup/sample/compose/hellocompose/HelloComposeTest.kt
@@ -1,0 +1,27 @@
+package com.squareup.sample.compose.hellocompose
+
+import androidx.compose.ui.test.assertIsDisplayed
+import androidx.compose.ui.test.junit4.createAndroidComposeRule
+import androidx.compose.ui.test.onNodeWithText
+import androidx.compose.ui.test.performClick
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import org.junit.Rule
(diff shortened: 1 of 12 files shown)


workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/RenderAsState.kt
```diff
@@ -185,7 +204,13 @@ private class WorkflowRuntimeState<PropsT, OutputT : Any, RenderingT>(
...
         snapshotState.value = it.snapshot
       }
-      .launchIn(workflowScope)
+      // We collect the renderings in the workflowScope to participate in structured concurrency,
+      // however we don't need to use its dispatcher – this coroutine is simply setting snapshot
+      // state values, which is threadsafe.
+      // Also, if the scope uses a non-immediate dispatcher, the initial states won't get set until
+      // the dispatcher dispatchers the collection coroutine, but our contract requires them to be
+      // set by the time this function returns.
+      .launchIn(workflowScope + Dispatchers.Unconfined)
```
steve-the-edwards, line 213: These seem to be important changes buried in this "migrate samples" PR. Could you please add some notation in the commit history?

steve-the-edwards (approved):

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/RenderAsState.kt
zach-klippenstein: Oops, I forgot to retarget this pr. It's based on this branch: https://github.com/square/workflow-kotlin/pull/503

rjrjr (approved):

PR #495, opened 2021-08-19
Introduce a sample for the ComposeWorkflow concept.

zach-klippenstein:
Part of #458.

~This is pretty weird and sketchy, probably don't want to merge, but wanted to get it up for discussion anyway.~

This is only a sample now.

### DIFF
--- a/samples/compose-samples/src/main/java/com/squareup/sample/compose/hellocomposeworkflow/ComposeWorkflow.kt
+++ b/samples/compose-samples/src/main/java/com/squareup/sample/compose/hellocomposeworkflow/ComposeWorkflow.kt
@@ -1,0 +1,71 @@
+package com.squareup.sample.compose.hellocomposeworkflow
+
+import androidx.compose.runtime.Composable
+import com.squareup.workflow1.RenderContext
+import com.squareup.workflow1.Sink
+import com.squareup.workflow1.StatefulWorkflow
+import com.squareup.workflow1.Workflow
+import com.squareup.workflow1.ui.ViewEnvironment
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+import com.squareup.workflow1.ui.compose.ComposeRendering
+
+/**
+ * A stateless [Workflow] that [renders][RenderingContent] itself as a [Composable] function.
+ * Effectively defines an inline [ComposeRendering].
+ *
+ * This workflow does not have access to a [RenderContext] since render contexts are only valid
+ * during render passes, and this workflow's [RenderingContent] method is invoked after the render
+ * pass, when view bindings are being shown.
+ *
+ * While this workflow is "stateless" in the usual workflow sense (it doesn't have a `StateT` type),
+ * since [RenderingContent] is a Composable function, it can use all the usual Compose facilities
+ * for state management.
+ *
+ * # CAVEAT
+ *
+ * We don't know if this is a useful/helpful pattern, or a pattern that will be confusing and cause
+ * trouble. However, it seems interesting, so we wanted to be able to point to it if a use case
+ * comes up.
+ */
+@WorkflowUiExperimentalApi
+public abstract class ComposeWorkflow<in PropsT, out OutputT : Any> :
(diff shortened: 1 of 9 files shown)


rjrjr:
What would you think about making this a sample?

rjrjr:
Yeah, I get your ambivalence. I really want to play with this; have no idea if it's a good idea.

Nice thing about making it a sample is that it stays at least a bit visible w/o support being implied. We can loop back and play with it, see if others do.

zach-klippenstein:
I'd be good with making it a sample.

samples/compose-samples/src/main/java/com/squareup/sample/compose/hellocomposeworkflow/ComposeWorkflow.kt
```diff
@@ -0,0 +1,65 @@
...
+
+  override fun asStatefulWorkflow(): StatefulWorkflow<PropsT, *, OutputT, ComposeRendering> =
+    ComposeWorkflowImpl(this)
+}
+
+/**
+ * Returns a [ComposeWorkflow] that renders itself using the given [render] function.
+ */
+@WorkflowUiExperimentalApi
+public inline fun <PropsT, OutputT : Any> Workflow.Companion.composed(
```
steve-the-edwards, line 51: What's the use case for this extension?

steve-the-edwards (approved):

samples/compose-samples/src/main/java/com/squareup/sample/compose/hellocomposeworkflow/ComposeWorkflow.kt
rjrjr: Same as `Workflow.Companion.stateless` and '.stateful`. Lets us create anonymous Workflow instances.

samples/compose-samples/src/main/java/com/squareup/sample/compose/hellocomposeworkflow/HelloComposeWorkflow.kt
```diff
@@ -0,0 +1,49 @@
...
+          .wrapContentSize(Alignment.Center)
+      )
+    }
+  }
+}
+
+@OptIn(WorkflowUiExperimentalApi::class)
+@Preview(showBackground = true)
+@Composable fun HelloComposeWorkflowPreview() {
+  // TODO(#458) Uncomment once preview support is imported.
```
rjrjr, line 47: Didn't you already merge this?

samples/compose-samples/src/main/java/com/squareup/sample/compose/hellocomposeworkflow/HelloComposeWorkflowActivity.kt
```diff
@@ -0,0 +1,40 @@
...
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+    val model: HelloComposeModel by viewModels()
+    setContentView(
+      WorkflowLayout(this).apply {
+        start(
+          renderings = model.renderings,
+          registry = ViewRegistry()
+        )
+      }
```
rjrjr, line 26:
```suggestion
      WorkflowLayout(this).apply { start(model.renderings) }
```

rjrjr (approved):

zach-klippenstein:
Test failure: 
```
com.squareup.sample.compose.launcher.SampleLauncherTest > allSamplesLaunch[test(AVD) - 7.0] FAILED 
	java.lang.AssertionError: Failed to launch sample Inline ComposeRendering
	at com.squareup.sample.compose.launcher.SampleLauncherTest.allSamplesLaunch(SampleLauncherTest.kt:31)
```

It's passing locally on my 29 emulator though, so might be a version-specific thing. Testing on a lower emulator…

zach-klippenstein:
Ah, it's probably because the screen size/density is different and the last sample is now hidden below the viewport. Need to scroll to it in the test.

PR #496, opened 2021-08-19
wip trying some more fixes

zach-klippenstein:


workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLifecycleOwner.kt
```diff
@@ -118,6 +119,17 @@ internal class RealWorkflowLifecycleOwner(
...
     if (enforceMainThread) LifecycleRegistry(this) else createUnsafe(this)
 
+  /**
+   * We can't rely on [localLifecycle] to know if we've actually attempted to destroy the lifecycle,
+   * because there's a case where we can be about to destroy our lifecycle but we're still in the
+   * INITIALIZED state so we just stay there (because that's an invalid transition). This flag lets
+   * us determine if we've done that and should skip work in onAttached.
+   *
+   * Maybe this means that [destroyOnDetach] is too eager, and should actually wait for _attach_ and
+   * _then_ detach?
```
zach-klippenstein, line 129: I don't think this is true. I'm not sure why these views are re-attaching after their lifecycle has been destroyed, but really that shouldn't affect the correctness of this class – once destroyed, always destroyed, the view's attachments are its own business.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLifecycleOwner.kt
```diff
@@ -229,18 +247,20 @@ internal class RealWorkflowLifecycleOwner(
         parentLifecycle?.removeObserver(this)
         parentLifecycle = null
 
-        // We can't change state anymore, so we don't care about watching for new parents.
-        view.get()?.removeOnAttachStateChangeListener(this)
```
zach-klippenstein, line 233: I think we can restore this, can't hurt to best-effort try to remove this listener.

PR #497, opened 2021-08-20
Fixes leak in WorkflowLifecycleOwner.installOn

rjrjr:
The lambda created by `WorkflowLifecycleOwner.installOn` was needlessly
holding a reference to the given view. The fix is to change the
`findParentLifecycle` function to expect a `View` argument, to be passed at
attach time.


rjrjr:
@zach-klippenstein I wonder if we really need that `WeakReference`?

rjrjr:
> @zach-klippenstein I wonder if we really need that `WeakReference`?

I'm not convinced that we do. I'm testing dropping that change and the others in favor of this one.

rjrjr:
I'm going to close this PR and instead put up a new branch that includes just the memory leak fixes, so that we can make an alpha19 release that doesn't bring along the more drastic state management work.

rjrjr:
Closed in favor of #498

Issue #472, opened 2021-08-17
Add LeakCanary in UI test runs 


zach-klippenstein:
We're lucky enough to have extensive internal testing that runs LeakCanary and so this library effectively gets tested via that, but it would be nice to have our own leak tests.

rjrjr:
SavedStateRegistry support is particularly plagued by such issues. e.g. see [this comment](https://github.com/square/workflow-kotlin/pull/486#discussion_r710556948) on #486.

rjrjr:
#561 introduces Leakcanary, but is blocked until https://issuetracker.google.com/issues/202190483 is fixed, or we can find a work around.

PR #498, opened 2021-08-20
Fixes memory leak in WorkflowLifecycleOwner.

rjrjr:
Hold a pointer to the `View` only while it is attached. Pass a `View` param to
the `findParentLifecycle` function to ensure it doesn't need to hold a
leak-prone ref either.

These fixes are a more focussed distillation of the handful that landed on
`main` in the last few days. There are no changes related to lifecycle, and
we no longer worry about nulling out the lambda.

Note that this PR will merge into a new `v1.0.0-alpha19` branch, which was cut from the `v1.0.0-alpha18` tag.


zach-klippenstein (approved):
You didn't need the `WeakReference`? I think I will have to tweak this more to make the VTSSRO stuff work, maybe that's what I needed the WR for. Anyway, fine with this if it plugs the leak for now.

zach-klippenstein:
For tests, see #472.

PR #503, opened 2021-08-21
Fix some bugs in renderAsState when the workflow scope dispatcher is non-immediate.

zach-klippenstein:
See the test and the new comments for more information.

### DIFF
--- a/workflow-ui/compose/src/androidTest/java/com/squareup/workflow1/ui/compose/RenderAsStateTest.kt
+++ b/workflow-ui/compose/src/androidTest/java/com/squareup/workflow1/ui/compose/RenderAsStateTest.kt
@@ -1,11 +1,15 @@
+@file:OptIn(ExperimentalCoroutinesApi::class)
+
 package com.squareup.workflow1.ui.compose
 
 import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.runtime.State
+import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.runtime.saveable.LocalSaveableStateRegistry
 import androidx.compose.runtime.saveable.SaveableStateRegistry
+import androidx.compose.runtime.setValue
 import androidx.compose.ui.test.junit4.StateRestorationTester
 import androidx.compose.ui.test.junit4.createComposeRule
 import androidx.test.ext.junit.runners.AndroidJUnit4
@@ -16,17 +20,26 @@ import com.squareup.workflow1.Workflow
 import com.squareup.workflow1.action
 import com.squareup.workflow1.parse
 import com.squareup.workflow1.readUtf8WithLength
+import com.squareup.workflow1.rendering
 import com.squareup.workflow1.stateless
 import com.squareup.workflow1.ui.compose.RenderAsStateTest.SnapshottingWorkflow.SnapshottedRendering
 import com.squareup.workflow1.writeUtf8WithLength
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.awaitCancellation
+import kotlinx.coroutines.isActive
+import kotlinx.coroutines.job
+import kotlinx.coroutines.test.TestCoroutineScope
 import okio.ByteString
 import okio.ByteString.Companion.decodeBase64
+import org.junit.Ignore
 import org.junit.Rule
 import org.junit.Test
 import org.junit.runner.RunWith
+import kotlin.test.assertFailsWith
 
 @RunWith(AndroidJUnit4::class)
-class RenderAsStateTest {
+internal class RenderAsStateTest {
 
   @get:Rule val composeRule = createComposeRule()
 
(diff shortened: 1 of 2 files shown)


workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/RenderAsState.kt
```diff
@@ -185,7 +204,13 @@ private class WorkflowRuntimeState<PropsT, OutputT : Any, RenderingT>(
         renderingState.value = it.rendering
         snapshotState.value = it.snapshot
       }
-      .launchIn(workflowScope)
+      // We collect the renderings in the workflowScope to participate in structured concurrency,
+      // however we don't need to use its dispatcher – this coroutine is simply setting snapshot
+      // state values, which is threadsafe.
+      // Also, if the scope uses a non-immediate dispatcher, the initial states won't get set until
+      // the dispatcher dispatchers the collection coroutine, but our contract requires them to be
```
zach-klippenstein, line 211:
```suggestion
      // the dispatcher dispatches the collection coroutine, but our contract requires them to be
```

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/RenderAsState.kt
```diff
@@ -185,7 +204,13 @@ private class WorkflowRuntimeState<PropsT, OutputT : Any, RenderingT>(
...
+      // set by the time this function returns.
```
zach-klippenstein, line 212:
```suggestion
      // set by the time this function returns and using the Unconfined dispatcher guarantees that.
```

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/RenderAsState.kt
```diff
@@ -185,7 +204,13 @@ private class WorkflowRuntimeState<PropsT, OutputT : Any, RenderingT>(
         renderingState.value = it.rendering
         snapshotState.value = it.snapshot
       }
-      .launchIn(workflowScope)
+      // We collect the renderings in the workflowScope to participate in structured concurrency,
+      // however we don't need to use its dispatcher – this coroutine is simply setting snapshot
```
zach-klippenstein, line 208:
```suggestion
      // however we don't need to use its dispatcher – this collector is simply setting snapshot
```

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/RenderAsState.kt
```diff
@@ -185,7 +204,13 @@ private class WorkflowRuntimeState<PropsT, OutputT : Any, RenderingT>(
...
+      // state values, which is threadsafe.
```
zach-klippenstein, line 209:
```suggestion
      // state values, which is thread safe.
```

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/RenderAsState.kt
```diff
@@ -162,6 +175,12 @@ private class WorkflowRuntimeState<PropsT, OutputT : Any, RenderingT>(
   private val renderingState = mutableStateOf<RenderingT?>(null)
   private val propsFlow = MutableStateFlow(initialProps)
 
+  /**
+   * The actual scope used to run the workflow. It has a child [Job] of the incoming scope so
+   * we can cancel the runtime without cancelling the incoming scope.
+   */
+  private val workflowScope = workflowScope + Job(parent = workflowScope.coroutineContext.job)
```
zach-klippenstein, line 182: The `job` extension property throws if the context has no job. In this case, we don't actually care if the parent doesn't have a job – in that case, this will be the root job, and that's fine.
```suggestion
  private val workflowScope = workflowScope + Job(parent = workflowScope.coroutineContext[Job])
```

workflow-ui/compose/src/androidTest/java/com/squareup/workflow1/ui/compose/RenderAsStateTest.kt
```diff
@@ -212,6 +225,107 @@ class RenderAsStateTest {
...
+
+    composeRule.setContent {
+      val initialRendering = workflow.renderAsState(
+        props = Unit, onOutput = {},
+        scope = scope
+      )
+      assertThat(initialRendering.value).isNotNull()
+    }
+
+    scope.cleanupTestCoroutines()
```
vgonda, line 243: This cleanup wont happen if the test fails on an earlier line. Same for tests below.

If you don't want to set up the scope with `@Before`/`@After`, one pattern I've used in tests before for clean up is `try`/`finally`

vgonda (approved):
I don't know much about workflow, but from your comments, it looks like this solution makes sense!

workflow-ui/compose/src/androidTest/java/com/squareup/workflow1/ui/compose/RenderAsStateTest.kt
zach-klippenstein: I think if the test fails this scope is already going to have been torn down? That might not be true though and it's better not to rely on it. 

- [ ] Fix coroutines cleanup

steve-the-edwards (approved):

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/RenderAsState.kt
```diff
@@ -162,6 +175,12 @@ private class WorkflowRuntimeState<PropsT, OutputT : Any, RenderingT>(
   private val renderingState = mutableStateOf<RenderingT?>(null)
   private val propsFlow = MutableStateFlow(initialProps)
 
+  /**
+   * The actual scope used to run the workflow. It has a child [Job] of the incoming scope so
+   * we can cancel the runtime without cancelling the incoming scope.
+   */
+  private val workflowScope = workflowScope + Job(parent = workflowScope.coroutineContext[Job])
```
rjrjr, line 182: Nice catch.

rjrjr (approved):

PR #507, opened 2021-08-22
Add a sample demonstrating returning an inline ComposeRendering from a workflow.

zach-klippenstein:
Kapture 2021-08-22 at 13 04 31

### DIFF
--- a/samples/compose-samples/src/androidTest/java/com/squareup/sample/compose/inlinerendering/InlineRenderingTest.kt
+++ b/samples/compose-samples/src/androidTest/java/com/squareup/sample/compose/inlinerendering/InlineRenderingTest.kt
@@ -1,0 +1,42 @@
+package com.squareup.sample.compose.inlinerendering
+
+import androidx.compose.ui.test.assertIsDisplayed
+import androidx.compose.ui.test.assertTextEquals
+import androidx.compose.ui.test.hasClickAction
+import androidx.compose.ui.test.junit4.createAndroidComposeRule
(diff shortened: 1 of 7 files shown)


samples/compose-samples/src/androidTest/java/com/squareup/sample/compose/inlinerendering/InlineRenderingTest.kt
```diff
@@ -0,0 +1,42 @@
...
+    composeRule.mainClock.autoAdvance = false
+
+    composeRule.onNode(hasClickAction())
+      .performClick()
+
+    composeRule.mainClock.advanceTimeByFrame()
+
+    // During the animation, both counter values will be present.
+    composeRule.onNode(hasClickAction())
+      .assertTextEquals("Counter: ", "0", "1")
```
steve-the-edwards, line 40: Wow. Interesting way to test this.

steve-the-edwards (approved):

samples/compose-samples/src/androidTest/java/com/squareup/sample/compose/inlinerendering/InlineRenderingTest.kt
zach-klippenstein: Yea, not sure it's the best way. Seemed a bit more explicit than individually asserting `onNodeWithText(digit).assertIsDisplayed()` for "0" and "1".

rjrjr (approved):

PR #517, opened 2021-08-23
Fix the compose artifacts: don't strip metadata!

zach-klippenstein:
Verified by publishing to maven local and consuming from a separate project.


rjrjr (approved):

zach-klippenstein:
This went green just now except for one UI test shard that timed out. Fixing in #519, but this PR isn't going to be tested by our UI tests anyway so merging now.

PR #544, opened 2021-09-07
Builds in NamedViewFactory properly.

### LINKED ISSUE #538: Is `NamedViewFactory` included automatically or isn't it?
https://workflow-community.slack.com/archives/CHTFPR277/p1630599181038900?thread_ts=1630549583.028900&cid=CHTFPR277

rjrjr:
`ViewRegistry.defaultViewFactories` has never worked very well, and it
completely failed `WorkflowRendering()`. We get rid of it, and instead ensure
that `NamedViewFactory` is always available via the same hook that makes
`AndroidViewRendering` work.

Fixes #538

### DIFF
--- a/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewRegistry.kt
+++ b/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewRegistry.kt
@@ -8,12 +8,6 @@ import android.view.ViewGroup
 import kotlin.reflect.KClass
 
 /**
- * [ViewFactory]s that are always available.
- */
-@WorkflowUiExperimentalApi
-internal val defaultViewFactories = ViewRegistry(NamedViewFactory)
-
-/**
  * The [ViewEnvironment] service that can be used to display the stream of renderings
  * from a workflow tree as [View] instances. This is the engine behind [AndroidViewRendering],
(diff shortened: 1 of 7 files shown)


rjrjr:
My God, I hate UI tests. Rolling the dice again…

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewRegistry.kt
```diff
@@ -137,6 +131,7 @@ public fun <RenderingT : Any>
   @Suppress("UNCHECKED_CAST")
   return getFactoryFor(rendering::class)
     ?: (rendering as? AndroidViewRendering<*>)?.viewFactory as? ViewFactory<RenderingT>
+    ?: (rendering as? Named<*>)?.let { NamedViewFactory as ViewFactory<RenderingT> }
```
zach-klippenstein, line 134: I just realized this wrapping pattern (even before this change) might introduce an extra view unnecessarily when the wrapped rendering's factory is a ComposeViewFactory. It shouldn't affect correctness but something we should probably check/test and to keep in mind as yall adopt compose. 

Filed: https://github.com/square/workflow-kotlin/issues/546

zach-klippenstein (approved):

PR #547, opened 2021-09-08
Introduces Wrapper

rjrjr:
We make a lot of use of renderings that wrap others to add meta information or
behavior, but which need to delegate for UI construction. This is tricky.

* We frequently forget to implement `Compatible`
* Things like `NamedViewFactory` or tricky `DecorativeViewFactory` implementations
  have to be added to `ViewRegistry`. Besides being fussy, these are platform specific.
* Introspective tasks like logging wind up with fragile code to peek
  through known wrappers.

The Wrapper interface and supporting extension functions should provide a
platform neutral way around this, and I hope pave the way to move
`ViewRegistry`(or a replacement) from `core-android` to `core-common`.

### DIFF
--- a/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewRegistry.kt
+++ b/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewRegistry.kt
@@ -128,10 +128,9 @@ public fun ViewRegistry(): ViewRegistry = TypedViewRegistry()
 @WorkflowUiExperimentalApi
 public fun <RenderingT : Any>
   ViewRegistry.getFactoryForRendering(rendering: RenderingT): ViewFactory<RenderingT> {
-  val unwrapped = unwrap(rendering)
   @Suppress("UNCHECKED_CAST")
-  return getFactoryFor(unwrapped::class)
-    ?: (unwrapped as? AndroidViewRendering<*>)?.viewFactory as? ViewFactory<RenderingT>
+  return getFactoryFor(rendering::class)
+    ?: (rendering as? AndroidViewRendering<*>)?.viewFactory as? ViewFactory<RenderingT>
     ?: throw IllegalArgumentException(
       "A ${ViewFactory::class.qualifiedName} should have been registered to display " +
         "${rendering::class.qualifiedName} instances, or that class should implement " +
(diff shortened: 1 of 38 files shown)


workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/Wrapper.kt
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * A rendering that wraps another, to which it delegates for UI construction duties.
+ */
+@WorkflowUiExperimentalApi
+public interface Wrapper<W : Any> : Compatible {
+  public val wrapped: W
+
+  override val compatibilityKey: String
+    get() = Compatible.keyFor(unwrap())
```
rjrjr, line 11: This is a mistake. Wrapped(Foo) should not be compatible with Foo.

rjrjr: Oh, no, it's fine. `Compatible` only devolves to the key if the types match.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewRegistry.kt
```diff
@@ -130,8 +130,7 @@ public fun <RenderingT : Any>
   ViewRegistry.getFactoryForRendering(rendering: RenderingT): ViewFactory<RenderingT> {
   @Suppress("UNCHECKED_CAST")
   return getFactoryFor(rendering::class)
-    ?: (rendering as? AndroidViewRendering<*>)?.viewFactory as? ViewFactory<RenderingT>
-    ?: (rendering as? Named<*>)?.let { NamedViewFactory as ViewFactory<RenderingT> }
+    ?: (unwrap(rendering) as? AndroidViewRendering<*>)?.viewFactory as? ViewFactory<RenderingT>
```
rjrjr, line 133: This is tricky, and I think wrong so far. A lot of wrappers exist precisely because they want to change how the view is built, like the BackButtonScreen touched by this PR. If a `BackButtonScreen` gets wrapped by something else, and in turn wraps and `AndroidViewRendering`, this would skip `BackButtonViewFactory` entirely.

This needs to be more visitor-like. Have to be careful not to over engineer it.

rjrjr: Maybe I'm over thinking this. `BackStackScreen` shouldn't implement `Wrapper` precisely because it doesn't entirely delegate UI work. And so we should unwrap `rendering` before calling `getFactoryFor`.

rjrjr: Yeah, that makes sense — on a non-Android UI trying to use `BackButtonScreen`, we would get a failed binding error, exactly the right thing to remind us that we need a non-Android implementation.

rjrjr:
It's more complicated than this. We want to work with the wrapped rendering when building and updating the view, but the parent needs to be able to call showRendering / getRendering with the wrapping type.

Still working on it.

PR #555, opened 2021-09-29
Hello TextController

rjrjr:
Three commits for your reviewing pleasure:

* Update of samples/todo-android in anticipation of TextController
* Introduces TextController, deprecates alternatives.
* Convert Todo sample to use TextController.

The tutorials are stll using the broken API, and can't be updated until the next release is published.

After updating the Todo sample to use it, I’m pretty happy with the API, even with these caveats that I just can’t see any practical way to address:

* It’s still pretty wiggy that this makes a workflow’s state mutable
* It’s sad that it makes participating workflows so very Android specific

Also, TextController is completely tied to classic EditText. I want follow up with a PR that adds a Compose sample that uses it. No, I don't know what that looks like yet. If anyone else is feeling motivated to get there first, go for it!


workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/TextController.kt
```diff
@@ -0,0 +1,166 @@
...
+   */
+  private val _textValue: MutableStateFlow<String> = MutableStateFlow(initialValue)
+
+  /**
+   * A [Flow] that emits the text value whenever it changes -- and only when it changes, the current value
+   * is not provided at subscription time. Workflows can safely observe changes by
+   * converting this value to a worker. (When using multiple instances, remember to provide unique
+   * key values to each `asWorker` call.)
+   *
+   * IF you can do processing that doesn't require running a `WorkflowAction` or triggering a render
```
Kernald, line 72: Nit: s/IF/if

Sripadmanabans:
I saw that you had change the mockito library in this PR to `com.nhaarman.mockitokotlin2:mockito-kotlin`. It looks like the library has been integrated into mocktio `https://github.com/mockito/mockito-kotlin`. Sorry if this is already common knowledge.

rjrjr:
> I saw that you had change the mockito library in this PR to `com.nhaarman.mockitokotlin2:mockito-kotlin`. It looks like the library has been integrated into mocktio `https://github.com/mockito/mockito-kotlin`. Sorry if this is already common knowledge.

News to me, thanks.

Kernald:
What's the idea for Compose? Just using `TextController.textValue` for both `value` and `onValueChange` for a `TextField`, still avoiding render passes if the workflow is not directly re-rendering on text changes?

rjrjr:
> What's the idea for Compose? Just using `TextController.textValue` for both `value` and `onValueChange` for a `TextField`, still avoiding render passes if the workflow is not directly re-rendering on text changes?

I guess? As I read this I realize that the thing I typed in the top comment about building in an `@Composable` function to `TextController` is a pretty terrible idea if we can avoid it.

I really don't yet know what I'm doing with Compose. I wanted to get this beachhead established, and then sort out the Compose use case calmly.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewLaunchWhenAttached.kt
```diff
@@ -0,0 +1,157 @@
...
+ * If the view is attached when this function is called, the coroutine will be started synchronously
+ * with the [CoroutineStart.UNDISPATCHED] mode, which means the suspend block will execute until its
+ * first suspension point. If the view is not attached when this function is called, the [block] will
+ * be queued to be executed as soon as the view is attached.
+ *
+ * When the view becomes detached (regardless of its initial attach state), the [CoroutineScope] that
+ * the [block] is running in will be canceled. The scope will remain cancelled _even if the view is
+ * later reattached_. The [block] _will not_ be re-started if the view is re-attached.
+ *
+ * ## Dispatcher
```
rjrjr, line 44: This is gibberish, delete it.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewLaunchWhenAttached.kt
```diff
@@ -0,0 +1,157 @@
...
+private fun View.ensureAttachedScope(): AttachedScope {
+  // Makes for clearer code below.
+  val view = this
+  return (view.getTag(R.id.view_attached_coroutine_scope) as? AttachedScope)
+    // If the scope is already cancelled, then this view was detached at some point while that scope
+    // was active.
+    ?.takeIf { it.coroutineScope.isActive }
+    ?: run {
+      // Create a new scope if the previous one is used up, or there wasn't one in the first place.
+      val lifecycleOwner = checkNotNull(lifecycleOwnerFromViewTreeOrContext(this)) {
```
rjrjr, line 94: Woo hoo, `lifecycleOwnerFromViewTreeOrContext`! Thanks, @zach-klippenstein !

rjrjr:
> still avoiding render passes if the workflow is not directly re-rendering on text changes?

BTW, we're actually _not_ avoiding render passes in the Todo sample, but we're still winning. `TextContoller.control` is idempotent, and the `TextController` instances are surviving the render passes. The workflow itself never changes the text value wrapped in `TextController`, so we don't stomp the UI in mid-edit.

samples/todo-android/app/src/main/java/com/squareup/sample/todo/TodoEditorWorkflow.kt
```diff
@@ -0,0 +1,118 @@
...
+
+  override fun snapshotState(state: TodoEditingSession): Snapshot? = null
+
+  override fun render(
+    renderProps: TodoList,
+    renderState: TodoEditingSession,
+    context: RenderContext
+  ): TodoEditorScreen {
+    // Monitor the title and each row for text changes.
+    context.runningWorker(renderState.title.onTextChanged.asWorker(), "title") { textChanged }
```
steve-the-edwards, line 44: nit:
```suggestion
    context.runningWorker(renderState.title.onTextChanged.asWorker(), "title", textChanged)
```

I actually don't know whether the compiler optimizes anything here in terms of callstack but any reason why we can't avoid another anonymous lambda wrapper?

samples/todo-android/app/src/main/java/com/squareup/sample/todo/TodoEditorWorkflow.kt
```diff
@@ -0,0 +1,118 @@
...
+    renderState.rows.forEach { row ->
+      context.runningWorker(row.textController.onTextChanged.asWorker(), "${row.id}") {
+        textChanged
+      }
```
steve-the-edwards, line 48: same nit:
```suggestion
      context.runningWorker(row.textController.onTextChanged.asWorker(), "${row.id}"), textChanged)
```

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/TextController.kt
```diff
@@ -0,0 +1,166 @@
...
+ *    ```
+ * 2. Create a matching property in your rendering type.
+ *    ```
+ *    data class Rendering(val text: TextController)
+ *    ```
+ * 3. In your `render` method, copy each [TextController] from your state to your rendering:
+ *    ```
+ *    return Rendering(state.text)
+ *    ```
+ * 4. In your `LayoutRunner`'s `showRendering` method, call [control] and pass your [EditText].
```
steve-the-edwards, line 32:
```suggestion
 * 4. In your `LayoutRunner`/`AndroidViewRendering`/`Screen`'s `showRendering` method, call [control] and pass your [EditText].
```

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/TextController.kt
```diff
@@ -0,0 +1,166 @@
...
+   * The current text value.
+   */
+  public var textValue: String
+    get() = _textValue.value
+    set(value) {
+      _textValue.value = value
+    }
+
+  /**
+   * Call this from your LayoutRunner's showRendering method. This method is idempotent: if it has
```
steve-the-edwards, line 90:
```suggestion
   * Call this from your LayoutRunner/AndroidViewRendering/Screen's showRendering method. This method is idempotent: if it has
```

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/TextController.kt
```diff
@@ -0,0 +1,166 @@
...
+      return
+    }
+
+    // If we're replacing a controller with a different one, cancel the previous subscription.
+    registeredController?.subscription?.cancel()
+
+    view.setText(_textValue.value)
+    val subscription = view.launchWhenAttached {
+      _textValue
+        .onEach { textValue ->
```
steve-the-edwards, line 115:
```suggestion
        .distinctUntilChanged()
        .onEach { textValue ->
```

I guess this necessitates another round of testing, but we might eke out a bit of optimization/debouncing on this side this way?

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/TextController.kt
```diff
@@ -0,0 +1,166 @@
...
+          // Only set the text if the actual text content has changed.
+          if (textValue != view.text.toString()) {
+            view.setText(textValue)
+          }
+        }
+        .launchIn(this)
+
+      view.listenForTextChangesUntilCancelled {
+        _textValue.value = it?.toString().orEmpty()
+      }
```
steve-the-edwards, line 125: Nit:
```suggestion
      view.listenForTextChangesUntilCancelled { updatedText ->
        _textValue.value = updatedText?.toString().orEmpty()
      }
```

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/TextController.kt
```diff
@@ -0,0 +1,166 @@
...
+  val subscription: Job
+)
+
+/**
+ * Suspends the coroutine until cancelled, calling [handler] any time a text change event is fired.
+ */
+private suspend fun TextView.listenForTextChangesUntilCancelled(
+  handler: (CharSequence?) -> Unit
+): Nothing {
+  suspendCancellableCoroutine<Nothing> { continuation ->
```
steve-the-edwards, line 143: I know I've reviewed before but what a brilliant use of structured concurrency. This is a great advertisement for the power of coroutines.

steve-the-edwards (approved):
Maybe I just haven't done enough programming this week and that's why I had so many nits - 😬 .

Anyway - love it overall :).

samples/todo-android/app/src/main/java/com/squareup/sample/todo/TodoEditorWorkflow.kt
rjrjr: Worker requires a function returning an action, can't pass the action directly.

rjrjr: Maybe ::textChanged, but that's just a cosmetic change.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/TextController.kt
rjrjr: That's already built into `MutableSharedFlow`.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/TextController.kt
rjrjr: Couldn't agree more. All kudos to @zach-klippenstein, I'm just a stenographer here.

rjrjr: In fact, I should fix the attribution on that commit.

samples/todo-android/app/src/main/java/com/squareup/sample/todo/TodoEditorWorkflow.kt
rjrjr: Yeah, no, that doesn't work. This is just how it is.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/TextController.kt
rjrjr: `AndroidViewRendering` doesn't have a `showRendering` method, and `Screen` doesn't exist. How about:

```
4. In your view code's `showRendering` method…
```

samples/todo-android/app/src/main/java/com/squareup/sample/todo/TodoEditorWorkflow.kt
steve-the-edwards: ah of course. I think I always blur one of the functional layers between the *handler* and the *action* that it returns (which is also a function). Could we overload `runningWorker` to also just take an action function in the cases where the handler doesn't have any side effects/work to do?

rjrjr: That's a pretty rare case, though. I don't think we want to simplify the ability to ignore the worker's output.

steve-the-edwards: Ack. Ok that makes sense.

PR #557, opened 2021-10-01
Introduces TextController.asMutableState.

rjrjr:
compose

### DIFF
--- a/samples/compose-samples/src/main/java/com/squareup/sample/compose/textinput/TextInputWorkflow.kt
+++ b/samples/compose-samples/src/main/java/com/squareup/sample/compose/textinput/TextInputWorkflow.kt
@@ -5,31 +5,25 @@ import com.squareup.sample.compose.textinput.TextInputWorkflow.State
 import com.squareup.workflow1.Snapshot
 import com.squareup.workflow1.StatefulWorkflow
 import com.squareup.workflow1.action
+import com.squareup.workflow1.ui.TextController
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
 
+@OptIn(WorkflowUiExperimentalApi::class)
 object TextInputWorkflow : StatefulWorkflow<Unit, State, Nothing, Rendering>() {
 
   data class State(
-    val textA: String = "",
-    val textB: String = "",
+    val textA: TextController = TextController(),
+    val textB: TextController = TextController(),
     val showingTextA: Boolean = true
   )
 
   data class Rendering(
-    val text: String,
-    val onTextChanged: (String) -> Unit,
+    val textController: TextController,
     val onSwapText: () -> Unit
   )
 
   private val swapText = action {
     state = state.copy(showingTextA = !state.showingTextA)
-  }
-
-  private fun changeText(text: String) = action {
-    state = if (state.showingTextA) {
-      state.copy(textA = text)
-    } else {
-      state.copy(textB = text)
-    }
   }
 
   override fun initialState(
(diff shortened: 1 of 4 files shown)


samples/compose-samples/src/main/java/com/squareup/sample/compose/textinput/TextInputWorkflow.kt
```diff
@@ -5,30 +5,38 @@ import com.squareup.sample.compose.textinput.TextInputWorkflow.State
 import com.squareup.workflow1.Snapshot
 import com.squareup.workflow1.StatefulWorkflow
 import com.squareup.workflow1.action
+import com.squareup.workflow1.ui.TextController
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
 
+@OptIn(WorkflowUiExperimentalApi::class)
 object TextInputWorkflow : StatefulWorkflow<Unit, State, Nothing, Rendering>() {
 
   data class State(
```
rjrjr, line 14: One wrinkle: I originally tried to use a `State` like this:

```kotlin
  data class State(
    val textA: TextController(""),
    val textB: TextController(""),
    val showingTextA: Boolean = true
  )
```

And then just swap which one was included in the rendering. That doesn’t work because `remember {}` only executes once, so the composition always holds the first `TextController` it saw.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/TextControllerAsMutableState.kt
```diff
@@ -0,0 +1,33 @@
...
+import com.squareup.workflow1.ui.TextController
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.launch
+import kotlin.reflect.KProperty
+
+@Composable public fun TextController.asMutableState(): MutableState<String> {
+  return remember { mutableStateOf(textValue) }.also { state ->
+    // Restart the effect if a different controller is passed in.
+    LaunchedEffect(this) {
```
rjrjr, line 20: Given how sticky `remember` is, will `LaunchedEffect` ever actually see a new `TextController` instance? 

I guess it will, but it won't matter because `state` will be the one that was created by the original `TextController`.

Very confusing, need to tweak this I think.

rjrjr: Seems like that should be fine? But when I had the rendering swap between different `TextController` instances, the swap button didn't work…

OOOoooh. `mutableState` persists, the new `LaunchedEffect ` starts, and the new `TextController` is immediately clobbered by it. Something like that.

Will play more.

rjrjr: Oh hell yeah, easy fix.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/TextControllerAsMutableState.kt
```diff
@@ -0,0 +1,29 @@
...
+import com.squareup.workflow1.ui.TextController
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.launch
+
+@Composable public fun TextController.asMutableState(): MutableState<String> {
+  return remember { mutableStateOf("") }.also { state ->
+    // Restart the effect if a different controller is passed in.
+    LaunchedEffect(this) {
+      state.value = textValue
```
zach-klippenstein, line 15: This won't run until after the first composition for the given text controller, which means you'll always get a frame with blank text. You need to initialize the state in the composition, ie when you create the mutable state.  I did it that way in the snippet for a reason. You should probably also pass the text controller as a key to remember so it's immediately updated if the controller changes as well.

rjrjr: I refuse to admit how hard I looked for a version of remember that took a key without finding it four lines down. Remember kids, it's important to get a full night's sleep.

rjrjr:
@laurakelly @maybeYouShouldBeTheLawyer @helios175 You're probably interested in this.

steve-the-edwards (approved):

rjrjr:
Flakiness tracked at #559, not related to this PR.

PR #558, opened 2021-10-02
Moves TextController from core-android to core-common.

rjrjr:
And make TextController.control(EditText) an extension function.


workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/TextControllerControlEditText.kt
```diff
@@ -0,0 +1,89 @@
...
+import kotlinx.coroutines.flow.onEach
+import kotlinx.coroutines.suspendCancellableCoroutine
+
+/**
+ * Call this from your view code's `showRendering` method. This method is idempotent: if it has
+ * already been called with a particular [EditText], and the view has not been detached since the
+ * last call, it will do nothing. If a different [TextController]'s [control] is called on the same
+ * [EditText], the old one will be disconnected and the new one will replace it.
+ *
+ * See [TextController] for more documentation.
```
rjrjr, line 18: Should move usage doc here

steve-the-edwards (approved):

PR #560, opened 2021-10-05
Eliminates WeakReference landmine in WorkflowLifecycleOwner.

rjrjr:
- [x] Verify against internal tests

We don't have good coverage of memory leaks here yet, but Square PoS sure does.

### DIFF
--- a/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/WorkflowLifecycleOwner.kt
+++ b/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/WorkflowLifecycleOwner.kt
@@ -18,7 +18,6 @@ import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
 import com.squareup.workflow1.ui.androidx.WorkflowAndroidXSupport.lifecycleOwnerFromViewTreeOrContext
 import com.squareup.workflow1.ui.androidx.WorkflowLifecycleOwner.Companion.get
 import com.squareup.workflow1.ui.androidx.WorkflowLifecycleOwner.Companion.installOn
-import java.lang.ref.WeakReference
 
 /**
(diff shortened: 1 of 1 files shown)


workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/WorkflowLifecycleOwner.kt
```diff
@@ -114,7 +113,7 @@ internal class RealWorkflowLifecycleOwner(
   /**
    * Weak reference ensures that we don't leak the view.
    */
```
steve-the-edwards, line 115: Need to remove/change the comment here (possibly elsewhere?).

What has changed that we are no longer concerned about the leak here? (Sorry if I should have already seen this pass by in a different discussion).

steve-the-edwards (approved):

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/WorkflowLifecycleOwner.kt
rjrjr: And if it's still leaking, what we should really do is add some LeakCanary tests of our own out here to repro.

rjrjr: > Nothing has changed

Well, this PR does add a lot of nulling out of the `View`, so that's a change.

steve-the-edwards: ack, yes OK, if we can repro with a test then SGTM!

rjrjr:
Couple of updates:

* Internal tests against a SNAPSHOT with this change reveals no leaks, so I feel good about merging it.
* #561 reveals leaks in the Compose samples _without_ this change. Sadly, this doesn't fix them.

PR #561, opened 2021-10-05
Hello leakcanary.

rjrjr:
Just plugging it in to start.


rjrjr:
Reveals leaks in the compose sample. Closing while I address that.

rjrjr:
Meh, easier to discuss if it stays open.

PR #564, opened 2021-10-06
Reference WelcomeWorkflow in the tutorial

aruh:
The example refers to a `HelloWorkflow` that doesn't exist in `samples/tutorial`. When using this reference, the code doesn't compile as the class doesn't exist.
I guess it existed previously and this reference was missed when changing the tutorial.

The reference is changed to the `WelcomeWorkflow` - the class that was created in this tutorial.


rjrjr:
Good catch, thanks!

steve-the-edwards (approved):

PR #566, opened 2021-10-06
Superstitiously specifying explicit profile for Instrumentation tests.

rjrjr:
For lack of better ideas, setting an explicit profile of `Galaxy Nexus` to see
if that gives us some stability. If nothing else @ychescale9 suggests it's
one of the faster ones.


rjrjr:
Well cool, this made _both_ shards spin for 30 minutes.

rjrjr:
This might actually be a win, he foolishly said out loud.

PR #567, opened 2021-10-06
Introduces Leakcanary

### LINKED ISSUE #472: Add LeakCanary in UI test runs

rjrjr:
### Hello Leakcanary.

Just plugging it in to start.

Closes #472.

### AndroidComposeTestRule work around and test cleanup.

Introduces `IdleAfterTestRule` as a work around for
https://issuetracker.google.com/issues/202190483, revealed by Leakcanary.

Adds some missing `@WorkflowUiExperimentalApi` on our internal test utilities.
They're already in a module named "internal", but let's be clearer about
their unsanctioned status.

Gets rid of some unnecessary `@JvmField` annotations on some test rules.


rjrjr:
CI failed wiht another flake that I am seeing only on API 21, #568.

steve-the-edwards:
Can we write our own ComposeTestRule that will do this workaround rule automatically so we can prevent forgetting to add that?

PR #575, opened 2021-10-28
Adding a dependency step prior to references to AndroidX ViewModel on Tutorial step 1

o-obushenko:
Since the following lines reference ViewModels from AndroidX, and the referenced dependencies didn't exist in build.gradle, the tutorial would not compile during this step. Because Tutorial2.md includes a step to add the dependencies to build.gradle, that convention will be followed to add lines to include the necessary dependencies during step 1.


steve-the-edwards:
@o-obushenko thanks I will merge this when you are able to get a chance to sign the CLA.

PR #577, opened 2021-11-05
Introduces Screen, WithEnvironment : Screen. Deprecates ViewFactory

### LINKED ISSUE #579: Does AndroidViewRendering / AndroidScreen break Compose integration?
The swizzle in `ViewRegistry.mapFactories` seems to have assumed assume that it was downstream of the `AndroidViewRendering` swizzle. It isn't.
    
Compose sample choked on AsIs with this, and spookily works just fine when I delete the check. So...is this transformation actually necessary? Has AndroidViewRendering been broken all along? Will have to check.

_Originally posted by @rjrjr in https://github.com/square/workflow-kotlin/pull/577#r744969308_

rjrjr:
First of several PRs replacing #202 and addressing #195, overhauling `workflow-ui`. Aimed at long living branch `ray/ui-update`.

# Step One (this PR): `Screen` and `WithEnvironment : Screen`

- Decouples `ViewEnvironment` and `ViewRegistry` from Android
- Introduces `ViewEnvironment.updateFrom` to simplify combining environments and their nested `ViewRegistry` entries.
- Introduces the `Screen` marker interface, similar to the one in `workflow-swift`
- Introduces the `WithEnvironment` rendering type as the preferred way to manage the `ViewEnvironment`, and makes `WorkflowLayout` use it
- Promotes `BackStackScreen` to workflow-ui:core-common
- Updates all non-Compose samples accordingly

This work is intended to be mostly backward compatible. My goal is that client code should require no more than updating some imports and ignoring a lot of `@Deprecated` warnings to upgrade.

| Was | Is |
| --- | --- |
| `RenderingT : Any`  |  `RenderingT : Screen` |
| `LayoutRunner`     | `ScreenViewRunner` |
| `ViewFactory<T>` | `ViewRegistry.Entry<T>`,  `ScreenViewFactory<T: Screen> : Entry<T>` |
| `AndroidViewRendering` | `AndroidScreen` |
| `WorkflowViewStub.update(Any)` | `WorkflowViewStub.show(Screen)` |
| `ViewRegistry.buildView(Any, ViewEnvironment, ...)` | `Screen.buildView(ViewEnvironment, ...)` |
| `WorkflowLayout.start(Flow<Any>)` | `WorkflowLayout.take(Flow<Screen>)` |

Yes, `Screen` is kind of a weird choice when what we really mean is `View`. But it's well established in `workflow-swift`, and doesn't seem to confuse anyone. More on this below.

# Step Two: `Overlay` replaces `ModalContainer`

`ModalContainer` and the Compose support are basically unchanged in this PR, and still built against the now deprecated machinery. They'll be addressed in follow ups, also merged to `ray/ui-update`. 

The Modal changes will be pretty drastic. The idea is to introduce another marker interface, `Overlay<T>`, and a corresponding `OverlayDialogRunner<OverlayT : Modal> : ViewRegistry.Entry<ModalT>`. `OverlayDialogRunner` will replace the abstract methods in `ModalContainer`, and also `ModalContainer.DialogRef`.

With that, we won't have to make a new `FrameLayout` subclass every time we want a new type of dialog. To show, say, an alert over a modal backstack panel over a body, we can render something like:

```kotlin
BodyAndOverlays(
  body = HomeScreen(),
  overlays: listOf(
    AlertModal("Oh Well", "Guess you screwed up", listOf(AlertButton("I sure did"))),
    CardModal(BackStackScreen(StepOne(), StepTwo())
  )
)
```

# Step Three: Compose

Hoping this lets us do something cleaner than extending `AndroidViewRendering<Nothing>`. Current strawman:

* `ScreenComposable<T: Screen> : ViewRegistry.Entry<T>`
* `Screen.buildView`
  * looks for `ScreenComposable` if it fails to find `ScreenViewFactory`
  * If a `ScreenComposable` is found, returns a `ScreenViewFactory` like today's `ComposeViewFactory` -- creates a `ComposeView` and uses it to host `ScreenComposable.content()`
* `@Composable Screen.BoxRendering(rendering: Screen, modifier: Modifier)`
  * Basically today's `WorkflowRendering`, which uses `Box()` but didn't document that fact
  * Combines `WorkflowViewStub` and `getViewFactoryForRendering()`, is that okay?
* `ComposeScreen : Screen` is like today's `ComposeRendering`, analog to `AndroidScreen`

This will probably require replacing the `ViewRegistry` interface with something that it's actually possible to wrap, in order to:

* Fix #579
* Avoid making `workflow-ui:android` depend upon the Compose runtime, which would be rude to library authors

# Terminology and concepts: Why "Screen"?

On the `main` branch, the only thing that I can map a rendering type to is a `ViewFactory`, which can only build `android.view.View`. The driving goal of the work kicked off by this PR is to make it possible for us to map rendering types to other things. Most immediately, we'd like to be able to map some rendering types directly to `android.app.Dialog`, without having to define an entire new classes of `android.view.View` to host them.

**`Screen`** is the marker interface for _something view-like_.
  - In classical Android, it maps to `android.view.View`
  - In Compose, it maps to a `@Composable` function that gets wrapped in `Box() {}`

`Modal` is a terrible name, so we'll replace it with **`Overlay`**. It is the marker interface for _something window-like_.
  - In classical Android, it maps to `android.app.Dialog` at the moment
  - In the past, it has mapped to a nested `FrameLayout` that acted window-like. That might happen again.
  - In Compose, dunno yet. Maybe it always falls back on the classic Android implementation. Maybe something entirely different.
 
The implementations that `Screen` and `Overlay` bind to are not interesting. The fact that we could easily implement both via `View`, or via `@Composable`, is not interesting. The key distinction is their behavior, and that they are definitely not interchangable.

We don't want `Rendering` as the marker interface for _something view-like_. The `RenderingT` type parameter of the `Workflow` interface exists already. Having an interface with the same name would be terribly misleading. A workflow can render `Screen`s, it can render `Overlay`s, or it can render anything else at all, even things completely unrelated to UI concerns.

We can't co-opt `View` as that marker interface either. In both Android and iOS, `View` is no longer just UI terminology, it indicates a concrete class. An Android developer reads `FooView` to mean a subclass of `android.view.View`. For the same reason, `Overlay` is more attractive than `Dialog` or `Window`.

`Screen` is nice because it isn't used yet, and yet it's familiar. "I was on the home screen, I was on the checkout screen." The term was in widespread use long before Workflow showed up, and when we started using it as our marker term in Swift, it was understood intuitively. And back to the "why not View" point, it is not uncommon in our code base for both `class FooScreen` and `class FooView` to exist, where the former is the model for the latter.

Both `Screen` and `Overlay` are examples of "UI renderings." UI rendering is likely to be a useful term in documentation, but we have no need for a common parent interface for them to extend. `Screen : Any`, there is no need for `Screen : UiRendering`. Solves no problems.

# Some Q & A:

_What’s the name of the component/idea that the Workflow runtime uses to convert the Container into the various Screen+Overlays that show up on screen?_

In `main` today, that’s `ViewFactory`. In this PR, it’s `ViewRegistry.Entry`. There is still only one type of `Entry` in this PR, but I'm looking to follow up quickly with at least two more:

    interface ScreenViewFactory<T: Screen> : ViewRegistry.Entry<T>

    interface OverlayDialogFactory<T : Overlay> : ViewRegistry.Entry<T>

    interface ScreenComposableFactory<T: Screen> : ViewRegistry.Entry<T>

_How does the Workflow runtime convert a Container (composed of Screens and Overlays) into Views+Dialogs? Like who does the mapping?_

Consider this example:

    class AlertModal(
      val title: String,
      val body: String,
      val buttons: List<AlertButton>
    ) : Overlay
    
    class CardModal<S: Screen>(val content: S) : Overlay
    
    class BodyAndOverlays<O: Overlay, B: Screen>(
      body: B,
      overlays: List<O>
    ): Screen

A workflow might render this:

    BodyAndOverlays(
      body = HomeScreen(),
      overlays: listOf(
        AlertModal("Oh Well", "Guess you screwed up", listOf(AlertButton("I sure did"))),
        CardModal(BackStackScreen(StepOne(), StepTwo())
      )
    )

To turn that into `View` and `Dialog` instances, in the new world:

- `BodyAndOverlays.buildView()` is called, probably by `WorklowViewStub.show()`. 

- `buildView()` finds the associated `ScreenViewFactory<BodyAndOverlays<*, *>>`, which probably inflates a `BodyAndOverlaysContainer : FrameLayout`, and calls a private `show(BodyAndOverlays, ViewEnvironment)` method on it.

- That method calls `WorklowViewStub.show(rendering.body)` to show the body, and something like [TBD] `Overlay.buildDialogRunner(coveringView: this)` for each entry in `rendering.overlays`


workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/AndroidViewEnvironment.kt
```diff
@@ -0,0 +1,113 @@
...
+}
+
+/**
+ * Default implementation for the `initializeView` argument of [ViewEnvironment.buildView],
+ * and for [DecorativeScreenViewFactory.initializeView]. Calls [showRendering] against
+ * [getRendering] and [environment].
+ */
+@WorkflowUiExperimentalApi
+public fun View.showFirstRendering() {
+  showRendering(getRendering()!!, environment!!)
```
rjrjr, line 112: I want to replace `showRendering(Any)` with `showRendering(Screen)`, but this PR is already big enough.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/LayoutScreenViewFactory.kt
```diff
@@ -0,0 +1,36 @@
...
+import kotlin.reflect.KClass
+
+/**
+ * A [ScreenViewFactory] that ties a [layout resource][layoutId] to a
+ * [ViewRunner factory][runnerConstructor] function. See [ScreenViewRunner] for
+ * details.
+ */
+@WorkflowUiExperimentalApi
+@PublishedApi
+internal class LayoutScreenViewFactory<RenderingT : Screen>(
```
rjrjr, line 16: Tempted to change all of these `RenderingT` param types to `ScreenT`. Downside is that we lose the hint that these are expected to be the `RenderingT` from a workflow. Probably won't mess.

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/ViewRegistry.kt
```diff
@@ -0,0 +1,111 @@
+@file:Suppress("FunctionName")
+
+package com.squareup.workflow1.ui
+
+import com.squareup.workflow1.ui.ViewRegistry.Entry
+import kotlin.reflect.KClass
+
+/**
+ * The [ViewEnvironment] service that can be used to display the stream of renderings
```
rjrjr, line 9: Needs updating, will get to it.

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/container/RootScreen.kt
```diff
@@ -0,0 +1,43 @@
+package com.squareup.workflow1.ui.container
```
rjrjr, line 1: I'm thinking this `container` sub-package is not useful.

rjrjr:
Spent the rest of the day tidying this up -- updating samples, muting deprecations, fixing docs. @steve-the-edwards, @Blisse, @mbrubin56-square, @helios175, I'd like to get LGs from at least a couple of you before merging. I don't know if it'll be green before I give up for the day, but it'll be close.

I think I'll update the `Compose` stuff on Monday. Shouldn't be a drastic change, just clean up a couple of warts now that `ViewRegistry` is more flexible. Thanks for leaving things so tidy, @zach-klippenstein.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/CompositionRoot.kt
```diff
@@ -84,23 +83,19 @@ public fun ViewRegistry.withCompositionRoot(root: CompositionRoot): ViewRegistry
...
   transform: (ViewFactory<*>) -> ViewFactory<*>
 ): ViewRegistry = object : ViewRegistry {
   override val keys: Set<KClass<*>> get() = this@mapFactories.keys
 
-  override fun <RenderingT : Any> getFactoryFor(
+  override fun <RenderingT : Any> getEntryFor(
     renderingType: KClass<out RenderingT>
-  ): ViewFactory<RenderingT> {
-    val factoryFor =
-      this@mapFactories.getFactoryFor(renderingType) ?: throw IllegalArgumentException(
```
rjrjr, line 99: This could never have worked, as this method is not and was not the method that swizzles in AndroidViewRendering and such.
    
Compose sample choked on AsIs with this, and spookily works just fine when I delete the check. So...is this transformation actually necessary? Has AndroidViewRendering been broken all along? Will have to check.

samples/containers/android/src/main/java/com/squareup/sample/container/BackButtonViewFactory.kt
```diff
@@ -1,30 +1,31 @@
...
-import com.squareup.workflow1.ui.DecorativeViewFactory
-import com.squareup.workflow1.ui.ViewFactory
+import com.squareup.workflow1.ui.DecorativeScreenViewFactory
+import com.squareup.workflow1.ui.ScreenViewFactory
 import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
 import com.squareup.workflow1.ui.backPressedHandler
 
 /**
- * [ViewFactory] that performs the work required by [BackButtonScreen].
+ * [_root_ide_package_.com.squareup.workflow1.ui.ScreenViewFactory] that performs the work required by [BackButtonScreen].
```
rjrjr, line 9: sigh

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewRunner.kt
```diff
@@ -0,0 +1,105 @@
...
+/**
+ * A delegate that implements a [showRendering] method to be called when a workflow
+ * rendering of type [RenderingT] : [Screen] is ready to be displayed in a view created
+ * by a [ScreenViewFactory].
+ *
+ * If you're using [AndroidX ViewBinding][ViewBinding] you likely won't need to
+ * implement this interface at all. For details, see the three overloads of [ScreenViewRunner.bind].
+ */
+@WorkflowUiExperimentalApi
+public fun interface ScreenViewRunner<RenderingT : Screen> {
```
rjrjr, line 22: This is as unchanged as possible from LayoutRunner.kt.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/BackStackContainer.kt
```diff
@@ -0,0 +1,233 @@
...
+ * heuristic fails you, use [ViewEnvironment.withBackStackStateKeyPrefix] to add unique names to
+ * the [ViewEnvironment] used to show each [BackStackScreen].
+ *
+ * There's a potential issue here where if our ID is changed to something else, then another
+ * [BackStackContainer] is added with our old ID, that container will overwrite our state. Since
+ * they'd both be using the same key, [SavedStateRegistry] would throw an exception. As long as this
+ * container is detached before its ID is changed, it shouldn't be a problem.
+ */
+@WorkflowUiExperimentalApi
+public open class BackStackContainer @JvmOverloads constructor(
```
rjrjr, line 60: Almost nothing changed here, beyond moving the class and replacing `Named` with `NamedScreen`. I don't know why github won't show a diff.

workflow-ui/core-android/src/test/java/com/squareup/workflow1/ui/TestViewFactory.kt
```diff
@@ -1,3 +1,5 @@
+@file:Suppress("DEPRECATION")
```
rjrjr: This file is unused, delete.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/CompositionRoot.kt
rjrjr: Tracked in #579

rjrjr:
Oh joy. `samples:hello-workflow-fragment:connectedDebugAndroidTest` failing consistently on API 21 only:

```
Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'androidx.activity.result.contract.ActivityResultContract$SynchronousResult androidx.activity.result.contract.ActivityResultContract.getSynchronousResult(android.content.Context, java.lang.Object)' on a null object reference
at androidx.activity.ComponentActivity$2.onLaunch(ComponentActivity.java:157)
at androidx.activity.OnBackPressedDispatcher$LifecycleOnBackPressedCancellable.onStateChanged(OnBackPressedDispatcher.java:233)
at androidx.lifecycle.LifecycleRegistry.forwardPass(LifecycleRegistry.java:265)
at androidx.lifecycle.LifecycleRegistry.sync(LifecycleRegistry.java:307)
at androidx.lifecycle.LifecycleRegistry.moveToState(LifecycleRegistry.java:148)
at androidx.lifecycle.LifecycleRegistry.handleLifecycleEvent(LifecycleRegistry.java:134)
at androidx.fragment.app.Fragment.performStart(Fragment.java:3026)
at androidx.fragment.app.FragmentStateManager.start(FragmentStateManager.java:589)
at androidx.fragment.app.FragmentStateManager.moveToExpectedState(FragmentStateManager.java:300)
at androidx.fragment.app.FragmentStore.moveToExpectedState(FragmentStore.java:112)
at androidx.fragment.app.FragmentManager.moveToState(FragmentManager.java:1647)
at androidx.fragment.app.FragmentManager.dispatchStateChange(FragmentManager.java:3128)
at androidx.fragment.app.FragmentManager.dispatchStart(FragmentManager.java:3079)
at androidx.fragment.app.FragmentController.dispatchStart(FragmentController.java:262)
at androidx.fragment.app.FragmentActivity.onStart(FragmentActivity.java:510)
at androidx.appcompat.app.AppCompatActivity.onStart(AppCompatActivity.java:246)
at android.app.Instrumentation.callActivityOnStart(Instrumentation.java:1220)
at androidx.test.runner.MonitoringInstrumentation.callActivityOnStart(MonitoringInstrumentation.java:723)
at android.app.Activity.performStart(Activity.java:5953)
at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2261)
... 10 more
```

rjrjr:
I'm seeing `samples:hello-workflow-fragment:connectedDebugAndroidTest` fail consistently on API 21 on `main` as well. Digging into that.

In the meantime, we should consider this PR green.

0legg:
> I'm seeing `samples:hello-workflow-fragment:connectedDebugAndroidTest` fail consistently on API 21 on `main` as well. Digging into that.
> 
> In the meantime, we should consider this PR green.

Sounds like the same beast we're fighting internally for almost two months.

rjrjr:
> > I'm seeing `samples:hello-workflow-fragment:connectedDebugAndroidTest` fail consistently on API 21 on `main` as well. Digging into that.
> > In the meantime, we should consider this PR green.
> 
> Sounds like the same beast we're fighting internally for almost two months.

I thought that at first, but it fails completely reliably, even after a `git clean -fdx`/ `killall -9 java`.

Is there anything in particular I can look at to confirm / deny harder? It'd be great for this little test to be a repro case.

rjrjr:
AAAAIIIGH why the hell is the test still running on CI? That suppress did the trick locally.

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/container/RootScreen.kt
```diff
@@ -0,0 +1,49 @@
...
+import com.squareup.workflow1.ui.ViewRegistry
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+import com.squareup.workflow1.ui.updateFrom
+
+/**
+ * Pairs a [screen] rendering with a [viewEnvironment] to support its display.
+ * Typically the rendering type (`RenderingT`) of the root of a UI workflow,
+ * but can be used at any point to modify the [ViewEnvironment] received from
+ * a parent view.
+ */
```
Blisse, line 15: * Introduces the Screen marker interface, similar to the one in workflow-swift
* Introduces the RootScreen rendering type as the preferred way to manage the ViewEnvironment, and makes WorkflowLayout use it

> What is RootScreen

I guess I'm trying more to understand what does "Root" mean in the context of Workflows here. In my mind Root almost means singular, but this seems more generic.

steve-the-edwards: Makes sense as a utility. I think that we need a new dominant metaphor for naming this as the 'tree' metaphor is lacking here when talking about transforming the ViewEnvironment.

steve-the-edwards (approved):
I have a huge bikeshed I'd like to paint as I wholeheartedly do not resonate with the `Screen` naming. This is especially a problem when we introduce `Modal` as that may begin to suggest that a `Modal` is anything less than a "full" "Screen" when that it is explicitly _not_ what it wants to represent. Can talk in slack or elsewhere.

rjrjr:
> I have a huge bikeshed I'd like to paint as I wholeheartedly do not resonate with the `Screen` naming. This is especially a problem when we introduce `Modal` as that may begin to suggest that a `Modal` is anything less than a "full" "Screen" when that it is explicitly _not_ what it wants to represent. Can talk in slack or elsewhere.

https://workflow-community.slack.com/archives/CHTFPR277/p1636572894038600

rjrjr:
Gonna try to summarize the bikeshed here.

On the `main` branch, the only thing that I can map a rendering type to is a `ViewFactory`, which can only build `android.view.View`. The driving goal of the work kicked off by this PR is to make it possible for us to map rendering types to other things. Most immediately, we'd like to be able to map some rendering types directly to `android.app.Dialog`, without having to define an entire new classes of `android.view.View` to host them.

**`Screen`** will indeed be the marker interface for _something view-like_
  - In classical Android, it maps to `android.view.View`
  - In Compose, it maps to a `@Composable` function that gets wrapped in `Box() {}`

`Modal` is a terrible name, and will probably be replaced with **`Overlay`**. It is the marker interface for _something window-like_.
  - In classical Android, it maps to `android.app.Dialog` at the moment
  - In the past, it has mapped to a nested `FrameLayout` that acted window-like. That might happen again.
  - In Compose, dunno yet. Maybe it always falls back on the classic Android implementation. Maybe something entirely different.
 
The implementations that `Screen` and `Overlay` bind to are not interesting. The fact that we could easily implement both via `View`, or via `@Composable`, is not interesting. The key disctinction is their behavior, and that they are definitely not interchangable.

We don't want `Rendering` as the marker interface for _something view-like_. The `RenderingT` type parameter of the `Workflow` interface exists already. A workflow can render `Screen`s, it can render `Overlay`s, or it can render anything else at all, even things completely unrelated to UI concerns.

We don't `View` as that marker interface either. In both Android and iOS, `View` is no longer just UI terminology, it indicates a concrete class. An Android developer reads `FooView` to mean a subclass of `android.view.View`.

`Screen` is nice because it isn't used yet, and yet it's familiar. "I was on the home screen, I was on the checkout screen." The term was in widespread use long before Workflow showed up, and when we started using it as our marker term in Swift, it was understood intuitively. And back to the "why not View" point, it is not uncommon in our code base for both `class FooScreen` and `class FooView` to exist, where the former is the model for the latter.

Both `Screen` and `Overlay` are examples of UI renderings. UI rendering is likely to be a useful term in documentation, but we have no need for a common parent interface for them to extend. `Screen : Any`, there is no need for `Screen : UiRendering`. Solves no problems.

Some Q & A:

_What’s the name of the component/idea that the Workflow runtime uses to convert the Container into the various Screen+Modals that show up on screen?_

In main, that’s `ViewFactory`. In the new PR, it’s `ViewRegistry.Entry`. There is still only one type of `Entry`, but I'm looking to follow up quickly with at least two more:

    interface ScreenViewFactory<T: Screen> : ViewRegistry.Entry<T>

    interface ModalDialogFactory<M : Modal> : ViewRegistry.Entry<M>

    interface ScreenComposableFactory<T: Screen> : ViewRegistry.Entry<T>

_How does the Workflow runtime convert a Container (composed of Screens and Modals) into Views+Dialogs? Like who does the mapping?_

Consider this example:

    class AlertModal(
      val title: String,
      val body: String,
      val buttons: List<AlertButton>
    ) : Overlay
    
    class CardModal<S: Screen>(val content: S) : Overlay
    
    class BodyAndOverlays<O: Overlay, B: Screen>(
      body: B,
      overlays: List<O>
    ): Screen

A workflow might render this:

    BodyAndOverlays(
      body = HomeScreen(),
      overlays: listOf(
        AlertModal("Oh Well", "Guess you screwed up", listOf(AlertButton("I sure did"))),
        CardModal(BackStackScreen(StepOne(), StepTwo())
      )
    )

To turn that into `View` and `Dialog` instances, in the new world:


- The `BodyAndOverlays` rendering gets passed to `ViewEnvironment.buildView(Screen)`, probably by `WorklowViewStub.show()`. 

- `buildView()` finds the associated `ScreenViewFactory<BodyAndOverlays<*, *>>`, which probably inflates a `BodyAndOverlaysContainer : FrameLayout`, and calls a private `show(BodyAndOverlays, ViewEnvironment)` method on it.

` That method calls `WorklowViewStub.show(rendering.body)` to show the body, and something like [TBD] `ViewEnvironment.buildDialogRunner(Overlay)` for each entry in `rendering.overlays`

rjrjr:
Maybe I should add the comment above to the commit message.

rjrjr:
Hmmmm. Probably this:

```kotlin
fun <T: Screen> ViewEnvironment.buildView(screenRendering: T): View
```

should be this:

```kotlin
fun <T: Screen> T.buildView(environment: ViewEnvironment): View
```

Makes the marker interface serve more of a purpose, and is an even closer parallel to how the Swift code works.

@steve-the-edwards @Blisse @mbrubin56-square WDYT?

steve-the-edwards:
> Hmmmm. Probably this:
> 
> ```kotlin
> fun <T: Screen> ViewEnvironment.buildView(screenRendering: T): View
> ```
> 
> should be this:
> 
> ```kotlin
> fun <T: Screen> T.buildView(environment: ViewEnvironment): View
> ```
> 
> Makes the marker interface serve more of a purpose, and is an even closer parallel to how the Swift code works.
> 
> @steve-the-edwards @Blisse @mbrubin56-square WDYT?



> Hmmmm. Probably this:
> 
> ```kotlin
> fun <T: Screen> ViewEnvironment.buildView(screenRendering: T): View
> ```
> 
> should be this:
> 
> ```kotlin
> fun <T: Screen> T.buildView(environment: ViewEnvironment): View
> ```
> 
> Makes the marker interface serve more of a purpose, and is an even closer parallel to how the Swift code works.
> 
> @steve-the-edwards @Blisse @mbrubin56-square WDYT?

I like it a lot. The `Screen` extension should know how to build its view with the help of the `viewEnvironment` not the other way around.

As for `RootScreen` - does it make any sense to be `EnvironmentRoot`? We could have a separate extension there that uses its own `viewEnvironment` to build the view.

```
fun <T: EnvironmentRoot> T.buildView(): View
```

rjrjr:
> As for `RootScreen` - does it make any sense to be `EnvironmentRoot`? We could have a separate extension there that uses its own `viewEnvironment` to build the view.
> 
> ```
> fun <T: EnvironmentRoot> T.buildView(): View
> ```

I like the extra extension. How about `WithEnvironment` as the class name?

rjrjr:
Actually…I can't find any place that we would ever actually call the `WithEnvironment.buildView()` overload, because `WorkflowViewStub`. Gonna refrain.

rjrjr:
Okey doke, I have…

- Moved ViewEnvironment.buildView() to Screen
- Renamed RootScreen to WithEnvironment
- Updated the PR title and first comment to reflect these changes, and overhauled the comment to include the bikeshed recap.

I'll plan to merge later today, barring any further feedback.

Blisse (approved):

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/AndroidViewEnvironment.kt
```diff
@@ -0,0 +1,101 @@
...
+ *    parent is destroyed), or
+ * 2. Have its
+ *    [WorkflowLifecycleOwner.destroyOnDetach][com.squareup.workflow1.ui.androidx.WorkflowLifecycleOwner.destroyOnDetach]
+ *    called, which will either schedule the
+ *    lifecycle to be destroyed if the view is attached, or destroy it immediately if it's detached.
+ *
+ * @throws IllegalArgumentException if no factory can be find for type [RenderingT]
+ */
+@WorkflowUiExperimentalApi
+public fun <RenderingT : Any>
```
rjrjr: These are nearly verbatim copies of existing extensions on `ViewRegistry`.

workflow-ui/core-android/src/androidTest/java/com/squareup/workflow1/ui/DecorativeScreenViewFactoryTest.kt
```diff
@@ -0,0 +1,153 @@
...
+
+import android.content.Context
+import android.view.View
+import android.view.ViewGroup
+import androidx.test.platform.app.InstrumentationRegistry
+import com.google.common.truth.Truth.assertThat
+import org.junit.Test
+
+@OptIn(WorkflowUiExperimentalApi::class)
+internal class DecorativeScreenViewFactoryTest {
```
rjrjr, line 11: copy / paste

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/DecorativeScreenViewFactory.kt
```diff
@@ -0,0 +1,179 @@
+package com.squareup.workflow1.ui
+
+import android.content.Context
+import android.view.View
+import android.view.ViewGroup
+import kotlin.reflect.KClass
+
+/**
```
rjrjr, line 8: copy / paste

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ManualScreenViewFactory.kt
```diff
@@ -0,0 +1,43 @@
+package com.squareup.workflow1.ui
+
+import android.content.Context
+import android.view.View
+import android.view.ViewGroup
+import kotlin.reflect.KClass
+
+/**
+ * A [ScreenViewFactory] that creates [View]s that need to be generated from code.
```
rjrjr, line 9: copy / paste of BuilderViewFactory

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewRunner.kt
```diff
@@ -0,0 +1,105 @@
...
+ * A delegate that implements a [showRendering] method to be called when a workflow rendering
+ * of type [RenderingT] is ready to be displayed in a view inflated from a layout resource
+ * by a [ViewRegistry]. (Use [ManualScreenViewFactory] if you want to build views from code rather
+ * than layouts.)
+ *
+ * If you're using [AndroidX ViewBinding][ViewBinding] you likely won't need to
+ * implement this interface at all. For details, see the three overloads of [ScreenViewRunner.bind].
+ */
+@WorkflowUiExperimentalApi
+public fun interface ScreenViewRunner<RenderingT : Screen> {
```
rjrjr, line 22: copy / paste of LayoutRunner

PR #581, opened 2021-11-09
Daisy-chaining support for `testRender`

### LINKED ISSUE #580: Support daisy-chaining of separate `testRender` sessions
There should be a middle ground between `testRender` and `launchForTestingFromStartWith` The former is a nicer API, but doesn’t work for changes when you want to run one or two render passes. Maybe we could daisy-chain `verifyActionResult`.
If we'll have it, the integration testing of workflow code will be much easier and won't require providing implementation and mocking a thousand of things.

From the first impression: while it's the same instance of `RealRenderTester` under the hood for any call in the chain, outside of the library we don't know that it's true (nor it's guaranteed). I'm thinking of making `verifyAction` and `verifyActionResult` to return some `RenderTestContinuation` (name WIP) interface, also implemented by `RealRenderTester`. This interface will expose two new functions:
- `testNextRender(): RenderTester`, which will use the same props.
- `testNextRenderWithProps(newProps: PropsT?): RenderTester`, which will override the props from the original call.

Both will perform the same under the hood: either copy itself (we're still having the same `RealRenderTester` under the hood),  or just clean all lists and reset all vars to the original state. The former could be safer if we have some race conditions, the latter is easier on memory.

Also, we could somehow short-circuit `expect*` part there, maybe by having a boolean flag `keepExpectations: Boolean = false` for all parameters of `RenderTestContinuation`.

0legg:
Resolves #580 .

### DIFF
--- a/workflow-testing/src/main/java/com/squareup/workflow1/testing/RealRenderTester.kt
+++ b/workflow-testing/src/main/java/com/squareup/workflow1/testing/RealRenderTester.kt
@@ -63,7 +63,7 @@ internal class RealRenderTester<PropsT, StateT, OutputT, RenderingT>(
   private val ranSideEffects: MutableList<String> = mutableListOf()
 ) : RenderTester<PropsT, StateT, OutputT, RenderingT>(),
   BaseRenderContext<PropsT, StateT, OutputT>,
-  RenderTestResult<PropsT, StateT, OutputT>,
+  RenderTestResult<PropsT, StateT, OutputT, RenderingT>,
   Sink<WorkflowAction<PropsT, StateT, OutputT>> {
 
   internal sealed class Expectation<out OutputT> {
@@ -117,7 +117,9 @@ internal class RealRenderTester<PropsT, StateT, OutputT, RenderingT>(
   }
 
   @OptIn(ExperimentalStdlibApi::class)
-  override fun render(block: (RenderingT) -> Unit): RenderTestResult<PropsT, StateT, OutputT> {
+  override fun render(
+    block: (RenderingT) -> Unit
+  ): RenderTestResult<PropsT, StateT, OutputT, RenderingT> {
     if (!explicitWorkerExpectationsRequired) {
       // Allow unexpected workers.
       expectWorker(description = "unexpected worker", exactMatch = false) { _, _, _ -> true }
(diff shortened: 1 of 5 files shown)


workflow-testing/src/main/java/com/squareup/workflow1/testing/RealRenderTester.kt
```diff
@@ -117,7 +118,9 @@ internal class RealRenderTester<PropsT, StateT, OutputT, RenderingT>(
   }
 
   @OptIn(ExperimentalStdlibApi::class)
-  override fun render(block: (RenderingT) -> Unit): RenderTestResult<PropsT, StateT, OutputT> {
+  override fun render(
+    block: (RenderingT) -> Unit
+  ): RenderTestResult<PropsT, StateT, OutputT, RenderingT> = apply {
```
rjrjr, line 123: Does this actually change anything?

rjrjr (commented):
So simple!

Like you hinted offline, maybe the new interface and the new method are overkill. The new tests would read just fine if you deleted the `testNextRender()` lines.

workflow-testing/src/main/java/com/squareup/workflow1/testing/RealRenderTester.kt
0legg: I need to plumb `RenderingT` through generics.

rjrjr: Wow, I spaced the param type. But I meant using `apply {` rather than just returning `this`.

0legg: Oh, probably it was because I meant it as a draft and left a room for some potential changes.

steve-the-edwards (approved):
I agree with @rjrjr - let's apply the action before every call to `render()` so we don't need the extra `testNextRender()` method in between.

rjrjr:
> I agree with @rjrjr - let's apply the action before every call to `render()` so we don't need the extra `testNextRender()` method in between.

We chatted offline and talked each other out of that.

* We need the new message so that we can have an overload that accepts updated props
* It also allows us to avoid creating and returning a new object that most callers will never use

It does seem like a good idea to drop the new interface, though.

workflow-testing/src/main/java/com/squareup/workflow1/testing/RealRenderTester.kt
rjrjr: I still don't like `apply` here.

- [ ] Nix on apply

workflow-testing/src/main/java/com/squareup/workflow1/testing/RealRenderTester.kt
```diff
@@ -259,20 +259,38 @@ internal class RealRenderTester<PropsT, StateT, OutputT, RenderingT>(
     processedAction = value
   }
 
-  override fun verifyAction(block: (WorkflowAction<PropsT, StateT, OutputT>) -> Unit) {
+  override fun verifyAction(
+    block: (WorkflowAction<PropsT, StateT, OutputT>) -> Unit
+  ): RenderTestResult<PropsT, StateT, OutputT, RenderingT> = apply {
     val action = processedAction ?: noAction()
     block(action)
   }
```
rjrjr, line 267:
```suggestion
  ): RenderTestResult<PropsT, StateT, OutputT, RenderingT> {
    val action = processedAction ?: noAction()
    block(action)
    return this
  }
```

rjrjr (approved):

rjrjr:
I like this change so much.

workflow-testing/src/main/java/com/squareup/workflow1/testing/RealRenderTester.kt
rjrjr: It's obfuscating. We're just implementing an interface method, how does `apply` make this more readable than simply ending a method with `return this`?

0legg: I'm on the other side of the spectrum here — why do we need to write explicitly `return this` if Kotlin has a good convention to do it (ok, I agree that `apply` could be misleading here because we don't mutate anything). Anyway, not a hill I would die on, and you're the mastermind behind this repo :)

Issue #524, opened 2021-08-24
Support RxJava3 


WhosNickDoglio:
There's an RxJava2 artifact are there plans for an RxJava3 artifact? 

Would you be happy to open a PR for it you're open to it

rjrjr:
Thanks for the offer, but I don't think we want to imply that we have the bandwidth to support RxJava3. 

The good news is that it should be straightforward to build such support without needing any changes in this repo, or access to any private APIs.

rjrjr:
cc @steve-the-edwards, @zach-klippenstein for other perspectives.

zach-klippenstein:
No, I don't think it makes sense for us to maintain that. Rx2 is basically just there to support migrations away from it anyway.

PR #586, opened 2021-11-14
Add RxJava3 support for Workflow

0legg:
I know, Kotlin Flows are the future.
I know, RxJava is the past.

But someone could want to have RxJava. And RxJava2 is [end-of-life](https://github.com/ReactiveX/RxJava#version-2x), so having RxJava3 bridge is a convenient thing.

And yes, I've just copied `workflow-rx2` and updated all imports. This is how migration works, right?


rjrjr:
Thank you, but the reasoning from #524 still applies -- YAGNI.

PR #587, opened 2021-11-17
Renames WithEnvironment to EnvironmentScreen.

### LINKED ISSUE #585: Rename WithEnvironment to EnvironmentScreen
Probably what the like named thing will be called in Swift, and just a better name.

rjrjr:
Fixes #585.


bencochran (approved):
I’m a fan! (but you already knew that 😇)

steve-the-edwards (approved):
I prefer `WithEnvironment` - but you already knew that 😝

PR #588, opened 2021-11-17
Introduces Overlay, and its first uses: BodyAndModalsScreen, AlertOverlay

rjrjr:
New platform-neutral rendering types:

- `Overlay` is a marker type that identifies a window-like rendering.

- `AlertOverlay` is the first use of `Overlay`, models Android's
  `AlertDialog`

- `BodyAndModalsScreen` has a body `Screen` covered by zero or more
  `Overlay`s that are given a modal treatment -- its body is shielded
  from events while any `Overlay` is present.

Android support code that uses them to manage `Dialog`s:

- `OverlayDialogFactory` is to `Overlay` as
  `ScreenViewRunner` and `ScreenViewFactory` are to `Screen`

- `DialogStack` uses `OverlayDialogFactory` to display `List<Overlay>`
  as a pile of `Dialog`s

- `BodyAndModalsContainer` is a `FrameLayout` that uses `DialogStack`
  to display `BodyAndModalScreen`

- `AlertOverlayDialogFactory` creates `AlertDialog` instances for
  `AlertOverlay`

Note that `BodyAndModalsContainer` and `AlertOverlayDialogFactory` are
`internal`, with default binings built into `Overlay.toDialogFactory()`
and `Screen.buildView`.

Also note that `AndroidViewEnvironment.kt` and the bulk of `AndroidScreen.kt`
have been moved to `ScreenViewFactory.kt`, which seemed like a less
surprising home for them.

Updates `hellobackbutton` sample to demonstrate the use of all of this.

Note the commented out deprecations. Turns out its impossible to suppress
deprecation warnings in `typealias` calls like the ones our sample apps are
still using, and those can't be ported until we have a replacement for
`ModalViewContainer`. Tracked in #589.


workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewFactory.kt
```diff
@@ -31,3 +37,87 @@ public interface ScreenViewFactory<in RenderingT : Screen> : ViewRegistry.Entry<
     container: ViewGroup? = null
   ): View
 }
+
+/**
```
rjrjr, line 41: Outside of making `getViewFactoryForRendering` non-public, no real changes in this code, just moved it from another location.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/BodyAndModalsContainer.kt
```diff
@@ -0,0 +1,134 @@
...
+import com.squareup.workflow1.ui.ManualScreenViewFactory
+import com.squareup.workflow1.ui.R
+import com.squareup.workflow1.ui.ScreenViewFactory
+import com.squareup.workflow1.ui.ViewEnvironment
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+import com.squareup.workflow1.ui.WorkflowViewStub
+import com.squareup.workflow1.ui.bindShowRendering
+
+@WorkflowUiExperimentalApi
+public class BodyAndModalsContainer @JvmOverloads constructor(
```
rjrjr, line 23: Meant to make this `internal`

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewFactory.kt
```diff
@@ -31,3 +37,87 @@ public interface ScreenViewFactory<in RenderingT : Screen> : ViewRegistry.Entry<
...
+  val entry = get(ViewRegistry).getEntryFor(rendering::class)
+
+  @Suppress("UNCHECKED_CAST", "DEPRECATION")
+  return (entry as? ScreenViewFactory<ScreenT>)
+    ?: (rendering as? AndroidScreen<*>)?.viewFactory as? ScreenViewFactory<ScreenT>
+    ?: (rendering as? AsScreen<*>)?.let { AsScreenViewFactory as ScreenViewFactory<ScreenT> }
+    ?: (rendering as? BackStackScreen<*>)?.let {
+      BackStackScreenViewFactory as ScreenViewFactory<ScreenT>
+    }
+    ?: (rendering as? BodyAndModalsScreen<*, *>)?.let {
```
rjrjr, line 112: Okay, I lied -- this line is new

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/RectProvider.kt
```diff
@@ -0,0 +1,37 @@
...
+import android.view.View
+
+/**
+ * Defines the desired bounds for a [Dialog][android.app.Dialog] built by
+ * [OverlayDialogFactory.buildDialog]. Custom containers can define their
+ * bounds policy via [DialogStack.createRectProvider].
+ *
+ * Use [ViewRectProvider] to monitor the bounds of a [View].
+ */
+public interface RectProvider {
```
helios175, line 13: Not disagreeing but what about this:
- a dialog might receive more than one rect:
  - coverarea (what to scrim)
  - anchor view (what to appear next to... like a dropdown)
  - the proper dialog area (which can be defined by the type of dialog, for example a bottom sheet might appear horizontally centered –or using all width– and to the bottom of the cover area.

All depends where you want to place the "this type of dialog should appear here according to cover area and anchor area". If that's an "OverlayRunner" concern (as it is for our in-house library) then the runner would receive the cover-area, anchor-area, etc, rects and on each updated value recalculate the final position for the dialog (and the scrim).

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/DialogRunner.kt
```diff
@@ -0,0 +1,151 @@
...
+import com.squareup.workflow1.ui.ViewEnvironment
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+import com.squareup.workflow1.ui.androidx.WorkflowLifecycleOwner
+import com.squareup.workflow1.ui.compatible
+
+@WorkflowUiExperimentalApi
+internal class DialogRunner<T : Overlay>(
+  initialRendering: T,
+  initialViewEnvironment: ViewEnvironment,
+  private val bounds: RectProvider,
```
helios175, line 25: This seems to say:
the creator of a dialog runner will say where it should appear as opposed to:
- I will tell it the "cover area", the "anchor area", etc, and let the runner and its idiosyncrasy to decide where to appear depending on those inputs.

helios175: As I see now, the particular implementations of overlays are not the dialog runners, but the overlay dialog factories. My comment still holds: These bounds might be cover-area, anchor-area and let the overlay dialog factory to update the final location depending on the inputs.

Something like:
```
Runner {
+ coverArea: RectProvider
+ anchorArea: RectProvider

  // when dialog is shown
  coverArea.onChange { coverRect = it }
  anchorArea.onChange { anchorRect = it }
  // when dialog is dismissed
  coverArea.onChange = null
  anchorArea.onChange = null
  
  protected coverRect: Rect by observable(...) { updateDialogLocation(...) }
  protected coverRect: Rect by observable(...) { updateDialogLocation(...) }
  fun updateDialogLocation() {
    someone has to decide how to calculate dialog position depending on cover + anchor
    that is not clear to me now, because Runner is fixed, and factory doesn't receiver cover + anchor
    I'd say that particular factories now the dialog they create
    and they also know what positioning they implement
    so they have the responsibility to update (or provide the cover+anchor -> bounds lambda at least).
  }
}
```

helios175: Question is: what's the conceptual difference between a runner and a factory, unless the runner contains the common code (instead of having a base factory class, which might be a good idea).

helios175: As I see now, the particular implementations of overlays are not the dialog runners, but the overlay dialog factories. My comment still holds: These bounds might be cover-area, anchor-area and let the overlay dialog factory to update the final location depending on the inputs.

Something like:
```kotlin
Runner {
+ coverArea: RectProvider
+ anchorArea: RectProvider

  // when dialog is shown
  coverArea.onChange { coverRect = it }
  anchorArea.onChange { anchorRect = it }
  // when dialog is dismissed
  coverArea.onChange = null
  anchorArea.onChange = null
  
  protected coverRect: Rect by observable(...) { updateDialogLocation(...) }
  protected coverRect: Rect by observable(...) { updateDialogLocation(...) }
  fun updateDialogLocation() {
    dialogOrNull
  }
}
```

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/ListenToCoordinates.kt
```diff
@@ -0,0 +1,101 @@
...
+ *       view changes position -> block()
+ *       view changes position -> block()
+ *       view is detached -> no more listening.
+ *
+ * @param block lambda to execute for the current position and each future change.
+ * @return a lambda that can be invoked to cancel the tracking.
+ */
+internal fun View.listenToCoordinates(
+  block: (Rect) -> Unit
+): () -> Unit {
```
steve-the-edwards, line 34: Couldn't we throw a simple alias on this so it is more identifiable as a `Canceller`?

```
typealias Cancel = () -> Unit
```

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/ListenToCoordinates.kt
```diff
@@ -0,0 +1,101 @@
...
+  }
+  // If we already have valid coordinates call back for the first time.
+  if (view.height > 0 && view.width > 0) {
+    layoutListener.onGlobalLayout()
+  }
+  // And listen for it.
+  view.viewTreeObserver.addOnGlobalLayoutListener(layoutListener)
+
+  // Return a lambda that can be called to cancel the coordinates tracking.
+  lateinit var cancelHandler: () -> Unit
```
steve-the-edwards, line 85: TIL you can `lateinit` a var in a function.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/BodyAndModalsContainer.kt
```diff
@@ -0,0 +1,134 @@
...
+import com.squareup.workflow1.ui.ManualScreenViewFactory
+import com.squareup.workflow1.ui.R
+import com.squareup.workflow1.ui.ScreenViewFactory
+import com.squareup.workflow1.ui.ViewEnvironment
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+import com.squareup.workflow1.ui.WorkflowViewStub
+import com.squareup.workflow1.ui.bindShowRendering
+
+@WorkflowUiExperimentalApi
+internal class BodyAndModalsContainer @JvmOverloads constructor(
```
steve-the-edwards, line 23: At the risk of another bikeshed why isn't this `BodyAndOverlaysContainer`? And the screen `BodyAndOverlaysScreen`?

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/DialogStack.kt
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ * Does the work of maintaining a set of [Dialog][android.app.Dialog]s
+ * to reflect lists of [Overlay].
+ */
+@WorkflowUiExperimentalApi
+public class DialogStack(
+  private val context: Context,
+  private val getParentLifecycleOwner: () -> LifecycleOwner?,
+  private val createRectProvider: (Overlay) -> RectProvider
```
steve-the-edwards, line 24: `createRectProvider` could use some kdoc explaining why we might want to override it.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/DialogStack.kt
```diff
@@ -0,0 +1,104 @@
...
+          DialogRunner(
+            overlay, overlayEnvironment, createRectProvider(overlay), context, dialogFactory
+          ).apply {
+            showRendering(overlay, overlayEnvironment)
+            beforeShowing()
+            show(getParentLifecycleOwner())
+          }
+        }
+      }
+    }
```
steve-the-edwards, line 76: I wish I could pretend to understand what this class is handling for us but I think i'm lost 😵

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/container/BodyAndModalsScreen.kt
```diff
@@ -0,0 +1,20 @@
...
+import com.squareup.workflow1.ui.Screen
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+
+/**
+ * A screen that may stack a number of modal [Overlay]s over a body.
+ * While modals are present, the body is expected to ignore any
+ * input events -- touch, keyboard, etc.
+ */
+@WorkflowUiExperimentalApi
+public class BodyAndModalsScreen<B : Screen, M : Overlay>(
```
steve-the-edwards, line 12: `BodyAndOverlaysScreen`?

steve-the-edwards (approved):
Thank-you for the final kdoc commit. I will likely need more to understand everything going on here, but hopefully @helios175 has a better understanding?

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/RectProvider.kt
rjrjr: I can see the container providing the cover area -- that's what `BodyAndModalsContainer` does in this PR. But if the dialog is deciding its type, and therefore its size, it's also deciding its scrim area: cover area - body area == scrim area. Making the container responsible for both seems like it makes things much harder to customize.

It also makes sense that the container would provide the anchor view. I think the mechanism here accomodates both of those cases. I meant to imply that with this example from the other day:

```kotlin
BodyAndModals(
  body = 
    BodyAndPopUp(
      body = HomeScreen(),
      popUp = TutorialPrompt(
        targetId = "magic tutorial tag value",
        content = NowClickChargeStep()
      )
    ),
  modals: listOf<Overlay>(
    AlertModal("Oh Well", "Guess you screwed up", listOf(AlertButton("I sure did"))),
    CardModal(BackStackScreen(StepOne(), StepTwo()),
    SheetModal(SomethingThatCoversTheBody)
  )
)
```

- `BodyAndModals`, in this PR, is a container that sets `RectProvider` based on its own bounds.
  - `AlertModal` ignores RectProvider and puts up an `AlertDialog`
  - `CardModal` does the center-the-modal-in-a-scrim thing
    - Today's implementation makes a dialog that covers the entire `RectProvider` area with a transparent scrim, and centers the `BackStackScreen` within it.
    - Perhaps it would be simpler if `BodyAndModalsContainer` did its own scrim effect, and `CardAndModal` just made a centered dialog.
  - `SheetModal` fills the entire `RectProvider`
- `BodyAndPopup` would use a new container, also a built around `DialogStack`, that sets `RectProvider` to the bounds of an anchor view
  - `TutorialPrompt` understands this convention, and interprets it appropriately.
  - Probably `TutorialPrompt` implements `interface PopUp(val: idOfAnchorView): Overlay`, and `BodyAndPopup.popUp` is of type `PopUp`

That seems like it covers all of your use cases. What am I missing? Do we ever really need both a cover area and an anchor area?

> This seems to say: the creator of a dialog runner will say where it should appear

So I guess what I'm actually proposing is that the owner of a dialog can identify a rectangle of interest. Different containers can have different conventions for what that means, honored by particular dialog factories / runners.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/ListenToCoordinates.kt
rjrjr: Yeah, I think this API does need work. But I don't actually use it at all in this PR, should hold off on refining it until it's actually driving something.

Hmm. As I type this, I'm thinking that `RectProvider` is actually redundant with the `ViewEnvironment`. In the example from my reply to @helios175, it seems like the container could just provide the cover area and target area hints that way. It would still use `listenToCoordinates` to know when to update the rect, but there's no benefit to building that callback API directly into `DialogFactory`.

I should really rebase this PR to remove `ListenToCoordinates` and `RectProvider`, and introduce them with the next PR when/if they're put into action.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/RectProvider.kt
rjrjr: As we discuss this, let's remember that our internal API can still be in the shape it is now, so long as it can be driven from this with a minimum of fuss. Maybe `DialogRunner` as I'm using it here is just a different thing than your class of that name that I'm cribbing from.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/BodyAndModalsContainer.kt
rjrjr: Well spotted, and not an accident. I've realized that modals are a subset of overlays, and this name is meant to respect that distinction -- a modal is an overlay whose presence implies that things beneath it are blocked from receiving any events. Their presence is a visual indicator that we are in a different mode of the app. Contrast this with tooltips and other pop-ups that are also naturally modeled via Overlay, and which should not have the side effect of modality.

Fell into this because the container itself actually has to participate -- dialogs simply can't do the job all by themselves, as we've re-learned painfully through BugSnag in the past year.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/DialogStack.kt
rjrjr: Yeah. But I've just ripped it out of this PR, and I suspect it will come back in a different form.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/DialogStack.kt
rjrjr: Rather than explain here I'll add comments, and we can see if they get the job done.

rjrjr: @steve-the-edwards PTAL

https://github.com/square/workflow-kotlin/pull/588/files#diff-9c32c5692c4fa791f87816ff6af4ee5e17cbde6276a8932c766891a1349d026fR46

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/DialogStack.kt
```diff
@@ -0,0 +1,146 @@
...
+        overlay.toDialogFactory(overlayEnvironment).let { dialogFactory ->
+          DialogRunner(
+            overlay, overlayEnvironment, context, dialogFactory
+          ).also { newRunner ->
+            // Has the side effect of creating the dialog if necessary.
+            newRunner.showRendering(overlay, overlayEnvironment)
+            // Custom behavior from the container to be fired when we show a new dialog.
+            // The default modal container uses this to flush its body of partially
+            // processed touch events that should have been blocked by the modal.
+            beforeShowing()
```
rjrjr, line 89: Probably will need to tweak this to pass the Dialog itself in the next PR.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/DialogStack.kt
```diff
@@ -0,0 +1,146 @@
...
+            // to host an androidx ViewTreeLifecycleOwner and SavedStateRegistryOwner
+            // for each dialog. These are generally expected these days, and absolutely
+            // required by Compose.
+            newRunner.show(getParentLifecycleOwner())
+          }
+        }
+      }
+    }
+
+    (runners - newRunners.toSet()).forEach { it.dismiss() }
```
rjrjr, line 100: The IDE prompted me to add the `toSet()` call as an optimization, and I'm very suggestible.

rjrjr:
Damn espresso tests are so flaky, I can't tell if I've made them more so.

rjrjr:
Simply cannot get a green build. #590 is awful, and clearly unrelated to this PR, we didn't change that app or anything it uses.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/BodyAndModalsContainer.kt
steve-the-edwards: Yes this makes total sense and is a great place to settle. We should document this.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/DialogRunner.kt
```diff
@@ -0,0 +1,150 @@
...
+      dialog.window?.decorView?.let(WorkflowLifecycleOwner::get)?.destroyOnDetach()
+      dialog.dismiss()
+    }
+  }
+
+  fun canShowRendering(rendering: Overlay): Boolean {
+    return compatible(this.rendering, rendering)
+  }
+
+  fun showRendering(
```
steve-the-edwards, line 84: I would change this to `bindRendering` as I'm thinking of the RecyclerView ViewHolder analogy. What you are doing is 'binding' the rendering to this dialog - nothing is actually _changing_ with respect to its shown/hidden state.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/DialogRunner.kt
```diff
@@ -0,0 +1,150 @@
...
+import androidx.lifecycle.LifecycleOwner
+import androidx.lifecycle.OnLifecycleEvent
+import com.squareup.workflow1.ui.Compatible
+import com.squareup.workflow1.ui.ViewEnvironment
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+import com.squareup.workflow1.ui.androidx.WorkflowLifecycleOwner
+import com.squareup.workflow1.ui.compatible
+
+@WorkflowUiExperimentalApi
+internal class DialogRunner<T : Overlay>(
```
steve-the-edwards, line 22: I would change this to `DialogHolder`. I realize we use `Runner` elsewhere but I also just realized that while I got the 'Runner' metaphor the 'ViewHolder' analogy (as in the RecyclerView sense) is much more powerful to me as an Android person with RecyclerView experience.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/DialogRunner.kt
```diff
@@ -0,0 +1,150 @@
...
+    // The dialog's views are about to be detached, and when that happens we want to transition
+    // the dialog view's lifecycle to a terminal state even though the parent is probably still
+    // alive.
+    dialogOrNull?.let { dialog ->
+      dialog.window?.decorView?.let(WorkflowLifecycleOwner::get)?.destroyOnDetach()
+      dialog.dismiss()
+    }
+  }
+
+  fun canShowRendering(rendering: Overlay): Boolean {
```
steve-the-edwards, line 80: `canBindRendering`?

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/DialogStack.kt
```diff
@@ -0,0 +1,146 @@
...
+import com.squareup.workflow1.ui.container.DialogRunner.KeyAndBundle
+
+/**
+ * Does the bulk of the work of maintaining a set of [Dialog][android.app.Dialog]s
+ * to reflect lists of [Overlay]. Can be used to create custom [Overlay]-based
+ * layouts if [BodyAndModalsScreen] or the default [View] bound to it are too restrictive.
+ * Provides a [LifecycleOwner] per managed dialog, and view persistence support.
+ */
+@WorkflowUiExperimentalApi
+public class DialogStack(
```
steve-the-edwards, line 21: `DialogStackContainer` we use the `Container` prefix for these logic classes that deal with interaction between groups of Views bound to Renderings no?

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/container/BodyAndModalsScreen.kt
```diff
@@ -0,0 +1,20 @@
+package com.squareup.workflow1.ui.container
+
+import com.squareup.workflow1.ui.Screen
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+
+/**
+ * A screen that may stack a number of modal [Overlay]s over a body.
+ * While modals are present, the body is expected to ignore any
+ * input events -- touch, keyboard, etc.
+ */
```
steve-the-edwards, line 10:
```suggestion
/**
 * A screen that may stack a number of modal [Overlay]s over a body.
 * While modals are present, the body is expected to ignore any
 * input events -- touch, keyboard, etc.
 * Passing an [Overlay] rendering as part of [modals] here marks that
 * Overlay as modal - that is, it should block input events around outside
 * of its content area. The logic for that input event blocking is handled
 * for users of this screen by the internal [BodyAndModalsContainer].
 */
```

steve-the-edwards (commented):
Definitely understand now! Still have more naming change requests that I think coudl be helpful.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/DialogRunner.kt
rjrjr: I don't know why exactly but `bind` bugs me. How about `canTakeRendering` and `takeRendering`?

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/DialogRunner.kt
rjrjr: Great call. When we use `Runner` elsewhere it really means something else entirely.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/DialogStack.kt
rjrjr: `Container` isn't good b/c that's consistently meant "a workflow view that displays other workflow views"

And as @helios175 pointed out, `Stack` isn't good because this is so completely different from `BackStack`.

How about `LayeredDialogHolders`.

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/container/BodyAndModalsScreen.kt
rjrjr: > it should block input events around outside of its content area

Oh, no no no no no. It should only block events within the bounds of the body. Blocking events outside of that area is a contract violation. And in fact turning on the "modal" flag for an Android Dialog implementation would be a bug precisely because of this.

I'm too tired to phrase this properly right now. Will come up with something before merging.

rjrjr: Actually…let me leave this as is and rephrase with the next PR, where I'll actually be implementing proper custom dialogs. There may be more nuances then.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/DialogStack.kt
rjrjr: Oh, but `DialogHolder` is a private type. So just `LayeredDialogs`.

rjrjr:
Update with name changes proposed during review:

- `DialogRunner` > `DialogHolder`
- `DialogStack` > `LayeredDialogs`
- `canShowRendering` > `canTakeRendering`
- `showRendering` > `takeRendering`

Still need to find a home for the essay explaining `Overlay` v. `Modal`. Guess I'll put it right on the `Overlay` interface.

rjrjr:
Updated the kdoc on `Overlay`:

```kotlin
/**
 * Marker interface implemented by window-like renderings that map to a layer above
 * a base [Screen][com.squareup.workflow1.ui.Screen].
 *
 * Note that an Overlay is not necessarily a modal window, though that is how
 * they are used in [BodyAndModalsScreen]. An Overlay can be any part of the UI
 * that visually floats in a layer above the main UI, or above other Overlays.
 * Possible examples include alerts, drawers, and tooltips.
 * 
 * Whether or not an Overlay's presence indicates that events are blocked from lower
 * layers is a separate concern.
 */
@WorkflowUiExperimentalApi
public interface Overlay
```

If no one complains I'll merge this when CI blesses it.

Issue #589, opened 2021-11-17
Deprecate AlertContainerScreen, AlertScreen, HasModals, etc.


rjrjr:

rjrjr:
Fixed by #595

Issue #204, opened 2019-08-26
Dialogs isn't dismissed when workflow stops rendering it without interaction on the dialog itself


zach-klippenstein:
If a workflow renders a dialog with a listener on the buttons that result in the dialog no longer being rendered, everything is fine.

However, if the workflow suddenly stops rendering the dialog _without_ the user having clicked on the dialog buttons, the dialog is not dismissed.

Repro steps:
1. Render a dialog with `AlertContainerScreen`.
2. After a timeout, stop rendering it. Dialog will still be around.

Repro project: https://github.com/square/droidcon-nyc-2019-workflow-game (win the game, then enter time travel mode and scrub past the "game over" dialog)

rjrjr:
Fixed by #595

Issue #138, opened 2020-07-22
Buttons on successive AlertScreen renderings are not respected


christiandeange:
When rendering two different `AlertScreen`s in back-to-back render passes, the button contents do not seem to be updated by the `AlertContainer` properly.

| First Rendering | Second Rendering |
|----------------|--------------------|
| <code>AlertScreen(message = "Starting Checkout...", onEvent = {})</code> | <code>AlertScreen(message = "Waiting for checkout...", buttons = mapOf(NEGATIVE to "Cancel"), onEvent = { ... })</code> |
| ![Screenshot_1595382241](https://user-images.githubusercontent.com/3208823/88193018-30950300-cc0b-11ea-8e0e-26727fd755cc.png) | ![Screenshot_1595382242](https://user-images.githubusercontent.com/3208823/88193020-312d9980-cc0b-11ea-9e8d-e2297953ca9f.png) |

At first I was able to work around this locally by using some reflection to manually force the buttons to be re-bound to the UI again:

```
val buttonPanel = dialog.findViewById<ViewGroup>(androidx.appcompat.R.id.buttonPanel)
dialog.javaClass.getDeclaredField("mAlert")
    .let { mAlert ->
      mAlert.isAccessible = true
      mAlert[dialog].javaClass.getDeclaredMethod("setupButtons", ViewGroup::class.java)
          .let { setupButtons ->
            setupButtons.isAccessible = true
            setupButtons(mAlert[dialog], buttonPanel)
          }
    }
```

Eventually I just decided to implement my own `AlertContainer`-like container to subvert this entirely.

rjrjr:
Fixed by #595

PR #595, opened 2021-12-02
Introduces ScreenOverlay and ModalScreenOverlayDialogFactory

### LINKED ISSUE #259: ModalViewContainer should not monopolize Dialog.setOnKeyListener
We use `Dialog.setOnKeyListener` for back button handling. If a subclass tries to set its own listener, it'll get stomped. Instead we should wrap `Dialog.window.callBack`.

rjrjr:
`ScreenOverlay` marks an `Overlay` with a `Screen` defining its content.

`ModalScreenOverlayDialogFactory` can show a `ScreenOverlay` as an
`android.app.Dialog`, taking care of updating the content view and managing
the Back button. It's the much simpler replacement for `ModalViewContainer`.

`ModalScreenOverlayDialogFactory` works with `BodyAndModalsContainer`
and `LayeredDialogs` to perform two important tricks:

 - `updateBounds(Dialog, Rect)`, which allows dialogs to be restrictred to a subset
   of the screen. Implemented via the private `ModalArea` value in `ViewEnvironment`

 - Touch and keyboard events are blocked by views that are covered by
   dialogs managed by a `ModalScreenOverlayDialogFactory`. Same for
   dialogs with a lower z order. We use another private `ViewEnvironment`
   value for that, `CoveredByModal`.

In sample code, `PanelOverlay` replaces `PanelContainerScreen`, and the Tic
Tac Toe sample is updated to use the new hotness. I think the diff of the
sample code really highlights how much the new marker interfaces improve our
composition story.

That said, the `ScrimScreen` bit in the sample is a little rough. In the old
code we magically swizzled the scrim into place automatically. This PR
seems big enough already, so I'll follow up with a clean up that does
something similar and deletes all the deprecated sample code.

Fixes #259, #138, #99, #204, #314, #589


samples/tictactoe/common/src/main/java/com/squareup/sample/gameworkflow/RunGameWorkflow.kt
```diff
@@ -69,90 +78,101 @@ class RealRunGameWorkflow(
...
+          )
         )
-      )
-    }
+      }
 
-    is MaybeQuittingForSure -> {
-      nestedAlertsScreen(
-        GamePlayScreen(renderState.playerInfo, renderState.completedGame.lastTurn),
-        maybeQuitScreen(),
```
rjrjr, line 122: This was a shameful hacky workaround for #138

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/AndroidDialogBounds.kt
```diff
@@ -0,0 +1,77 @@
...
+import com.squareup.workflow1.ui.environment
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.launchIn
+import kotlinx.coroutines.flow.onEach
+
+@WorkflowUiExperimentalApi
+public fun Dialog.setBounds(bounds: Bounds) {
```
helios175, line 20: Nice!

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/AndroidDialogBounds.kt
```diff
@@ -0,0 +1,77 @@
...
+  window?.let { window ->
+    val windowRect = WindowManager.LayoutParams()
+      .apply {
+        copyFrom(window.attributes)
```
helios175, line 24: Also you can reassign the same attributes. I can confirm it works:
`attributes = attributes.apply { ... }`
I know it might be smelly but...

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/AndroidDialogBounds.kt
```diff
@@ -0,0 +1,77 @@
...
+@WorkflowUiExperimentalApi
+public fun <D : Dialog> D.maintainBounds(
+  environment: ViewEnvironment,
+  onBoundsChange: (D, Bounds) -> Unit = { d, b -> d.setBounds(b) }
+) {
+  maintainBounds(environment[ModalArea].bounds, onBoundsChange)
+}
+
+@WorkflowUiExperimentalApi
+public fun <D : Dialog> D.maintainBounds(
```
helios175, line 53: 😘👌

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/AndroidDialogBounds.kt
```diff
@@ -0,0 +1,77 @@
...
+  bounds: StateFlow<Bounds>,
+  onBoundsChange: (D, Bounds) -> Unit = { d, b -> d.setBounds(b) }
+) {
+  val window = requireNotNull(window) { "Dialog must be attached to a window." }
+  window.setFlags(FLAG_NOT_TOUCH_MODAL, FLAG_NOT_TOUCH_MODAL)
```
helios175, line 58: Wondering if this needs to be here? (I mean: one method one responsibility)

samples/containers/android/src/main/java/com/squareup/sample/container/panel/PanelOverlayDialogFactory.kt
```diff
@@ -0,0 +1,50 @@
...
+    return Dialog(context, R.style.PanelDialog).also { dialog ->
+      dialog.setContentView(contentView)
+
+      val typedValue = TypedValue()
+      context.theme.resolveAttribute(android.R.attr.windowBackground, typedValue, true)
+      if (typedValue.type in TypedValue.TYPE_FIRST_COLOR_INT..TypedValue.TYPE_LAST_COLOR_INT) {
+        dialog.window!!.setBackgroundDrawable(ColorDrawable(typedValue.data))
+      }
+
+      dialog.maintainBounds(contentView) { _, bounds ->
```
helios175, line 29: Quick thought: what do you think of separating positioning from creation?
Here you are:
1) creating the dialog
2) positioning it for the first time
3) maintaining positions

How to do that?
Paraphrasing Market: the thing that controls the dialog (in Market the runner)i
- implements buildDialog
- receives the values (or flow) for cover area, anchor view and what not.
A probably base code would react to any of those changes by doing:
{ if dialog showing -> updateDialogLocation }
- concrete runner/controller class would only override updateDialogLocation.
In that case: you're not leveraging the great maintain bounds below.
But, you could still do:
maintainBounds(coverAreaFlow) { updateDialogLocation }
maintainBounds(anchorAreaFlow) { updateDialogLocation }

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/AndroidDialogBounds.kt
rjrjr: I'll hold my nose to avoid allocs.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/AndroidDialogBounds.kt
rjrjr: Nice catch. At the very least I should call it out in the kdoc. I _think_ it's the right thing to do because dialogs are modal by default, and so restricting the bounds will be pointless. I'll confirm that.

WDYT?

samples/containers/android/src/main/java/com/squareup/sample/container/panel/PanelOverlayDialogFactory.kt
rjrjr: I'd still want to pass the `StateFlow<Bounds>` through the `ViewEnvironment`, so that that whole notion of positioning doesn't have to be built into `LayeredDialogs` and `OverlayDialogFactory` -- different containers can create different policies.

But the more specialized `ModalScreenOverlayDialogFactory` could certainly offer an `abstract fun updateBounds(bounds: Bounds)`, would be more foolproof. And our app can build an `AnchorOverlayDialogFactory` for that use case, etc.

Sold.

rjrjr: Probably still use `Dialog.maintainBounds` to implement that, but now it doesn't have to be public. I like that.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/AlertOverlayDialogFactory.kt
```diff
@@ -42,25 +43,45 @@ internal object AlertOverlayDialogFactory : OverlayDialogFactory<AlertOverlay> {
...
-        }
+      // Force all the buttons to be created so that we can show
+      // and hide them at will during updates. They get created
+      // when show is called, so we can't mess with them before that
+      // (e.g. to fix the dismiss-on-click nonsense),
+      // and we can't create new ones after that.
+      //
+      // https://github.com/square/workflow-kotlin/issues/138
+      // Why " "? An empty string means no button.
+      alertDialog.setButton(button.toId(), " ") { _, _ -> }
```
rjrjr, line 54: @christiandeange This the fix for #138, remember that? Can you believe how hacky?

rjrjr: I should move this part to `buildDialog` instead of doing it here in `updateDialog`, less confusing.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/AndroidDialogBounds.kt
rjrjr: Moved it from here to `buildDialog`, called it out on the kdoc for `ModalScreenOverlayDialogFactory`.

rjrjr:
> That said, the ScrimScreen bit in the sample is a little rough. In the old
> code we magically swizzled the scrim into place automatically, and I think in
> prod I would still do the same, but the sample is complex enough already.

I think I actually need to address that in this PR. Panel is a pretty high fidelity
sample, used all over the place. Now is the time to figure out how to make it usable.

rjrjr:
Or maybe the thing to do is undelete the old PanelContainer and get this merged, since the PR is already too big. Then follow up with updating all the other samples that use the old modal stuff.

samples/containers/android/src/main/java/com/squareup/sample/container/panel/PanelOverlayDialogFactory.kt
```diff
@@ -0,0 +1,52 @@
...
+import android.util.TypedValue
+import android.view.View
+import com.squareup.sample.container.R
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+import com.squareup.workflow1.ui.container.Bounds
+import com.squareup.workflow1.ui.container.ModalScreenOverlayDialogFactory
+import com.squareup.workflow1.ui.container.setBounds
+
+@OptIn(WorkflowUiExperimentalApi::class)
+internal object PanelOverlayDialogFactory : ModalScreenOverlayDialogFactory<PanelOverlay<*>>(
```
rjrjr, line 14: Holding off on kdoc for the Panel stuff until I polish it up in the next PR.

rjrjr:
Green enough, failure is due to ##590.

samples/containers/android/src/main/java/com/squareup/sample/container/SampleContainers.kt
```diff
@@ -1,12 +1,11 @@
...
-import com.squareup.sample.container.panel.ScrimContainer
-import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+import com.squareup.sample.container.panel.PanelOverlayDialogFactory
 import com.squareup.workflow1.ui.ViewRegistry
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
 
 @OptIn(WorkflowUiExperimentalApi::class)
 val SampleContainers = ViewRegistry(
-    BackButtonViewFactory, OverviewDetailContainer, PanelContainer, ScrimContainer
+  BackButtonViewFactory, OverviewDetailContainer, PanelOverlayDialogFactory, ScrimContainer
```
rjrjr, line 10: Oh! I thought `PanelContainerScreen` was more broadly used, but it was actually only in the Tic Tac Toe sample.

- [ ] delete it

rjrjr: And maybe the old scrim screen too.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/ScreenOverlayOnBackPressed.kt
```diff
@@ -0,0 +1,40 @@
...
+ * The function called to handle back button events in [Dialog][android.app.Dialog]s built
+ * by [ModalScreenOverlayDialogFactory]. The default implementation uses the
+ * [Activity][android.app.Activity]'s
+ * [OnBackPressedDispatcher][androidx.activity.OnBackPressedDispatcher].
+ *
+ * This is a hook to accommodate apps that have a back button handling scheme
+ * that predates `OnBackPressedDispatcher`.
+ */
+@WorkflowUiExperimentalApi
+public object ScreenOverlayOnBackPressed : ViewEnvironmentKey<Handler>(
```
rjrjr, line 19: - [ ] Rename this `BodyAndModalsOnBackPressed`, since that's where it's used.

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/container/Bounds.kt
```diff
@@ -0,0 +1,51 @@
+package com.squareup.workflow1.ui.container
+
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+
+@WorkflowUiExperimentalApi
+public class Bounds(
```
rjrjr, line 6: - [ ] kdoc

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/container/ModalArea.kt
```diff
@@ -0,0 +1,18 @@
...
+import com.squareup.workflow1.ui.ViewEnvironmentKey
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+
+/**
+ * Reports the the area of the screen whose events should be blocked by any modal [Overlay].
+ */
+@WorkflowUiExperimentalApi
+public class ModalArea(
```
rjrjr, line 12: internal?

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/container/CoveredByModal.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import com.squareup.workflow1.ui.ViewEnvironmentKey
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+
+/**
+ * True in views managed by [BodyAndModalsScreen] when their events are being blocked
+ * by a modal [Overlay].
+ */
+@WorkflowUiExperimentalApi
+public object CoveredByModal : ViewEnvironmentKey<Boolean>(type = Boolean::class) {
```
rjrjr, line 11: internal?

rjrjr:
Flip flopping on panels again. Since they are only used in the TTT sample, gonna clean them up in this PR and be done with it.

rjrjr:
Update deletes some more unused stuff.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/BodyAndModalsContainer.kt
```diff
@@ -26,32 +29,80 @@ internal class BodyAndModalsContainer @JvmOverloads constructor(
...
 
-  private val dialogs = LayeredDialogs(this)
+  private val dialogs = LayeredDialogs(view = this, modal = true)
+
+  private val bounds = MutableStateFlow(Bounds())
+  private val boundsRect = Rect()
+  private val boundsListener = OnGlobalLayoutListener {
+    getGlobalVisibleRect(boundsRect)
+    if (!boundsRect.matches(bounds.value)) bounds.value = boundsRect.toBounds()
+  }
```
steve-the-edwards, line 43: here we are just extracting the 'Global Visible Rect' right? - what does that represent?

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/container/Bounds.kt
```diff
@@ -0,0 +1,52 @@
...
+    } ?: false
+  }
+
+  override fun hashCode(): Int {
+    var result = left
+    result = 31 * result + top
+    result = 31 * result + right
+    result = 31 * result + bottom
+    return result
+  }
```
steve-the-edwards, line 47: Is this the reason we create the `Bounds` class and don't use `Rect`? couldn't some of this just be extensions on `Rect`? What am I missing?

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/DialogHolder.kt
```diff
@@ -93,6 +131,7 @@ internal class DialogHolder<T : Overlay>(
     @Suppress("UNCHECKED_CAST")
     this.rendering = rendering as T
     this.environment = environment
+    allowEvents = !modal || !environment[CoveredByModal]
```
steve-the-edwards, line 134: This is another aside - but since we take such care in locking up this modal functionality (and in some cases making it tighter to avoid certain edge cases) - why are we using `Dialog` types at all? Just for the separate window? Couldn't we just create our own `Window`? I guess i get it for `AlertDialog` since that gives us some simpled themed UI.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/ModalScreenOverlayDialogFactory.kt
```diff
@@ -0,0 +1,100 @@
...
+import java.lang.IllegalStateException
+import kotlin.reflect.KClass
+
+/**
+ * Convenient base class for building [ScreenOverlay] UIs that are compatible
+ * with [View.backPressedHandler], and which honor the [ModalArea] constraint
+ * placed in the [ViewEnvironment] by the standard [BodyAndModalsScreen] container.
+ *
+ * Ironically, [Dialog] instances are created with [FLAG_NOT_TOUCH_MODAL], to ensure
+ * they do not interfere with events outside of [ModalArea].
```
steve-the-edwards, line 28:
```suggestion
 * Ironically, [Dialog] instances are created with [FLAG_NOT_TOUCH_MODAL], to ensure
 * they will pass on events outside of [ModalArea] to layers below.
```

steve-the-edwards (approved):

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/container/Bounds.kt
rjrjr: This is a java module, `workflow-ui:core-common`, not `:core-android`. We keep the built in rendering types and `ViewEnvironment` entries platform neutral.

That said, you could argue that this should just be an Android implementation detail rather than part of common. But even then, `Rect` freaks me out because it's mutable and we're passing it through a `StateFlow`. App code could stomp it.

Buuuuut…if this were in `:core-android`, I think the environment stuff could be internal. We'd only have to worry about library code using `Rect` badly, which seems managable.

WDYT? It's pretty silly for this to be anywhere but Android, isn't it?

samples/containers/android/src/main/java/com/squareup/sample/container/panel/PanelOverlayDialogFactory.kt
rjrjr: Gonna push back. This is sample code, not for general use. I'm even tempted to rename it just `PanelDialogFactory`. Regardless, inclined to hold off on painting this part of the bikeshed until the next PR, with improved Panel fu.

samples/containers/android/src/main/java/com/squareup/sample/container/panel/PanelOverlayDialogFactory.kt
```diff
@@ -0,0 +1,55 @@
...
+  override fun buildDialogWithContentView(contentView: View): Dialog {
+    val context = contentView.context
+    return Dialog(context, R.style.PanelDialog).also { dialog ->
+      dialog.setContentView(contentView)
+
+      val typedValue = TypedValue()
+      context.theme.resolveAttribute(android.R.attr.windowBackground, typedValue, true)
+      if (typedValue.type in TypedValue.TYPE_FIRST_COLOR_INT..TypedValue.TYPE_LAST_COLOR_INT) {
+        dialog.window!!.setBackgroundDrawable(ColorDrawable(typedValue.data))
+      }
```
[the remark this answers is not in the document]
rjrjr: Sample code, so don't worry too much. And I'm afraid it's pretty common Android boilerplate -- I'd even argue idiomatic. Compose can't come fast enough.

But yes, that's a better name, and a comment can help. Not inclined to write a dissertation, though, b/c it's just how Android life is.

samples/tictactoe/common/src/main/java/com/squareup/sample/gameworkflow/RunGameWorkflow.kt
```diff
@@ -69,90 +78,106 @@ class RealRunGameWorkflow(
...
+        val emptyGameScreen = GamePlayScreen()
+
+        RunGameRendering(
+          gameScreen = emptyGameScreen,
+          namePrompt = object : ScreenOverlay<Screen> {
+            override val content = NewGameScreen(
+              renderState.defaultXName,
+              renderState.defaultOName,
+              onCancel = context.eventHandler
+              { setOutput(CanceledStart) },
```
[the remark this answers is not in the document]
rjrjr: Gonna blame JetBrains for that, thanks for catching it.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/AndroidDialogBounds.kt
```diff
@@ -0,0 +1,69 @@
...
+    var job: Job? = null
+
+    override fun onAttachedToWindow() {
+      job = bounds.onEach { onBoundsChange(this@maintainBounds, it) }
+        .launchIn(scope)
+    }
+
+    override fun onDetachedFromWindow() {
+      job?.cancel()
+      job = null
```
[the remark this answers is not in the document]
rjrjr: Because I'm still really bad at Coroutines. Thanks.

Note that we use the same technique in `WorkflowLayout.takeWhileAttached`, will fix that too.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/BodyAndModalsContainer.kt
rjrjr: It's the bounds of the view in global coordinates -- the screen's coordinate system, not the parent view's. But I just noticed that this method returns false "If the view is completely clipped or translated out". I imagine the contents of the `Rect` are nonsense in that case.

Not even sure what to do if that happens. Probably should close all the dialogs.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/DialogHolder.kt
rjrjr: I don't think this is a `Dialog` issue, I think it's a `Window` issue. And `Window` is a PITA to use directly. Also remember that we're allowing app developers to build their own. Don't want to force them to use `Window` for that.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/ModalScreenOverlayDialogFactory.kt
rjrjr: That's misleading, though. I don't think a non-modal window is involved in handling such events, that's the window manager I'm pretty sure.

How about:

```kotlin
 * Ironically, [Dialog] instances are created with [FLAG_NOT_TOUCH_MODAL], to ensure
 * that events outside of [ModalArea] reach views in lower windows.
```

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/container/Bounds.kt
steve-the-edwards: Gotcha. I didn't catch that nuance. I don't know what will be best but I think starting narrow with Android as it's our only use case right now makes sense.

samples/containers/android/src/main/java/com/squareup/sample/container/panel/PanelOverlayDialogFactory.kt
steve-the-edwards: Ya - I can't settle on whether I think of this as _just_ sample code or as paradigmatic samples that act like documentation. But your point stands and is likely the most important one. Keep it simple and don't let perfect be the enemy of good.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/BodyAndModalsContainer.kt
steve-the-edwards: Ok. Sounds good.

samples/containers/android/src/main/java/com/squareup/sample/container/panel/PanelOverlayDialogFactory.kt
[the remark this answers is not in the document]
rjrjr:
```kotlin
      // Welcome to Android. Nothing workflow-related here, this is just how one
      // finds the window background color for the theme. I sure hope it's better in Compose.
      val maybeWindowColor = TypedValue()
      context.theme.resolveAttribute(android.R.attr.windowBackground, maybeWindowColor, true)
      if (maybeWindowColor.type in TypedValue.TYPE_FIRST_COLOR_INT..TypedValue.TYPE_LAST_COLOR_INT) {
        dialog.window!!.setBackgroundDrawable(ColorDrawable(maybeWindowColor.data))
      }
```

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/AndroidDialogBounds.kt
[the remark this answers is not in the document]
rjrjr: Does this look right?

```kotlin
  window.callback = object : Window.Callback by window.callback {
    var scope : CoroutineScope? = null

    override fun onAttachedToWindow() {
      scope  = CoroutineScope(Dispatchers.Main.immediate).also {
        bounds.onEach { b -> onBoundsChange(this@maintainBounds, b) }
          .launchIn(it)
      }
    }

    override fun onDetachedFromWindow() {
      scope?.cancel()
      scope = null
    }
  }
```

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/BodyAndModalsContainer.kt
rjrjr:
```kotlin
  // The bounds of this view in global (display) coordinates, as reported
  // by getGlobalVisibleRect.
  //
  // Made available to managed ModalScreenOverlayDialogFactory instances
  // via the ModalArea key in ViewEnvironment. When this updates their
  // updateBounds methods will fire. They should resize themselves to
  // avoid covering peers of this view.
  private val bounds = MutableStateFlow(Rect())
  private val boundsRect = Rect()

  private val boundsListener = OnGlobalLayoutListener {
    if (getGlobalVisibleRect(boundsRect) && boundsRect != bounds.value) {
      bounds.value = Rect(boundsRect)
    }
    // Should we close the dialogs if getGlobalVisibleRect returns false?
    // https://github.com/square/workflow-kotlin/issues/599
  }
```

rjrjr: Note that I punted handling a `false` return in any interesting why. We've been shipping the equivalent code for more than a year and it's not been an issue. I'd like to dig into that case separately and carefully -- see if it's actually an issue at all.

rjrjr:
Updated to honor the feedback from @steve-the-edwards, especially around eliminating `Bounds` and making the two new `ViewEnvironment.Key`s private. PTAL

samples/tictactoe/common/src/main/java/com/squareup/sample/gameworkflow/RunGameWorkflow.kt
```diff
@@ -69,90 +78,106 @@ class RealRunGameWorkflow(
...
+        // already going. TakeTurnsWorkflow.render is immediately called,
+        // and the GamePlayScreen it renders is immediately returned.
+        val takeTurnsScreen = context.renderChild(
+          takeTurnsWorkflow,
+          props = renderState.resume
+            ?.let { TakeTurnsProps.resumeGame(renderState.playerInfo, it) }
+            ?: TakeTurnsProps.newGame(renderState.playerInfo)
+        ) { stopPlaying(it) }
+
+        RunGameRendering(takeTurnsScreen)
```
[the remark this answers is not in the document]
rjrjr: Missed this because mf#*$&*#ng GitHub hid half your comments! Will do.

samples/tictactoe/common/src/main/java/com/squareup/sample/mainworkflow/TicTacToeWorkflow.kt
```diff
@@ -52,49 +56,56 @@ class TicTacToeWorkflow(
...
-        panelsMod[0] = stubAuthBackStack + panels[0]
-        childRendering.copy(beneathModals = childRendering.beneathModals.copy(modals = panelsMod))
+          BodyAndModalsScreen(gameRendering.gameScreen, allModals)
+        }
       }
     }
+
+    // Add the scrim. Dim it only if there is a panel showing.
+    val dim = bodyAndModals.modals.any { modal -> modal is PanelOverlay<*> }
+    return bodyAndModals.mapBody { body -> ScrimScreen(body, dimmed = dim) }
```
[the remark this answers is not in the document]
rjrjr: I've been working on those in the next PR, I think you're really going to like it.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/AlertOverlayDialogFactory.kt
```diff
@@ -24,43 +25,62 @@ internal object AlertOverlayDialogFactory : OverlayDialogFactory<AlertOverlay> {
...
+        for (button in Button.values()) {
+          // Force all the buttons to be created so that we can show
+          // and hide them at will during updates. They get created
+          // when show is called, so we can't mess with them before that
+          // (e.g. to fix the dismiss-on-click nonsense),
+          // and we can't create new ones after that.
+          //
+          // https://github.com/square/workflow-kotlin/issues/138
+          //
+          // Why " "? An empty string means no button.
```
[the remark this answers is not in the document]
rjrjr: We want to be able to update the alert while it's showing, including to maybe show more buttons than were there originally. The API for Android's `AlertDialog` makes you think you can do that, but it actually doesn't work. So we force `AlertDialog.Builder` to show every possible button; then we hide them all; and then we manage their visibility ourselves at update time.

We also don't want Android to tear down the dialog without our say so -- again, we might need to update the thing. But there is a dismiss call built in to click handers put in place by `AlertDialog`. So, when we're preflighting every possible button, we put garbage click handlers in place. Then we replace them with our own, again at update time, by setting each live button's click handler directly, without letting `AlertDialog` interfere.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/AndroidDialogBounds.kt
```diff
@@ -0,0 +1,69 @@
...
+}
+
+@WorkflowUiExperimentalApi
+internal fun <D : Dialog> D.maintainBounds(
+  bounds: StateFlow<Bounds>,
+  onBoundsChange: (D, Bounds) -> Unit
+) {
+  val window = requireNotNull(window) { "Dialog must be attached to a window." }
+  window.callback = object : Window.Callback by window.callback {
+    val scope = CoroutineScope(Dispatchers.Main.immediate)
```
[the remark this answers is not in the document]
rjrjr: How about a public `object UiDispatcher : ViewEnvironmentKey<Dispatcher>` that defaults to `Dispatchers.Main.immediate`?

rjrjr:
@steve-the-edwards Oops, I was too quick to ask PTAL, I missed some of your comments. Will ping again soon.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/AndroidDialogBounds.kt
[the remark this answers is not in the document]
rjrjr: Let me do that in a separate PR, so I can fix `WorkflowLayout` at the same time. #600

samples/tictactoe/common/src/main/java/com/squareup/sample/gameworkflow/RunGameWorkflow.kt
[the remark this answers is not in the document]
rjrjr: I'm not sure what you're suggesting here.

rjrjr:
> @steve-the-edwards Oops, I was too quick to ask PTAL, I missed some of your comments. Will ping again soon.

Actually, okay to review. Tracking the most interesting thing I missed separately, #600.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/AlertOverlayDialogFactory.kt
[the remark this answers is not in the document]
rjrjr:
```kotlin
    return AlertDialog.Builder(context, initialEnvironment[AlertDialogThemeResId])
      .create().apply {
        for (button in Button.values()) {
          // We want to be able to update the alert while it's showing, including to maybe
          // show more buttons than were there originally. The API for Android's `AlertDialog`
          // makes you think you can do that, but it actually doesn't work. So we force
          // `AlertDialog.Builder` to show every possible button; then we hide them all;
          // and then we manage their visibility ourselves at update time.
          //
          // We also don't want Android to tear down the dialog without our say so --
          // again, we might need to update the thing. But there is a dismiss call
          // built in to click handers put in place by `AlertDialog`. So, when we're
          // preflighting every possible button, we put garbage click handlers in place.
          // Then we replace them with our own, again at update time, by setting each live
          // button's click handler directly, without letting `AlertDialog` interfere.
          //
          // https://github.com/square/workflow-kotlin/issues/138
          //
          // Why " "? An empty string means no button.
          setButton(button.toId(), " ") { _, _ -> }
        }
      }
```

samples/tictactoe/common/src/main/java/com/squareup/sample/gameworkflow/RunGameWorkflow.kt
```diff
@@ -69,90 +78,106 @@ class RealRunGameWorkflow(
...
+            renderState.playerInfo,
+            renderState.completedGame.lastTurn
+          ),
+          alerts = listOf(
+            maybeQuitScreen(
+              message = "Do you really want to concede the game?",
+              positive = "I Quit",
+              negative = "No",
+              confirmQuit = context.eventHandler {
+                (state as? MaybeQuitting)?.let { oldState ->
```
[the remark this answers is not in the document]
rjrjr: 📝

samples/tictactoe/common/src/main/java/com/squareup/sample/gameworkflow/RunGameWorkflow.kt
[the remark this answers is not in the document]
steve-the-edwards: I guess I was suggesting to name the arg here still

```suggestion
        RunGameRendering(gameScreen = takeTurnsScreen)
```

but its not a big deal.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/AlertOverlayDialogFactory.kt
```diff
@@ -24,43 +25,69 @@ internal object AlertOverlayDialogFactory : OverlayDialogFactory<AlertOverlay> {
...
+          // makes you think you can do that, but it actually doesn't work. So we force
+          // `AlertDialog.Builder` to show every possible button; then we hide them all;
+          // and then we manage their visibility ourselves at update time.
+          //
+          // We also don't want Android to tear down the dialog without our say so --
+          // again, we might need to update the thing. But there is a dismiss call
+          // built in to click handers put in place by `AlertDialog`. So, when we're
+          // preflighting every possible button, we put garbage click handlers in place.
+          // Then we replace them with our own, again at update time, by setting each live
+          // button's click handler directly, without letting `AlertDialog` interfere.
```
steve-the-edwards, line 41: Ahh - now I understand. This is very helpful.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/AndroidDialogBounds.kt
```diff
@@ -0,0 +1,78 @@
...
+        bounds.onEach { b -> onBoundsChange(this@maintainBounds, b) }
+          .launchIn(it)
+      }
+    }
+
+    override fun onDetachedFromWindow() {
+      scope?.cancel()
+      scope = null
+    }
+  }
```
steve-the-edwards, line 74: 👍🏻 .

Other than the injected immediate main dispatcher - but I see thats covered by #600 .

If we want debugging help we could add a name into the context for the scope.

```suggestion
  window.callback = object : Window.Callback by window.callback {
    var scope: CoroutineScope? = null

    override fun onAttachedToWindow() {
      val id = window.decorView.id
      scope = CoroutineScope(Dispatchers.Main.immediate + CoroutineName(id)).also {
        bounds.onEach { b -> onBoundsChange(this@maintainBounds, b) }
          .launchIn(it)
      }
    }

    override fun onDetachedFromWindow() {
      scope?.cancel()
      scope = null
    }
  }
```

steve-the-edwards (approved):
Awesome!

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/AndroidDialogBounds.kt
rjrjr: s/classname/toString()/, 10x more useful. I was shying away from that since it can be big, and that burned us in `workflowAction {}`. But I think that's okay for infrequent events like creating a dialog.

rjrjr: Meh, I'm over engineering this.

Gonna merge as is and add a note to #600 to consider useful scope name at the same time.

Issue #516, opened 2021-08-23
Terminal sample is busted


rjrjr:
Launch it, type any character, crash:

```
Exception in thread "main" java.lang.StringIndexOutOfBoundsException: begin 0, end 13, length 12
        at java.base/java.lang.String.checkBoundsBeginEnd(String.java:3319)
        at java.base/java.lang.String.substring(String.java:1874)
        at com.squareup.sample.hellotodo.EditTextWorkflowKt.insertCharAt(EditTextWorkflow.kt:100)
        at com.squareup.sample.hellotodo.EditTextWorkflowKt.access$insertCharAt(EditTextWorkflow.kt:1)
        at com.squareup.sample.hellotodo.EditTextWorkflow$onKeystroke$1.invoke(EditTextWorkflow.kt:72)
        at com.squareup.sample.hellotodo.EditTextWorkflow$onKeystroke$1.invoke(EditTextWorkflow.kt:68)
```

Arrow keys and return work okay, including return checking / unchecking items in the list.

If you arrow back a bit and type a few chars, they aren't echoed. Instead they act like the right arrow until you reach the end of the line and then you crash as above.

PR #605, opened 2021-12-10
Fixes index out of bounds crash in hello-todo-terminal sample

JsFlo:
Fixes index out of bounds crash when attempting to add characters or moving the cursor to the ends of the string being edited.

* uses `setOutput` to cause the parent workflow to update the string when adding characters
* coerces cursor position within the 0..text.length

Fixes issue: https://github.com/square/workflow-kotlin/issues/516


rjrjr:
Thank you!

rjrjr (approved):

steve-the-edwards (approved):
Thanks!

PR #607, opened 2021-12-10
End CHANGELOG.md

rjrjr:
Because the [releases page](https://github.com/square/workflow-kotlin/releases) is so much nicer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 Change Log
 ==========
 
+> This file is no longer updated, please visit [Releases](https://github.com/square/workflow-kotlin/releases) instead.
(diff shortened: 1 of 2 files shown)


dhavalshreyas:
Do we need `git tag v0.1.0` still?

dhavalshreyas (approved):

steve-the-edwards (approved):

rjrjr:
> Do we need `git tag v0.1.0` still?

Yeah, because we don't like to push the tag until push to Sonatype / Maven Central has actually worked, and thats when we know we have the right SHA to tag. Too easy to tag the wrong thing if we use the web page I think.

PR #614, opened 2021-12-18
Add separate methods for verifying state and output to RenderTestResult

japplin:
Not sure if verifyActionResult should be deprecated, deleted, or replaced with a version that doesn't use the WorkflowOutput. Maybe the correct thing is some combination of the three. I went with just deprecating for a first try since it's the least ambiguos, ie. one team member may prefer using the two seperate methods as they allow using `it` it could be written as single line lambdas, another member may prefer using the single assertion for both output and state, by only providing one option it removes this avenue for interpretation.


workflow-testing/src/main/java/com/squareup/workflow1/testing/RenderTestResult.kt
```diff
@@ -38,10 +38,59 @@ public interface RenderTestResult<PropsT, StateT, OutputT, RenderingT> {
...
   ): RenderTestResult<PropsT, StateT, OutputT, RenderingT>
 
+  /**
+   * Asserts that the render pass handled either a workflow/worker output or a rendering event,
+   * "executes" the action with the state passed to [testRender], then invokes [block] with the
+   * resulting state.
+   *
+   * If the workflow didn't process any actions, `newState` will be the initial state.
+   *
+   * Note that by using this method, you're also testing the implementation of your action. This can
```
japplin, line 57: I think the documentation for these 3 methods could be simplified, I took the approach of copy pasting the verifyActionResult docs to each of them since I wasn't sure which part of those docs were really important to maintain.

steve-the-edwards:
> Not sure if verifyActionResult should be deprecated, deleted, or replaced with a version that doesn't use the WorkflowOutput. Maybe the correct thing is some combination of the three. I went with just deprecating for a first try since it's the least ambiguos, ie. one team member may prefer using the two seperate methods as they allow using `it` it could be written as single line lambdas, another member may prefer using the single assertion for both output and state, by only providing one option it removes this avenue for interpretation.

I could see having the other convenience test functions for simpler lambdas as you say but I wouldn't want to deprecate the original `verifyActionResult` as each action cascade can produce an `Output` and advance the `State` so I want to keep that association strong.

I need to think more about this as well - what if having the convenience functions you thought you could use them both and then the action would be applied multiple times? For that reason alone I'm not sure we would want to include them as that would be confusing.

steve-the-edwards:
> > Not sure if verifyActionResult should be deprecated, deleted, or replaced with a version that doesn't use the WorkflowOutput. Maybe the correct thing is some combination of the three. I went with just deprecating for a first try since it's the least ambiguos, ie. one team member may prefer using the two seperate methods as they allow using `it` it could be written as single line lambdas, another member may prefer using the single assertion for both output and state, by only providing one option it removes this avenue for interpretation.
> 
> I could see having the other convenience test functions for simpler lambdas as you say but I wouldn't want to deprecate the original `verifyActionResult` as each action cascade can produce an `Output` and advance the `State` so I want to keep that association strong.
> 
> I need to think more about this as well - what if having the convenience functions you thought you could use them both and then the action would be applied multiple times? For that reason alone I'm not sure we would want to include them as that would be confusing.

... Thinking more, you would get a `RenderTestResult` so you wouldn't be able to chain them. So just have to decide that we are ok without encouraging that you verify the Output and State in the same test 🤔

japplin:
@steve-the-edwards Great questions it's possible to chain these methods together and they don't affect each other, I pushed 2 more tests verifying that behavior. I also realize that I left out some of the initial context around this change that happened in slack. The primary motivation for this change was to allow assertions on the output specifically instead of on the WorkflowOutput wrapper, an additional benefit provided by this solution is that the api becomes more obvious since it's using terms that are used elsewhere in workflows: state and output. Whereas the current result naming isn't used anywhere else in the api so it might not be obvious at first that method is used to verify the state or output.

internal square slack link: https://square.slack.com/archives/CE470RL5T/p1639676043437600

workflow-testing/src/main/java/com/squareup/workflow1/testing/RenderTestResult.kt
```diff
@@ -38,10 +38,59 @@ public interface RenderTestResult<PropsT, StateT, OutputT, RenderingT> {
...
+   * instead and write separate unit tests for your action implementations.
+   */
+  public fun verifyActionOutput(
+    block: (newState: OutputT) -> Unit
+  ): RenderTestResult<PropsT, StateT, OutputT, RenderingT>
+
+  /**
+   * Asserts that the render pass handled either a workflow/worker output or a rendering event,
+   * "executes" the action with the state passed to [testRender], then verifies that no output
+   * occurred.
```
steve-the-edwards, line 84:
```suggestion
   * Asserts that the render pass handled a rendering event,
   * "executes" the action with the state passed to [testRender], then verifies that no output
   * occurred.
```

workflow-testing/src/main/java/com/squareup/workflow1/testing/RenderTestResult.kt
```diff
@@ -38,10 +38,59 @@ public interface RenderTestResult<PropsT, StateT, OutputT, RenderingT> {
...
+   * instead and write separate unit tests for your action implementations.
+   */
+  public fun verifyActionState(
+    block: (newState: StateT) -> Unit
+  ): RenderTestResult<PropsT, StateT, OutputT, RenderingT>
+
+  /**
+   * Asserts that the render pass handled either a workflow/worker output or a rendering event,
+   * "executes" the action with the state passed to [testRender], then invokes [block] with the
+   * resulting output.
```
steve-the-edwards, line 68:
```suggestion
  /**
   * Asserts that the render pass handled either a workflow/worker output, 
   * "executes" the action with the state passed to [testRender], then invokes [block] with the
   * resulting output.
```

workflow-testing/src/main/java/com/squareup/workflow1/testing/RenderTestResult.kt
```diff
@@ -38,10 +38,59 @@ public interface RenderTestResult<PropsT, StateT, OutputT, RenderingT> {
    * be useful if your actions are anonymous. If they are a sealed class or enum, use [verifyAction]
    * instead and write separate unit tests for your action implementations.
    */
+  @Deprecated(
```
steve-the-edwards, line 41: Not sure we need to deprecate this but maybe update the kdoc to say

"... invokes [block] with the resulting state and output. Note that the output is wrapped in [WorkflowOutput] to differentiate from the case of no Output emitted (`output == null`) or `null` emitted (`WorkflowOutput(null)`)."

workflow-testing/src/main/java/com/squareup/workflow1/testing/RenderTestResult.kt
```diff
@@ -38,10 +38,59 @@ public interface RenderTestResult<PropsT, StateT, OutputT, RenderingT> {
...
+  public fun verifyActionState(
+    block: (newState: StateT) -> Unit
+  ): RenderTestResult<PropsT, StateT, OutputT, RenderingT>
+
+  /**
+   * Asserts that the render pass handled either a workflow/worker output or a rendering event,
+   * "executes" the action with the state passed to [testRender], then invokes [block] with the
+   * resulting output.
+   *
+   * If the workflow didn't process any actions, or actions process didn't emit and outputs then
```
steve-the-edwards, line 70:
```suggestion
   * If the workflow didn't process any actions, or actions processed didn't emit any outputs then
```

workflow-testing/src/test/java/com/squareup/workflow1/testing/RealRenderTesterTest.kt
```diff
@@ -1025,6 +1026,100 @@ internal class RealRenderTesterTest {
     }
   }
 
+  @Test fun `verifyActionOutput throws when no output`() {
+    val workflow = Workflow.stateless<Unit, Nothing, Sink<TestAction>> {
+      actionSink.contraMap { it }
+    }
+    val testResult = workflow.testRender(Unit)
+      .render()
+    assertFails {
```
steve-the-edwards, line 1035: lets tighten this up to `assertFailsWith(IllegalStateException("Expected an output but no output was found."))`

workflow-testing/src/test/java/com/squareup/workflow1/testing/RealRenderTesterTest.kt
```diff
@@ -1025,6 +1026,100 @@ internal class RealRenderTesterTest {
...
+  }
+
+  @Test fun `verifyNoActionOutput throws given an output`() {
+    val workflow = Workflow.stateless<Unit, Int, Sink<TestAction>> {
+      actionSink.send(action { setOutput(1) })
+      actionSink.contraMap { it }
+    }
+    val testRender = workflow.testRender(Unit)
+      .render()
+    assertFails {
```
steve-the-edwards, line 1057: Same thing here to tighten up the test

steve-the-edwards (approved):
A few changes but otherwise you have convinced me :)

workflow-testing/src/test/java/com/squareup/workflow1/testing/RealRenderTesterTest.kt
japplin: I even found a small bug updating this test, it was logging the output wrapper when a value was present instead of the output value directly.

workflow-testing/src/test/java/com/squareup/workflow1/testing/RealRenderTesterTest.kt
japplin: I updated this one to an AssertionError instead of IllegalStateException since all there were other test method which were throwing AssertionErrors.

japplin:
@steve-the-edwards I updated this pr, would be able to approve running the github checks for this pr? I'm seeing:
Screen Shot 2021-12-21 at 1 45 45 PM

steve-the-edwards:
> @steve-the-edwards I updated this pr, would you be able to approve running the github checks for this pr? I'm seeing: 

Thanks ya, did so. The ktlint failures will need to be fixed. The CI does sometimes timeout - it is run on both API 21 and API 24 and as you can see, one passed the other didn't. It is flakey though so check it out but it may just be a flake.

PR #617, opened 2022-01-05
Upgrade the compile SDK to 31.

vRallev:
There's no reason no to, and it's actually required by newer Compose versions.


rjrjr (approved):
Thanks!

PR #621, opened 2022-01-06
wip: Replaces View.showRendering et al with a wrapper interface

### LINKED ISSUE #413: DecorativeViewFactory should support wrapping Context

### LINKED ISSUE #606: Consider creating `interface WorkflowView<RenderingT: Screen>`
We _could_ create a facade interface that just wrapped a plain old View. I don't think it's worth the effort. Feature developers basically never bump into this stuff, thanks to `WorkflowViewStub` and such. That will be even more true when the UI overhaul is done, I think.

But just to scratch the nerdsnipe itch, here's a sketch of what that could be.

```kotlin
fun <RenderingT> ViewRegistry.buildView(...): WorkflowView<RenderingT>

interface WorkflowView<RenderingT> {
  fun asView(): View
  
  fun start()

  val rendering: RenderingT
  val environment: ViewEnvironment

  fun update(
    rendering: RenderingT,
    environment: ViewEnvironment
  )
  // ...
}

fun <RenderingT> View.asWorkflowView(): WorkflowView<T>? {
  // Looks for the tag and confirms the type is right? 
}
```

Still thinking this would be a distraction, but I'm wavering. Dammit Steve, are you proud that you've done this to me?

_Originally posted by @rjrjr in https://github.com/square/workflow-kotlin/pull/602#discussion_r766956989_

rjrjr:
Closes #606, fixes #413


workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/DecorativeScreenViewFactory.kt
```diff
@@ -5,118 +5,6 @@ import android.view.View
...
- * @param viewStarter An optional wrapper for the function invoked when [View.start]
- * is called, allowing for last second initialization of a newly built [View].
- * See [ViewStarter] for details.
- *
- * @param doShowRendering called to apply the [ViewShowRendering] function for
- * [InnerT], allowing pre- and post-processing. Default implementation simply
- * uses [map] to extract the [InnerT] instance from [OuterT] and makes the function call.
- */
 @WorkflowUiExperimentalApi
 public class DecorativeScreenViewFactory<OuterT : Screen, InnerT : Screen>(
```
rjrjr, line 9: Expecting to delete this.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/DecorativeViewFactory.kt
```diff
@@ -5,8 +5,111 @@ import android.view.View
 import android.view.ViewGroup
 import kotlin.reflect.KClass
 
+/**
```
rjrjr, line 8: This is still deprecated, but its seems obnoxious to delete its copious kdoc when there is no direct analog in the new world.

rjrjr:
Starting to feel like a dead end, [deets](https://github.com/square/workflow-kotlin/issues/606#issuecomment-1009255530).

PR #629, opened 2022-01-12
Undoes port of AlertContainerScreen to Screen

rjrjr:
It is entirely legacy, with no direct counterpart in the new world. Giving it a foot in both worlds just makes it harder to adopt the workflow update.


steve-the-edwards (approved):

rjrjr:
To verify during Gradle-induced JCenter outage, running `./gradlew clean build connectedCheck` locally

PR #630, opened 2022-01-12
Simpler, safer ViewEnvironment and ViewRegistry management.

### LINKED ISSUE #395: Merging ViewRegistries inside ViewEnvironments is a nightmare.
There are a couple of issues:

 * When you `plus()` two `ViewEnvironments`, entries are completely replaced, there are no merge hooks
 * When you `plus()` two `ViewRegistries`, if they have any redundant entries an exception is thrown

For the former, perhaps we can build optional merge hooks into `ViewEnvironmentKey`?

For the latter, I'm hard pressed to think of any bugs we have caught by throwing, but I can come up of lots of examples where we had to struggle mightily to work around it. How about if we document that entries in the registry on the right in a `ViewRegistry.plus()` call will clobber those on the left? We could offer a strict variant, or even default to it, but I can't really imagine when I'd use it.

rjrjr:
`ViewEnvironment(Map<ViewEnvironmentKey<*>, Any>)` makes it easy to build a
`ViewEnvironment` whose with values that don't match the type of the
`ViewEnvironmentKey`s. We deprecate it with an eye toward making it
non-public down the road. In its place we introduce `ViewEnvironment.EMPTY`,
and guide people to use it and the type safe `plus` operators.

We introduce a pattern of providing more `plus` operators with new environment
value types.

For `ViewRegistry`, in addition to the `plus` operator we provide a set of
`merge` methods, which finish the long overdue job of making it simpler to
mess with `ViewEnvironment` without stomping on `ViewRegistry` entries, and
to override them. `EnvironmentScreen.withRegistry` and `.withEnvironment` use
these.

Fixes #395.

### DIFF
--- a/samples/containers/android/src/main/java/com/squareup/sample/container/overviewdetail/OverviewDetailConfig.kt
+++ b/samples/containers/android/src/main/java/com/squareup/sample/container/overviewdetail/OverviewDetailConfig.kt
@@ -1,7 +1,8 @@
 package com.squareup.sample.container.overviewdetail
 
-import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+import com.squareup.workflow1.ui.ViewEnvironment
 import com.squareup.workflow1.ui.ViewEnvironmentKey
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
 
 /**
  * [com.squareup.workflow1.ui.ViewEnvironment] value that informs views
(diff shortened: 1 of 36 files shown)


samples/containers/android/src/main/java/com/squareup/sample/container/overviewdetail/OverviewDetailConfig.kt
```diff
@@ -34,3 +35,7 @@ enum class OverviewDetailConfig {
     override val default = None
   }
 }
+
+@WorkflowUiExperimentalApi
+operator fun ViewEnvironment.plus(config: OverviewDetailConfig): ViewEnvironment =
```
rjrjr, line 40: Introduced all these and then forgot to use them.

rjrjr:
Yikes! Merged this w/o review and w/o testing, too many tabs open at once. Reverting.

PR #631, opened 2022-01-12
Simpler, safer ViewEnvironment and ViewRegistry management.

### LINKED ISSUE #395: Merging ViewRegistries inside ViewEnvironments is a nightmare.

rjrjr:
`ViewEnvironment(Map<ViewEnvironmentKey<*>, Any>)` makes it easy to build a
`ViewEnvironment` whose with values that don't match the type of the
`ViewEnvironmentKey`s. We deprecate it with an eye toward making it
non-public down the road. In its place we introduce `ViewEnvironment.EMPTY`,
and guide people to use it and the type safe `plus` operators.

We introduce a pattern of providing more `plus` operators with new environment
value types.

For `ViewRegistry`, in addition to the `plus` operator we provide a set of
`merge` methods, which finish the long overdue job of making it simpler to
mess with `ViewEnvironment` without stomping on `ViewRegistry` entries, and
to override them. `EnvironmentScreen.withRegistry` and `.withEnvironment` use
these.

Fixes #395. Replaces #630, reverted due to accidental merge.

### DIFF
--- a/samples/compose-samples/src/main/java/com/squareup/sample/compose/textinput/App.kt
+++ b/samples/compose-samples/src/main/java/com/squareup/sample/compose/textinput/App.kt
@@ -1,4 +1,5 @@
 @file:OptIn(WorkflowUiExperimentalApi::class)
+@file:Suppress("FunctionName")
 
 package com.squareup.sample.compose.textinput
 
@@ -11,10 +12,9 @@ import com.squareup.workflow1.ui.ViewRegistry
 import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
 import com.squareup.workflow1.ui.compose.WorkflowRendering
 import com.squareup.workflow1.ui.compose.renderAsState
+import com.squareup.workflow1.ui.plus
 
-private val viewRegistry = ViewRegistry(TextInputViewFactory)
-
-private val viewEnvironment = ViewEnvironment(mapOf(ViewRegistry to viewRegistry))
+private val viewEnvironment = ViewEnvironment.EMPTY + ViewRegistry(TextInputViewFactory)
 
 @Composable fun TextInputApp() {
   MaterialTheme {
--- a/workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/ViewRegistry.kt
+++ b/workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/ViewRegistry.kt
@@ -76,9 +76,7 @@ public interface ViewRegistry {
   public val keys: Set<KClass<*>>
 
   /**
-   * This method is not for general use, use [WorkflowViewStub] instead.
-   *
-   * Returns the [ViewFactory] that was registered for the given [renderingType], or null
+   * Returns the [Entry] that was registered for the given [renderingType], or null
    * if none was found.
    */
   public fun <RenderingT : Any> getEntryFor(
@@ -89,6 +87,10 @@ public interface ViewRegistry {
     override val default: ViewRegistry get() = ViewRegistry()
   }
 }
+
+@WorkflowUiExperimentalApi public inline operator fun <reified RenderingT : Any> ViewRegistry.get(
+  renderingType: KClass<out RenderingT>
+): Entry<RenderingT>? = getEntryFor(renderingType)
 
 @WorkflowUiExperimentalApi
 public fun ViewRegistry(vararg bindings: Entry<*>): ViewRegistry =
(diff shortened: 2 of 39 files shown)


steve-the-edwards (commented):
Have not finished reviewing because paused thinking about why we can't have the + operator come in via an interface. Just fiddling with the variance to see if I can get it to work a la 

```

interface ViewEnvironmental

operator fun <T: ViewEnvironmental> ViewEnvironment.plus(value: T): ViewEnvironment =
  this + ViewEnvironment(mapOf(value::class to value))

```

which currently doesn't compile due to the variance of `value::class`

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/ViewRegistry.kt
```diff
@@ -102,10 +104,67 @@ public fun ViewRegistry(vararg bindings: Entry<*>): ViewRegistry =
...
+
+/**
+ * Merges the [ViewRegistry] of the receiver with [registry]. If there are conflicting entries,
+ * those in [registry] are preferred.
+ */
+@WorkflowUiExperimentalApi
+public infix fun ViewEnvironment.merge(registry: ViewRegistry): ViewEnvironment {
+  val oldReg = this[ViewRegistry]
+
+  val union = (oldReg.keys + registry.keys).asSequence()
```
rjrjr, line 149: Oops, this is a dup of the merge() method above. Should call that from here.

samples/compose-samples/src/main/java/com/squareup/sample/compose/textinput/App.kt
```diff
@@ -1,4 +1,5 @@
 @file:OptIn(WorkflowUiExperimentalApi::class)
+@file:Suppress("FunctionName")
```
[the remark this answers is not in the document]
rjrjr: It's complaining about the capitalized function name. We don't _have_ to fail it, the build was green. I just hate IDE warnings.

rjrjr:
> Have not finished reviewing because paused thinking about why we can't have the + operator come in via an interface. Just fiddling with the variance to see if I can get it to work a la
> 
> ```
> 
> interface ViewEnvironmental
> 
> operator fun <T: ViewEnvironmental> ViewEnvironment.plus(value: T): ViewEnvironment =
>   this + ViewEnvironment(mapOf(value::class to value))
> ```
> 
> which currently doesn't compile due to the variance of `value::class`

It's a fun problem space, isn't it? Really seems like `ViewEnvironment` could be spun out into its own microlibrary -- `com.block.polymap`.

I keep resisting the urge to create a general purpose mechanism for things to declare themselves mergable, b/c I strongly suspect that `ViewRegistry` will be the only use case forever.

steve-the-edwards:
> Have not finished reviewing because paused thinking about why we can't have the + operator come in via an interface. ...

Blarg we'd be sacrificing one convenience/legibility helper for another. I think i'm rocking the boat too much here, but the idea would be to take your pattern of making the companion object the `ViewEnvironmentKey` (nice because it has only one initialization at class init time) and instead move it to an interface where we could then enforce the operations etc.

I think i'm trying to get back to the `ViewWrapper` that I (unhelpfully) nerd-sniped you into with the failed #606 .

Ok will stop trying to make this harder than it is 😊

rjrjr:
Latest force-push makes the tests green.

`TicTacToeEspressoTest#showRenderingTagStaysFresh` was failing because it inadvertently exposed a wateful allocations -- when we started calling the new `merge` functions on every render pass, we were creating new instances of `CompositeViewRegistry` for no reason. I've added / restored some checks to return the existing `ViewEnvironment` or `ViewRegistry` when merging with an empty one.

I'll add some explicit coverage for that, seems worth being formal about it.

rjrjr:
@steve-the-edwards PTAL, I've added a commit that adds the missing test coverage for that optimization, and adds a few more such `if (this.isEmpty()) return that` cases.

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/ViewRegistry.kt
```diff
@@ -102,10 +104,70 @@ public fun ViewRegistry(vararg bindings: Entry<*>): ViewRegistry =
...
+}
+
+/**
+ * Merges the [ViewRegistry] of the receiver with [registry]. If there are conflicting entries,
+ * those in [registry] are preferred.
+ */
 @WorkflowUiExperimentalApi
-public operator fun ViewRegistry.plus(binding: Entry<*>): ViewRegistry =
-  this + ViewRegistry(binding)
+public infix fun ViewEnvironment.merge(registry: ViewRegistry): ViewEnvironment {
```
steve-the-edwards, line 153: Beware I think once we start having mixed typed operators things can sometimes get confusing - this is the case for `CoroutineContext` which maps scope, dispatcher, name etc. in a very similar way to this. 90% of the time things just 'work' and work in a quick and satisfying way, but sometimes when reasoning about it it can get confusing how things overlap.

steve-the-edwards (approved):

PR #644, opened 2022-01-25
Check if runtime CoroutineScope cancelled after nextOutput()

### LINKED ISSUE #642: nextRendering() is called in runtime if Workflow's scope is cancelled.
Right now in `renderWorkflowIn` we check if the `CoroutineScope` `isActive` at the top of the loop:

```
  scope.launch(start = ATOMIC) {
    while (isActive) {
      // It might look weird to start by consuming the output before getting the rendering below,
      // but remember the first render pass already occurred above, before this coroutine was even
      // launched.
      val output = runner.nextOutput()

      renderingsAndSnapshots.value = runner.nextRendering()
      output?.let { onOutput(it.value) }
    }
  }
```

But since `nextOutput()` is suspending we could resume and have had an `action` `cancel()` the `CoroutineScope` of the root workflow. We should pro-actively check this before calling `nextRendering()`.

Something like:

```
  scope.launch(start = ATOMIC) {
    while (isActive) {
      // It might look weird to start by consuming the output before getting the rendering below,
      // but remember the first render pass already occurred above, before this coroutine was even
      // launched.
      val output = runner.nextOutput()

      if (!isActive) return@launch
      renderingsAndSnapshots.value = runner.nextRendering()
      output?.let { onOutput(it.value) }
    }
  }
```

steve-the-edwards:
Fixes #642

### DIFF
--- a/workflow-runtime/src/main/java/com/squareup/workflow1/RenderWorkflow.kt
+++ b/workflow-runtime/src/main/java/com/squareup/workflow1/RenderWorkflow.kt
@@ -138,6 +138,10 @@ public fun <PropsT, OutputT, RenderingT> renderWorkflowIn(
       // launched.
       val output = runner.nextOutput()
 
+      // After resuming from runner.nextOutput() our coroutine could now be cancelled, check so we
+      // don't waste doing an un-needed rendering.
+      if (!isActive) return@launch
+
       // After receiving an output, the next render pass must be done before emitting that output,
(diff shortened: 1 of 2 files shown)


steve-the-edwards:
Passing at least one API version of CI instrumentation tests (the other with the flaky SSL timeout?).

Also fully passed register CI test suite when included via snapshot.

Ready to merge IMHO. Would love a blessing from @zach-klippenstein

PR #645, opened 2022-01-25
Launch renderWorkflowIn CoroutineScope with Default; Fixes #643

### LINKED ISSUE #643: Remove `(start = ATOMIC)` from `renderWorkflowIn` `launch`
We do the first rendering synchronously and have no finally or also that we need to account for happening as we used to. This should not be needed anymore and we can use the `DEFAULT` start scheduling.

steve-the-edwards:
Closes #643


RBusarow (approved):

rjrjr (approved):

steve-the-edwards:
Passing at least one API version of CI instrumentation tests (the other with the flaky SSL timeout?).

Also fully passed register CI test suite when included via snapshot.

Ready to merge IMHO. Would love a blessing from @zach-klippenstein

PR #646, opened 2022-01-26
fix TextInputTest flakes

RBusarow:
The previous method of `expectValue(EditableText, AnnotatedString(text))` intermittently fails when I run it locally against an IRL Pixel 3a XL.

The node in the assertion is already being matched with `hasSetTextAction()`, so that's already established.  It should be fine to just assert against the text.

I'm tempted to say that the flakes are device-dependant, with different phones creating slight variations on `AnnotatedString`, but my Pixel isn't exactly changing in between test runs.  What we need is an expert in both Compose and Text.


vgonda (approved):
From a Compose standpoint, this makes sense to me.

I'm not a Text expert, so maybe also look for a review who knows more about that.

steve-the-edwards (approved):
LGTM but don't have the Text/Compose expertise to confirm your thoughts.

rjrjr (approved):
¯\_(ツ)_/¯

PR #652, opened 2022-01-28
Works around https://github.com/square/leakcanary/issues/2297

rjrjr:


steve-the-edwards (approved):

rjrjr:
Updates fix lint and api/Check and such.

PR #654, opened 2022-01-31
Merges main into ray/ui-update

rjrjr:
Keeping `ray/ui-update` caught up with `main`. I don't usually ask for reviews on these, but lately `main` has been more of a moving target than usual. @steve-the-edwards and @RBusarow, could you sanity check that the recent work you're familiar with looks like it's properly preserved here? I'm doing the same.

Strongly recommend hiding whitespace changes, there are tons of files with nothing but.


workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLayout.kt
```diff
@@ -47,25 +51,89 @@ public class WorkflowLayout(
...
+
+  @Deprecated(
+    "Use show",
+    ReplaceWith(
+      "show(asScreen(newRendering).withEnvironment(environment))",
+      "com.squareup.workflow1.ui.container.withEnvironment",
+      "com.squareup.workflow1.ui.asScreen"
+    )
+  )
+  public fun update(
```
rjrjr, line 96: Maybe should rename this to `show` on `main` before the next release, so that there will be one less bit of swirl when the overhaul lands.

README.md
```diff
@@ -64,7 +64,7 @@ your `repositories` block, and then add dependencies on the following artifacts:
     <td>Your Android app uses modals (popups).</td>
   </tr>
   <tr>
-    <td nowrap><code>com.squareup.workflow1:workflow-ui-backstack-android:x.y.z</code></td>
+    <td nowrap><code>com.squareup.workflow1:workflow-ui-container-android:x.y.z</code></td>
```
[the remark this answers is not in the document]
rjrjr: Actually, I just deleted the two dying modules.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLayout.kt
rjrjr: Too late, I forgot that we've already released it that way.

vgonda (approved):
I have no complaints on the Compose related diffs

steve-the-edwards (approved):
LGTM!

PR #648, opened 2022-01-27
migrate dependency definitions to version catalogs

RBusarow:
fixes #633

There were a few version changes/consolidations:

|  module                                               | old version   | new version  | version name         |
| :---------------------------------------------------: | :-----------: | :----------: | :------------------: |
| `androidx.activity:activity-compose`                  | `1.3.1`       | `1.3.0`      | `androidx-activity`  |
| `androidx.compose.ui:ui-test-junit4`                  | `1.0.1`       | `1.1.0-rc01` | `androidx-compose`   |
| `androidx.databinding:viewbinding`                    | `4.2.1`       | `7.0.0`      | `androidTools`                  |
| `androidx.lifecycle:lifecycle-viewmodel-savedstate`   | `1.1.0`       | `2.4.0`      | `androidx-lifecycle` |
| `androidx.test:core`                                  | `1.3.0`       | `1.4.0`      | `androidx-test`      |

PR #656, opened 2022-02-01
ModalContainer provides SavedStateRegistries to each of its layers.

### LINKED ISSUE #469: Consider making ModalContainer wrap all its renderings with Named to differentiate layers

### LINKED ISSUE #470: Support nesting ModalContainers under BackStackContainers.
This currently doesn't work because `ModalContainer` doesn't set the `ViewTreeSavedStateRegistryOwner` on its modals, so if views in those modals ask for one, they'll get the wrong one or nothing. However, I think it would already be so weird to put a modal container inside a backstack that it's not worth putting the effort into wiring it up until we have a real use case.

Note that `WorkflowViewStub` has to solve the same problem, and it does so.

rjrjr:
Without this `BackStackContainer` is the only thing that can host `ComposeView` in `Dialog` windows.

This fixes #469, and also fixes #470.

Co-authored-by: Ray Ryan <ray@squareup.com>


rjrjr:
This sometimes fails when opening a modal (on tablet, at least). e.g., run the Tic Tac Toe sample, complete a game, and click Exit. I suspect it's masked in cases where the modal is shown immediately at launch.

```
E/TicTacToeComponent: java.lang.IllegalStateException: You can consumeRestoredStateForKey only after super.onCreate of corresponding component
        at androidx.savedstate.SavedStateRegistry.consumeRestoredStateForKey(SavedStateRegistry.java:77)
        at com.squareup.workflow1.ui.androidx.StateRegistryAggregator$lifecycleObserver$1.onStateChanged(StateRegistryAggregator.kt:89)
        at androidx.lifecycle.LifecycleRegistry.addObserver(LifecycleRegistry.java:196)
        at com.squareup.workflow1.ui.androidx.StateRegistryAggregator.attachToParentRegistry(StateRegistryAggregator.kt:145)
        at com.squareup.workflow1.ui.backstack.ViewStateCache.attachToParentRegistry(ViewStateCache.kt:149)
        at com.squareup.workflow1.ui.backstack.BackStackContainer.onAttachedToWindow(BackStackContainer.kt:183)
        at android.view.View.dispatchAttachedToWindow(View.java:17381)
        at android.view.ViewGroup.dispatchAttachedToWindow(ViewGroup.java:3319)
        at android.view.ViewGroup.dispatchAttachedToWindow(ViewGroup.java:3326)
        at android.view.ViewGroup.dispatchAttachedToWindow(ViewGroup.java:3326)
        at android.view.ViewGroup.dispatchAttachedToWindow(ViewGroup.java:3326)
        at android.view.ViewRootImpl.performTraversals(ViewRootImpl.java:1658)
        at android.view.ViewRootImpl.doTraversal(ViewRootImpl.java:1386)
        at android.view.ViewRootImpl$TraversalRunnable.run(ViewRootImpl.java:6733)
        at android.view.Choreographer$CallbackRecord.run(Choreographer.java:911)
        at android.view.Choreographer.doCallbacks(Choreographer.java:723)
        at android.view.Choreographer.doFrame(Choreographer.java:658)
        at android.view.Choreographer$FrameDisplayEventReceiver.run(Choreographer.java:897)
        at android.os.Handler.handleCallback(Handler.java:789)
        at android.os.Handler.dispatchMessage(Handler.java:98)
        at android.os.Looper.loop(Looper.java:164)
        at android.app.ActivityThread.main(ActivityThread.java:6541)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:240)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:767)
```

samples/containers/android/src/main/java/com/squareup/sample/container/overviewdetail/OverviewDetailContainer.kt
```diff
@@ -53,11 +53,8 @@ class OverviewDetailContainer(view: View) : LayoutRunner<OverviewDetailScreen> {
     if (rendering.detailRendering == null && rendering.selectDefault != null) {
       rendering.selectDefault!!.invoke()
     } else {
-      // Since we have two sibling backstacks, we need to give them each different
-      // SavedStateRegistry key prefixes.
-      val overviewViewEnvironment = viewEnvironment
-        .withBackStackStateKeyPrefix(OverviewBackStackKey) + (OverviewDetailConfig to Overview)
-      overviewStub!!.update(rendering.overviewRendering, overviewViewEnvironment)
+      val overviewViewEnvironment = viewEnvironment + (OverviewDetailConfig to Overview)
+      overviewStub!!.update(Named(rendering.overviewRendering, "overview"), overviewViewEnvironment)
```
rjrjr, line 57: Explain this.

rjrjr:
This is green (enough), but I'm still proving it out against our massive app. Hoping to open it up for review tomorrow.

workflow-ui/container-android/src/main/java/com/squareup/workflow1/ui/modal/ModalContainer.kt
```diff
@@ -48,24 +52,46 @@ public abstract class ModalContainer<ModalRenderingT : Any> @JvmOverloads constr
...
+    onWillSave = { aggregator ->
+      dialogs.forEach { ref ->
+        ref.stateRegistryOwner.let {
+          aggregator.saveRegistryController(it.key, it.controller)
+        }
+      }
+    },
+    onRestored = { aggregator ->
+      dialogs.forEach { ref ->
+        // We're only allowed to restore from an INITIALIZED state, but this callback can also be
```
rjrjr, line 69: We have copy/pasted this fix in both places we use the aggregator. Can we build it in? I think we'd have to change it to depend on `KeyedStateRegistryOwner` directly. Maybe that's fine? Or vice versa, make the aggregator an implementation detail of `KSRO`.

workflow-ui/container-android/src/main/java/com/squareup/workflow1/ui/modal/ModalContainer.kt
```diff
@@ -75,6 +101,12 @@ public abstract class ModalContainer<ModalRenderingT : Any> @JvmOverloads constr
               dialogView,
               findParentLifecycle = { parentLifecycleOwner?.lifecycle }
             )
+            // Ensure that each dialog has its own SavedStateRegistry so views in each dialog layer
+            // don't clash with other layers.
+            ref.stateRegistryOwner = KeyedStateRegistryOwner.installAsSavedStateRegistryOwnerOn(
+              view = dialogView,
+              key = Compatible.keyFor(newScreen, i.toString())
```
rjrjr, line 108: Can this be built into `createStateRegistryKeyForContainer`? Add an optional `index: Int? = null` parameter, e.g.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/StateRegistryAggregator.kt
```diff
@@ -82,9 +82,9 @@ internal class StateRegistryAggregator(
 
       // These properties are guaranteed to be non-null because this observer is only registered
       // while attached, and these properties are always non-null while attached.
-      val restoredState =
-        parentRegistryOwner!!.savedStateRegistry.consumeRestoredStateForKey(parentKey!!)
-      restoreFromBundle(restoredState)
+      parentRegistryOwner!!.savedStateRegistry.takeIf { it.isRestored }?.let {
+        restoreFromBundle(it.consumeRestoredStateForKey(parentKey!!))
```
rjrjr, line 86: Probably should always call `restoreFromBundle` here, with `null` if necessary, so that its prune call always happens.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/WorkflowAndroidXSupport.kt
```diff
@@ -34,10 +39,52 @@ public object WorkflowAndroidXSupport {
    * is found in the view tree.
    */
   @WorkflowUiExperimentalApi
-  public fun stateRegistryOwnerFromViewTreeOrContext(view: View): SavedStateRegistryOwner? =
+  public fun requireStateRegistryOwnerFromViewTreeOrContext(view: View): SavedStateRegistryOwner =
```
rjrjr, line 42: Get rid of `require` prefix.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/StateRegistryAggregator.kt
```diff
@@ -0,0 +1,300 @@
...
+    val lifecycleOwner = checkNotNull(WorkflowLifecycleOwner.get(view)) {
+      "Expected container view to have set a WorkflowLifecycleOwner on its immediate " +
+        "child $view."
+    }
+    val registryOwner = KeyedStateRegistryOwner(key, lifecycleOwner)
+    children.put(key, registryOwner)?.let {
+      throw IllegalArgumentException("$key is already in use, it cannot be used register $view")
+    }
+    ViewTreeSavedStateRegistryOwner.get(view)?.let {
+      "$view already has ViewTreeSavedStateRegistryOwner: $it"
```
vgonda, line 216: Did you mean to throw this? It looks like a floating, unused string right now.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/StateRegistryAggregator.kt
```diff
@@ -0,0 +1,300 @@
...
+    }
+  }
+
+  private inline fun doIfRestored(block: (MutableMap<String, Bundle>) -> Unit) {
+    states?.let(block)
+  }
+
+  private fun saveToBundle() = Bundle().apply {
+    doIfRestored { states ->
+      // onWillSave(this@StateRegistryAggregator)
```
vgonda, line 279: I'm not sure what this comment is supposed to communicate. Similar to `// onRestored(this)` below.

vgonda (commented):
I have no arguments about any of this, but I also don't feel I understand it enough to 👍 

If you end up setting up a meeting, I'd love to join (if I can be pinned down. I'm traveling this week, so I know that's a challenge).

workflow-ui/container-android/src/main/java/com/squareup/workflow1/ui/backstack/ViewStateCache.kt
```diff
@@ -131,9 +104,8 @@ internal constructor(
           }
           viewStates += savedKey to ViewStateFrame(savedKey, saved)
 
-          // Registry state (note reverse order as saved above)
-          val owner = ViewTreeSavedStateRegistryOwner.get(oldViewMaybe) as KeyedStateRegistryOwner
-          registryHolder.saveRegistryController(savedKey, owner.controller)
+          // Save registry state (note reverse order as saved above)
```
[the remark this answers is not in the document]
rjrjr: Me either. :) I think this might be a vestige of when this method was more complex, will clean up.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/StateRegistryAggregator.kt
rjrjr: Oops! I was experimenting w/whether I could delete those params. I did, meant to delete this.

workflow-ui/container-android/src/androidTest/java/com/squareup/workflow1/ui/backstack/test/ViewStateCacheTest.kt
```diff
@@ -66,7 +66,9 @@ internal class ViewStateCacheTest {
     // Set some state on the first view that will be saved.
     firstView.viewState = "hello world"
     // Cache expects there to be a registry owner on old views.
-    KeyedStateRegistryOwner.installAsSavedStateRegistryOwnerOn(firstView, "first")
+    StateRegistryAggregator({}) {}.installChildSavedStateRegistryControllerOn(
```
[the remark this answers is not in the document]
rjrjr: More temprorary cruft that I forgot to delete -- those are the same two `onSave / onRestore` params you asked about.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/StateRegistryAggregator.kt
```diff
@@ -0,0 +1,300 @@
...
+   * The set of [KeyedStateRegistryOwner] instances created by
+   * [installChildSavedStateRegistryControllerOn], which have not
+   * yet been retired via [saveChildSavedStateRegistryOwner] or [prune].
+   */
+  private val children = mutableMapOf<String, KeyedStateRegistryOwner>()
+
+  /**
+   * Used to observe the parent registry's lifecycle to know when it becomes `CREATED`
+   * and is ready for us to restore ourselves. This observer is only registered between
+   * calls to [attachToParentRegistry] and [detachFromParentRegistry], and will only
```
steve-the-edwards, line 78: Is this fragile in the sense that if we called `setContentView` from _within_ `onCreate()` which would attach the view to the window, which would `attachToParentRegistry` we could miss the `ON_CREATE`?

I'm trying to think through how this works on config change and process restoration ... why is this working? Are we not calling `setCOntentView` via `onCreate`? Is this installed in `onCreate` but the 'ON_CREATE` lifecycle event happens after `onCreate` is complete?

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/StateRegistryAggregator.kt
```diff
@@ -0,0 +1,300 @@
...
+          "This is most easily remedied by giving your container Screen rendering a unique " +
+          "Compatible.compatibilityKey, perhaps by wrapping it with Named.",
+        e
+      )
+    }
+
+    // Even if the parent lifecycle is in a state further than CREATED, new observers are sent all
+    // the lifecycle events required to catch the observer up to the current state, so this handles
+    // both the cases where we're ready to immediately restore, and where we have to wait.
+    parentLifecycle.addObserver(lifecycleObserver)
```
steve-the-edwards, line 162: Ahh. tada. this is the answer to my query above I guess. Good to know.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/StateRegistryAggregator.kt
```diff
@@ -0,0 +1,300 @@
...
+    parentRegistryOwner?.lifecycle?.removeObserver(lifecycleObserver)
+    parentRegistryOwner = null
+    parentKey = null
+    children.clear()
+  }
+
+  /**
+   * Puts a new [ViewTreeSavedStateRegistryOwner] in place on [view], registered
+   * with its [WorkflowLifecycleOwner]. **This method must be called before [view]
+   * is attached to a window.**
```
steve-the-edwards, line 182: This _is_ quite the dance eh? SO how about them Fragments? (/joking!)

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/StateRegistryAggregator.kt
```diff
@@ -0,0 +1,300 @@
...
+   *
+   * Clean up requirements after making this call are nuanced:
+   *
+   *  - all instances are dropped after [detachFromParentRegistry] is called
+   *
+   *  - if [view] is dropped but may be restored before [detachFromParentRegistry] is called,
+   *    as when pushing and popping a back stack, call [saveChildRegistryOwner] to capture
+   *    its state tear down its [ViewTreeSavedStateRegistryOwner]. The saved state
+   *    will be restored to the next [view] passed to [installChildSavedStateRegistryOwnerOn]
+   *    with the same [key]
```
steve-the-edwards, line 192: Got lost in the grammar of the first sentence?  "call X to capture its state tear down its ..."?

vgonda:
> 

😭

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/StateRegistryAggregator.kt
steve-the-edwards, line 216: Is this supposed to be an exception? Isn't it just a naked string here?

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/StateRegistryAggregator.kt
```diff
@@ -0,0 +1,300 @@
...
+    }
+  }
+
+  /**
+   * If this object has been restored from its parent registry, restores the child
+   * [KeyedStateRegistryOwner.controller] that was previously [saved][saveChildSavedStateRegistryOwner]
+   * with the same [KeyedStateRegistryOwner.key].
+   *
+   * This method can only be called once per key per instance of this class.
+   * After a key has been restored, its data will be removed from this object.
```
steve-the-edwards, line 249: Is this part important? Why do we remove it? IS that what AndroidX is doing as well?

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/StateRegistryAggregator.kt
```diff
@@ -0,0 +1,300 @@
...
+    (children.keys - keysToKeep).forEach { children.remove(it) }
+
+    doIfRestored { states ->
+      val deadKeys = states.keys - keysToKeep
+      states -= deadKeys
+    }
+  }
+
+  private inline fun doIfRestored(block: (MutableMap<String, Bundle>) -> Unit) {
+    states?.let(block)
```
steve-the-edwards, line 274: Don't we have `isRestored` also? We are kind of duplicating that here.

steve-the-edwards (approved):

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/WorkflowAndroidXSupport.kt
```diff
@@ -34,10 +38,52 @@ public object WorkflowAndroidXSupport {
...
+   * There's a potential issue here where if [containerView]'s ID is changed to something else, then
+   * another container is added with the old ID, the new container will overwrite the old one's
+   * state. Since they'd both be using the same key, [SavedStateRegistry] would throw an exception.
+   * However, as long as the first container is detached before its ID is changed this shouldn't be
+   * a problem.
+   */
+  @OptIn(WorkflowUiExperimentalApi::class)
+  public fun createStateRegistryKeyForContainer(containerView: View): String {
+    val nameSuffix = containerView.getRendering<Any>()
+      ?.let { Compatible.keyFor(it) }?.let { "-$it" }
```
helios175, line 81: If `Compatible.keyFor(it)` returns something it won't have a trailing `"-"` right?

helios175:
There's a bunch of
```
com.squareup.workflow1.ui.compose.ComposeViewTreeIntegrationTest > composition_state_is_not_restored_after_screen_is_removed_from_backstack[test(AVD) - 5.0.2] FAILED 
	java.lang.IllegalStateException: You can consumeRestoredStateForKey only after super.onCreate of corresponding component
	at androidx.savedstate.SavedStateRegistry.consumeRestoredStateForKey(SavedStateRegistry.java:77)

com.squareup.workflow1.ui.compose.ComposeViewTreeIntegrationTest > composition_state_is_restored_after_config_change[test(AVD) - 5.0.2] FAILED 
	java.lang.IllegalStateException: You can consumeRestoredStateForKey only after super.onCreate of corresponding component
	at androidx.savedstate.SavedStateRegistry.consumeRestoredStateForKey(SavedStateRegistry.java:77)

com.squareup.workflow1.ui.compose.ComposeViewTreeIntegrationTest > composition_is_disposed_when_navigated_away_dispose_on_destroy_strategy[test(AVD) - 5.0.2] FAILED 
	java.lang.IllegalStateException: You can consumeRestoredStateForKey only after super.onCreate of corresponding component
	at androidx.savedstate.SavedStateRegistry.consumeRestoredStateForKey(SavedStateRegistry.java:77)
```
in the instrumentation tests. Misconfigured or wrong assumptions?

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/StateRegistryAggregator.kt
rjrjr: This is a lie, I deleted it. We actually do call this method more than once.

workflow-ui/container-android/src/main/java/com/squareup/workflow1/ui/backstack/ViewStateCache.kt
```diff
@@ -131,9 +104,8 @@ internal constructor(
...
+          stateRegistryAggregator.saveChildSavedStateRegistryOwner(savedKey)
```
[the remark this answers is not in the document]
rjrjr: Left over comment, no longer meaningful. Deleted.

workflow-ui/container-android/src/main/java/com/squareup/workflow1/ui/backstack/ViewStateCache.kt
```diff
@@ -37,30 +39,7 @@ internal constructor(
...
-        // We're only allowed to restore from an INITIALIZED state, but this callback can also be
-        // invoked while the owner is already CREATED.
-        // https://github.com/square/workflow-kotlin/issues/570
-        ?.takeIf { it.lifecycle.currentState == Lifecycle.State.INITIALIZED }
-        ?.let { owner ->
-          aggregator.restoreRegistryControllerIfReady(owner.key, owner.controller)
-        }
-    }
-  )
+  private val stateRegistryAggregator = StateRegistryAggregator()
```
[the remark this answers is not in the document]
rjrjr: How about `WorkflowSavedStateRegistryAggregator`? More parallel to `WorkflowLifecycleOwner`.

samples/containers/android/src/main/java/com/squareup/sample/container/overviewdetail/OverviewDetailContainer.kt
```diff
@@ -53,11 +53,13 @@ class OverviewDetailContainer(view: View) : LayoutRunner<OverviewDetailScreen> {
     if (rendering.detailRendering == null && rendering.selectDefault != null) {
       rendering.selectDefault!!.invoke()
     } else {
-      // Since we have two sibling backstacks, we need to give them each different
-      // SavedStateRegistry key prefixes.
-      val overviewViewEnvironment = viewEnvironment
-        .withBackStackStateKeyPrefix(OverviewBackStackKey) + (OverviewDetailConfig to Overview)
```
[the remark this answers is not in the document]
rjrjr: The latter. `OverviewBackStackKey` was a hacky hint that was read only by `BackStackContainer`. Now that we have general support tied to `Compatible.compatibilityKey`, we can use `Named` for this kind of issue.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/StateRegistryAggregator.kt
rjrjr: Yup, need to add a test covering it.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/StateRegistryAggregator.kt
rjrjr: If we check `isRestored` here we'd be duplicating the `states == null` check instead.

workflow-ui/container-android/src/main/java/com/squareup/workflow1/ui/modal/ModalContainer.kt
```diff
@@ -75,6 +75,13 @@ public abstract class ModalContainer<ModalRenderingT : Any> @JvmOverloads constr
               dialogView,
               findParentLifecycle = { parentLifecycleOwner?.lifecycle }
             )
+            // Ensure that each dialog has its own ViewTreeSavedStateRegistryOwner,
+            // so views in each dialog layer don't clash with other layers.
+            ref.savedStateRegistryKey = Compatible.keyFor(newScreen, i.toString())
```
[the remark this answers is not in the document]
rjrjr: `buildDialog` is an abstract method. By passing the id through it we're relying on subclasses to handle it properly. I did it this way to have one less mistake they could make. WDYT?

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/KeyedStateRegistryOwner.kt
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * The container should use a [StateRegistryAggregator] to manage its set of
+ * [KeyedStateRegistryOwner] instances, which will save and restore them
+ * via its own [ViewTreeSavedStateRegistryOwner].
+ *
+ * To create an instance, call [StateRegistryAggregator.installChildSavedStateRegistryControllerOn].
+ *
+ * @param key The key used to save and restore this controller from a [SavedStateRegistry].
+ * @param lifecycleOwner The [LifecycleOwner] that will be delegated to by this instance. Note that
+ * [SavedStateRegistryOwner] extends [LifecycleOwner].
```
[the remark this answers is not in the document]
rjrjr: Yeah, I'm not sure that this is useful information (and I think it was a mistake that the androidx team made this choice, can't see this redundant lifecycle is useful, or even safe to rely on). I'll delete the comment.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/KeyedStateRegistryOwner.kt
```diff
@@ -0,0 +1,33 @@
...
+ */
+@WorkflowUiExperimentalApi
+internal class KeyedStateRegistryOwner internal constructor(
```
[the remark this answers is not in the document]
rjrjr: Done

workflow-ui/container-android/src/main/java/com/squareup/workflow1/ui/backstack/BackStackContainer.kt
```diff
@@ -47,17 +46,11 @@ import com.squareup.workflow1.ui.start
...
- * [compatibility key][com.squareup.workflow1.ui.Compatible.compatibilityKey] of its rendering.
- * This method isn't guaranteed to give a unique registry key, but it should be good enough: If you
- * need to nest multiple [BackStackContainer]s under the same `SavedStateRegistry`, just wrap each
- * [BackStackScreen] with a [Named], or give each [BackStackContainer] a unique view ID.
- *
- * There's a potential issue here where if our ID is changed to something else, then another
- * [BackStackContainer] is added with our old ID, that container will overwrite our state. Since
- * they'd both be using the same key, [SavedStateRegistry] would throw an exception. As long as this
- * container is detached before its ID is changed, it shouldn't be a problem.
+ * requirement, [BackStackContainer] tries to generate a best-effort unique key by using
```
rjrjr, line 49: stale

workflow-ui/container-android/src/main/java/com/squareup/workflow1/ui/modal/ModalContainer.kt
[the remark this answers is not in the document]
steve-the-edwards: Gotcha. I guess its a bit awkward either way. Anyway - this is fine enough :)

workflow-ui/container-android/src/main/java/com/squareup/workflow1/ui/modal/ModalContainer.kt
```diff
@@ -48,24 +47,25 @@ public abstract class ModalContainer<ModalRenderingT : Any> @JvmOverloads constr
     WorkflowLifecycleOwner.get(this)
   }
 
+  /**
+   * Provides a `ViewTreeSavedStateRegistryOwner` for each dialog,
+   * tied to that for this container view.
```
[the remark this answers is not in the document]
rjrjr:
```kotlin
  /**
   * Provides a new `ViewTreeSavedStateRegistryOwner` for each dialog,
   * which will save to `ViewTreeSavedStateRegistryOwner` of this container view.
   */
  private val stateRegistryAggregator = WorkflowSavedStateRegistryAggregator()
```

rjrjr:
> There's a bunch of
> 
> ```
> com.squareup.workflow1.ui.compose.ComposeViewTreeIntegrationTest > composition_state_is_not_restored_after_screen_is_removed_from_backstack[test(AVD) - 5.0.2] FAILED 
> 	java.lang.IllegalStateException: You can consumeRestoredStateForKey only after super.onCreate of corresponding component
> 	at androidx.savedstate.SavedStateRegistry.consumeRestoredStateForKey(SavedStateRegistry.java:77)
> 
> com.squareup.workflow1.ui.compose.ComposeViewTreeIntegrationTest > composition_state_is_restored_after_config_change[test(AVD) - 5.0.2] FAILED 
> 	java.lang.IllegalStateException: You can consumeRestoredStateForKey only after super.onCreate of corresponding component
> 	at androidx.savedstate.SavedStateRegistry.consumeRestoredStateForKey(SavedStateRegistry.java:77)
> 
> com.squareup.workflow1.ui.compose.ComposeViewTreeIntegrationTest > composition_is_disposed_when_navigated_away_dispose_on_destroy_strategy[test(AVD) - 5.0.2] FAILED 
> 	java.lang.IllegalStateException: You can consumeRestoredStateForKey only after super.onCreate of corresponding component
> 	at androidx.savedstate.SavedStateRegistry.consumeRestoredStateForKey(SavedStateRegistry.java:77)
> ```
> 
> in the instrumentation tests. Misconfigured or wrong assumptions?

The caught me in a couple of actual bugs on the third SHA. Phew!

workflow-ui/compose/src/androidTest/java/com/squareup/workflow1/ui/compose/ComposeViewTreeIntegrationTest.kt
```diff
@@ -436,13 +446,117 @@ internal class ComposeViewTreeIntegrationTest {
...
+      // Use the same compatibility key across layers – these screens are in different modals, so
+      // they won't conflict.
+      compatibilityKey = screen.toString()
+    ) {
+      var counter by rememberSaveable { mutableStateOf(0) }
+      BasicText(
+        "Counter[$layer][$screen]: $counter",
+        Modifier
+          .clickable { counter++ }
+          .testTag("[$layer][$screen]")
```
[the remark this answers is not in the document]
rjrjr: Wow, nice catch. Thanks.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/WorkflowAndroidXSupport.kt
rjrjr: Nice catch.

Starting to second guess whether this method is even useful. Zach put it in place before I made the `Compatible.keyFor` usage more consistent -- he originally was doing `is Named<*>` checks. I'm going to try deleting this.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/StateRegistryAggregator.kt
rjrjr: Is this better?

```
   *  - if [view] is dropped but may later be replaced with a new instance, as when
   *   pushing and popping a back stack, call [saveAndPruneChildRegistryOwner]
   *   to capture its state, and to stop requesting updates from it.
   *   The saved state will be restored to the next [view] passed to
   *   [installChildRegistryOwnerOn] with the same [key]
```

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/StateRegistryAggregator.kt
rjrjr: It's SO handy. My single favorite thing about `Lifecycle`.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/KeyedStateRegistryOwner.kt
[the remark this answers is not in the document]
rjrjr: Actually, this _is_ kind of useful, if only to explain why the hell we're doing this at all. Rephrased it a bit:

```
 * @param lifecycleOwner The [LifecycleOwner] that will be delegated to by this instance.
 * (Required because [SavedStateRegistryOwner] extends [LifecycleOwner] for no clear reason.)
```

workflow-ui/container-android/src/main/java/com/squareup/workflow1/ui/modal/ModalContainer.kt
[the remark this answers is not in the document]
rjrjr: FWIW:

```kotlin
          // This is the unique id we'll use to make stateRegistryAggregator put a
          // SavedStateRegistryOwner on the dialog's decorView, as required by Compose.
          // We put it in this sketchy lateInit field on DialogRef rather than passing
          // it through abstract buildDialog method so that subclasses can't screw it up.
          ref.savedStateRegistryKey = Compatible.keyFor(modal, i.toString())
```

workflow-ui/container-android/src/androidTest/java/com/squareup/workflow1/ui/modal/test/ModalViewContainerLifecycleActivity.kt
```diff
@@ -45,7 +45,11 @@ internal class ModalViewContainerLifecycleActivity : AbstractLifecycleTestActivi
       override val compatibilityKey: String get() = name
     }
 
-    data class RecurseRendering(val wrapped: LeafRendering) : TestRendering()
+    data class RecurseRendering(
```
rjrjr, line 48: Unneeded

rjrjr:
Cool, cool, this passes locally:

```
* What went wrong:
Execution failed for task ':workflow-ui:container-android:ktlintAndroidTestSourceSetCheck'.
> A failure occurred while executing org.jlleitschuh.gradle.ktlint.worker.ConsoleReportWorkAction
   > KtLint found code style violations. Please see the following reports:
     - /home/runner/work/workflow-kotlin/workflow-kotlin/workflow-ui/container-android/build/reports/ktlint/ktlintAndroidTestSourceSetCheck/ktlintAndroidTestSourceSetCheck.json
```

rjrjr:
> Cool, cool, this passes locally:
> 
> ```
> * What went wrong:
> Execution failed for task ':workflow-ui:container-android:ktlintAndroidTestSourceSetCheck'.
> > A failure occurred while executing org.jlleitschuh.gradle.ktlint.worker.ConsoleReportWorkAction
>    > KtLint found code style violations. Please see the following reports:
>      - /home/runner/work/workflow-kotlin/workflow-kotlin/workflow-ui/container-android/build/reports/ktlint/ktlintAndroidTestSourceSetCheck/ktlintAndroidTestSourceSetCheck.json
> ```

Oh, b/c I hadn't pushed. Ray tired.

rjrjr:
@helios175 @vgonda @steve-the-edwards @RBusarow Everything passes locally. This should be ready for a final look through if you're so inclined.

I believe I've addressed all the feedback. The broken tests revealed one or two real problems, but no drastic changes were required. I've renamed a bunch of classes and methods, and was able to delete `createStateRegistryKeyForContainer()`, which is nice.

rjrjr:
Yay, I think I added a flaky test:

```
com.squareup.sample.TicTacToeEspressoTest > fullJourney[test(AVD) - 7.0] FAILED 
	androidx.test.espresso.NoMatchingViewException: No views in hierarchy found matching: with id is <com.squareup.sample.tictacworkflow:id/login_email>
```

rjrjr:
Publishing this branch as `1.6.0-SSR3-SNAPSHOT` while I do one more round of internal smoke testing.

rjrjr:
Maybe the answer is to add a non-optional `findParentLifecycle` function to `buildView`?

rjrjr:
Or maybe it's just fine to build it in. The only place we customize the `findParentLifecycle` function is when dialog builders apply it to `Dialog.window.decorView`, otherwise it's always the default function that looks in the view's parent. Redundant calls will just clobber each other with an identical function.

Wondering if there would be any remaining uses of `ViewStarter` with this change.

rjrjr:
> Wondering if there would be any remaining uses of `ViewStarter` with this change.

Yes, many, via `DecorativeViewFactory`.

rjrjr:
Going to track this separately, #673

workflow-ui/container-android/src/main/java/com/squareup/workflow1/ui/modal/ModalContainer.kt
```diff
@@ -167,12 +200,25 @@ public abstract class ModalContainer<ModalRenderingT : Any> @JvmOverloads constr
...
+    /**
+     * The unique id of the `ViewTreeSavedStateRegistryOwner` that will be placed
+     * on the dialog's decor view by [stateRegistryAggregator].
+     */
+    internal lateinit var savedStateRegistryKey: String
+
+    public fun copy(
+      modalRendering: ModalRenderingT,
+      viewEnvironment: ViewEnvironment,
+    ): DialogRef<ModalRenderingT> = DialogRef(modalRendering, viewEnvironment, dialog, extra).also {
```
steve-the-edwards, line 218: Are we trying to enforce any behaviour here (i.e. specify `modalRendering` and `viewEnvironment` when making a new copy instance?). Otherwise we could just do a typical copy function like:

```
copy(
  modalRendering: ModalRenderingT = this.modalRendering,
  viewEnvironment: ViewEnvironment = this.viewEnvironment,
  dialog: Dialog = this.dialog,
  extra: Any? = this.extra,
```

steve-the-edwards (approved):
Still LGTM

workflow-ui/container-android/src/main/java/com/squareup/workflow1/ui/modal/ModalContainer.kt
rjrjr: It won't matter in practice b/c this class is eliminated by the overhaul, but yeah, good point. I'll make it normal.

rjrjr: Also note that that class itself is actually `protected`, so all of these `public`s are mistakes.

rjrjr: Oh, actually required b/c it's no longer a `data` class.

rjrjr:
Internal tests are green. I'm making the `DialogRef.copy()` fix that @steve-the-edwards suggested, merging this, and then I'll publish 1.6.0.

rjrjr:
God damn it:

```kotlin
com.squareup.workflow1.ui.compose.ComposeViewTreeIntegrationTest > composition_is_restored_in_multiple_modals_backstacks_after_config_change[test(AVD) - 10] FAILED 
	java.lang.AssertionError: Failed to assert the following: (Text + EditableText = [Counter[0][0]: 1])
	The node is no longer in the tree, last known semantics:
```

rjrjr:
These are green at SHA 6e6bbf7e

rjrjr:
@RBusarow Looks like the only interesting change since the green build is #648, the move to the version catalogs. Yaaaay. No specific version change there is jumping out at me yet. Tweaked a few, no improvement.

workflow-ui/container-android/src/main/java/com/squareup/workflow1/ui/modal/ModalContainer.kt
```diff
@@ -132,9 +153,21 @@ public abstract class ModalContainer<ModalRenderingT : Any> @JvmOverloads constr
...
-      // Some other class wrote state, but we're not allowed to skip
-      // the call to super. Make a no-op call.
       ?: super.onRestoreInstanceState(super.onSaveInstanceState())
+    // ?: Some other class wrote state, but we're not allowed to skip
+    // the call to super. Make a no-op call.
+  }
+
+  override fun onAttachedToWindow() {
+    super.onAttachedToWindow()
+    val parentRegistry = stateRegistryOwnerFromViewTreeOrContext(this)
```
rjrjr, line 163: DRY. #681

PR #658, opened 2022-02-02
wrap Activity and Compose rules in LeakCanary checks on both sides

### LINKED ISSUE #657: Make LeakCanary more strict
Reported by @pyricau:  At the moment we are only checking for leaks after the activity is destroyed, which means there is a huge window we're missing. We should check before that as well:

```kotlin
@get:Rule val rules = RuleChain.outerRule(DetectLeaksAfterTestSuccess()).around(scenarioRule).around(DetectLeaksAfterTestSuccess())!!
```

RBusarow:
I'm open to bike-shedding the name. 😄 

fixes #657

### DIFF
--- a/workflow-ui/internal-testing-android/src/main/java/com/squareup/workflow1/ui/internal/test/DetectLeaksAfterTestSuccess.kt
+++ b/workflow-ui/internal-testing-android/src/main/java/com/squareup/workflow1/ui/internal/test/DetectLeaksAfterTestSuccess.kt
@@ -2,6 +2,7 @@ package com.squareup.workflow1.ui.internal.test
 
 import leakcanary.AppWatcher
 import leakcanary.LeakAssertions
+import org.junit.rules.RuleChain
 import org.junit.rules.TestRule
 import org.junit.runner.Description
 import org.junit.runners.model.Statement
@@ -37,3 +38,15 @@ public class DetectLeaksAfterTestSuccess(
     }
   }
 }
+
+/**
+ * Invokes [LeakAssertions.assertNoLeaks] before performing the tear-down logic of the receiver
+ * [TestRule], then again *after* the other rule's logic.
+ *
+ * https://github.com/square/workflow-kotlin/issues/657
+ */
+public fun TestRule.wrapInLeakCanary(): RuleChain = requireNotNull(
+  RuleChain.outerRule(DetectLeaksAfterTestSuccess())
+    .around(this)
+    .around(DetectLeaksAfterTestSuccess())
+)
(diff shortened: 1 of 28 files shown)


workflow-ui/internal-testing-android/src/main/java/com/squareup/workflow1/ui/internal/test/DetectLeaksAfterTestSuccess.kt
```diff
@@ -37,3 +38,15 @@ public class DetectLeaksAfterTestSuccess(
...
   }
 }
+
+/**
+ * Invokes [LeakAssertions.assertNoLeaks] before performing the tear-down logic of the receiver
+ * [TestRule], then again *after* the other rule's logic.
+ *
+ * https://github.com/square/workflow-kotlin/issues/657
+ */
+public fun TestRule.wrapInLeakCanary(): RuleChain = requireNotNull(
```
RBusarow, line 48: The actual rule is just above the fold, so GitHub won't let me comment on the line I'm interested in...

https://github.com/square/workflow-kotlin/blob/72ba8c7bedbd0118164db82b49bc849aa119e90c/workflow-ui/internal-testing-android/src/main/java/com/squareup/workflow1/ui/internal/test/DetectLeaksAfterTestSuccess.kt#L29-L36

If we're executing this twice per test, do we want to skip the call to `clearWatchedObjects()` the first time?

rjrjr: I have no idea. @py?

pyricau: No, should be fine. But maybe use clearAfter timestamp instead

pyricau:
```
 try { 
   base.evaluate()
   val heapDumpUptimeMillis = SystemClock.uptimeMillis()
   LeakAssertions.assertNoLeaks(tag) 
 } finally { 
   // Otherwise upstream test failures will be reported as leaks. 
   // https://github.com/square/leakcanary/issues/2297 
   AppWatcher.objectWatcher.clearObjectsWatchedBefore(heapDumpUptimeMillis) 
 } 
 ```

RBusarow: I made this change -- thanks!

vgonda (approved):
LGTM

But I'm just learning about all this, so you probably want another 👍

steve-the-edwards (approved):

pyricau:
One thing to note: all the changes here (and in internal forks of the rule) have been backported to LeakCanary which will soon be released as 2.9. You'll need to adapt the code a bit but should be able to delete this custom rule.

https://github.com/square/leakcanary/blob/main/leakcanary-android-instrumentation/src/main/java/leakcanary/DetectLeaksAfterTestSuccess.kt

rjrjr:
@RBusarow Any plans to update this to LC 2.9 and get it merged?

PR #662, opened 2022-02-15
661: Save JMH Benchmark Results

### LINKED ISSUE #661: Update JMH Benchmarks and ensure running/healthy
See https://workflow-community.slack.com/archives/CHTFPR277/p1575584077139800

steve-the-edwards:
Closes #661 .

Save results to a file (for now). Its not perfect, but its a least a historical record.

Also updated JMH version.

### DIFF
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -105,8 +105,8 @@ object Dependencies {
 
   object Jmh {
     const val gradlePlugin = "me.champeau.gradle:jmh-gradle-plugin:0.5.3"
-    const val core = "org.openjdk.jmh:jmh-core:1.32"
-    const val generator = "org.openjdk.jmh:jmh-generator-annprocess:1.32"
+    const val core = "org.openjdk.jmh:jmh-core:1.34"
+    const val generator = "org.openjdk.jmh:jmh-generator-annprocess:1.34"
   }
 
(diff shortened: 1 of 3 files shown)


steve-the-edwards:
CI Timeout on API 21 not concerning as typical flake unrelated to this.

rjrjr (approved):

rjrjr:
> CI Timeout on API 21 not concerning as typical flake unrelated to this.

@steve-the-edwards @RBusarow is it time to delete this shard? I can't remember the last time it provided any signal that the other one didn't.

RBusarow:
> > CI Timeout on API 21 not concerning as typical flake unrelated to this.
> 
> @steve-the-edwards @RBusarow is it time to delete this shard? I can't remember the last time it provided any signal that the other one didn't.

That depends on how much faith we have in the abstractions provided by Androidx, doesn't it?  Clearly there's some logic which is api-level sensitive, like that test I wrote which only fails on 31.

RBusarow (approved):

rjrjr:
> > > CI Timeout on API 21 not concerning as typical flake unrelated to this.
> > 
> > 
> > @steve-the-edwards @RBusarow is it time to delete this shard? I can't remember the last time it provided any signal that the other one didn't.
> 
> That depends on how much faith we have in the abstractions provided by Androidx, doesn't it? Clearly there's some logic which is api-level sensitive, like that test I wrote which only fails on 31.

Sounds like we should replace it with an API 31 shard. Wonder if that's even possible.

PR #670, opened 2022-02-17
introduce `IdlingDispatcher` and `IdlingDispatcherRule`

RBusarow:
This hooks coroutines into Espresso's `IdlingRegistry`.  It's adapted from [Dispatch](https://github.com/RBusarow/Dispatch).

This is somewhat less effective than when using the full Dispatch library, but it does have value when only wrapping `Dispatchers.Main[.immediate]`.  Any coroutines which are _started_ from `Main` (or immediate) will be tracked, even if they then dispatch to something else using `withContext`.


workflow-ui/internal-testing-android/src/test/java/com/squareup/workflow1/ui/internal/test/IdlingDIspatcherTest.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    assertThat(idler.isIdle()).isTrue()
+
+    val job = launch(idler) {
+      withContext(b) {
+        lock.await()
+        fail("unreachable")
+      }
+    }
+
```
rjrjr, line 63: Would `isIdle()` be `false` before the `yield()`? Should we add such an assertion?

workflow-ui/internal-testing-android/src/test/java/com/squareup/workflow1/ui/internal/test/IdlingDIspatcherTest.kt
```diff
@@ -0,0 +1,82 @@
...
+  inner class NamedDispatcher(private val name: String) : CoroutineDispatcher() {
+
+    override fun dispatch(
+      context: CoroutineContext,
+      block: Runnable
+    ) {
+      block.run()
+    }
+
+    override fun toString(): String = "RecordingDispatcher ($name)"
```
rjrjr, line 80:
```suggestion
    override fun toString(): String = "NamedDispatcher ($name)"
```

rjrjr (approved):
So cool. I bet we can get rid of `TicTacToeComponent#countingIdlingResource` now.

I wonder if this will actually solve [this problem](https://github.com/square/workflow-kotlin/pull/656#issuecomment-1042442255), where I suspect espresso is peaking at the UI before `BackStackContainer#onRestoreInstanceState` gets to run.

workflow-ui/compose/src/androidTest/java/com/squareup/workflow1/ui/compose/ComposeViewFactoryTest.kt
```diff
@@ -36,6 +37,7 @@ internal class ComposeViewFactoryTest {
   @get:Rule val rules: RuleChain =
     RuleChain.outerRule(DetectLeaksAfterTestSuccess())
       .around(IdleAfterTestRule)
+      .around(IdlingDispatcherRule)
```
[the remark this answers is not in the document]
rjrjr: Probably time to write a `WatchLeaksAndIdleBetweenDispatches<R: Rule>`, where `R` would give us access to the wrapped `ComposeRule` or `ScenarioRule`.

rjrjr:
> So cool. I bet we can get rid of `TicTacToeComponent#countingIdlingResource` now.
> 
> I wonder if this will actually solve [this problem](https://github.com/square/workflow-kotlin/pull/656#issuecomment-1042442255), where I suspect espresso is peeking at the UI before `BackStackContainer#onRestoreInstanceState` gets to run.

If it doesn't, we could create a wrapping `ViewGroup` or `FrameLayout` that ties in to save / restore.

workflow-ui/internal-testing-android/src/main/java/com/squareup/workflow1/ui/internal/test/IdlingDispatcher.kt
```diff
@@ -0,0 +1,64 @@
...
+ * idle.
+ */
+public class IdlingDispatcher(
+  private val delegate: CoroutineDispatcher
+) : CoroutineDispatcher() {
+
+  /**
+   * The [CountingIdlingResource] which is responsible for [Espresso] functionality.
+   */
+  public val counter: CountingIdlingResource = CountingIdlingResource("IdlingResource for $this")
```
rjrjr, line 31: Where are we actually bringing this to Espresso's attention?

RBusarow: In the rule.

rjrjr: Need a :derp: emoji. Oh, I guess we have one 🤦

workflow-ui/compose/src/androidTest/java/com/squareup/workflow1/ui/compose/ComposeViewFactoryTest.kt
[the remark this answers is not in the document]
RBusarow: The order's different because...  I honestly don't know?  But nice catch -- I'll fix.

I agree that a composite rule seems to make sense.  Once this is merged, I can address the leaks found by the LeakCanary rule changes and combine all the logic.

rjrjr:
> > So cool. I bet we can get rid of `TicTacToeComponent#countingIdlingResource` now.
> > I wonder if this will actually solve [this problem](https://github.com/square/workflow-kotlin/pull/656#issuecomment-1042442255), where I suspect espresso is peeking at the UI before `BackStackContainer#onRestoreInstanceState` gets to run.
> 
> If it doesn't, we could create a wrapping `ViewGroup` or `FrameLayout` that ties in to save / restore.

I think it did the trick! I can't repro that failure locally anymore!

workflow-ui/internal-testing-android/src/test/java/com/squareup/workflow1/ui/internal/test/IdlingDIspatcherTest.kt
RBusarow: `isIdle()` is `true` before the yield.  The assertion is made before `launch { ... }`.

Unless we launch with `start = UNDISPATCHED` or use `Dispatchers.Unconfined`, the runnable is queued for dispatch.  And since everything is executing on a single thread, it won't actually be dispatched until we call `yield()` or `job.join()`.

workflow-ui/internal-testing-android/src/test/java/com/squareup/workflow1/ui/internal/test/IdlingDIspatcherTest.kt
RBusarow: Copy pasta 😄 

thanks!

workflow-ui/internal-testing-android/src/main/java/com/squareup/workflow1/ui/internal/test/IdlingDispatcherRule.kt
```diff
@@ -0,0 +1,27 @@
...
+import kotlinx.coroutines.test.setMain
+import org.junit.rules.TestWatcher
+import org.junit.runner.Description
+
+@OptIn(ExperimentalCoroutinesApi::class)
+public object IdlingDispatcherRule : TestWatcher() {
+
+  private lateinit var dispatcher: IdlingDispatcher
+
+  override fun starting(description: Description?) {
```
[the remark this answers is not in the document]
RBusarow: I've never used the description argument for anything but reads.  Do you have an example of modifying it?  What would I add?

RBusarow:
> > So cool. I bet we can get rid of TicTacToeComponent#countingIdlingResource now.
> 
> Ah, but this is not true. Looks like we need both. That's fine.

Yeah, I'm afraid so.  Because `RealAuthService` uses `delay(...)`, the dispatcher temporarily goes idle.  The ergonomics of this get a little weird, especially when dealing with `Dispatchers.Main`.  We still delegate all actual execution to the original dispatcher, which means that the original dispatcher is handling delays and we can't track them.

It's been almost two years since I thought about this stuff.  There may be options for how we might bring two idling resources down to one.  I'll probably poke around later and I'll make another PR if I'm able to come up with something.

workflow-ui/internal-testing-android/src/test/java/com/squareup/workflow1/ui/internal/test/IdlingDIspatcherTest.kt
RBusarow: I updated the test a bit to add clarification.  The short answer to your question, based upon the previous state, is that it wouldn't be false, and then it would be again. :)

rjrjr (approved):

steve-the-edwards (approved):

PR #671, opened 2022-02-17
Sacrificing some chickens to the espresso gods

rjrjr:
Two commits so far:

###     Drops AVD 21, bumps 27 to 29

21 has long been a source of nothing but noise, and I suspect that 29 is the first that actually honors the commands we use to turn of animation.

We have at least one test that only fails on 32, but that's not available yet. 30 and 31 are reputed to be [too slow to use](https://github.com/ReactiveCircus/android-emulator-runner/issues/222).

###    Drops com.android.tools:desugar_jdk_libs

Suspecting that it's the cause of a frequent CI failure building `samples:dungeon:common:compileKotlin`:

```
Caused by: org.gradle.api.GradleException: Could not read directory path '/Users/runner/work/workflow-kotlin/workflow-kotlin/samples/dungeon/common/build/kotlin/compileKotlin/caches-jvm/inputs'.
```

At any rate, it's the only interesting thing about that build, and doesn't appear to be necessary any longer.

### DIFF
--- a/.github/workflows/kotlin.yml
+++ b/.github/workflows/kotlin.yml
@@ -63,8 +63,9 @@ jobs:
       fail-fast: false
       matrix:
         api-level:
-          - 21
-          - 24
+          - 29
+      # Unclear that older versions actually honor command to disable animation.
+      # Newer versions are reputed to be too slow: https://github.com/ReactiveCircus/android-emulator-runner/issues/222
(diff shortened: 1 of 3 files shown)


RBusarow (approved):

steve-the-edwards (approved):
SGTM!

PR #675, opened 2022-02-21
Truth in advertising: findParentLifecycle cannot return null

rjrjr:
The `findParentLifecycle` function param of `WorkflowLifecycleInstallOn` had a nullable return type, but its only call site was wrapped in a fatal null check. That's just silly.


workflow-ui/container-android/src/main/java/com/squareup/workflow1/ui/modal/ModalContainer.kt
```diff
@@ -48,7 +48,7 @@ public abstract class ModalContainer<ModalRenderingT : Any> @JvmOverloads constr
    * this container's modals. Only valid after the the view has been attached.
    */
   private val parentLifecycleOwner by lazy(mode = LazyThreadSafetyMode.NONE) {
-    WorkflowLifecycleOwner.get(this)
+    WorkflowLifecycleOwner.get(this)!!
```
rjrjr, line 51: Should put a more helpful error message here.

PR #676, opened 2022-02-22
Added "New Board" button to Game Over dialog of Dungeon Crawler sample

### LINKED ISSUE #205: Add a "New Board" button to the Game Over dialog in the Dungeon Crawler sample.
_Originally posted by @zach-klippenstein in https://github.com/square/workflow/pull/874#issuecomment-574342417_

lpizzinidev:
resolves #205

### DIFF
--- a/samples/dungeon/app/src/main/java/com/squareup/sample/dungeon/GameSessionWorkflow.kt
+++ b/samples/dungeon/app/src/main/java/com/squareup/sample/dungeon/GameSessionWorkflow.kt
@@ -7,6 +7,7 @@ import com.squareup.sample.dungeon.GameSessionWorkflow.Props
 import com.squareup.sample.dungeon.GameSessionWorkflow.State
 import com.squareup.sample.dungeon.GameSessionWorkflow.State.GameOver
 import com.squareup.sample.dungeon.GameSessionWorkflow.State.Loading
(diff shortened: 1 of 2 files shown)


samples/dungeon/app/src/main/java/com/squareup/sample/dungeon/GameSessionWorkflow.kt
```diff
@@ -44,6 +47,7 @@ class GameSessionWorkflow(
     object Loading : State(), Screen
     data class Running(val board: Board) : State()
     data class GameOver(val board: Board) : State()
+    object NewBoard : State()
```
steve-the-edwards, line 50: The way this is architected, the `DungeonAppWorkflow` is the parent of this `GameSessionWorkflow` and so we don't want to have this as a `State` here where we run `DungeonAppWorkflow` as a child because we are then reproducing this parent and child for each game session.

Instead this should be an `Output` (need to add a type as this currently has `Nothing` as Output type).

Then in `DungeonAppWorkflow` when we run this as a child we would handle the Output and set the state back to `ChoosingBoard`.

That way we won't just keep building an infinite tree of Workflows while re-playing sessions.

steve-the-edwards (changes requested):

lpizzinidev:
@steve-the-edwards 
Thanks for the feedback!
I've tried to fix the problem, let me know if further adjustments are needed.

PR #677, opened 2022-02-22
use burrunan/gradle-cache-action in CI

RBusarow:
This is another attempt to mitigate the random dependency resolution failures during CI builds.

[burrunan/gradle-cache-action](https://github.com/burrunan/gradle-cache-action) basically turns GitHub Actions' allocated Azure storage into a mini remote build cache.  This cache includes the downloaded dependencies, which should mean that we don't have to download them from Maven Central every time.  And if we're not trying to download them from Maven Central, then we can't get failures from Maven Central.

Because this is just freebie caching, the requests are rate-limited by Azure.  I rarely see errors in other projects -- typically only when I'm running a matrix of jobs.  When this rate-limiting happens, the action will just execute the build locally as though it's a cache miss.

### DIFF
--- a/.github/workflows/kotlin.yml
+++ b/.github/workflows/kotlin.yml
@@ -22,15 +22,31 @@ jobs:
           java-version: 11.0.7
 
       ## Actual task
-      - name: Assemble with gradle — make sure everything builds
-        run: ./gradlew assemble --no-daemon --stacktrace
+      - uses : burrunan/gradle-cache-action@v1
+        name : Assemble with gradle — make sure everything builds
+        with :
+          gradle-dependencies-cache-key : |
+            gradle/libs.versions.toml
+          arguments : |
+            assemble --no-daemon --stacktrace
+          concurrent : true
+          gradle-build-scan-report : false
+          gradle-distribution-sha-256-sum-warning : false
 
       # This should ideally be done as a Check job below, but it needs to be done as a separate
       # step after running assemble. Heckin' ridikalus.
       # Probably fixed in dokka 1.4.10, but we can't move to kotlin 1.4 yet.
       #  https://github.com/square/workflow/issues/1152.
-      - name: Run dokka to validate kdoc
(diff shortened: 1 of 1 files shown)


RBusarow:
The first run of CI came up green and the cache was uploaded for the instrumented tests.

> ##[debug]Uploading chunk of size 33554432 bytes at offset 0 with content range: bytes 0-33554431/*
##[debug]Uploading chunk of size 33554432 bytes at offset 33554432 with content range: bytes 33554432-67108863/*
##[debug]Uploading chunk of size 33554432 bytes at offset 67108864 with content range: bytes 67108864-[100](https://github.com/square/workflow-kotlin/runs/5292745143?check_suite_focus=true#step:10:100)663295/*
##[debug]Uploading chunk of size 33554432 bytes at offset 100663296 with content range: bytes 100663296-134217727/*
##[debug]Uploading chunk of size 33554432 bytes at offset 134217728 with content range: bytes 134217728-167772159/*
##[debug]Uploading chunk of size 33554432 bytes at offset 167772160 with content range: bytes 167772160-201326591/*
##[debug]Uploading chunk of size 33554432 bytes at offset 201326592 with content range: bytes 201326592-23488[102](https://github.com/square/workflow-kotlin/runs/5292745143?check_suite_focus=true#step:10:102)3/*
##[debug]Uploading chunk of size 33554432 bytes at offset 234881024 with content range: bytes 234881024-268435455/*
##[debug]Uploading chunk of size 9330308 bytes at offset 268435456 with content range: bytes 268435456-277765763/*
##[debug]Commiting cache

https://github.com/square/workflow-kotlin/runs/5292745143?check_suite_focus=true

Going to re-kick now and see if the cache is used.

steve-the-edwards (approved):
Amazing - so cool @RBusarow

RBusarow:
Another minor win here is that any continued resolution failures will happen sooner.

Previously, we'd build the artifacts for one app, then test, then build for the next, then test, etc.

With this version, we build everything first. 👍

PR #680, opened 2022-02-22
Merge remote-tracking branch 'origin/main' into ray/ui-update

rjrjr:
The big change here was the introduction of `WorkflowSavedStateRegistryAggregator`. We re-apply it to `BackStackContainer` (moved from its old location so git didn't do it for us), and update the new dialog support to use it. The latter means copying the code from `ModalContainer`, and spreading it across `BodyAndModalsContainer`, `LayerDialogs`, and `DialogHolder`.

Recall that `BodyAndModalsContainer` is a `FrameLayout` that delegates most of its work to `LayerDialogs`, which in turn has a collection of `DialogHolder` instances wrapping each `Dialog` it manages. `DialogHolder` is analagous to `ModalContainer.DialogRef` in the old code.

```
* origin/main: (22 commits)
  use gradle-cache-action in CI
  Truth in advertising: findParentLifecycle cannot return null
  Added comments.
  Created ParcelableTextController.
  Update tutorial workflow_version to 1.6.0
  Finish releasing v1.6.0.
  Releasing v1.6.0.
  Tightens up the API related to ViewTreeSavedStateRegistryOwner.
  Better use of Compatible for ease of container uniquing.
  ModalContainer provides SavedStateRegistries to each of its layers.
  revert Espresso version to 3.3.0
  Drops com.android.tools:desugar_jdk_libs
  Drops AVD 21, bumps 27 to 29
  misc cleanup
  add definitions for macro-benchmark and profile installer
  update the api file
  add viewmodel-ktx as a dependency in ui-core-android
  update Jmh to 1.34
  introduce `IdlingDispatcher` and `IdlingDispatcherRule`
  661: Save JMH Benchmark Results
  ...
```


workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/DialogHolder.kt
```diff
@@ -28,10 +29,12 @@ import com.squareup.workflow1.ui.compatible
 internal class DialogHolder<T : Overlay>(
   initialRendering: T,
   initialViewEnvironment: ViewEnvironment,
+  index: Int,
   private val modal: Boolean,
   private val context: Context,
   private val factory: OverlayDialogFactory<T>
 ) {
+  val savedStateRegistryKey = Compatible.keyFor(initialRendering, index.toString())
```
rjrjr, line 37: This is like the field of the same name in `ModalContainer.DialogRef`.

workflow-ui/core-android/src/test/java/com/squareup/workflow1/ui/androidx/WorkflowSavedStateRegistryAggregatorTest.kt
```diff
@@ -0,0 +1,715 @@
+<<<<<<<< HEAD:workflow-ui/core-android/src/test/java/com/squareup/workflow1/ui/container/StateRegistryAggregatorTest.kt
```
rjrjr: God damn it.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/LayeredDialogs.kt
```diff
@@ -34,7 +36,21 @@ public class LayeredDialogs(
   public constructor(
     view: View,
     modal: Boolean
-  ) : this(view.context, modal, { WorkflowLifecycleOwner.get(view) })
+  ) : this(
+    context = view.context,
+    modal = modal,
+    getParentLifecycleOwner = {
+      checkNotNull(WorkflowLifecycleOwner.get(view)) {
```
rjrjr, line 43: It doesn't really have to be a `WorkflowLifecycleOwner`, use the androidx call instead.

rjrjr:
Well hell, flaky test. This was green on the last run, and I don't see how the last tweak could have broken it.

```
com.squareup.sample.TicTacToeEspressoTest > canGoBackInModalViewAndSeeRestoredViewState[test(AVD) - 10] FAILED 
	androidx.test.espresso.NoMatchingViewException: No views in hierarchy found matching: with id is <com.squareup.sample.tictacworkflow:id/second_factor>
```

rjrjr:
Uh oh, this is a bad surprise:

```
com.squareup.sample.TicTacToeEspressoTest > dialogSurvivesConfigChange[test(AVD) - 10] FAILED 
	java.lang.OutOfMemoryError: Failed to allocate a 1335864 byte allocation with 705056 free bytes and 688KB until OOM, target footprint 16777216, growth limit 16777216
	at shark.internal.UnsortedByteEntries.append(UnsortedByteEntries.kt:32)
```

RBusarow (approved):

PR #678, opened 2022-02-22
auto-publish a snapshot after pushing to `main` or `ray/ui-update`

RBusarow:
Note @rjrjr that we'll need to change the ui branch's version to something like `1.7.0-ui-SNAPSHOT` in order to avoid overwrites.

PR #682, opened 2022-02-23
use `--no-daemon` for both Gradle commands in `publish-snapshot` action

RBusarow:
see https://github.com/square/workflow-kotlin/pull/678#discussion_r812456381

### DIFF
--- a/.github/workflows/publish-snapshot.yml
+++ b/.github/workflows/publish-snapshot.yml
@@ -22,7 +22,9 @@ jobs:
           java-version: 11.0.7
 
       - name : Publish Snapshots
-        run: ./gradlew clean build && ./gradlew publish --no-parallel --no-daemon
+        run: |
+          ./gradlew clean build --no-daemon
+          ./gradlew publish --no-parallel --no-daemon
         env :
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-snapshot.yml
```diff
@@ -22,7 +22,9 @@ jobs:
           java-version: 11.0.7
 
       - name : Publish Snapshots
-        run: ./gradlew clean build && ./gradlew publish --no-parallel --no-daemon
+        run: |
+          ./gradlew clean build --no-daemon
+          ./gradlew publish --no-parallel --no-daemon
```
rjrjr, line 27: Can we count on the publish call being prevented w/o the &&?

RBusarow: Yep.  The different lines are executed sequentially.  Sadly, the [documentation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun) isn't very clear about this, but that's how it works.

steve-the-edwards (approved):

PR #684, opened 2022-02-25
Fixes view structure assertion in OverviewDetailContainer

rjrjr:


gradle/libs.versions.toml
```diff
@@ -57,7 +57,7 @@ mockito-core = "3.3.3"
 mockito-kotlin = "3.2.0"
 
 mockk = "1.11.0"
-robolectric = "4.5.1"
+robolectric = "4.6.1"
```
rjrjr, line 60: Didn't really need this, but might as well.

samples/containers/android/build.gradle.kts
```diff
@@ -24,7 +30,11 @@ dependencies {
   implementation(libs.kotlinx.coroutines.android)
   implementation(libs.kotlinx.coroutines.core)
 
-  testImplementation(libs.junit)
-  testImplementation(libs.truth)
-  testImplementation(libs.kotlinx.coroutines.test)
+  androidTestImplementation(libs.androidx.activity.core)
+  androidTestImplementation(libs.androidx.activity.core)
```
RBusarow, line 34: Is "dual activity" the next hot trend?

rjrjr: More is better, everyone knows that.

rjrjr:
Amazing. The inadvertent bump from Robolectric 4.5.1 to 4.6.1 broke `IdlingDIspatcherTest`. We should really try to eliminate Robolectric from our deps.

```
com.squareup.workflow1.ui.internal.test.IdlingDIspatcherTest > should return to idle if a coroutine is cancelled before completion FAILED
    java.lang.ExceptionInInitializerError at IdlingDIspatcherTest.kt:45
        Caused by: java.lang.RuntimeException at IdlingDIspatcherTest.kt:45
            Caused by: java.lang.reflect.InvocationTargetException at IdlingDIspatcherTest.kt:45
                Caused by: java.lang.NullPointerException at IdlingDIspatcherTest.kt:45
```

RBusarow (approved):

PR #687, opened 2022-02-26
Compose overhaul WIP, is it "too pure"?

rjrjr:
_This is mostly a note to Monday me, with extra formatting effort so that it can serve as the basis of a commit message some day soon._

The existing Compose support is completely tied to `ViewFactory`, which makes sense, because there was nothing else. It works _very_ well.

In this rewrite, I've tried to follow the patterns introduced by `Screen` and `Overlay`, now that `ViewRegistry.Entry` allows us to have multiple kinds of factories:

| Marker Interface | Extension backed by…      | … a ViewRegistry.Entry | Compile time binding interface |
| ---              | ---                       | ---                    | --- |
| `Screen`         | `Screen.buildView`        | `ScreenViewFactory`     | `AndroidScreen` |
| `Overlay`        | `Overlay.toDialogFactory` | `OverlayDialogFactory` | `AndroidOverlay`|

This PR updates that table like so:

| Marker Interface | Extension backed by…      | … a ViewRegistry.Entry | Compile time binding interface |
| ---              | ---                       | ---                    | --- |
| `Screen`         | `Screen.buildView`        | `ScreenViewFactory`     | `AndroidScreen` |
|                  | `Screen.toComposableFactory` | `ScreenComposableFactory`| `ComposeScreen` |
| `Overlay`        | `Overlay.toDialogFactory` | `OverlayDialogFactory` | `AndroidOverlay`|

- There is a new `Screen.toComposableFactory` extension, backed by…

- `ScreenComposableFactory`, a new `ViewRegistry.Entry` that can turn a `Screen` into a `@Composable` function, and…

- `ComposeScreen`, like `AndroidScreen` / `AndroidViewRendering`, an interface implemented by renderings that lets them define their `@Composable fun Content()` at compile time.

I have no idea yet if it works.

I'm concerned that there is going to be a bootstrap problem. With the existing stuff, you're basically forced to create both a `@Composeable` entry point and a `ViewFactory` entry point, because everything `ViewFactory` is the only show in town. (It's not a burden in practice, the View-specific bits all happen for free.) Because of this, we're able to map every rendering into both a `@Composable` and a `View` automatically, regardless of whether its base binding is `View`-based or Compose-based.

In the new world, by separating `ScreenViewFactory` and `ScreenComposableFactory`, I'm losing some of the  every-rendering-provides-both-factories nature.

It's okay in the Compose > Classic direction. `Screen.toComposableFactory` is able to do the trick. It can find an existing `ScreenComposableFactory`, or wrap a `ScreenViewFactory` with one.

But I can't seamlessly go in the other direction, Classic > Compose.  `Screen.buildView` can't find and automatically wrap a `ScreenComposableFactory` with a `ComposeView` unless we do one of the following:

 1. Provide a Compose specific `ScreenViewFactoryFinder`, and require Compose apps to set it up. (`ScreenViewFactoryFinder` is the customizable strategy object that backs `Screen.buildView`.)

 2. Make all of workflow-ui depend on Compose, and build knowledge of it into the default `ScreenViewFactoryFinder`

 3. Forget about the new `ScreenComposableFactory : ViewRegistry.Entry`, and just port the current stuff from `ViewFactory` to `ScreenViewFactory`. Make `CompseScreen` an abstract subclass of `AndroidScreen`.

Option three would be a lot easier, really. It's a little ugly on the implementation side, but I'm hard pressed to see any actual downside.

### DIFF
--- a/workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposeScreen.kt
+++ b/workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposeScreen.kt
@@ -1,0 +1,26 @@
+package com.squareup.workflow1.ui.compose
+
+import androidx.compose.runtime.Composable
+import com.squareup.workflow1.ui.Screen
+import com.squareup.workflow1.ui.ViewEnvironment
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+
+/**
+ * Interface implemented by a rendering class to allow it to drive an Android UI
+ * via an appropriate `@Composable` [Content] function.
+ */
+@WorkflowUiExperimentalApi
+public interface ComposeScreen<S : ComposeScreen<S>> : Screen {
+
+  /**
+   * The composable content of this rendering. This method will be called
+   * any time a new rendering is emitted, or the [viewEnvironment]
+   * changes.
+   *
+   * This function will either serve as the root of a
+   * [ComposeView][androidx.compose.ui.platform.ComposeView], or else
+   * be called as the child of a [Box][androidx.compose.foundation.layout.Box]
+   */
+  @Suppress("FunctionName")
+  @Composable public fun Content(viewEnvironment: ViewEnvironment)
+}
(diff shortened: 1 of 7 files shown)


rjrjr:
So what bothers me about this? I guess the fact that Compose UI and classic are so permanently intertwined. But…they are in real life too, so I should just get over that and stop creating work for myself.

PR #697, opened 2022-03-11
introduce `artifacts-check` plugin

RBusarow:
Have you ever copy/pasted a module in order to create a new one, but forgot to update the artifact ID in `gradle.properties`? 🖐️ 

Have you ever published from a new machine, then realized (too late) that the library now targets a different JDK level? 🖐️ 

Have you ever migrated from one publishing plugin/script to another without any way to be sure the same artifacts are going to the same places?

This plugin is similar to [Kotlinx binary compatibility plugin](https://github.com/Kotlin/binary-compatibility-validator), except it records the particulars of each Maven artifact instead of the api surface.  This is done at the root project level, so that it can look at everything at once in order to find duplicate artifact IDs or descriptions.

The **`artifactsDump`** task is just like `apiDump`.  It should be re-run any time there's an intentional change.  Its output is `artifacts.json` in the project root, which is added to git.

The **`artifactsCheck`** task is similar to `apiCheck`.  It parses the current state of things and compares that to the canonical `artifacts.json`.  The task fails if there are any changes, with an output something like this:
```
> Task :artifactsCheck FAILED
   Artifact definitions don't match.  If this is intended, run `./gradlew artifactsDump` and commit changes.

        This artifact is changed:

            :trace-encoder -
                        old packaging - aar
                        new packaging - jar
```

`artifactsCheck` is hooked into the basic `check` task by convention.  We don't use `check` in CI, but if we did it would automatically include this new task.  

`artifactsCheck` will also automatically run before `publish` or `publishToMavenLocal`.

Finally, I'm adding this _before_ replacing the existing [publishing groovy script](https://github.com/square/workflow-kotlin/blob/9f6a2d2705fa4c22626a4e74141a329901109a2b/.buildscript/configure-maven-publish.gradle) with a convention plugin.  The new plugin may use the new [Base plugin](https://github.com/vanniktech/gradle-maven-publish-plugin#base-plugin) which eliminates the need for a `gradle.properties` in each module.  This `artifacts-check` plugin will ensure that we don't suddenly start shipping compose-ui with the testing-utils artifact ID.

### DIFF
--- a/.github/workflows/kotlin.yml
+++ b/.github/workflows/kotlin.yml
@@ -78,7 +78,7 @@ jobs:
           gradle-dependencies-cache-key : |
             gradle/libs.versions.toml
           arguments : |
-            test apiCheck lint ktlintCheck jmhJar --no-daemon --stacktrace --continue
+            test apiCheck artifactsCheck lint ktlintCheck jmhJar --no-daemon --stacktrace --continue
           concurrent : true
           gradle-build-scan-report : false
           gradle-distribution-sha-256-sum-warning : false
--- a/artifacts.json
+++ b/artifacts.json
@@ -1,0 +1,114 @@
+[
+  {
+    "gradlePath": ":internal-testing-utils",
+    "group": "com.squareup.workflow1",
+    "artifactId": "workflow-internal-testing-utils",
+    "description": "Workflow internal testing utilities",
+    "packaging": "jar",
+    "javaVersion": "1.8"
+  },
+  {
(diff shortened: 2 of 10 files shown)


steve-the-edwards (approved):
Nice. did you just decide to write this?

RBusarow:
> Nice. did you just decide to write this?

Thank you. :)

I've wanted to write a plugin like this for a while, so this was kind of a fun little proof-of-concept side project.

rjrjr (approved):

PR #701, opened 2022-03-15
Allow repeatOnLifecycle state to be configurable.

c-soon:
While most workflow use cases is likely to be inside an Activity and can specify renderings
to be updated while in the Lifecycle.State.STARTED state, there are certain use cases
where we want the workflow to depend on the application's lifecycle (using ProcessLifecycleOwner.)
which will not work when calling repeatOnLifecycle with Lifecycle.State.STARTED as this state event
is tied to Activities.

### DIFF
--- a/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLayout.kt
+++ b/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLayout.kt
@@ -73,15 +73,18 @@ public class WorkflowLayout(
    *
    * @param [lifecycle] the lifecycle that defines when and how this view should be updated.
    * Typically this comes from `ComponentActivity.lifecycle` or  `Fragment.lifecycle`.
+   * @param [repeatOnLifecycle] the lifecycle state in which renderings should be actively
+   * updated. Defaults to STARTED, which is appropriate for Activity and Fragment.
    */
   public fun start(
     lifecycle: Lifecycle,
     renderings: Flow<Any>,
+    repeatOnLifecycle: Lifecycle.State = Lifecycle.State.STARTED,
     environment: ViewEnvironment = ViewEnvironment()
   ) {
     // Just like https://medium.com/androiddevelopers/a-safer-way-to-collect-flows-from-android-uis-23080b1f8bda
     lifecycle.coroutineScope.launch {
-      lifecycle.repeatOnLifecycle(Lifecycle.State.STARTED) {
+      lifecycle.repeatOnLifecycle(repeatOnLifecycle) {
         renderings.collect { update(it, environment) }
       }
     }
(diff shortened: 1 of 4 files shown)


workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLayout.kt
```diff
@@ -73,15 +73,18 @@ public class WorkflowLayout(
    *
    * @param [lifecycle] the lifecycle that defines when and how this view should be updated.
    * Typically this comes from `ComponentActivity.lifecycle` or  `Fragment.lifecycle`.
+   * @param [repeatOnLifecycle] the lifecycle state in which renderings should be actively
```
rjrjr, line 76:
```suggestion
   * @param [repeatOnLifecycle] the lifecycle state in which renderings should be actively. Defaults to STARTED, which is appropriate for Activity and Fragment.
```

rjrjr (approved):
You'll need to update the api files too.

```
./gradlew workflow-ui:core-android:apiDump
```

rjrjr:
Thanks!

c-soon:
> You'll need to update the api files too.
> 
> ```
> ./gradlew workflow-ui:core-android:apiDump
> ```

Oops, thank you! Updated!

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLayout.kt
c-soon: Updated.

c-soon: updated!

c-soon:
@rjrjr fyi this looks like its pending a maintainer to approve something? 

```First-time contributors need a maintainer to approve running workflows.```

rjrjr:
> First-time contributors need a maintainer to approve running workflows.

That already happened, it's what's allowing CI to run.

rjrjr:
The CI failures do look bogus, though. Trying again…

c-soon:
Looks like it executed now, for some reason the 3 failed executions didn't run before, I'll address the failures i think i need to update the usages in the samples to named params

rjrjr:
Oh, of course. Thanks.

steve-the-edwards:
I approved them to run - thats why they started running. Sorry for the confusion!

c-soon:
> I approved them to run - thats why they started running. Sorry for the confusion!

Ah, thank you! Does this need to be done every time I force push an update? I see the same message again.

rjrjr:
> > I approved them to run - thats why they started running. Sorry for the confusion!
> 
> Ah, thank you! Does this need to be done every time I force push an update? I see the same message again.

That's weird, I think a recent development. @RBusarow, does this have anything to do with the new action to kill off previous runs? Any clue to share?

RBusarow:
I believe this is a relatively new setting.

image

It's meant to prevent people/bots from PRing malicious actions and having them auto-execute.  I guess you technically don't count as a contributor until the first PR's merged.

rjrjr:
> It's meant to prevent people/bots from PRing malicious actions and having them auto-execute. I guess you technically don't count as a contributor until the first PR's merged.

Gross. I guess it's been this way for a while, then. I never noticed before that I have to approve over and over again.

We had to turn this on after attracting some bitcoin creeps who messed with our actions. :/

PR #591, opened 2021-11-19
UI Update

rjrjr:
Long living branch to collect the changes inspired by #195. Individual PRs are being reviewed and merged here. This will not be merged into main until that work is complete.

PR #703, opened 2022-03-15
Introduces ScreenViewFactory.updateView, drastically reduces cleverness.

### LINKED ISSUE #413: DecorativeViewFactory should support wrapping Context

### LINKED ISSUE #551: ViewEnvironment argument of DecorativeViewFactory.map is redundant
Nothing you can do there that you can't do with `initializeView`. Get rid of it and simplify things a bit.

### LINKED ISSUE #598: Either hide ShowRenderingTag or don't make it a data class

### LINKED ISSUE #626: Replace DecorativeScreenViewFactory with a builder
It should be possible to build something like transforms suggested by #606 (spiked in #621) as extensions on `ScreenViewFactory`. Then we can either get rid of that thing or make it private.

rjrjr:
The legacy workflow-ui world is built around a set of extension methods on `View`, backed by a `WorkflowViewState` object hanging off of a tag. This is overly clever, needlessly throws away a good bit of compile time safety, and makes it much harder than it should be to implement wrapper renderings.

In particular, every `ViewFactory` is required to call `View.bindShowRendering` to put all this machinery in place. When a screen rendering wraps another, with the intention of delegating to its registered `ViewFactory`, that requires creating a wrapper `ViewFactory` that makes a new `bindShowRendering` call replacing and wrapping the `WorkflowViewState` put in place by the delegate. That's so tricky it required the introduction of `DecorativeViewFactory`, which is still pretty tricky to use.

To solve all this, we do two things: 

- deprecate all the existing `View` extension methods and abandon `WorkflowViewState`, replacing it with `ScreenViewHolder`
- add an `updateView(View)` method to `ScreenViewFactory`

`ScreenViewHolder` is inverted from `WorkflowViewState` -- instead of belonging to the `View`, nestled in a tag, the `View` belongs to it. Most of the old `View` extension methods are replaced by methods on `ScreenViewHolder`. It is very similar to `DialogHolder`, but is part of the public API.

The change to `ScreenViewFactory`, giving it an `update` method, makes it much more like `OverlayDialogFactory`. I toyed a bit with the idea of trying to unify `ScreenViewFactory` / `OverlayDialogFactory` and `ScreenViewHolder` / `DialogHolder`, but it didn't seem worth the added complexity, especially around parameter types.

Reviewers should also note that the **`buildView` method no longer has access to the rendering**, a choice I'm still second guessing. So far I haven't hit a use case that required it, beyond a couple of changes to expected strings in `BackStackContainerPersistenceTest`. Perhaps that's reason enough to bring it back?

It is also worth calling out that `ScreenViewFactory` still is not involved in the `canShow` path. That was a very conscious choice. Making that call dynamic would add a lot of flexibility, but we've never needed that flexibility. The downside of adding it is that we would lose our 1:1 relationship between factories and rendering types, and suddenly would need a resolution mechanism. The cost benefit analysis there seems clear to me.

One of the deprecated `View` extensions did not move to `ScreenViewHolder`: `View.Start` is replaced by `ScreenViewFactory.start`. This extension method is what calls `ScreenViewFactory.buildView`. It creates and returns a `ScreenViewHolder` that wraps the new view and accept updates for it, which it does by calling `ScreenViewFactory.update`, as you'd expect. `ScreenViewFactory.start` is also the new home for the `ViewStarter` mechanism that is used by `WorkflowViewStub` and other containers to call `WorkflowLifecycleOwner.installOn`.

We are not completely out of the `View.setTag` business. `ScreenViewFactory` implementations must be stateless -- it is common for separate instances to field calls to `buildView` and `updateView`, or for a single factory instance to build and update multiple views. `View.setTag` is the most practical way for `buildView` to set up state that `updateView` can rely on. This technique is used as an implementation detail of several standard `ScreenViewFactory` implementations, but is never a public concern. In particular, the factories built by the various `ScreenViewRunner.bind` methods rely on this technique to keep a `View` paired with its runner.

And `ScreenViewRunner` is still very much needed under the new scheme, so that we don't force every implementation of `ScreenViewFactory` to solve this problem of maintaining per-view state themselves. But `ScreenViewRunner` is now a convenience, rather than a fundamental mechanism.

By making `ScreenViewFactory` responsible for both building and updating, it becomes trivial for one factory to delegate to another -- the update method no longer lives in a fragile `View` tag, and so no longer needs to be replaced. A wrapper factory can just delegate to another one. This means there is no longer a problem to be solved by `DecorativeScreenViewFactory`, so that class is deleted. `ManualScreenViewFactory` is also gone, replaced by a `bindBuiltView()` function that returns `Pair<View, ScreenViewRunner>`.

The `ViewStarter` mechanism is also much simpler now, also no longer relying on any state stored in `View` tags. The implementation of `ScreenViewFactory.start()` remains a bit complex, but only because of the work it has to do to keep the legacy `View.start()` method working, in partnership with `AsScreenViewFactory()`. Once we delete the deprecated View code, most of the code in both of those can be deleted.

Fixes #413, fixes #598, fixes #626, fixes #551, fixes #443.

For an illustration of how all this fits together, here's a sequence diagram of two calls to `WorkflowViewStub.show`.

WorkflowViewStub show

```
# https://sequencediagram.org/

title WorkflowViewStub.show()

participant "WorkflowViewStub" as stub
participant Screen
participant "ViewEnvironment" as env
participant "ScreenViewFactoryFinder" as finder
participant "ViewRegistry" as registry
participant "ScreenViewFactory" as factory
participant "ScreenViewHolder" as holder

[->stub:show(screen, env)

stub->Screen:toView(env)
activate Screen
Screen->Screen:toViewFactory(env)
activate Screen
Screen->env:get(ScreenViewFactoryFinder)
Screen<--env:finder
Screen->finder:getViewFactoryForRendering(this)
activate finder
finder->registry:getEntryFor(screen)
finder<--registry:factory
Screen<--finder:factory
deactivate finder
deactivate Screen

Screen->factory:start(this, env)
activate factory
factory->factory:buildView()
factory->holder:init(this, view, screen, env)
factory<--holder:
factory->factory:updateView(view, screen, env)
Screen<--factory:holder
deactivate factory

stub<--Screen:holder
deactivate Screen


[->stub:show(screen, env)
activate stub
stub->holder:canShow(screen)
activate holder
holder->holder: compatible(screen, this.screen)
stub<--holder:true
deactivate holder
stub->holder:show(screen, env)
activate holder
holder->factory: update(this.view, this.screen, this.env)
deactivate holder
deactivate stub
```


rjrjr:
For an illustration of how all this fits together, here's a sequence diagram of two calls to `WorkflowViewStub.show`.

WorkflowViewStub show

```
# https://sequencediagram.org/

title WorkflowViewStub.show()

participant "WorkflowViewStub" as stub
participant Screen
participant "ViewEnvironment" as env
participant "ScreenViewFactoryFinder" as finder
participant "ViewRegistry" as registry
participant "ScreenViewFactory" as factory
participant "ScreenViewHolder" as holder

[->stub:show(screen, env)

stub->Screen:toView(env)
activate Screen
Screen->Screen:toViewFactory(env)
activate Screen
Screen->env:get(ScreenViewFactoryFinder)
Screen<--env:finder
Screen->finder:getViewFactoryForRendering(this)
activate finder
finder->registry:getViewFactoryForRendering(screen)
finder<--registry:factory
Screen<--finder:factory
deactivate finder
deactivate Screen

Screen->factory:start(this, env)
activate factory
factory->factory:buildView()
factory->holder:init(this, view, screen, env)
factory<--holder:
factory->factory:updateView(view, screen, env)
Screen<--factory:holder
deactivate factory

stub<--Screen:holder
deactivate Screen


[->stub:show(screen, env)
activate stub
stub->holder:canShow(screen)
activate holder
holder->holder: compatible(screen, this.screen)
stub<--holder:true
deactivate holder
stub->holder:show(screen, env)
activate holder
holder->factory: update(this.view, this.screen, this.env)
deactivate holder
deactivate stub
```

workflow-ui/core-android/src/androidTest/java/com/squareup/workflow1/ui/container/BackStackContainerPersistenceTest.kt
```diff
@@ -93,7 +93,7 @@ internal class BackStackContainerPersistenceTest {
     scenario.onActivity {
       assertThat(it.consumeLifecycleEvents())
         .containsExactly(
-          "nested onViewCreated viewState=",
+          "onViewCreated viewState=",
```
vgonda, line 96: Can you add a note on the reason for this change? It's not clear to me.

workflow-ui/core-android/src/androidTest/java/com/squareup/workflow1/ui/container/ViewStateCacheTest.kt
```diff
@@ -172,13 +167,26 @@ internal class ViewStateCacheTest {
     }
   }
 
+  private val fakeFactory = ScreenViewFactory<NamedScreen<*>>(
+    { _, _, _ -> TODO() }, { _, _, _ -> }
```
vgonda, line 171: Hehe this looks silly and I love it

vgonda (commented):
I made it about halfway through this morning (including the main compose stuff) and I'll continue tomorrow.

Everything looks good to me, so far! Feels a little bit like a rubber stamp, but I'm understanding more and more each time I look at stuff. Really, I should just find a task to pair on so I can really wrap my head around things.

workflow-ui/core-android/src/androidTest/java/com/squareup/workflow1/ui/container/BackStackContainerPersistenceTest.kt
rjrjr: `ScreenViewFactory.buildView` doesn't have access to the rendering. That's actually a pretty big change that I should call out more explicitly. My reasoning was that I should make the method as small as possible, and try not to add it back unless I bumped into an existing case that couldn't be accomodated. So far I haven't (outside of this test).

I'm still a bit uncertain about the choice. I like the fact that it removes any ambiguity about whether `buildView` is expected to call `updateView` itself -- can't do that if you can't see the rendering.

workflow-ui/core-android/src/androidTest/java/com/squareup/workflow1/ui/container/ViewStateCacheTest.kt
rjrjr: Oops, I meant to make that a bit more explicit. :)

workflow-ui/core-android/src/androidTest/java/com/squareup/workflow1/ui/container/BackStackContainerPersistenceTest.kt
rjrjr: I've updated the commit message to call this out, and also to discuss why `ScreenViewFactory` does not get to answer `canShow`.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewFactory.kt
```diff
@@ -5,85 +5,162 @@ import android.view.View
...
-        "ScreenViewFactory that created it."
-    }
-    viewStarter?.let { givenStarter ->
-      val doStart = view.starter
-      view.starter = { newView ->
-        givenStarter.startView(newView) { doStart.invoke(newView) }
+    val legacyStarter: ((View) -> Unit)? = holder.view.starterOrNull
+
+    if (legacyStarter != null) {
+      var shown = false
```
RBusarow, line 96: nit:  This variable is duplicated in the `else` block and could be pulled up out of the `if` statement.

rjrjr: That was actually intentional, b/c I wanted to customize the exception message.

rjrjr: Like, the duplication felt cleaner than letting each `shown` bleed into a wider scope.

RBusarow: Got it.

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/Screen.kt
```diff
@@ -2,6 +2,19 @@ package com.squareup.workflow1.ui
...
+ *
+ * It is expected that view systems will make the rendering driving a view component
+ * available in the [ViewEnvironment], as the value for the [Screen] key. In the
+ * case of wrapper renderings like [NamedScreen], the [ViewEnvironment] entry should
+ * include all wrappers.
  */
 @WorkflowUiExperimentalApi
-public interface Screen
+public interface Screen {
+  public companion object : ViewEnvironmentKey<Screen>(Screen::class) {
```
RBusarow, line 13: Now that I've seen this change, it makes a lot of sense.  It's very clever!

RBusarow (approved):
This is still a bit of a rubber stamp from me as well, but it looks great!

rjrjr:
Chatting w/@helios175, and we think it was a mistake to take `rendering` away from buildView, will add it back.

rjrjr:
More from @helios175 chat: really feels like we have one more moving part than we should, `ScreenViewRunner` and `ScreenViewHolder` are kind of redundant. Also ugly that `ScreenViewRunner`, when it exists, is still in harm's way in a View tag.

Can we add another type param to `ScreenViewFactory` that embodies the custom data that Runner passes around w/o screwing up usability? Worth trying.

rjrjr:
And a nit: probably `Screen.toViewHolder(): ScreenViewHolder` is a better name than `Screen.toView(): ScreenViewHolder()`

rjrjr:
> Can we add another type param to ScreenViewFactory that embodies the custom data that Runner passes around w/o screwing up usability? Worth trying.

So far, yeah, no. The driving use case for this work is to make things like the following easy, and that's lost as soon as `ScreenViewFactory` tries to parameterize the stuff that comes out of `buildView` and goes into `updateView`. I'll sleep on it, but I'm not hopeful.

```kotlin
internal fun NamedScreenViewFactory(
  initialRendering: NamedScreen<*>,
  initialViewEnvironment: ViewEnvironment
): ScreenViewFactory<NamedScreen<*>> {
  val wrappedFactory = initialRendering.wrapped.toViewFactory(initialViewEnvironment)

  return ScreenViewFactory(
    buildView = wrappedFactory::buildView,
    updateView = { view, rendering, environment ->
      wrappedFactory.updateView(view, rendering.wrapped, environment)
    }
  )
}
```

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/EnvironmentScreenViewFactory.kt
```diff
@@ -1,18 +1,24 @@
...
+  initialViewEnvironment: ViewEnvironment
+): ScreenViewFactory<EnvironmentScreen<*>> {
+  val realFactory = initialRendering.screen.toViewFactory(
+    initialViewEnvironment merge initialRendering.viewEnvironment
+  )
+
+  return ScreenViewFactory(
+    buildView = realFactory::buildView,
+    updateView = { view, rendering, environment ->
+      realFactory.updateView(view, rendering.screen, environment)
```
rjrjr, line 21: OMG this is totally broken. Clearly need to add some tests.

rjrjr:
I awake COMPLETELY NERD SNIPED by this question. Plan is to:

* Fix the egregious bug introduced in `EnvironmentScreenViewFactory` and add the test that should have caught it
* Merge what we have because it's approved and clearly shippable
* Timebox exploring improvements

rjrjr:
Well cool. That missing test for `EnvironmentScreenViewFactory` reveals that this PR is fundamentally flawed: it is not possible for `EnvironmentScreen.environment` to reach `ScreenViewHolder.environment`.

Back to the drawing board.

rjrjr:
Writing the missing test against #591 branch and learning that `EnvironmentScreen` doesn't work very well. :D

rjrjr:
I finally have a version of this that fixes `EnvironmentScreenAndroidIntegrationTest`. It's different enough that I'm going to close this PR and float a new one.

PR #704, opened 2022-03-17
699: Add simulated performance config to Poetry

steve-the-edwards:


samples/containers/poetry/src/main/java/com/squareup/sample/poetry/views/MayBeLoadingScreen.kt
```diff
@@ -0,0 +1,20 @@
+package com.squareup.sample.poetry.views
+
+import com.squareup.sample.container.overviewdetail.OverviewDetailScreen
+import com.squareup.sample.container.panel.ScrimContainerScreen
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+import com.squareup.workflow1.ui.modal.HasModals
+
+@OptIn(WorkflowUiExperimentalApi::class)
+class MayBeLoadingScreen(
```
RBusarow, line 9: nit: `MaybeLoadingScreen`

steve-the-edwards: Haha thats a great point - I wasn't even thinking of this in that descriptive sense but more of the prescriptive. This 'may' 'be loading'. English is funny :).

samples/containers/poetry/src/main/java/com/squareup/sample/poetry/PoemWorkflow.kt
```diff
@@ -1,111 +1,274 @@
...
  * Renders a [Poem] as a [OverviewDetailScreen], whose overview is a [StanzaListRendering]
  * for the poem, and whose detail traverses through [StanzaRendering]s.
  */
-object PoemWorkflow : StatefulWorkflow<Poem, SelectedStanza, ClosePoem, OverviewDetailScreen>() {
-  object ClosePoem
-  public const val NO_SELECTED_STANZA = -1
+object PoemWorkflow : StatefulWorkflow<Config, State, Output, PoemRenderingWithLoading>() {
+  const val NO_SELECTED_STANZA = -1
+
+  data class Config(
```
rjrjr, line 49: I'm not happy with all this complexity leaking into the poetry sample itself. This makes it much less useful as an example of how to deal with complex composition. If there's no better alternative, I'd rather see it forked.

rjrjr: Perhaps you could take a manual DI approach? e.g. Make `PoemWorkflow` an interface with the original code as its default implementation, and give `PoemActivity` a component interface like `TicTacToeComponent`.

Need to be thoughtful about whether that too would muddy the example too much. At least it would be more similar to how a production app would be built.

samples/containers/poetry/src/main/java/com/squareup/sample/poetry/PoemsBrowserWorkflow.kt
```diff
@@ -2,54 +2,219 @@ package com.squareup.sample.poetry
...
+  }
 
   override fun initialState(
-    props: List<Poem>,
+    props: Config,
     snapshot: Snapshot?
-  ): SelectedPoem {
-    return snapshot?.bytes?.parse { source -> source.readInt() }
-      ?: NO_POEM_SELECTED
+  ): State {
```
rjrjr, line 53: If you do choose to go the DI route, the benchmark implementation could just return a null `Snapshot`.

samples/containers/poetry/src/main/java/com/squareup/sample/poetry/PoemsBrowserWorkflow.kt
```diff
@@ -2,54 +2,219 @@ package com.squareup.sample.poetry
...
     context: RenderContext
-  ): OverviewDetailScreen {
-    val poems: OverviewDetailScreen =
-      context.renderChild(PoemListWorkflow, renderProps) { selected -> choosePoem(selected) }
-        .copy(selection = renderState)
-        .let { OverviewDetailScreen(BackStackScreen(it)) }
-
-    return if (renderState == NO_POEM_SELECTED) {
-      poems
+  ): MayBeLoadingScreen {
```
rjrjr, line 73: The loading screen is another example of my concern about this sample becoming too confusing to be useful as such. I'd like it to stay focused on BackStackScreen and OverviewDetailScreen.

samples/containers/poetry/src/main/java/com/squareup/sample/poetry/PoemWorkflow.kt
rjrjr: Another nice effect of the DI approach is that it paves the way to multiple scenarios in the future, w/o requiring more and more branching here.

samples/containers/poetry/src/main/java/com/squareup/sample/poetry/SimulatedPerfConfig.kt
```diff
@@ -0,0 +1,19 @@
+package com.squareup.sample.poetry
+
+data class SimulatedPerfConfig(
```
rjrjr, line 3: kdoc

samples/containers/poetry/src/main/java/com/squareup/sample/poetry/PoemsBrowserWorkflow.kt
rjrjr: It should be possible to add loading to the benchmark version w/o requiring it to leak into the "public API" of the sample. E.g. take a gatekeeper approach:

- the benchmark versions of this workflow injects a relay that controls whether the loading screen is showing
- an outer-most workflow used only by the benchmark implementation monitors that relay, and wraps its children's renderings in a `MayBeLoadingScreen`.

steve-the-edwards:
@rjrjr great points. This sacrifices the clarity of the original sample. I will find a better way to abstract and polymorphise.

benchmarks/performancepoetry/src/main/java/com/squareup/benchmarks/performance/poetry/PerformancePoemWorkflow.kt
```diff
@@ -0,0 +1,244 @@
...
+    public data class Selected(val stanzaIndex: Int) : State()
+  }
+
+  @OptIn(WorkflowUiExperimentalApi::class)
+  public data class PoemRenderingWithLoading(
+    val screen: OverviewDetailScreen,
+    // This won't be used if we are reporting through Output.
+    val isLoading: Boolean = false
+  ) : AndroidViewRendering<PoemRenderingWithLoading> {
+    // Note that this ViewFactory does not handle the loading state at all. It just ignores it.
```
rjrjr, line 68: Maybe give a hint about where it _is_ handled?

benchmarks/performancepoetry/src/main/java/com/squareup/benchmarks/performance/poetry/PerformancePoemWorkflow.kt
```diff
@@ -0,0 +1,244 @@
...
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.flow.MutableStateFlow
+
+/**
+ * Renders a [Poem] as a [OverviewDetailScreen], whose overview is a [StanzaListRendering]
+ * for the poem, and whose detail traverses through [StanzaRendering]s.
+ */
+public class PerformancePoemWorkflow(
+  private val simulatedPerfConfig: SimulatedPerfConfig = SimulatedPerfConfig.NO_SIMULATED_PERF,
+  private val isLoading: MutableStateFlow<Boolean>
```
rjrjr, line 48: This would be more than a little sketchy in real life, maybe document what you're up to. I assume `isLoading` can be toggled by both this workflow and its parent?

benchmarks/performancepoetry/src/main/java/com/squareup/benchmarks/performance/poetry/PerformancePoemWorkflow.kt
```diff
@@ -0,0 +1,244 @@
...
+) : PoemWorkflow, StatefulWorkflow<Poem, State, ClosePoem, OverviewDetailScreen>() {
+
+  public sealed class State {
+    // N.B. This state is a smell. We include it as possible in order to be able to config similar
+    // to real-world usage. It should be folded into [NoSelection] at the very least.
```
rjrjr, line 53:
```suggestion
    // N.B. This state is a smell. We include it to be able to mimic smells
    // we encounter in real life. Best practice would be to fold it into
    // [NoSelection] at the very least.
```

benchmarks/performancepoetry/src/main/java/com/squareup/benchmarks/performance/poetry/PerformancePoemWorkflow.kt
```diff
@@ -0,0 +1,244 @@
...
+    context: RenderContext
+  ): OverviewDetailScreen {
+    return when (renderState) {
+      Initializing -> {
+        context.runningWorker(Worker.from { Unit }, "initializing") {
+        action {
+          state = Selected(NO_SELECTED_STANZA)
+        }
+      }
+        isLoading.value = true
```
rjrjr, line 104: OMFG

It'd be good to put a big "DO NOT TRY THIS AT HOME" disclaimer somewhere, calling out how Very Very BAD it is to have a render method with side effects like this.

benchmarks/performancepoetry/src/main/java/com/squareup/benchmarks/performance/poetry/PerformancePoemWorkflow.kt
```diff
@@ -0,0 +1,244 @@
...
+        OverviewDetailScreen(overviewRendering = BackStackScreen(BlankScreen))
+      }
+      else -> {
+        val (stanzaIndex, currentStateIsLoading) = when (renderState) {
+          is ComplexCall -> renderState.payload to true
+          is Selected -> renderState.stanzaIndex to false
```
rjrjr, line 110:
```suggestion
          is ComplexCall -> Pair(renderState.payload,  true)
          is Selected -> Pair(renderState.stanzaIndex, false)
```

Pet peeve, I find `to` confusing outside of Map initialization.

benchmarks/performancepoetry/src/main/java/com/squareup/benchmarks/performance/poetry/PerformancePoemWorkflow.kt
```diff
@@ -0,0 +1,244 @@
...
+    class SelectPrevious(override val simulatedPerfConfig: SimulatedPerfConfig) : Action()
+    class SelectNext(override val simulatedPerfConfig: SimulatedPerfConfig) : Action()
+    class HandleStanzaListOutput(
+      override val simulatedPerfConfig: SimulatedPerfConfig,
+      val selection: Int
+    ) : Action()
+
+    class ExitPoem(override val simulatedPerfConfig: SimulatedPerfConfig) : Action()
+
+    override fun Updater.apply() {
```
rjrjr, line 198: Are you sure you can't move the `isLoading.value =` calls here? The first one could move to `initialState`.

benchmarks/performancepoetry/src/main/java/com/squareup/benchmarks/performance/poetry/PerformancePoemsBrowserWorkflow.kt
```diff
@@ -0,0 +1,145 @@
...
+public class PerformancePoemsBrowserWorkflow(
+  private val simulatedPerfConfig: SimulatedPerfConfig,
+  private val poemWorkflow: PoemWorkflow,
+  private val isLoading: MutableStateFlow<Boolean>
+) :
+  PoemsBrowserWorkflow,
+  StatefulWorkflow<List<Poem>, State, Unit, OverviewDetailScreen>() {
+
+  public sealed class State {
+    // N.B. This state is a smell. We include it as possible in order to be able to config similar
```
rjrjr, line 35: Same verbiage suggestion as above.

benchmarks/performancepoetry/src/main/java/com/squareup/benchmarks/performance/poetry/PerformancePoetryActivity.kt
```diff
@@ -38,14 +41,34 @@ public class PerformancePoetryActivity : AppCompatActivity() {
...
     renderWorkflowIn(
-      workflow = PoemsBrowserWorkflow,
+      workflow = workflow,
       scope = viewModelScope,
-      prop = Poem.allPoems,
       savedStateHandle = savedState
     )
   }
+
+  public class Factory(
```
rjrjr, line 57: I almost suggested assisted inject. 😊

samples/containers/poetry/src/main/java/com/squareup/sample/poetry/PoemWorkflow.kt
```diff
@@ -1,114 +1,10 @@
...
+import com.squareup.workflow1.Workflow
 
-typealias SelectedStanza = Int
-
-/**
- * Renders a [Poem] as a [OverviewDetailScreen], whose overview is a [StanzaListRendering]
- * for the poem, and whose detail traverses through [StanzaRendering]s.
- */
-object PoemWorkflow : StatefulWorkflow<Poem, SelectedStanza, ClosePoem, OverviewDetailScreen>() {
+interface PoemWorkflow : Workflow<Poem, ClosePoem, OverviewDetailScreen> {
```
rjrjr, line 8: Please keep the kdoc. And maybe note that we use the interface to allow other implementations.

samples/containers/poetry/src/main/java/com/squareup/sample/poetry/PoemsBrowserWorkflow.kt
```diff
@@ -1,55 +1,7 @@
...
-    sink.writeInt(state)
-  }
-
-  private fun choosePoem(index: SelectedPoem) = action("goToPoem") {
-    state = index
-  }
-
-  private val clearSelection = choosePoem(NO_POEM_SELECTED)
-}
+interface PoemsBrowserWorkflow : Workflow<List<Poem>, Unit, OverviewDetailScreen>
```
rjrjr, line 7:
```suggestion
/**
 * Provides an overview / detail treatment of a list of [Poem]s.
 *
 * (Defining this as an interface allows us to use other implementations
 * in other contexts -- check out our benchmarking module!)
 */
interface PoemsBrowserWorkflow : Workflow<List<Poem>, Unit, OverviewDetailScreen>
```

samples/containers/poetry/src/main/java/com/squareup/sample/poetry/RealPoemWorkflow.kt
```diff
@@ -0,0 +1,119 @@
...
+import com.squareup.workflow1.parse
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+import com.squareup.workflow1.ui.backstack.BackStackScreen
+import com.squareup.workflow1.ui.backstack.toBackStackScreen
+
+/**
+ * Renders a [Poem] as a [OverviewDetailScreen], whose overview is a [StanzaListRendering]
+ * for the poem, and whose detail traverses through [StanzaRendering]s.
+ */
+class RealPoemWorkflow : PoemWorkflow,
```
rjrjr, line 28:
```suggestion
/**
 * Default implementation of [PoemWorkflow].
 */
class RealPoemWorkflow : PoemWorkflow,
```

samples/containers/poetry/src/main/java/com/squareup/sample/poetry/RealPoemsBrowserWorkflow.kt
```diff
@@ -0,0 +1,59 @@
...
+import com.squareup.workflow1.Snapshot
+import com.squareup.workflow1.StatefulWorkflow
+import com.squareup.workflow1.action
+import com.squareup.workflow1.parse
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+import com.squareup.workflow1.ui.backstack.BackStackScreen
+
+typealias SelectedPoem = Int
+
+class RealPoemsBrowserWorkflow(
```
rjrjr, line 15:
```suggestion
/**
 * Default implementation of [PoemsBrowserWorkflow]. Note the use of
 * a child [PoemWorkflow] to render the selected [Poem], and in particular
 * how the child's overview pane (the list of stanzas) is combined with
 * the list of poems, giving us back stack navigation on both sides of 
 * the splitter.
 */
class RealPoemsBrowserWorkflow(
```

samples/containers/poetry/src/main/java/com/squareup/sample/poetry/SimulatedPerfConfig.kt
rjrjr: Can this move to the benchmark module?

rjrjr (approved):

benchmarks/performancepoetry/src/main/java/com/squareup/benchmarks/performance/poetry/PerformancePoemWorkflow.kt
steve-the-edwards: will move it out to the action.

benchmarks/performancepoetry/src/main/java/com/squareup/benchmarks/performance/poetry/PerformancePoemWorkflow.kt
steve-the-edwards: yes it can. Perhaps I need to step back though if this is smelly to you. I thought this was your suggestion wrt how Gatekeepers would overlay all children via an injected signal?

steve-the-edwards: I can extend `OverviewDetailScreen` to an `MaybeLoadingOverviewDetailScreen` and handle it that way if we think this is preferable and better practice?

benchmarks/performancepoetry/api/performancepoetry.api
```diff
@@ -1,3 +1,145 @@
+public final class com/squareup/benchmarks/performance/poetry/MaybeLoadingGatekeeperWorkflow : com/squareup/workflow1/StatefulWorkflow {
```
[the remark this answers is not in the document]
steve-the-edwards: I just didn't know where to disable it for these modules - but I found it now :)

samples/containers/poetry/src/main/java/com/squareup/sample/poetry/PoemsBrowserWorkflow.kt
steve-the-edwards: This is where I thought you were suggesting the injected loading trigger @rjrjr ? But now it seems like you're having second thoughts about that - should we move it back to the rendering?

benchmarks/performancepoetry/src/main/java/com/squareup/benchmarks/performance/poetry/PerformancePoemWorkflow.kt
rjrjr: I'm just saying that it's sketchy to have `MutableStateFlow` floating around as public API rather than an implementation detail. I'd expect the lower level workflows to be injecting an interface that lets them request the loading UI be shown or hidden, but that doesn't give them access to the stream of loading values.

In real life, we'd tie that to a token in case of overlapping show/hide requests. Probably overkill for this, but just for the record:

```kotlin
interface Loader {
  fun startLoading(token: String)
  fun stopLoading(token: String)
}
```

```kotlin
class RealLoader: Loader {
  private val loaders = mutableSetOf<String>()
  private val _showIt = MutableStateFlow(false)

  public val showIt(): StateFlow<Boolean> = _showIt

  override fun startLoading(token: String) {
    loaders.add(token)
    _showIt.value = true
  }


  override fun stopLoading(token: String) {
    loaders.remove(token)
    _showIt.value = loaders.isNotEmpty()
  }  
}
```

Alos, I don't think we'd want to make `OverviewDetailScreen` non-final.

samples/containers/poetry/src/main/java/com/squareup/sample/poetry/PoemsBrowserWorkflow.kt
rjrjr: I'm saying that the Gatekeeper approach was suggested as a way to keep all knowledge of the loading stuff out of the non-benchmark workflow. Loading gets implemented as a wrapper around `PoemWorkflow`, without its knowledge.

* The root UI of the Benchmark version uses something like `MaybeLoading<PoemWorkflow>`, or `MaybeLoading<*>` as its `RenderingT`. It runs a `PoemWorkflow` as a child.
* `PerformancePoemWorkflow` is in on the joke. It injects the `Loader` service interface I just suggested to make it known when it enters or exits a loading state.

steve-the-edwards: Ack - ok that makes a lot of sense. I will add a note saying that we should wrap it in an interface but that isn't shown here for the sake of simplicity.

steve-the-edwards: - [ ] Add note about smelliness of shared mutable state.

benchmarks/performancepoetry/src/main/java/com/squareup/benchmarks/performance/poetry/PerformancePoemWorkflow.kt
steve-the-edwards: This was dead code that I no longer used.

benchmarks/performancepoetry/src/main/java/com/squareup/benchmarks/performance/poetry/PerformancePoemWorkflow.kt
steve-the-edwards: Added the note to make that clear and hint at this type of a solution.

benchmarks/performancepoetry/src/main/java/com/squareup/benchmarks/performance/poetry/PerformancePoemWorkflow.kt
steve-the-edwards: Actually, moved it into the Worker and action so it would not be run multiple times in render().

benchmarks/performancepoetry/src/main/java/com/squareup/benchmarks/performance/poetry/PerformancePoemWorkflow.kt
```diff
@@ -0,0 +1,235 @@
...
+
+  @OptIn(WorkflowUiExperimentalApi::class)
+  override fun render(
+    renderProps: Poem,
+    renderState: State,
+    context: RenderContext
+  ): OverviewDetailScreen {
+    return when (renderState) {
+      Initializing -> {
+        context.runningWorker(Worker.from { Unit }, "initializing") {
```
rjrjr, line 88:
```suggestion
        // Again, then entire `Initializing` state is a smell, and 
        // `Worker.from { Unit }` is the most odiferous bit.
        // Don't try this at home.
        context.runningWorker(Worker.from { Unit }, "initializing") {
```

benchmarks/performancepoetry/src/main/java/com/squareup/benchmarks/performance/poetry/PerformancePoemWorkflow.kt
```diff
@@ -0,0 +1,235 @@
...
+          is Selected -> Pair(renderState.stanzaIndex, false)
+          Initializing -> throw IllegalStateException("No longer initializing.")
+        }
+
+        if (currentStateIsLoading) {
+          context.runningWorker(
+            Worker.from {
+              isLoading.value = true
+              delay(simulatedPerfConfig.complexityDelay)
+              stanzaIndex
```
rjrjr, line 109: Seems like you're not actually using the `stanzaIndex` value in the action below?

rjrjr (approved):
Nice! Thanks for iterating.

PR #705, opened 2022-03-17
Fixes rapid BackStackContainer.update calls.

rjrjr:
Turns out that `TransitionManager.go()` is a no-op if there is already a transition in progress, so a new screen shown before the previous one was done animating would never be displayed. Even better, a redundant call to display that last screen again would cause a crash in `WorkflowSavedStateRegistryAggregator`, when we would try to register the name already logged for the lost view.


workflow-ui/container-android/src/main/java/com/squareup/workflow1/ui/backstack/BackStackContainer.kt
```diff
@@ -134,6 +134,7 @@ public open class BackStackContainer @JvmOverloads constructor(
           .addTransition(Slide(inEdge).addTarget(newTarget))
           .setInterpolator(AccelerateDecelerateInterpolator())
 
+        TransitionManager.endTransitions(this)
```
RBusarow, line 137: Geez...  This is pretty easy to miss.  Nice catch!

RBusarow (approved):

PR #706, opened 2022-03-18
LegacyComposeViewTreeIntegrationTest is more purely legacy.

rjrjr:
I had cheated a bit on this, wrapping various things so that I could avoid the deprecated methods on WorkflowViewStub. Doing so caused testing difficulties in #703 that seem purely related to the tests themselves, not a reflection of problems in real life.

So we revert the wrapping in `LegacyComposeViewTreeIntegrationTest` and give it its own `Activity` to run against, full of comfortaing deprecation warnings.


rjrjr:
Cool! This didn't actually fix what I thought it would fix in #703, so there must be a real bug in there somewhere. I'm still inclined to merge this, since it keeps the legacy test more purely legacy.

vgonda:
> > Because this is only used for testing, does it need to be in the `main` source set?
> 
> It's the `main` source set of a testing module, so it's only ever used from non-`main` source of production modules. This is idiomatic.

Gotcha! I missed that

RBusarow (approved):

PR #709, opened 2022-03-22
Gives bind* methods better names and better home.

rjrjr:
All the `ScreenLayoutRunner.bind` methods are now `ScreenViewFactory.for*` methods.


samples/containers/hello-back-button/src/main/java/com/squareup/sample/hellobackbutton/HelloBackButtonLayoutRunner.kt
```diff
@@ -2,6 +2,7 @@ package com.squareup.sample.hellobackbutton
 
 import android.view.View
 import android.widget.TextView
+import com.squareup.sample.hellobackbutton.R.layout
```
RBusarow, line 5: Some day, they'll fix this in Android Studio.

RBusarow (approved):
These names are a big improvement!

samples/containers/hello-back-button/src/main/java/com/squareup/sample/hellobackbutton/HelloBackButtonLayoutRunner.kt
rjrjr: I'll take that bet.

PR #710, opened 2022-03-23
Adds missing integration tests EnvironmentScreen, NamedScreen.

rjrjr:
The other default `ScreenViewFactory` implementations are well covered by all the espresso tests. Really, so is `NamedScreen`, but it was so easy to write.

Note that `EnvironmentScreen` is broken, so one assertion is commented out.

### DIFF
--- a/workflow-ui/core-android/src/test/java/com/squareup/workflow1/ui/container/ScreenViewFactoryTestUtil.kt
+++ b/workflow-ui/core-android/src/test/java/com/squareup/workflow1/ui/container/ScreenViewFactoryTestUtil.kt
@@ -1,0 +1,62 @@
+@file:OptIn(WorkflowUiExperimentalApi::class)
+
+package com.squareup.workflow1.ui.container
+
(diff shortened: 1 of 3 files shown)


workflow-ui/core-android/src/test/java/com/squareup/workflow1/ui/container/ScreenViewFactoryTestUtil.kt
```diff
@@ -0,0 +1,84 @@
...
+    return mockView().also { view ->
+      view.bindShowRendering(initialRendering, initialViewEnvironment) { _, environment ->
+        lastEnv = environment
+      }
+      lastView = view
+    }
+  }
+}
+
+internal class AlternativeWrappedFactory : ScreenViewFactory<WrappedScreen> {
```
rjrjr, line 64: D'oh. Don't need this to be a seprate class, can just use different instances of the other one.

RBusarow (approved):

PR #712, opened 2022-03-29
ScreenViewHolder replaces View.* extensions and DecorativeViewFactory

### LINKED ISSUE #413: DecorativeViewFactory should support wrapping Context

### LINKED ISSUE #551: ViewEnvironment argument of DecorativeViewFactory.map is redundant

### LINKED ISSUE #598: Either hide ShowRenderingTag or don't make it a data class

### LINKED ISSUE #626: Replace DecorativeScreenViewFactory with a builder

rjrjr:
Replaces #703.

The legacy workflow-ui world was built around a set of extension methods on `View`, backed by a `WorkflowViewState` object hanging off of a tag. This was overly clever, needlessly threw away a good bit of compile time safety, and made it much harder than it should be to implement wrapper renderings.

In particular, every `ViewFactory` was required to call `View.bindShowRendering` to put all this machinery in place. When a screen rendering wrapped another, with the intention of delegating to its registered `ViewFactory`, that required creating a wrapper `ViewFactory` that made a new `bindShowRendering` call replacing and wrapping the `WorkflowViewState` put in place by the delegate. That's so tricky it required the introduction of `DecorativeViewFactory`, which was still pretty tricky to use.

To solve all this, we reverse the relationship between `View` and `WorkflowViewState`, with the latter being replaced by `ScreenViewHolder`. Instead of returing a raw `View`, `ScreenViewFactory.buildView` now returns the view in a `ScreenViewHolder`. The call to `ScreenViewHolder()` replaces the old call to `bindShowRendering`, and the compiler makes it impossible to forget.

`ScreenViewHolder` is a box that holds:

- a `View`
- the `(Screen, ViewEnvironment) -> Unit` function that updates it (still `fun interface ScreenViewRunner` for ease of implementation)
- and the latest `ViewEnvironment` that function received

It is worth noting that `ScreenViewHolder.showing` `.canShow` and `.show()` are all static extensions, built around the new `Showing: ViewEnvironmentKey<Screen>`. `ScreenViewHolder` itself is "write only", with an `in ScreenT: Screen` type parameter -- you can push strongly typed `Screen` renderings into it, but you can't pull them out again.

This minimalist `ScreenViewHolder` interface is trivial to wrap, so all the complexity that inspired `DecorativeViewFactory` is gone. It is replaced by the new `ScreenViewFactory.unwrapping` extension functions, whose use is demontrated in `NamedScreenViewFactory`, `EnvironmentScreenViewFactory` and `BackButtonScreen`.

The `ViewStarter` mechanism is also much simpler now, also no longer relying on any state stored in `View` tags. The implementation of `ScreenViewFactory.start()` remains a bit complex, but only because of the work it has to do to keep the legacy `View.start()` method working, in partnership with `AsScreenViewFactory()`. Once we delete the deprecated View code, most of the code in both of those can be deleted.

Fixes #413, fixes #598, fixes #626, fixes #551, fixes #443.

WorkflowViewStub show()

```
# https://sequencediagram.org/
title WorkflowViewStub.show()

participant "WorkflowViewStub" as stub
participant Screen
participant "ViewEnvironment" as env
participant "ScreenViewFactoryFinder" as finder
participant "ViewRegistry" as registry
participant "ScreenViewFactory" as factory
participant "ScreenViewHolder" as holder

[->stub:show(screen, env)

stub->Screen:buildView(env)
activate Screen
Screen->Screen:toViewFactory(env)
activate Screen
Screen->env:get(ScreenViewFactoryFinder)
Screen<--env:finder
Screen->finder:getViewFactoryForRendering(this)
activate finder
finder->registry:getEntryFor(screen)
finder<--registry:factory
Screen<--finder:factory
deactivate finder
deactivate Screen

Screen->factory:start(this, env)
activate factory
factory->factory:buildView(screen, env)
activate factory
factory->holder:init(screen, env, runner)
holder-->factory:holder
deactivate factory
factory->holder:show(screen, env)
activate holder
holder->holder:this.environment = env + (Showing to screen)
holder->holder:this.runner.show(screen, environment)
deactivate holder
Screen<--factory:holder
deactivate factory

stub<--Screen:holder
deactivate Screen


[->stub:show(screen, env)
activate stub
stub->holder:canShow(screen)
activate holder
holder->holder: compatible(this.environment[Showing], screen)
stub<--holder:true
deactivate holder
stub->holder:show(screen, env)
activate holder
holder->holder:this.environment = env + (Showing to screen)
holder->holder:this.runner.show(screen, environment)
deactivate holder
deactivate stub
```


rjrjr:
I've opened this open for review again because it's FINALLY GREEN. It's a bit sprawling, though, because a bunch of method renames that were previously in a separate PR have been squashed into this one.

I'm used up for today, I'll see if I can make it easier to digest in the morning.

workflow-ui/compose/src/androidTest/java/com/squareup/workflow1/ui/compose/NoTransitionBackStackContainer.kt
```diff
@@ -21,28 +20,25 @@ import com.squareup.workflow1.ui.container.R
...
         .apply {
           id = R.id.workflow_back_stack_container
           layoutParams = LayoutParams(MATCH_PARENT, MATCH_PARENT)
-          bindShowRendering(
-            initialRendering,
-            initialEnv,
-            { newRendering, newViewEnvironment -> update(newRendering, newViewEnvironment) }
-          )
         }
+      ScreenViewHolder(initialEnvironment, view) { rendering, environment ->
```
rjrjr, line 39: Notice how `bindShowRendering` calls migrate to `ScreenViewHolder` instantiation.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/WorkflowRendering.kt
```diff
@@ -184,29 +187,27 @@ private fun <R : Screen> ScreenViewFactory<R>.asComposeViewFactory() =
...
-
-              // Mirrors the check done in ViewRegistry.buildView.
-              checkNotNull(view.getShowRendering<Any>()) {
-                "View.bindShowRendering should have been called for $view, typically by the " +
-                  "ScreenViewFactory that created it."
-              }
-
+          originalFactory.start(rendering, viewEnvironment, context, container = null)
+            .let { viewHolder ->
+              // Put the viewHolder in a tag so that we can find it in the update lambda, below.
```
rjrjr, line 192: This is our only remaining use of `View.setTag`. I don't see an alternative.

workflow-ui/container-android/src/androidTest/java/com/squareup/workflow1/ui/modal/test/ModalViewContainerLifecycleActivity.kt
```diff
@@ -56,20 +48,15 @@ internal class ModalViewContainerLifecycleActivity : AbstractLifecycleTestActivi
...
-        container.bindShowRendering(
-          initialRendering,
-          initialViewEnvironment
-        ) { rendering, env ->
-          stub.show(asScreen(TestModals(listOf(rendering.wrapped))), env)
-        }
       }
-    },
+      ScreenViewHolder(environment, frame) { rendering, viewEnvironment ->
+        stub.show(asScreen(TestModals(listOf(rendering.wrapped))), viewEnvironment)
```
rjrjr, line 57: Compare this with #703, where we were forced to cast `(stub as WorkflowViewStub)`

workflow-ui/container-android/src/main/java/com/squareup/workflow1/ui/modal/ModalViewContainer.kt
```diff
@@ -71,14 +72,13 @@ public open class ModalViewContainer @JvmOverloads constructor(
...
-      viewEnvironment = initialViewEnvironment,
+    val viewHolder = wrappedRendering.buildView(
+      initialViewEnvironment = initialViewEnvironment,
       contextForNewView = this.context,
       container = this
     )
-    view.start()
 
-    return buildDialogForView(view)
+    return buildDialogForView(viewHolder.view)
```
rjrjr, line 81: Should change `buildDialogForView` to accept `ScreenViewHolder` instead.

Probably should update `OverlayDialogFactory` to work more like `ScreenViewFactory` -- make `DialogHolder` public, and eliminate the separate `update` method.

rjrjr: Oh, derp, wrong class -- this is the legacy `buildDialogForView`. But these notes probably apply to the new stuff.

workflow-ui/core-android/src/androidTest/java/com/squareup/workflow1/ui/container/ViewStateCacheTest.kt
```diff
@@ -130,35 +139,22 @@ internal class ViewStateCacheTest {
...
     // "Navigate" back to the first screen, restoring state.
     val firstViewRestored = createTestView(firstRendering)
-    cache.update(listOf(firstRendering), oldViewMaybe = secondView, newView = firstViewRestored)
+    cache.update(listOf(firstRendering), oldHolderMaybe = secondView, newHolder = firstViewRestored)
 
     // Check that the state was NOT restored.
     assertThat(firstViewRestored.viewState).isEqualTo("")
   }
 
-  @Test fun throws_when_view_not_bound() {
```
rjrjr, line 149: This mistake is now impossible.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/NamedScreenViewFactory.kt
```diff
@@ -1,9 +1,17 @@
...
  * [ScreenViewFactory] that allows views to display instances of [NamedScreen]. Delegates
  * to the factory for [NamedScreen.wrapped].
  */
 @WorkflowUiExperimentalApi
-internal object NamedScreenViewFactory : ScreenViewFactory<NamedScreen<*>>
-by DecorativeScreenViewFactory(NamedScreen::class, { named -> named.wrapped })
+internal
+fun <WrappedT : Screen> NamedScreenViewFactory(): ScreenViewFactory<NamedScreen<WrappedT>> {
+  return forBuiltView { initialNamedScreen, initialEnvironment, context, container ->
+    initialNamedScreen.wrapped.toViewFactory(initialEnvironment)
```
rjrjr, line 13: Less concise, but I hope also less mysterious.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/RealScreenViewHolder.kt
```diff
@@ -0,0 +1,25 @@
+package com.squareup.workflow1.ui
+
+import android.view.View
+
+/**
```
rjrjr, line 5: snip, this is redundant noise on an internal impl class.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewFactoryFinder.kt
```diff
@@ -8,22 +8,22 @@ import com.squareup.workflow1.ui.container.EnvironmentScreen
...
+ *    by ScreenViewFactory(
+ *      buildView = { environment, context, _ ->
  *        MyBackStackContainer(context)
- *          .apply {
- *            layoutParams = (LayoutParams(MATCH_PARENT, MATCH_PARENT))
- *            bindShowRendering(initialRendering, initialEnv, ::update)
- *          }
+ *          .apply { layoutParams = (LayoutParams(MATCH_PARENT, MATCH_PARENT)) }
+ *      },
+ *      updateView = { view, rendering, environment ->
```
rjrjr, line 25: Stale, pretty sure I missed this in the kdoc update.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewFactoryFinder.kt
```diff
@@ -57,18 +57,20 @@ public interface ScreenViewFactoryFinder {
     @Suppress("UNCHECKED_CAST")
     return (entry as? ScreenViewFactory<ScreenT>)
       ?: (rendering as? AndroidScreen<*>)?.viewFactory as? ScreenViewFactory<ScreenT>
-      ?: (rendering as? AsScreen<*>)?.let { AsScreenViewFactory as ScreenViewFactory<ScreenT> }
+      ?: (rendering as? AsScreen<*>)?.let {
+        AsScreenViewFactory(it, environment) as ScreenViewFactory<ScreenT>
```
rjrjr, line 61: - [ ] `AsScreenViewFactory` should be updated to use param type correctly, see `NamedScreenViewFactory` and `EnvironmentScreenViewFactory`.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewShowRendering.kt
```diff
@@ -29,6 +29,7 @@ public typealias ViewShowRendering<RenderingT> =
  * @see DecorativeViewFactory
  */
 @WorkflowUiExperimentalApi
+@Deprecated("Replaced by ScreenViewFactory.updateView")
```
rjrjr, line 32: Stale.

- [ ] Need to search for lingering `updateView` references.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/EnvironmentScreenViewFactory.kt
```diff
@@ -1,18 +1,26 @@
...
-by DecorativeScreenViewFactory(
-  type = EnvironmentScreen::class,
-  unwrap = { withEnvironment, inheritedEnvironment ->
-    Pair(
-      withEnvironment.screen,
-      inheritedEnvironment merge withEnvironment.viewEnvironment
-    )
+internal fun <WrappedT : Screen> EnvironmentScreenViewFactory():
+  ScreenViewFactory<EnvironmentScreen<WrappedT>> {
+  return forBuiltView { initialEnvScreen, initialEnvironment, context, container ->
```
rjrjr, line 14: Not crazy about this name. `forBuilder`? `forConstructor`? `fromCode`?

workflow-ui/core-android/src/test/java/com/squareup/workflow1/ui/container/EnvironmentScreenAndroidIntegrationTest.kt
```diff
@@ -7,51 +7,41 @@ import com.squareup.workflow1.ui.ViewEnvironment.Companion.EMPTY
...
-    assertThat(view.environment!![SomeEnvValue]).isEqualTo("hi")
+    assertThat(holder.environment[SomeEnvValue]).isEqualTo("hi")
 
-    view.start()
-    view.showRendering(wrappedScreen.withEnvironment(EMPTY + (SomeEnvValue to "bye")), EMPTY)
+    holder.show(wrappedScreen.withEnvironment(EMPTY + (SomeEnvValue to "bye")), EMPTY)
 
     assertThat(wrappedScreen.viewFactory.lastEnv!![SomeEnvValue]).isEqualTo("bye")
-
-    // TODO To be fixed or obviated by https://github.com/square/workflow-kotlin/pull/703
```
rjrjr, line 54: TODONE

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewFactory.kt
```diff
@@ -177,12 +284,28 @@ public fun interface ViewStarter {
...
+ * @param unwrap a function to extract [WrappedT] instances from [WrapperT]s.
+ */
 @WorkflowUiExperimentalApi
 public inline
 fun <reified WrapperT : Screen, WrappedT : Screen> ScreenViewFactory<WrappedT>.unwrapping(
   crossinline unwrap: (wrapperScreen: WrapperT) -> WrappedT,
 ): ScreenViewFactory<WrapperT> = unwrapping(unwrap) { _, ws, e, su -> su(unwrap(ws), e) }
 
+/**
+ * Transforms a [ScreenViewFactory] of [WrappedT] into one that can handle
```
rjrjr, line 300: - [ ] Should update this to include the use cases from `DecorativeScreenViewFactory`.

workflow-ui/container-android/src/main/java/com/squareup/workflow1/ui/modal/ModalViewContainer.kt
rjrjr: Tracked in #713

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewFactoryFinder.kt
rjrjr: Or maybe those two can be `object` instead of `fun`

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewFactory.kt
```diff
@@ -93,3 +284,201 @@ public fun interface ViewStarter {
...
+ *       val onBackPressed: (() -> Unit)? = null
+ *    ) : Screen, Compatible {
+ *      override val compatibilityKey = Compatible.keyFor(wrapped)
+ *    }
+ *
+ *    fun <W: Screen> BackButtonViewFactory<W>() =
+ *      ScreenViewFactory.forBuiltView<BackButtonScreen<W>> {
+ *        initialRendering, initialEnv, context, container ->
+ *          initialRendering.wrapped.toViewFactory(initialEnv)
+ *            .unwrapping<BackButtonScreen<W>, W>(
```
rjrjr, line 418: These four lines happen a lot. I wonder if we can do better.

```kotlin
fun <WrapperT, WrappedT> ScreenViewFactory.forScreenWrapper(
  unwrap: (WrapperT) -> WrappedT,
  mumbled: (Mumble, Mumble) -> Mumble.
)
```

Not in this PR.

workflow-ui/core-android/src/androidTest/java/com/squareup/workflow1/ui/WorkflowViewStubLifecycleActivity.kt
```diff
@@ -20,20 +21,15 @@ internal class WorkflowViewStubLifecycleActivity : AbstractLifecycleTestActivity
...
     leafViewBinding(LeafRendering::class, lifecycleLoggingViewObserver { it.name }),
-    ManualScreenViewFactory(RecurseRendering::class) { initialRendering,
-      initialViewEnvironment,
-      contextForNewView, _ ->
-      FrameLayout(contextForNewView).also { container ->
-        val stub = WorkflowViewStub(contextForNewView)
+    forBuiltView<RecurseRendering> { _, initialEnvironment, context, _ ->
+      val stub = WorkflowViewStub(context)
+      val frame = FrameLayout(context).also { container ->
         container.addView(stub)
```
steve-the-edwards, line 27: nit: I prefer 
```suggestion
      val frame = FrameLayout(context).apply {
        addView(stub)
```

as adding in the `container` named arg is a little harder to immediately think of the `FrameLayout` just created.

steve-the-edwards (commented):
Halfway through. will break until tomorrow. Looks good so far! Really like the shape of the holder more!

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewFactoryFinder.kt
rjrjr: Oh, it's different b/c it wraps any old legacy screen.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewFactory.kt
rjrjr: Tracked in #714.

workflow-ui/core-android/src/androidTest/java/com/squareup/workflow1/ui/WorkflowViewStubLifecycleActivity.kt
rjrjr: For that matter it's silly to use a lambda at all since `frame` exists:

```kotlin
      val frame = FrameLayout(context)
      frame.addView(stub) 
```

rjrjr: Found a few more of these, tidying.

workflow-ui/core-android/src/androidTest/java/com/squareup/workflow1/ui/container/fixtures/BackStackContainerLifecycleActivity.kt
```diff
@@ -34,14 +33,14 @@ internal class BackStackContainerLifecycleActivity : AbstractLifecycleTestActivi
    * Default rendering always shown in the backstack to simplify test configuration.
    */
   object BaseRendering : Screen, ScreenViewFactory<BaseRendering> {
-    override val type: KClass<in BaseRendering> = BaseRendering::class
```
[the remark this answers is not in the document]
rjrjr: Unnecessary, the factory function is already doing that for us.

workflow-ui/core-android/src/androidTest/java/com/squareup/workflow1/ui/container/fixtures/NoTransitionBackStackContainer.kt
```diff
@@ -17,26 +16,26 @@ import com.squareup.workflow1.ui.container.BackStackScreen
...
+    addView(newHolder.view)
   }
 
   companion object : ScreenViewFactory<BackStackScreen<*>>
-  by ManualScreenViewFactory(
-    type = BackStackScreen::class,
-    viewConstructor = { initialRendering, initialEnv, context, _ ->
-      NoTransitionBackStackContainer(context)
+  by ScreenViewFactory.forBuiltView(
+    buildView = { _, initialEnvironment, context, _ ->
```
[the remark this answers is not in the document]
rjrjr: Combination of rushing and avoiding long lines, and don't care very much about consistency in non-sample code.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewFactory.kt
```diff
@@ -1,89 +1,279 @@
...
 
 /**
- * Factory for [View] instances that can show renderings of type [RenderingT] : [Screen].
+ * The function that updates a [View] instance built by a [ScreenViewFactory].
+ * Each [ScreenViewRunner] instance is paired with the single [View] instance,
+ * its neighbor in a [ScreenViewHolder].
  *
- * Two concrete [ScreenViewFactory] implementations are provided:
+ * Use [forLayoutResource], [forViewBinding], etc., to create a [ScreenViewFactory]
+ * from a [ScreenViewRunner].
```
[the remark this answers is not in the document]
rjrjr: Much better, thanks.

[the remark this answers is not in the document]
rjrjr:
```
 * Use [forLayoutResource], [forViewBinding], etc., to create a [ScreenViewFactory].
 * These helper methods take the resource, view, and/or view binding as arguments
 * along with a factory to create a [showRendering][ScreenViewRunner.showRendering]
 * function.
```

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewFactory.kt
```diff
@@ -1,89 +1,279 @@
...
+    @WorkflowUiExperimentalApi
+    public inline fun <reified ScreenT : Screen> forBuiltView(
+      crossinline buildView: (
+        initialRendering: ScreenT,
+        initialEnvironment: ViewEnvironment,
+        context: Context,
+        container: ViewGroup?
+      ) -> ScreenViewHolder<ScreenT>,
+    ): ScreenViewFactory<ScreenT> {
+      return object : ScreenViewFactory<ScreenT> {
```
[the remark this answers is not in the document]
rjrjr: Mainly trying to minimize the size of the PR. The separate classes existed already. It's all private API so I resisted the urge to tidy too much.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewFactory.kt
```diff
@@ -1,89 +1,279 @@
...
- * See [ViewStarter] for details.
+ * - It is more common to use [WorkflowViewStub.show] than to call this method directly
+ * - Call [ScreenViewFactory.start] to create and initialize a new [View]
+ * - If you don't particularly need to mess with the [ScreenViewFactory] before creating
+ *   a view, use [Screen.buildView] instead of this method.
+ */
+@WorkflowUiExperimentalApi
+public fun <ScreenT : Screen> ScreenT.toViewFactory(
+  environment: ViewEnvironment
+): ScreenViewFactory<ScreenT> {
```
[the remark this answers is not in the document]
rjrjr: That's impossible, `ViewEnvironment` never returns `null`.

```kotlin
  public companion object : ViewEnvironmentKey<ScreenViewFactoryFinder>(
    ScreenViewFactoryFinder::class
  ) {
    override val default: ScreenViewFactoryFinder
      get() = object : ScreenViewFactoryFinder {}
  }
```

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewFactory.kt
```diff
@@ -1,89 +1,279 @@
...
+  return environment[ScreenViewFactoryFinder].getViewFactoryForRendering(environment, this)
+}
+
+/**
+ * It is more common to use [WorkflowViewStub.show] than to call this method directly.
```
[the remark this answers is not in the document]
rjrjr: I really want to emphasize that `WorkflowViewStub` is the 99% case, though. I think this waters down that message.

```
 * It is more common to use [WorkflowViewStub.show] than to call this method directly.
 *
 * Creates a [ScreenViewHolder] wrapping a [View] able to display a stream
 * of [ScreenT] renderings, starting with [initialRendering].
 *
 * To add more initialization behavior (typically a call to [WorkflowLifecycleOwner.installOn]),
 * provide a [viewStarter].
```

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewFactory.kt
```diff
@@ -1,89 +1,279 @@
...
- * [View.bindShowRendering] when constructing the view
+ * By default "initialize" makes the first call to [ScreenViewHolder.show].
+ * To add more initialization behavior (typically a call to [WorkflowLifecycleOwner.installOn]),
+ * provide a [viewStarter].
  */
+@Suppress("DEPRECATION")
 @WorkflowUiExperimentalApi
-public fun <ScreenT : Screen> ScreenT.buildView(
-  viewEnvironment: ViewEnvironment,
+public fun <ScreenT : Screen> ScreenViewFactory<ScreenT>.start(
```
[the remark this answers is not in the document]
rjrjr: Maybe use the same name for the Screen method.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewFactory.kt
```diff
@@ -1,89 +1,279 @@
...
+ * default "initialize" makes the first call to [ScreenViewHolder.show].
+ * To add more initialization behavior (typically a call to [WorkflowLifecycleOwner.installOn]),
+ * provide a [viewStarter].
+ *
+ * This method is purely shorthand for calling [Screen.toViewFactory] and then
+ * [ScreenViewFactory.start]. You might wish to make those calls separately if you
+ * need to treat the [ScreenViewFactory] before using it, e.g. via [unwrapping].
+ */
+@WorkflowUiExperimentalApi
+public fun <ScreenT : Screen> ScreenT.buildView(
```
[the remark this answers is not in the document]
rjrjr: I'll inline this.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewFactory.kt
```diff
@@ -93,3 +283,201 @@ public fun interface ViewStarter {
...
+ *
+ * One general note: when creating a wrapper rendering, you're very likely to want it
+ * to implement [Compatible], to ensure that checks made to update or replace a view
+ * are based on the wrapped item. Each wrapper example below illustrates this.
+ *
+ * This a simpler variant of the like named function that takes two arguments, for
+ * use when there is no need to customize the [view][ScreenViewHolder.view] or
+ * the [environment][ScreenViewHolder.environment].
+ *
+ * ## Examples
```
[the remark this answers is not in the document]
rjrjr: I'm more concerned about discoverability than maintenance.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/RealScreenViewHolder.kt
```diff
@@ -0,0 +1,20 @@
...
+
+@WorkflowUiExperimentalApi
+internal class RealScreenViewHolder<ScreenT : Screen>(
+  initialEnvironment: ViewEnvironment,
+  override val view: View,
+  viewRunner: ScreenViewRunner<ScreenT>
+) : ScreenViewHolder<ScreenT> {
+
+  private var _environment: ViewEnvironment = initialEnvironment
+  override val environment: ViewEnvironment get() = _environment
```
helios175, line 13: nit: I think you could use `private set`:
```
override var environment: ViewEnvironment = initialEnvironment
  private set
```

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewFactory.kt
```diff
@@ -93,3 +283,201 @@ public fun interface ViewStarter {
...
+ *
+ * @param unwrap a function to extract [WrappedT] instances from [WrapperT]s.
+ *
+ * @param showWrapperScreen a function invoked when an instance of [WrapperT] needs
+ * to be shown in a [View] built to display instances of [WrappedT]. Allows
+ * pre- and post-processing of the [View].
+ */
+@WorkflowUiExperimentalApi
+public inline
+fun <reified WrapperT : Screen, WrappedT : Screen> ScreenViewFactory<WrappedT>.unwrapping(
```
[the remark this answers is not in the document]
rjrjr: How about `contraMap`? 😆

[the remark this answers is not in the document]
rjrjr: `toUnwrappingViewFactory`?

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewFactory.kt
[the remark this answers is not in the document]
helios175: I was about to ask the purpose of `ScreenViewRunner` if it was a way to keep `Holder` as static/fixed as possible, and place the polymorphism in the `Runner`. Like a "strategy". From the comments I think that's the reason. :)

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewFactory.kt
```diff
@@ -93,3 +283,201 @@ public fun interface ViewStarter {
...
+ * Transforms a [ScreenViewFactory] of [WrappedT] into one that can handle
+ * instances of [WrapperT].
+ *
+ * One general note: when creating a wrapper rendering, you're very likely to want it
+ * to implement [Compatible], to ensure that checks made to update or replace a view
+ * are based on the wrapped item. Each wrapper example below illustrates this.
+ *
+ * Also see the simpler variant of this function that takes only an [unwrap] argument.
+ *
+ * ## Example
```
helios175, line 382: It sounded to me like the two examples were different in shape, but they look the same. Same structure, same calls? (not saying it's bad, just asking if I'm missing any difference)

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewFactory.kt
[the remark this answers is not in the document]
steve-the-edwards: My concern was that "it is more common" is declarative rather than imperative for code that doesn't exist yet. I want to make it clear why it _should_ be more common. Though reading back I realize my redraft didn't do that either - ha.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewFactory.kt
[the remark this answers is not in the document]
steve-the-edwards: Honestly this made me think of `contraMap` and the err friction around that. Sure, I think `toUnwrappingViewFactory` makes sense.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewHolder.kt
```diff
@@ -0,0 +1,91 @@
...
+ * [environment] should always hold a reference to the [Screen] most recently shown
+ * in [view], with the key [Showing]. [ScreenViewHolder.showing] provides easy access
+ * to it. Note that the shown [Screen] may not be of type [ScreenT], if this
+ * [ScreenViewHolder] is wrapped by another one. (See [ScreenViewFactory.unwrapping].)
+ *
+ * Do not call [runner] directly. Use [ScreenViewHolder.show] instead. Or most commonly,
+ * allow [WorkflowViewStub.show] to call it for you.
+ */
+@WorkflowUiExperimentalApi
+public interface ScreenViewHolder<in ScreenT : Screen> {
```
helios175, line 21: Is the plan to make this type as sealed as possible? I'm asking because the polymorphism is extracted out to the runner. 
If the answer is yes this could be a class, or a sealed class at least, or something. Definitely not important.

But maybe worth mentioning in the docs:
- holder: pretty static / fixed / controlled / not-instantiable from outside, etc
- runner: the "strategy" (strategy pattern) to implement a particular holder.

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/ViewEnvironment.kt
```diff
@@ -69,3 +69,13 @@ public abstract class ViewEnvironmentKey<T : Any>(
     return "${this::class.simpleName}(${type.simpleName})"
   }
 }
+
+@WorkflowUiExperimentalApi
+public inline fun <reified T : Any> ViewEnvironmentKey(
+  crossinline produceDefault: () -> T,
```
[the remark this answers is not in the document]
helios175: Let me guess: Compose-inspired API?

helios175 (approved):
I've been wandering around the code for 1h+ and everything makes sense so far.
This is awesome!

Some comments around making sure we understand the purpose (as in Matrix's Agent Smith purpose) of each class, and making sure `!purpose(Holder).intersects(purpose(Runner))`

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/EnvironmentScreenViewFactory.kt
steve-the-edwards: I liked `forXYZ` so much more than `bind` but now I'm jumping back on here to say I do love @RBusarow 's `from__` suggestion! or even `buildFrom__`

`buildFromCode` `buildFromLayoutResource` `buildFromBinding`

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewHolder.kt
rjrjr: I kept bouncing back and forth btw. using an interface or just making `RealScreenViewHolder` public (and dropping the `Real` prefix, obv.). I forget what finally drove me back to the interface, something related to ease of wrapping. Unclear to me that it's still necessary, I'll take one more look at dropping

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewFactory.kt
rjrjr: You know, you're right. I originally had dropped the second one, but then tired me thought it actually covered more ground.

It really doesn't, deleting it.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/RealScreenViewHolder.kt
rjrjr: I can't, the public API is `val`. But if I drop the interface, then hell yeah.

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/ViewEnvironment.kt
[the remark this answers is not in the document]
rjrjr: #715.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/AndroidViewRegistry.kt
```diff
@@ -38,7 +38,7 @@ public fun <RenderingT : Any> ViewRegistry.getFactoryFor(
   return getEntryFor(renderingType) as? ViewFactory<RenderingT>
 }
 
-@Deprecated("Use Screen.buildView")
+@Deprecated("Use Screen.toView")
```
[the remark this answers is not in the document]
rjrjr: It became `buildView`, but now I've gone ahead and inlined it! Will fix.

workflow-ui/core-android/src/androidTest/java/com/squareup/workflow1/ui/container/fixtures/NoTransitionBackStackContainer.kt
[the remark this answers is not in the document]
rjrjr: In the sample code, agree with you that `ScreenViewFactory.forBuiltView` is much better.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewFactory.kt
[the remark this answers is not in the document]
rjrjr: I think we used to say something like _It is very rare to call this method directly, instead use [WorkflowViewStub.show]_, WDYT?

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewHolder.kt
rjrjr: I _think_ I can pull this off, but I'd prefer to merge what I have and tackle it in a follow up. In the meantime I'll try to punch up the docs as you suggest.

rjrjr: The important thing about the interface approach is that it makes it easy to wrap a `ScreenViewHolder` and delegate to the `environment` of the wrapped instance, which is crucial for our crazy `showing` semantics. I suppose we could achieve that another way, but I don't really see what the benefit would be.

This is the case that drove me to write `EnvironmentScreenAndroidIntegrationTest` last week.

PR #718, opened 2022-04-08
Fixes memory leak in ViewStateCache.

rjrjr:
`ViewStateCache` was `Parcelable`, in an attempt to simplify `View.onSaveInstanceState()` chores for client container views. That became a source of memory leaks once we added a `WorkflowSavedStateRegistryAggregator` to it, so we've dialed back the convenience.

`ViewStateCache` now offers `save(): Saved` and `restore(saved: Saved)`, where `Saved` is a simple `Parcelable`. Client views are responsible for managing their own subclasses of `View.BaseSavedState`, which is sad but not terribly unusual.

### DIFF
--- a/workflow-ui/container-android/api/container-android.api
+++ b/workflow-ui/container-android/api/container-android.api
@@ -33,39 +33,46 @@ public final class com/squareup/workflow1/ui/backstack/BackStackContainer$Compan
 	public fun getType ()Lkotlin/reflect/KClass;
 }
 
-public final class com/squareup/workflow1/ui/backstack/ViewStateCache : android/os/Parcelable {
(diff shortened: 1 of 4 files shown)


rjrjr:
Before putting this PR up for review, I'm publishing it as `1.7.0-SNAPSHOT` so that I can verify the leak fix.

steve-the-edwards (approved):

rjrjr:
> Before putting this PR up for review, I'm publishing it as `1.7.0-SNAPSHOT` so that I can verify the leak fix.

That was dumb, it got stomped immediately. Making `1.7.0-rjrjr-SNAPSHOT` this time.

PR #719, opened 2022-04-11
Add Baseline and Perf Tests to Perf Poetry

steve-the-edwards:
Use `WorkflowInterceptor` to trace each render pass as well as particular nodes within it.

We use a "Fresh Rendering Ratio" which is the # of render passes for a node that had a different RenderState / the total # of render passes.


workflow-runtime/src/main/java/com/squareup/workflow1/internal/WorkflowRunner.kt
```diff
@@ -57,9 +58,11 @@ internal class WorkflowRunner<PropsT, OutputT, RenderingT>(
...
   fun nextRendering(): RenderingAndSnapshot<RenderingT> {
-    val rendering = rootNode.render(workflow, currentProps)
-    val snapshot = rootNode.snapshot(workflow)
-    return RenderingAndSnapshot(rendering, snapshot)
+    return Tracing.trace({ "Render Pass" }) {
+      val rendering = rootNode.render(workflow, currentProps)
+      val snapshot = rootNode.snapshot(workflow)
+      RenderingAndSnapshot(rendering, snapshot)
+    }
   }
```
steve-the-edwards, line 66: Here is where we trace the 'render pass'. Still open to feedback about other places we might want to have pluggable tracing.

I'm also thinking about possibly considering a 'cumulative' render time in the benchmark some way. So if we keep the scenarios the same we can see improvements from lessening churn.

workflow-runtime/src/main/java/com/squareup/workflow1/internal/WorkflowNode.kt
```diff
@@ -100,7 +101,9 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
     workflow: StatefulWorkflow<PropsT, *, OutputT, RenderingT>,
     input: PropsT
   ): RenderingT =
-    renderWithStateType(workflow as StatefulWorkflow<PropsT, StateT, OutputT, RenderingT>, input)
+    Tracing.traceId({ identifier }) {
+      renderWithStateType(workflow as StatefulWorkflow<PropsT, StateT, OutputT, RenderingT>, input)
+    }
```
steve-the-edwards, line 106: This is tracing we add for any and all Workflow Nodes that we care about measuring the time for their render.

steve-the-edwards:
Still more work to do to refine these and establish the best scenarios, but this is a good basis and gets us setup in the application 'complex-poetry' where we should focus our performance testing in.

workflow-core/src/main/java/com/squareup/workflow1/Tracing.kt
```diff
@@ -0,0 +1,85 @@
...
+    override fun beginIdSection(label: () -> WorkflowIdentifier) {
+      tracerArray.forEach { it.beginIdSection(label) }
+    }
+
+    override fun endIdSection(label: () -> WorkflowIdentifier) {
+      tracerArray.forEach { it.endIdSection(label) }
+    }
+
+    public inline fun <T> trace(
+      noinline label: () -> String,
```
pyricau, line 65: This isn't free, right? Every Tracing.trace() call is now creating an instance of the lambda that generates the string. When that string is a string static at compile time, it's actually cheaper to reference that directly instead of a generating lambda.

workflow-core/src/main/java/com/squareup/workflow1/Tracing.kt
```diff
@@ -0,0 +1,85 @@
...
+    override fun beginSection(label: () -> String) {
+      tracerArray.forEach { it.beginSection(label) }
+    }
+
+    override fun endSection(label: () -> String) {
+      tracerArray.forEach { it.endSection(label) }
+    }
+
+    private val tracers = ArrayList<Tracer>()
+    @Volatile private var tracerArray = emptyArray<Tracer>()
```
pyricau, line 83: Very confused at first by the need of both an array list and also an array. I think this seems to be related to Volatile here, i.e. the Array is immutable and its ref is volatile so any update to the ref propagates through CPU caches and that means beginSection and endSection don't need any lock. And keeping the list makes it easy to create a new array?

That's definitely a bit surprising, you'll want to document this well.

workflow-core/src/main/java/com/squareup/workflow1/Tracing.kt
```diff
@@ -0,0 +1,85 @@
...
+
+    override fun beginSection(label: () -> String) {
+      tracerArray.forEach { it.beginSection(label) }
+    }
+
+    override fun endSection(label: () -> String) {
+      tracerArray.forEach { it.endSection(label) }
+    }
+
+    private val tracers = ArrayList<Tracer>()
```
pyricau, line 82: note: YAGNI? Do you really need to support N Tracer impls, vs just the one (or none) ?

workflow-core/src/main/java/com/squareup/workflow1/Tracing.kt
```diff
@@ -0,0 +1,85 @@
...
+    public fun endSection(label: () -> String)
+  }
+
+  public companion object : Tracer {
+
+    /**
+     * Note that by default we have nothing installed so nothing gets run! Only install something
+     * in an app you want to profile.
+     */
+    @JvmStatic public fun install(tracer: Tracer) {
```
pyricau, line 24: note: from an API standpoint, I generally avoid having the "install" and "log" public APIs in the same place. Install is called once and you forget about it. trace is called many times from different places, and when you trace you don't want to see noise about install.

pyricau: See `logcat {}` for an example of separating out install.

workflow-core/src/main/java/com/squareup/workflow1/Tracing.kt
```diff
@@ -0,0 +1,85 @@
...
+      require(tracer != this) { "Don't install Tracing on itself." }
+      synchronized(tracers) {
+        tracers.add(tracer)
+        tracerArray = tracers.toTypedArray()
+      }
+    }
+
+    @JvmStatic public fun uninstall(tracer: Tracer) {
+      synchronized(tracers) {
+        require(tracers.remove(tracer)) { "Can't uninstall tracer not installed." }
```
pyricau, line 34: Add `$tracer` to error msg

workflow-core/src/main/java/com/squareup/workflow1/Tracing.kt
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  public companion object : Tracer {
+
+    /**
+     * Note that by default we have nothing installed so nothing gets run! Only install something
+     * in an app you want to profile.
+     */
+    @JvmStatic public fun install(tracer: Tracer) {
+      require(tracer != this) { "Don't install Tracing on itself." }
```
pyricau, line 25: Add `$tracer` to error msg

workflow-core/src/main/java/com/squareup/workflow1/Tracing.kt
```diff
@@ -0,0 +1,85 @@
...
+ * Tracing facade to install later.
+ *
+ * This follows the static service locator that Timber demonstrates.
+ * See [Timber][https://github.com/JakeWharton/timber/blob/trunk/timber/src/main/java/timber/log/Timber.kt].
+ */
+public class Tracing private constructor() {
+
+  public interface Tracer {
+    public fun beginIdSection(label: () -> WorkflowIdentifier)
+    public fun endIdSection(label: () -> WorkflowIdentifier)
```
pyricau, line 13: I'm not sure I understand how this is going to be implemented. Do you have an example? It looks like this PR does not have any actual tracing code hooked up?

RELEASING.md
```diff
@@ -8,7 +8,7 @@
 1. Confirm that the kotlin build is green before committing any changes
    (Note we exclude benchmarks, but you can check those too!)
    ```bash
-   ./gradlew build connectedCheck ./gradlew connectedCheck -x :benchmarks:dungeon-benchmark:connectedCheck
+   ./gradlew build connectedCheck ./gradlew connectedCheck -x :benchmarks:dungeon-benchmark:connectedCheck -x :benchmarks:performance-poetry:complex-benchmark:connectedCheck
```
rjrjr, line 11: I just pushed a conflicting change b/c this is an invalid command line, and has a redundant bit:
```suggestion
   ./gradlew build && ./gradlew connectedCheck -x :benchmarks:dungeon-benchmark:connectedCheck -x :benchmarks:performance-poetry:complex-benchmark:connectedCheck
```

workflow-runtime/src/main/java/com/squareup/workflow1/internal/WorkflowNode.kt
rjrjr: You sure we can't accomplish this with the existing interceptor API? This seems like a very big new back door addressing the same kind of use case. It's not clear to me that this is any more expressive.

workflow-core/src/main/java/com/squareup/workflow1/Tracing.kt
steve-the-edwards: That's a great point. This is more needed when using the `WorkflowIdentifier` and not with the string.

workflow-core/src/main/java/com/squareup/workflow1/Tracing.kt
steve-the-edwards: Your logic sounds right - to be honest this was cribbed from Timber's service locator and I didn't think too deeply about it.

workflow-core/src/main/java/com/squareup/workflow1/Tracing.kt
steve-the-edwards: ya its a good point. I think one is probably fine.

workflow-core/src/main/java/com/squareup/workflow1/Tracing.kt
steve-the-edwards: It does. There is tracing hooked up in the complex-poetry app.

RELEASING.md
steve-the-edwards: Whoops thank you

workflow-runtime/src/main/java/com/squareup/workflow1/internal/WorkflowNode.kt
steve-the-edwards: We likely could reproduce something close the use for each `WorkflowNode` but the other example - that of the internal `WorkflowRunner` "Render Pass" timing wouldn't be possible I don't think. I discussed this with @RBusarow but maybe we should sync up again briefly to discuss it.

steve-the-edwards (commented):
- [ ] Rework to use `WorkflowInterceptor`

benchmarks/performance-poetry/complex-benchmark/src/main/java/com/squareup/benchmarks/performance/poetry/complex/benchmark/ComplexPoetryResults.txt
```diff
@@ -0,0 +1,106 @@
...
+RenderPass-30Ms   min 0.2,   median 0.3,   max 0.5
+RenderPass-31Ms   min 0.4,   median 0.8,   max 1.0
+RenderPass-3Ms   min 0.0,   median 0.1,   max 0.2
+RenderPass-4Ms   min 0.3,   median 0.9,   max 1.1
+RenderPass-5Ms   min 0.1,   median 0.3,   max 0.5
+RenderPass-6Ms   min 0.1,   median 0.2,   max 0.4
+RenderPass-7Ms   min 0.2,   median 0.6,   max 0.7
+RenderPass-8Ms   min 0.1,   median 0.2,   max 0.4
+RenderPass-9Ms   min 0.4,   median 0.6,   max 1.3
+RenderPassTotalMs   min 0.2,   median 0.2,   max 0.8
```
steve-the-edwards, line 46: This is the average of averages. The average of each of 32 render pass durations for each of 20 test iterations.

rjrjr (approved):

benchmarks/performance-poetry/complex-poetry/src/main/java/com/squareup/benchmarks/performance/complex/poetry/PerformancePoetryComponent.kt
```diff
@@ -37,9 +45,121 @@ class PerformancePoetryComponent : ViewModel() {
...
-      loadingGatekeeperForPoems
+      loadingGatekeeperForPoems,
+      performanceTraceInterceptor
+    )
+
+  companion object {
+    // Update this value if render passes for the scenario under test change.
+    // Can do so with `adb logcat | grep PerformancePoetryComponent` and look through the
+    // Render Pass Stats.
+    const val TOTAL_RENDER_PASSES_OPEN_RAVEN_NAVIGATE = 32
```
rjrjr, line 158: Should you throw or something if the list grows bigger than this, so we'll notice?

rjrjr: And/or we could make this a much bigger number (1000 say), so that it won't matter so much if it climbs a bit.

I'm presuming you're trying to avoid a hit for expanding the array in mid-benchmark.

benchmarks/dungeon-benchmark/src/main/java/com/squareup/sample/dungeon/benchmark/DungeonGatherBaselineProfile.kt
```diff
@@ -23,24 +23,21 @@ import kotlin.time.ExperimentalTime
  */
 @RunWith(AndroidJUnit4::class)
 @OptIn(ExperimentalBaselineProfilesApi::class)
-public class WorkflowBaselineProfiles {
+class DungeonGatherBaselineProfile {
```
pyricau, line 26: I don't understand what's up with dungeons and poetry. Can we use real names that explain the differences?

benchmarks/performance-poetry/complex-benchmark/build.gradle.kts
```diff
@@ -0,0 +1,62 @@
...
+  defaultConfig {
+    minSdk = 26
+    targetSdk = 32
+
+    testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
+  }
+
+  buildTypes {
+    // This benchmark buildType is used for benchmarking, and should function like your
+    // release build (for example, with minification on). It"s signed with a debug key
```
pyricau, line 30: `It''s`

benchmarks/performance-poetry/complex-benchmark/build.gradle.kts
```diff
@@ -0,0 +1,62 @@
...
+    create("benchmark") {
+      isDebuggable = true
+      signingConfig = getByName("debug").signingConfig
+      matchingFallbacks.add("release")
+      proguardFile("baseline-proguard-rules.pro")
+    }
+  }
+
+  // kotlin_builtins duplication between 1.5.20 and 1.6.10?
+  packagingOptions.exclude("**/*.kotlin_*")
```
pyricau, line 41: ?

benchmarks/performance-poetry/complex-benchmark/build.gradle.kts
```diff
@@ -0,0 +1,62 @@
...
+
+  targetProjectPath = ":benchmarks:performance-poetry:complex-poetry"
+  experimentalProperties["android.experimental.self-instrumenting"] = true
+}
+
+dependencies {
+  implementation(project(":benchmarks:performance-poetry:complex-poetry"))
```
pyricau, line 48: so what's a complex benchmark? What's a not complex one? Are there more interesting words we can use to describe the difference?

pyricau: Unless... this is about "ComplexPoetry". Why couldn't that have been just Poetry?!

benchmarks/performance-poetry/complex-benchmark/src/main/java/com/squareup/benchmarks/performance/poetry/complex/benchmark/ComplexPoetryBenchmarks.kt
```diff
@@ -0,0 +1,132 @@
...
+      List(TOTAL_RENDER_PASSES_OPEN_RAVEN_NAVIGATE) { i ->
+        TraceSectionMetric("$it-$i")
+      }
+    } + List(TOTAL_RENDER_PASSES_OPEN_RAVEN_NAVIGATE) { i ->
+      TraceSectionMetric("RenderPass-$i")
+    } + TraceSectionMetric("RenderPassTotal")
+    benchmarkRule.measureRepeated(
+      packageName = PACKAGE_NAME,
+      metrics = traceMetricsList,
+      iterations = 20,
```
pyricau, line 84: not sure 20 is enough. What's the distribution looking like?

benchmarks/performance-poetry/complex-benchmark/src/main/java/com/squareup/benchmarks/performance/poetry/complex/benchmark/ComplexPoetryResults.txt
```diff
@@ -0,0 +1,106 @@
+
+April 13
+Nexus 6
+
+Commit SHA - 7fe2fe3cbe2ef6908ce60a7876122089aaae8503
```
pyricau, line 5: Really want to commit this?

benchmarks/performance-poetry/complex-benchmark/src/main/java/com/squareup/benchmarks/performance/poetry/complex/benchmark/ComplexPoetryResults.txt
pyricau: I'm confused, what exactly is the average of average is all these values? The middle value says median.

Also you might want to sort these properly from 1 to 30 ?

I got confused what  `-30Ms` might mean and then just realized it's likely the unit, ms? `ms` should always be written lowercase and with a space, although I get that here its more like a class name. I'd suggest changing to `RenderPass 9 (ms)`

benchmarks/performance-poetry/complex-benchmark/src/main/java/com/squareup/benchmarks/performance/poetry/complex/benchmark/ComplexPoetryResults.txt
```diff
@@ -0,0 +1,106 @@
...
+WorkflowIdentifier(com.squareup.benchmarks.performance.complex.poetry.PerformancePoemsBrowserWorkflow)-3Ms   min 0.0,   median 0.0,   max 0.0
+WorkflowIdentifier(com.squareup.benchmarks.performance.complex.poetry.PerformancePoemsBrowserWorkflow)-4Ms   min 0.2,   median 0.6,   max 0.9
+WorkflowIdentifier(com.squareup.benchmarks.performance.complex.poetry.PerformancePoemsBrowserWorkflow)-5Ms   min 0.1,   median 0.2,   max 0.4
+WorkflowIdentifier(com.squareup.benchmarks.performance.complex.poetry.PerformancePoemsBrowserWorkflow)-6Ms   min 0.0,   median 0.1,   max 0.2
+WorkflowIdentifier(com.squareup.benchmarks.performance.complex.poetry.PerformancePoemsBrowserWorkflow)-7Ms   min 0.1,   median 0.2,   max 0.2
+WorkflowIdentifier(com.squareup.benchmarks.performance.complex.poetry.PerformancePoemsBrowserWorkflow)-8Ms   min 0.0,   median 0.1,   max 0.1
+WorkflowIdentifier(com.squareup.benchmarks.performance.complex.poetry.PerformancePoemsBrowserWorkflow)-9Ms   min 0.3,   median 0.4,   max 1.0
+Traces: Iteration 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19
```
pyricau, line 106: What this log `Traces: Iteration 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19` telling us exactly?

pyricau: I just realized this might not be a file you created, but instead generated by macrobenchmark?

benchmarks/performance-poetry/complex-poetry/src/main/java/com/squareup/benchmarks/performance/complex/poetry/PerformancePoetryComponent.kt
```diff
@@ -37,9 +45,121 @@ class PerformancePoetryComponent : ViewModel() {
...
+          )
+          Timber.i("~~~~~~ Workflow Render Pass Stats End ~~~~~~")
+          totalFresh += sessionRenderStats[rootSessionId]!!.nodesRenderedDifferent
+          totalStale += sessionRenderStats[rootSessionId]!!.nodesRenderedTheSame
+          Timber.i("--------- Workflow Renderings Cumulative Stats ---------")
+          Timber.i("Total Render Passes So Far: $totalRenderPasses")
+          Timber.i(
+            "Total Fresh Rendering Ratio: ${
+            totalFresh.toDouble() / (totalFresh + totalStale).toDouble()
+            }"
```
pyricau, line 139: note: beware of complex timber logs slowing down the performance which you're trying to measure.

benchmarks/performance-poetry/complex-poetry/src/main/java/com/squareup/benchmarks/performance/complex/poetry/PerformancePoetryComponent.kt
pyricau: no I think this has to do with the number of sections to look for in the final perfetto file, so you do want to know the exact number.

Alternatively you could also ditch macrobenchmark's provided results and query perfetto with the trace directly, which we then give you the list of all sections and wouldn't need to anticipate that. I wish this was possible directly within macrobenchmark but it's not yet, so you'd have to locate the trace file and do this after the fact.

benchmarks/performance-poetry/complex-poetry/src/main/java/com/squareup/benchmarks/performance/complex/poetry/PerformancePoetryComponent.kt
```diff
@@ -37,9 +45,121 @@ class PerformancePoetryComponent : ViewModel() {
...
+    ): R {
+      val root = session.parent == null
+      val tracedId = NODES_TO_TRACE.firstOrNull { it == session.identifier } != null
+      if (root) {
+        // Setup stats for render pass tracking.
+        rootSessionId = session.sessionId
+        sessionRenderStats[rootSessionId] = RenderStats()
+        // Cumulative Section that has total render pass time.
+        Trace.beginSection("RenderPassTotal")
+        Trace.beginSection("RenderPass-$totalRenderPasses")
```
pyricau, line 77: Are you certain these will always close in the proper order and there won't be any additional sections that intersect (e.g. the ending doesn't happen from a posted main thread message)?

If not you could use async sections.

benchmarks/performance-poetry/complex-poetry/src/main/java/com/squareup/benchmarks/performance/complex/poetry/PerformancePoetryComponent.kt
```diff
@@ -1,12 +1,20 @@
 package com.squareup.benchmarks.performance.complex.poetry
 
+import android.annotation.SuppressLint
+import android.os.Trace
```
pyricau, line 4: Use the Android X trace instead. Also check out Trace.kt in POS to learn about some of the gotchas

benchmarks/performance-poetry/complex-poetry/src/main/java/com/squareup/benchmarks/performance/complex/poetry/PerformancePoetryComponent.kt
steve-the-edwards: @pyricau is right that we should keep this exact for now.

benchmarks/performance-poetry/complex-benchmark/build.gradle.kts
steve-the-edwards: lol. This is a 'Complex' version of the Poetry app.

benchmarks/performance-poetry/complex-benchmark/src/main/java/com/squareup/benchmarks/performance/poetry/complex/benchmark/ComplexPoetryResults.txt
steve-the-edwards: where else do you want to store the results?

benchmarks/performance-poetry/complex-benchmark/src/main/java/com/squareup/benchmarks/performance/poetry/complex/benchmark/ComplexPoetryResults.txt
steve-the-edwards: These are the test results generated by macrobenchmark. I copied them here to save them. This just shows we did 20 iterations (in the results this is also a link to the trace file of each iteration).

benchmarks/performance-poetry/complex-poetry/src/main/java/com/squareup/benchmarks/performance/complex/poetry/PerformancePoetryComponent.kt
steve-the-edwards: I thought about having 2 separate versions - one where we run to measure the 'fresh rendering ratio' and count the render passes and another where we trace the timing. WDYT?

benchmarks/performance-poetry/complex-poetry/src/main/java/com/squareup/benchmarks/performance/complex/poetry/PerformancePoetryComponent.kt
steve-the-edwards: In my testing they were nested properly by the nature of the WOrkflowInterceptor - but do you have any reason to doubt that based on your work with teh Workflow Interceptor?

benchmarks/performance-poetry/complex-poetry/src/main/java/com/squareup/benchmarks/performance/complex/poetry/PerformancePoetryComponent.kt
steve-the-edwards: I thought I was using that - obviously the import messed up.

steve-the-edwards:
- [ ] Put the render pass and stale/fresh node counting into a UI test that can be verified.

benchmarks/README.md
```diff
@@ -1,22 +1,27 @@
 # benchmarks
 
 This module contains benchmarks. Used to measure and help improve Workflow performance. These tests
-should be run on physical devices.
+should be run on physical devices, or the closest approximation to physical devices you can get.
```
pyricau, line 4: For the deterministic tests, that's shouldn't be a requirement anymore, right?

benchmarks/performance-poetry/complex-poetry/src/androidTest/java/com/squareup/benchmarks/performance/complex/poetry/RenderPassTest.kt
```diff
@@ -0,0 +1,126 @@
...
+  companion object {
+    val COMPLEX_INITIALIZING = Scenario(
+      title = "the 'Raven navigation with initializing state scenario'",
+      config = SimulatedPerfConfig(
+        isComplex = true,
+        useInitializingState = true,
+        complexityDelay = 100L
+      ),
+      expectedPasses = 61,
+      expectedRatio = 0.11891891891891893
```
pyricau, line 94: Wondering if you'd want to store the ratio as 2 numbers instead? idk feels weird to store a truncated number.

pyricau: In fact, if you stored 2 numbers, you could then assert against each, and report whether there are less things fresh or more overall things.

pyricau (approved):

PR #723, opened 2022-04-13
Pt. 2: Bump coroutines to 1.6.1

bnvinay92:


RBusarow:
Is there a specific reason to make this update now, like a transitive dependency causing a version conflict?

And did we have a specific reason for using the `native-mt` version?

bnvinay92:
> Is there a specific reason to make this update now, like a transitive dependency causing a version conflict?
> 
> And did we have a specific reason for using the `native-mt` version?

This and okio (bumped here #721) are external dependencies stopping :workflow-core from being published as a amp library. This is the primary reason.

ahmedre:
I believe native-mt is deprecated in favor of the vanilla coroutines package due to the new memory model - via [here](https://github.com/Kotlin/kotlinx.coroutines/issues/462#issuecomment-999557659):

> 1.6.0-native-mt has been released.
>
> 1.6.x will be the last series of native-mt releases and we are effectively decommissioning native-mt in the favour of the [new memory model](https://blog.jetbrains.com/kotlin/2021/08/try-the-new-kotlin-native-memory-manager-development-preview/) that is supported out-of-the-box in 1.6.0 coroutines release.
>
> Coroutines 1.7.0 will be released without native-mt companion and we are gradually stopping the support of native-mt during the course of 1.6.x lifecycle. We still are ready to accept PRs or release versions with the new Kotlin if there is a need to, but new coroutine features won't get native-mt support.

PR #730, opened 2022-04-20
Fixes BackStackContainer view persistence

rjrjr:
And adds missing integration test to keep it working.


workflow-ui/core-android/src/androidTest/java/com/squareup/workflow1/ui/container/BackStackContainerTest.kt
```diff
@@ -29,10 +36,70 @@ internal class BackStackContainerTest {
...
+          }
+        ) { _, _ -> /* Noop */ }
       }
   }
 
+  @Test fun savedStateParcelingWorks() {
+    scenario.onActivity { activity ->
+      val workflowLayout = WorkflowLayout(activity)
+      activity.setContentView(workflowLayout)
+
```
rjrjr, line 53: not needed

RBusarow (approved):

PR #735, opened 2022-04-22
Pt. 1: Use multiplatform gradle plugin for all kotlin/java-only gradle modules

bnvinay92:
Java libraries were unable to add Gradle project dependencies (for ex. `api(project(":workflow-core"))`) on multiplatform modules because they were having issues disambiguating variants. This PR 

1. Migrates all of these java modules to use the multiplatform gradle plugin
2. Migrates :workflow-runtime to use the kotlinx.benchmark gradle plugin

Allowing us to raise KMP migration PRs per module.


workflow-runtime/build.gradle.kts
```diff
@@ -1,14 +1,10 @@
...
 plugins {
-  `java-library`
-  kotlin("jvm")
+  kotlin("multiplatform")
   id("org.jetbrains.dokka")
-  // Benchmark plugins.
-  id("me.champeau.gradle.jmh")
-  // If this plugin is not applied, IntelliJ won't see the JMH definitions for some reason.
-  idea
+  id("org.jetbrains.kotlinx.benchmark") version "0.4.2"
```
bnvinay92, line 7: Need to move this to the version catalog

workflow-runtime/build.gradle.kts
```diff
@@ -18,35 +14,51 @@ java {
...
+      val main by getting
+      val workflowNode by creating {
+        kotlinOptions {
+          val compileKotlinJvm: KotlinCompile by tasks
+          freeCompilerArgs += "-Xfriend-paths=${compileKotlinJvm.destinationDir}"
+        }
+        defaultSourceSet {
+          dependencies {
+            implementation(main.compileDependencyFiles + main.output.classesDirs)
+            implementation("org.jetbrains.kotlinx:kotlinx-benchmark-runtime:0.4.2")
```
bnvinay92, line 29: Will move this to the version catalog

workflow-runtime/build.gradle.kts
```diff
@@ -18,35 +14,51 @@ java {
...
 
-  // These dependencies will be available on the classpath for source inside src/jmh.
-  "jmh"(libs.kotlin.jdk6)
-  "jmh"(libs.jmh.core)
-  "jmh"(libs.jmh.generator)
+benchmark {
+  targets {
+    register("jvmWorkflowNode") {
+      this as JvmBenchmarkTarget
+      jmhVersion = "1.35"
```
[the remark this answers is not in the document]
RBusarow: Yep.  Revert the line you removed from the "plugins" section, then you can reference here it like:

```kotlin
jmhVersion = libs.versions.jmh.get()
```

Note that since the catalog is generated, you have to do a gradle sync before the IDE recognizes the new definition.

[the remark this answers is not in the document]
bnvinay92: Thanks

gradle.properties
```diff
@@ -1,6 +1,7 @@
 org.gradle.jvmargs='-Dfile.encoding=UTF-8'
 org.gradle.parallel=true
 android.useAndroidX=true
+kotlin.mpp.stability.nowarn=true
```
steve-the-edwards, line 4: Maybe throw a comment in here for why we want to quiet this warning.

samples/containers/common/build.gradle.kts
```diff
@@ -1,17 +1,27 @@
...
-  `java-library`
-  kotlin("jvm")
+  kotlin("multiplatform")
 }
 
-dependencies {
-  implementation(project(":workflow-ui:container-common"))
-  implementation(project(":workflow-core"))
+kotlin {
+  jvm { withJava() }
```
steve-the-edwards, line 6: I get we have equivalent outcomes here, but I do want to review with @rjrjr and @RBusarow whether or not now is the time to make the jump to the KMP DSL for dependencies here, as it will have some learning curve.

workflow-core/src/jvmMain/kotlin/com/squareup/workflow1/HasWorkerType.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import kotlin.reflect.KType
+
+/**
+ * So that we don't have to make [WorkerWorkflow] public. Xfriend-paths does not seem to work as a
+ * compiler arg on multiplatform projects.
+ */
+public interface HasWorkerType {
+  public val workerType: KType
+}
```
steve-the-edwards, line 11: Reviewers N.B. this change.

workflow-runtime/build.gradle.kts
```diff
@@ -18,35 +14,51 @@ java {
...
-    val compileKotlin: KotlinCompile by tasks
-    freeCompilerArgs += "-Xfriend-paths=${compileKotlin.destinationDir}"
+kotlin {
+  jvm {
+    compilations {
+      val main by getting
+      val workflowNode by creating {
+        kotlinOptions {
+          val compileKotlinJvm: KotlinCompile by tasks
+          freeCompilerArgs += "-Xfriend-paths=${compileKotlinJvm.destinationDir}"
```
steve-the-edwards, line 24: if this isn't working do we still need it?

workflow-testing/src/jvmMain/kotlin/com/squareup/workflow1/testing/WorkerWorkflowHelper.kt
```diff
@@ -15,4 +16,4 @@ import kotlin.reflect.KType
  * See https://youtrack.jetbrains.com/issue/KT-20760.
  */
 internal fun Workflow<*, *, *>.workerWorkflowWorkerTypeOrNull(): KType? =
-  (this as? com.squareup.workflow1.WorkerWorkflow)?.workerType
+  (this as? HasWorkerType)?.workerType
```
steve-the-edwards, line 19: Need to update the kdoc above to explain how this is done now through the interface.

steve-the-edwards (commented):
I am OK with this change (few comments excluded) but I want to sync with @RBusarow and @rjrjr about the impact of adopting the KMP gradle DSL.

workflow-runtime/build.gradle.kts
bnvinay92: It works across compilations within a module, but not across modules.

PR #737, opened 2022-04-23
Introduces ViewEnvironmentKey.combine, makes ViewRegistry use it.

rjrjr:
`ViewEnvironmentKey.combine` is called from `ViewEnvironment.plus` when both operands have keys of the same type. `ViewRegistry.Companion` is already a `ViewEnvironmentKey`, and now it implements `combine`, calling `ViewRegistry.merge`.

This changes `ViewEnvironment.plus` to merge the values of `ViewRegistry` entries instead of replacing the registry on the left with the one on the right, which is the only thing we've ever actually wanted to do. This allows us to eliminate `ViewEnvironment.merge`.

We have never seen a use case for completely stomping the `ViewRegistry` on the left, but have done it by accident a lot. If the need really does arise, we can add a `ViewEnvironment.minus` operator.

Also eliminates the `ViewEnvironmentKey()` factory function, which was just silly.

### DIFF
--- a/workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/ViewEnvironment.kt
+++ b/workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/ViewEnvironment.kt
@@ -19,19 +19,30 @@ public class ViewEnvironment
 constructor(
   public val map: Map<ViewEnvironmentKey<*>, Any> = emptyMap()
 ) {
-  @Suppress("UNCHECKED_CAST")
-  public operator fun <T : Any> get(key: ViewEnvironmentKey<T>): T = map[key] as? T ?: key.default
+  public operator fun <T : Any> get(key: ViewEnvironmentKey<T>): T = getOrNull(key) ?: key.default
 
-  @Suppress("DEPRECATION")
-  public operator fun <T : Any> plus(pair: Pair<ViewEnvironmentKey<T>, T>): ViewEnvironment =
-    ViewEnvironment(map + pair)
+  public operator fun <T : Any> plus(pair: Pair<ViewEnvironmentKey<T>, T>): ViewEnvironment {
+    val (newKey, newValue) = pair
+    val newPair = getOrNull(newKey)
+      ?.let { oldValue -> newKey to newKey.combine(oldValue, newValue) }
+      ?: pair
+    @Suppress("DEPRECATION")
+    return ViewEnvironment(map + newPair)
+  }
 
   @Suppress("DEPRECATION")
   public operator fun plus(other: ViewEnvironment): ViewEnvironment {
     if (this == other) return this
     if (other.map.isEmpty()) return this
-    if (this.map.isEmpty()) return other
-    return ViewEnvironment(map + other.map)
+    if (map.isEmpty()) return other
+    val newMap = map.toMutableMap()
+    other.map.entries.forEach { (key, value) ->
+      @Suppress("UNCHECKED_CAST")
+      newMap[key] = getOrNull(key as ViewEnvironmentKey<Any>)
+        ?.let { oldValue -> key.combine(oldValue, value) }
+        ?: value
+    }
+    return ViewEnvironment(newMap)
   }
 
   override fun toString(): String = "ViewEnvironment($map)"
@@ -40,6 +51,9 @@ constructor(
     (other as? ViewEnvironment)?.let { it.map == map } ?: false
 
   override fun hashCode(): Int = map.hashCode()
+
+  @Suppress("UNCHECKED_CAST")
+  private fun <T : Any> getOrNull(key: ViewEnvironmentKey<T>): T? = map[key] as? T
 
   public companion object {
     @Suppress("DEPRECATION")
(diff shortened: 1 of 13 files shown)


workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/ViewEnvironment.kt
```diff
@@ -41,6 +49,9 @@ constructor(
 
   override fun hashCode(): Int = map.hashCode()
 
+  @Suppress("UNCHECKED_CAST")
+  private fun <T : Any> getOrNull(key: ViewEnvironmentKey<T>): T? = map[key] as? T
```
0legg, line 53: Worth to make it `getOrDefault`, as all callers are using some sort of fallback.

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/ViewEnvironment.kt
```diff
@@ -19,19 +19,27 @@ public class ViewEnvironment
...
     if (this == other) return this
     if (other.map.isEmpty()) return this
     if (this.map.isEmpty()) return other
-    return ViewEnvironment(map + other.map)
+    val newMap = map.toMutableMap()
+    other.map.entries.forEach { (key, value) ->
+      @Suppress("UNCHECKED_CAST")
+      newMap[key] =
+        getOrNull(key as ViewEnvironmentKey<Any>)?.let { key.combine(it, value) } ?: value
+    }
```
0legg, line 41: We could try to use Java8 merge method
```suggestion
    other.map.forEach { key, value ->
      newMap.merge(key, value) { valueLeft, valueRight -> key.combine(valueLeft, valueRight) }
    }
```

0legg (approved):

steve-the-edwards (approved):

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/ViewEnvironment.kt
rjrjr: This is non-Android, though, and there's a chance of it someday becoming KMP.

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/ViewEnvironment.kt
rjrjr: But the main call wants to run a lambda only in the non-null case.

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/ViewEnvironment.kt
```diff
@@ -19,19 +19,27 @@ public class ViewEnvironment
...
-  @Suppress("UNCHECKED_CAST")
-  public operator fun <T : Any> get(key: ViewEnvironmentKey<T>): T = map[key] as? T ?: key.default
+  public operator fun <T : Any> get(key: ViewEnvironmentKey<T>): T = getOrNull(key) ?: key.default
 
-  @Suppress("DEPRECATION")
-  public operator fun <T : Any> plus(pair: Pair<ViewEnvironmentKey<T>, T>): ViewEnvironment =
-    ViewEnvironment(map + pair)
+  public operator fun <T : Any> plus(pair: Pair<ViewEnvironmentKey<T>, T>): ViewEnvironment {
+    val (newKey, newValue) = pair
+    val newPair = getOrNull(newKey)?.let { newKey to newKey.combine(it, newValue) } ?: pair
```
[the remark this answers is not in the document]
rjrjr: I think you meant this:

```kotlin
    val newPair =
      getOrNull(newKey)?.let { oldValue -> newKey to newKey.combine(oldValue, newValue) } ?: pair
```

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/ViewEnvironment.kt
```diff
@@ -19,19 +19,27 @@ public class ViewEnvironment
...
   public operator fun plus(other: ViewEnvironment): ViewEnvironment {
     if (this == other) return this
     if (other.map.isEmpty()) return this
     if (this.map.isEmpty()) return other
-    return ViewEnvironment(map + other.map)
+    val newMap = map.toMutableMap()
+    other.map.entries.forEach { (key, value) ->
+      @Suppress("UNCHECKED_CAST")
+      newMap[key] =
+        getOrNull(key as ViewEnvironmentKey<Any>)?.let { key.combine(it, value) } ?: value
```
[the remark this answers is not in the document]
rjrjr:
```kotlin
      newMap[key] = getOrNull(key as ViewEnvironmentKey<Any>)
        ?.let { oldValue -> key.combine(oldValue, value) }
        ?: value
```

rjrjr:
Updated with better naming in the two lambdas calling `combine`.

rjrjr:
Update deletes a redundant `this`.

RBusarow (approved):

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/ViewEnvironment.kt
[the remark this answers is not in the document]
steve-the-edwards: yes - sorry got lambda confused with the pair vs named argument syntax

PR #740, opened 2022-04-25
Convert :workflow-core and :workflow-runtime to KMP

bnvinay92:
Rundown:
- Migrates from existing jmh gradle plugin to kotlinx.benchmark
- Bumps coroutines to 1.6.1 with the new native memory model. In hindsight, should have just used 1.5.1-native-mt given timelines for adoption.
- Replaces the gradle plugin in all non-android projects with the kotlin multiplatform gradle plugin and DSL (with just the jvm source sets). This is because they were having trouble consuming the multiplatform projects as dependencies. Tried to look for a workaround with variant attributes but couldn't really find one as ideally this should work out of the box. There's some merit to consistency across the non-android gradle scripts though.
- Changed the `repositoriesMode` setting from `RepositoriesMode.FAIL_ON_PROJECT_REPOS` to `RepositoriesMode.PREFER_PROJECT`. The multiplatform gradle plugin implicitly adds an `ivy()` repo to download the kotlin native compiler. Looking for a workaround for this, or at least some way to allow only multiplatform modules on preferring project repos.
- Introduced `HasWorkerType` as a way for `:workflow-testing` to consume `WorkerWorkflow`. Previously `:workflow-core` and `:workflow-testing` modules were friends but that doesn't work with the new plugin so doing this to avoid making `WorkerWorkflow` public. 
- `WorkflowIdentifier` was using `type: KAnnotatedElement` to represent either a KClass or a KType. Explicitly introduced a sealed class `WorkflowIdentifierType` to emulate this. Equality is now based on the serialized version of this sealed class, which allows save and restore without reflection. RenderTesters though, depend on KClass and KType for checking subtypes so `WorkflowIdentifierType` keeps a reference to these, but they are lost after restoration breaking :workflow-testing code for restored workflows. This is ok imo because the only time we check if workflows are subtypes of each other using their KClass/KType is in :workflow-testing, otherwise consumers have access to the workflows themselves if they wish to evaluate relationships for whatever reason.


workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/SimpleLoggingWorkflowInterceptor.kt
```diff
@@ -5,6 +5,8 @@ import com.squareup.workflow1.WorkflowInterceptor.WorkflowSession
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Job
 
+// expect internal fun printlnErr(text: String): Unit
```
nikolaykasyanov, line 8: Leftover?

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/SimpleLoggingWorkflowInterceptor.kt
```diff
@@ -113,7 +115,7 @@ public open class SimpleLoggingWorkflowInterceptor : WorkflowInterceptor {
     println(text)
   }
 
-  protected open fun logError(text: String): Unit = System.err.println(text)
+  protected open fun logError(text: String): Unit = println("ERROR: $text")
```
nikolaykasyanov, line 118: Not sure how important this is, but it should be possible to implement multiplatform printing to stderr.

nikolaykasyanov (commented):
🎉 🚀 ❤️

PR #746, opened 2022-04-28
Pt. 1: Everything uses the kmp gradle plugin

bnvinay92:
Rundown:
- Migrates from existing jmh gradle plugin to kotlinx.benchmark
- Replaces the gradle plugin in all non-android projects with the kotlin multiplatform gradle plugin and DSL (with just the jvm source sets). This is because they were having trouble consuming the multiplatform projects as dependencies. Tried to look for a workaround with variant attributes but couldn't really find one as ideally this should work out of the box.
- Introduced `HasWorkerType` as a way for `:workflow-testing` to consume `WorkerWorkflow`. Previously `:workflow-core` and `:workflow-testing` modules were friends but that doesn't work with the new plugin so doing this to avoid making `WorkerWorkflow` public.


samples/containers/common/build.gradle.kts
```diff
@@ -1,18 +1,27 @@
 plugins {
-  `java-library`
-  kotlin("jvm")
+  kotlin("multiplatform")
 }
 
-dependencies {
-  implementation(project(":workflow-ui:container-common"))
-  implementation(project(":workflow-ui:core-android"))
```
bnvinay92, line 8: Dropped this dependency. Did not seem necessary.

workflow-runtime/build.gradle.kts
```diff
@@ -18,35 +14,51 @@ java {
...
-    val compileKotlin: KotlinCompile by tasks
-    freeCompilerArgs += "-Xfriend-paths=${compileKotlin.destinationDir}"
+kotlin {
+  jvm {
+    compilations {
+      val main by getting
+      val workflowNode by creating {
+        kotlinOptions {
+          val compileKotlinJvm: KotlinCompile by tasks
+          freeCompilerArgs += "-Xfriend-paths=${compileKotlinJvm.destinationDir}"
```
steve-the-edwards, line 24: If we don't think friend-path arguments work wiht the multiplatform plugin, why do we still have this here?

steve-the-edwards (commented):
I need to check if the baseline-prof.txt will get picked up when packaged this way, but it shouldn't make any difference, as it will still end up in the `main` of all build variants anyawy.

workflow-runtime/build.gradle.kts
RBusarow: I looked into this a bit.  `friend-paths` is working here, and it's necessary.  In order for it to work, we also need the dependency upon `main.output.classesDirs` which is added below.  Why is that necessary?  No idea.

We can technically kind of do the same thing with the testing module, but of course that's a published artifact, and it's pointing to a different published artifact.  We're pointing Gradle at actual files and I can't imagine how it would work using Maven coordinates.

bnvinay92: Managed to get friend paths working. Raised another PR.

PR #747, opened 2022-04-28
replace use of `SendChannel.offer` with `SendChannel.trySend`

RBusarow:
This is necessary for compatibility with coroutines 1.6.x, where `offer` has a hard deprecation.

### DIFF
--- a/trace-encoder/src/main/java/com/squareup/tracing/TraceEncoder.kt
+++ b/trace-encoder/src/main/java/com/squareup/tracing/TraceEncoder.kt
@@ -1,12 +1,10 @@
 package com.squareup.tracing
 
-import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers.IO
 import kotlinx.coroutines.ObsoleteCoroutinesApi
 import kotlinx.coroutines.channels.Channel.Factory.UNLIMITED
-import kotlinx.coroutines.channels.ClosedSendChannelException
 import kotlinx.coroutines.channels.SendChannel
 import kotlinx.coroutines.channels.actor
 import kotlinx.coroutines.channels.consumeEach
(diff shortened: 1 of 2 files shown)


trace-encoder/src/main/java/com/squareup/tracing/TraceEncoder.kt
```diff
@@ -109,23 +109,10 @@ public class TraceEncoder(
...
   }
 
   private fun getTimestampNow(): Long = start.elapsedNow
-
-  /**
-   * Like [SendChannel.offer] but won't throw if the channel is closed.
-   */
-  private fun <T> SendChannel<T>.safeOffer(value: T) {
-    try {
-      trySend(value)
```
RBusarow, line 122: This line was originally `offer`, but got updated to `trySend` last time coroutines were updated.  The `try/catch` no longer does anything since `trySend` doesn't throw.

RBusarow:
Ktlint's failing for a file which no longer exists:
```
home/runner/work/workflow-kotlin/workflow-kotlin/workflow-ui/core-android/src/androidTest/java/com/squareup/workflow1/ui/container/BackStackContainerPersistenceTest.kt:1:1 class BackStackContainerPersistenceLifecycleTest should be declared in a file named BackStackContainerPersistenceLifecycleTest.kt (cannot be auto-corrected) (filename)
```
...which means that the Ktlint plugin's cache isn't being invalidated properly.  This isn't an issue with our CI caching -- I've seen it happen with this plugin locally as well.

As a band-aid, I've added `--rerun-tasks` to the step in CI so that it creates a new build cache, which will then be synced to Azure.  I'll remove the `--rerun-tasks` after that succeeds.

rjrjr (approved):

RBusarow:
> Ktlint's failing for a file which no longer exists:
> [...]

`--rerun-tasks` didn't actually work, but I ran the job once with `clean` and then reverted that change.  It passed, and seems to have fixed the caching issue.

PR #751, opened 2022-04-29
Re-enable render counting tests; Make Deterministic again

### LINKED ISSUE #745: renderPass benchmarks are no longer deterministic
Somehow merging the UI overhaul branch into main has made the benchmarks nondeterministic. The main thing I had to change was how the Loading UI worked, so that's the primary suspect.

I've commented them out until we can dig into it.

> @steve-the-edwards Looks like something is not deterministic, and I have to assume it's something I did. Maybe we can pair tomorrow.
> 
> ```
> java.lang.AssertionError: Congrats! You have improved the the number of Render Passes (lower better) by 4 (6.90%) in the 'Raven navigation (no initializing state) scenario'! Please update the expected value for your config. The value is now 54.
>```

_Originally posted by @rjrjr in https://github.com/square/workflow-kotlin/pull/744#issuecomment-1112489377_

steve-the-edwards:
Closes #745


benchmarks/performance-poetry/complex-poetry/src/androidTest/java/com/squareup/benchmarks/performance/complex/poetry/RenderPassTest.kt
```diff
@@ -174,7 +174,7 @@ class RenderPassTest {
       scenario: String,
       by: String? = null
     ) = fail(
-      "Uh Oh! You have worsened the $subject ${by?.let { "by $by " }}in $scenario!" +
+      "Uh Oh! You have worsened the $subject ${by?.let { "by $by " } ?: ""}in $scenario!" +
         " The value is now $value. This likely results in worse performance." +
```
rjrjr, line 178: For each of these messages, I found myself wishing it showed both the before and the after.

```suggestion
        " The value is now $value (was $oldValue}). This likely results in worse performance." +
```

steve-the-edwards: Thanks ya - that makes sense.

RBusarow (approved):

PR #753, opened 2022-05-03
Add the dependency-guard plugin

RBusarow:
https://github.com/dropbox/dependency-guard

This records every dependency (and its version) used by the runtime classpaths of our published artifacts.


rjrjr (approved):

buildSrc/src/main/java/dependency-guard.gradle.kts
```diff
@@ -0,0 +1,31 @@
...
+  // so we can't look them up now using the `configurations` property.
+  // Instead we can just hard-code "releaseRuntimeClasspath" for any module which has AGP applied.
+  // This is actually pretty robust, since if this configuration ever changes,
+  // dependency-guard will fail when trying to look it up.
+  extensions.findByType<TestedExtension>() != null -> listOf("releaseRuntimeClasspath")
+  // If we got here, we're either in an empty "parent" module without a build plugin
+  // (and no configurations), or we're in a vanilla Kotlin module.  In this case, we can just look
+  // at configuration names.
+  else -> configurations
+    .map { it.name }
```
ZacSweers, line 17: this is eagerly configurating your configurations, you want to use `.matching` and do the name check in there

buildSrc/src/main/java/dependency-guard.gradle.kts
```diff
@@ -0,0 +1,31 @@
...
+
+val configurationNames = when {
+  // record the root project's *build* classpath
+  project == rootProject -> listOf("classpath")
+  // Android variants and their configurations are added later in the configuration phase,
+  // so we can't look them up now using the `configurations` property.
+  // Instead we can just hard-code "releaseRuntimeClasspath" for any module which has AGP applied.
+  // This is actually pretty robust, since if this configuration ever changes,
+  // dependency-guard will fail when trying to look it up.
+  extensions.findByType<TestedExtension>() != null -> listOf("releaseRuntimeClasspath")
```
ZacSweers, line 12: This is also eager and could be made lazy, fyi. There's an `extensions.configure(...)` API that acts as a callback instead

buildSrc/src/main/java/dependency-guard.gradle.kts
```diff
@@ -0,0 +1,31 @@
...
+  // If we got here, we're either in an empty "parent" module without a build plugin
+  // (and no configurations), or we're in a vanilla Kotlin module.  In this case, we can just look
+  // at configuration names.
+  else -> configurations
+    .map { it.name }
+    // anything ending with 'runtimeClasspath' but not 'testRuntimeClasspath'
+    .filter { it.matches("^(?!test)\\w*[rR]untimeClasspath$".toRegex()) }
+}
+
+if (configurationNames.isNotEmpty()) {
```
ZacSweers, line 22: If you make the above lazy, you can just lazily apply this plugin and add to its configuration names with the existing extension if the plugin is already applied

PR #755, opened 2022-05-04
Makes `workflow-ui/compose/README.md` informative.

### LINKED ISSUE #458: [Design] Overall design for officially supporting Compose from Workflow

rjrjr:
`README.md` is now the permanent home of the [blog post](https://developer.squareup.com/blog/jetpack-compose-support-in-workflow/) that described in depth how this stuff would work, so that we can keep it up to date.

Closes #458

### DIFF
--- a/.markdownlint.rb
+++ b/.markdownlint.rb
@@ -6,8 +6,8 @@
 # Need to explicitly run all rules, so the per-rule configs below aren't used as an allowlist.
 all
 
-# Increase default line length from 80 to 100.
-rule 'MD013', :line_length => 100, :code_blocks => false, :tables => false
+# We don't care about line length, because it prevents us from pasting in text from sane tools.
+exclude_rule 'MD013'
 
 # Enable inline HTML.
 exclude_rule 'MD033'
--- a/workflow-ui/compose/README.md
+++ b/workflow-ui/compose/README.md
@@ -1,4 +1,445 @@
 # compose
 
-This module hosts the workflow-ui compose integration. For more info
-on this please see https://developer.squareup.com/blog/jetpack-compose-support-in-workflow.
+This module hosts the workflow-ui compose integration, and this file describes in detail how that integration works and why.
+It was originally published as a [blog post](https://developer.squareup.com/blog/jetpack-compose-support-in-workflow).
+
+## Timeline and Process
+
+Compose entered beta in the first half of 2020. Since we were all locked in our homes with no social lives, it was the perfect time to start exploring what integration between Compose and Workflows would look like. This was very experimental work — Compose APIs were changing drastically every two weeks. To say the least, it was not “ready for production.” However, it was important to suss out what sort of integration points were available to us, what API shapes felt natural, and where the rough edges were. In addition to figuring out our own adoption story, we have also been able to contribute a lot of feedback to Google (see our [case study](https://developer.android.com/stories/apps/square-compose)), and some of the features we initially wrote specifically for workflow integration ended up making it into the library (e.g. automatic subcomposition linking in child `View`s).
+
+## Goals and Non-Goals
+
+At Square, when we start a project, we like to enumerate and distinguish goals and explicit non-goals to delineate the scope of the project.
+
+### Goals
+
+- Allow Workflow screens to be written in Compose.
+- Allow interop in both directions using the usual idioms:
+  - Non-Compose Workflow screens to nest Compose-based Workflow screens.
+  - Compose-based Workflow screens to nest both Compose-based and non-Compose Workflow screens.
+- Provide a way for Compose-based apps to host a Workflow runtime and display Compose apps, e.g. a parallel entry point to how non-Compose apps can use `WorkflowLayout`.
+- Ensure that data flow through `CompositionLocal`s is propagated to child Compose-based screens, regardless of whether there are non-Workflow screens sitting in between them.
+
+### Non-Goals
+
+- Convert existing screens in our apps to Compose.
+- Provide design system components in Compose. (This is planned, but as a separate project that depends on this one.)
+- Anything with our own internal declarative UI toolkit, Mosaic (sunsetting it, integrating with it, or otherwise).
+
+## Major Components
+
+There are a few major areas this project needs to focus on to support Compose from Workflows: navigation, `ViewFactory` support, and hosting.
+
+### Navigation support
+
+Workflow isn’t just a state management library — Workflow UI includes navigation containers for things like [backstacks](https://github.com/square/workflow-kotlin/blob/v1.0.0-alpha18/workflow-ui/backstack-common/src/main/java/com/squareup/workflow1/ui/backstack/BackStackScreen.kt) and [modals](https://github.com/square/workflow-kotlin/blob/v1.0.0-alpha18/workflow-ui/modal-common/src/main/java/com/squareup/workflow1/ui/modal/HasModals.kt) (Support for complex navigation logic was one of our main drivers in writing the library — we outgrew things like Jetpack Navigation a long time ago.).
+
+Because these containers define “lifecycles” for parts of the UI, they need to communicate that to the Compose primitives through the AndroidX concepts of [`LifecycleOwner`](https://developer.android.com/reference/androidx/lifecycle/LifecycleOwner) and [`SavedStateRegistry`](https://developer.android.com/reference/kotlin/androidx/savedstate/SavedStateRegistry). When a composition is hosted inside an Android `View`, the [`AbstractComposeView`](https://developer.android.com/reference/kotlin/androidx/compose/ui/platform/AbstractComposeView) that bridges the two reads the [`ViewTreeLifecycleOwner`](https://developer.android.com/reference/androidx/lifecycle/ViewTreeLifecycleOwner) to find the nearest `Lifecycle` responsible for that view.
+
+> AndroidX recently introduced the concept of `ViewTree*` helpers — these are static getters and setters that set `View` tags, and search up the view hierarchy for those tags, to allow views to communicate in an ambient way with their children. [`ViewTreeLifecycleOwner`](https://developer.android.com/reference/androidx/lifecycle/ViewTreeLifecycleOwner), for example, allows any view to find the nearest `LifecycleOwner` by looking up the `View` tree. AndroidX `Fragment`s and `ComponentActivity`s set the `ViewTreeLifecycleOwner`, `ViewTreeSavedStateRegistry`, and other owners on their root views to support this.
+
+The `Lifecycle` is then observed, both to know when it is safe to restore state, and to know when to dispose the composition because the navigation element is going away. The view also reads the [`ViewTreeSavedStateRegistry`](https://developer.android.com/reference/androidx/savedstate/ViewTreeSavedStateRegistryOwner), wraps it in a [`SaveableStateRegistry`](https://developer.android.com/reference/kotlin/androidx/compose/runtime/saveable/SaveableStateRegistry), and provides it to the composition via the `LocalSaveableStateRegistry`. As per the [`SavedStateRegistry` contract](https://developer.android.com/reference/androidx/savedstate/SavedStateRegistry#consumeRestoredStateForKey(java.lang.String)), the registry is asked to restore the composition state as soon as the `Lifecycle` moves to the `CREATED` state. Any `rememberSaveable` calls in the composition will use this mechanism to save and restore their state.
+
+In order for this wiring to all work with Workflows, the Workflow navigation containers must correctly publish `Lifecycle`s and `SavedStateRegistry`s for their child views. The container already manages state saving and restoration via the Android `View` “hierarchy state” mechanism that all `View` classes participate in, so it’s not much of a stretch for them to support this new AndroidX stuff as well. The tricky part is that the sequencing of these different state mechanisms is picky and a little complicated, and we ideally want the Workflow code to support this stuff even if the Workflow view root is hosted in an environment that doesn’t (e.g. a non-AndroidX `Activity`).
+
+> None of the AndroidX integrations described in this section actually have anything to do with Compose specifically. They are required for any code that makes use of the AndroidX `ViewTree*Owners` from within a Workflow view tree. Compose just happens to rely on this infrastructure, so Workflow has to support it in order to support Compose correctly.
+
+#### `Lifecycle`
+
+For `LifecycleOwner` support, we need to think of anything that can ask the `ViewRegistry` for a view as a `LifecycleOwner`. This is because all such containers know when they are going to stop showing a particular child view (e.g. because the rendering type has changed, or a rendering is otherwise incompatible with the current one, and a new view must be created and bound). When that happens, they need to move the `Lifecycle` to the `DESTROYED` state to ensure the composition will be disposed.
+
+We can provide an API for this so that containers only need to make a single call to dispose their lifecycle, and everything else “just works.” And luckily, most developers building features with Workflow will never write a container directly but instead use [`WorkflowViewStub`](https://github.com/square/workflow-kotlin/blob/v1.0.0-alpha18/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowViewStub.kt), which we will make do the right thing automatically.
+
+[`WorkflowLifecycleOwner`](https://github.com/square/workflow-kotlin/blob/v1.0.0-alpha18/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLifecycleOwner.kt) is the class we use to nest `Lifecycle`s. A `WorkflowLifecycleOwner` is a `LifecycleOwner` with a few extra semantics. `WorkflowLifecycleOwner`s form a tree. The lifecycle of a `WorkflowLifecycleOwner` will follow its parent, changing its own state any time the parent state changes, until either the parent enters the `DESTROYED` state or the `WorkflowLifecycleOwner` is explicitly destroyed. Thus, a tree of `WorkflowLifecycleOwner`s will be synced to the root `Lifecycle` (probably an `Activity`), but a container can set the state of an entire subtree to `DESTROYED` early – this will happen whenever the container is about to replace a view. When a container can show different views over its lifetime, it must install a `WorkflowLifecycleOwner` on each view it creates and destroy that owner when its view is about to be replaced. A `WorkflowLifecycleOwner`s automatically finds and observes its parent `Lifecycle` by the usual method — searching up the view tree.
+
+#### `SavedStateRegistry`
+
+`SavedStateRegistry` is a bit more complicated because of the sequencing of lifecycle and “view instance state” calls with the `SavedStateRegistry` ones.
+
+Before all this AndroidX stuff, here’s how view state saving and restoration worked:
+
+`View` is instantiated. Constructor probably performs some initialization, e.g. setting default `EditText` values. An ID should be set.
+`View` is added as a child of a `ViewGroup` and attached to a window.
+After the hosting `Activity` moves to the `STARTED` state, `onRestoreInstanceState` is called for every view in the hierarchy (even the `View`’s children, if it has any). `EditText`s, for example,  use this callback to restore any previously-entered text. Because this callback happens after initialization, it looks to the app user like the text was just restored — they never see the initial value.
+`View` gets arbitrarily-many calls to `onSaveInstanceState`. The last one of these before the view is destroyed is what may be used to restore the view later.
+
+The old mechanism depends on `View`s having their IDs set. These IDs are used to associate state with particular views, since there is no other way to match view instances between different processes.
+
+Here’s how the view restoration system works with AndroidX’s `SavedStateRegistry`:
+
+`View` is instantiated. Because the view hasn’t been attached to a parent yet, it can’t use the `ViewTree*Owner` functions.
+`View` is eventually added to a `ViewGroup`, and attached to the window. Now the view has a parent, so the `onAttached` callback can search up the tree for the `ViewTreeLifecycleOwner`. It also looks for the `ViewTreeSavedStateRegistryOwner` — it can’t use it yet though.
+One or more `SavedStateProvider`s are registered on the registry associated with arbitrary string keys — these providers are simply functions that will be called arbitrarily-many times to provide saved values when the system needs to save view state.
+The `Lifecycle` is observed, as long as the view remains attached.
+When the lifecycle state moves to `CREATED`, the `SavedStateRegistry` can be queried. The view’s initialization logic can now call `consumeRestoredStateForKey` to read back any previously-saved values associated with string keys. If there were no values available, null will be returned and the view should fallback to some default value.
+When the view goes away, the `SavedStateProvider`s should be unregistered.
+
+Note the difference in when the restoration happens relative to the lifecycle states. The following table summarizes the differences between the instance state mechanism and `SavedStateRegistry`.
+
+| Instance state                                                                                                                                                    | `SavedStateRegistry`                                                                                                                                                                                                                                |
+|-------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| All views participate in this mechanism, and can’t opt-out.                                                                                                       | Views must opt-in by getting access to a `SavedStateRegistry` either via a `ViewTreeSavedStateRegistry` or some other way.                                                                                                                          |
+| Save/restore callbacks are built into the `View` base class as overridable methods. _Restoration is “push-based” — views are told when to restore themselves._    | Views must register saved state providers explicitly in order to get the save callbacks. _Restoration is “pull-based” — views must request previously-saved values at the appropriate time. Registry API requires coordination with Lifecycle API._ |
+| Restored after lifecycle `STARTED`.                                                                                                                               | Restored after lifecycle `CREATED`.                                                                                                                                                                                                                 |
+| Identified by hierarchy of view IDs.                                                                                                                              | Identified by string keys.                                                                                                                                                                                                                          |
+| IDs only need to be unique within their parent view.                                                                                                              | Keys must be unique within a given registry _(this scope is not as clearly defined, but usually means within the navigation “frame”)_.                                                                                                              |
+| Saved state is represented as `Parcelable`s, only not really because well-behaved views should actually return their state as subclasses of a special base class. | Saved state is represented as `Bundle`s. _Compatible with `Parcelable`s but much more convenient API for every-day use._                                                                                                                            |
+
+These differences make tying these together and supporting both from a single container a little complicated.
+Every container must support both of these mechanisms, but ideally using a single source of truth for saved state.
+Because containers and `WorkflowViewStub`s can exist anywhere in a view tree, they must be able to identify themselves to the different state mechanisms appropriately. It turns out that the legacy instance state approach of using view IDs is capable of supporting the registry string key approach, but it’s not really feasible the other way around. So the source of truth needs to be the instance state, and the registry state is stored in and restored from that.
+Because the `SavedStateRegistry` contract says that it must be consumable as soon as the lifecycle is in the `CREATED` state, containers must also be able to control the lifecycle to ensure that it isn’t moved to that state until they’ve had a chance to actually seed the registry with restored data.
+
+The last two points form a cycle: we don’t get the `onRestoreInstanceState` callback until we’re in the `STARTED` state, but we can’t advance our children’s lifecycle past the `CREATED` state until we have read the registry state out of the instance state and seeded the registry. So the sequence we need to implement is:
+
+1. Create a `Lifecycle` to provide to our children, but hold it in the `INITIALIZED` state. Create a `SavedStateRegistry` to provide to our children as well.
+1. In `onRestoreInstanceState`, read both the instance state for our children, and the registry state.
+1. Seed our `SavedStateRegistry` with the restored state.
+1. Advance state to `CREATED`. Children will consume state from the registry.
+1. Advance state to `STARTED`. (This can actually be done in a single step with the previous one, the Lifecycle APIs ensure every state is hit.)
+1. Invoke children’s `onRestoreInstanceState` methods.
+
+We have looked at a few ways of implementing this:
+
+1. Single source of state: `SavedStateRegistry`. `BackStackContainer` only uses `SavedStateRegistry` to save and restore state, and implements support for classic view state on top of the newer registry APIs. This approach is infeasible because it requires emulating the way that state is namespaced with `View` IDs with string keys, and that is surprisingly difficult.
+1. Single source of state: view state. `BackStackContainer` only uses view state, and implements support for the newer registry APIs on top of classic view state. While easier to implement than (1), it requires changing `WorkflowLifecycleOwner` to give the container more control over the lifecycle to comply with `SavedStateRegistry`'s contract about when states can be restored.
+1. Use both. `BackStackContainer` uses the classic view state hooks to manage classic view state, and uses `SavedStateRegistry` hooks to manage registry state. This allows each mechanism to keep its advantage, and doesn't require emulating one's behavior with the other.
+
+### `ViewFactory` support
+
+Again, Workflow UI is built around the [`ViewFactory`](https://github.com/square/workflow-kotlin/blob/v1.0.0-alpha18/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewFactory.kt) interface, functions that build and update classic Android View instances for each type of view model rendered by a Workflow tree. Because Compose supports seamless integration from and to the classic Android `View` world, technically we don’t really _need_ to do anything to allow people to write Compose code inside `ViewFactory`s, at least to get 90% support. However, by providing some more convenient APIs, we not only remove some boilerplate, but also create the opportunity for some simplifications. There are also some edge cases that require a little more effort that we actually _do_ need to build support into Workflows for.
+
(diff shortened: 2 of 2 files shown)


steve-the-edwards (approved):

rjrjr:
Cool, the markdown lint was actually useful!

rjrjr:
Useful is strong. Really, that linter is left over from https://github.com/square/workflow, where it's mainly about the `docs/` directory. A smart person would have just deleted it, but instead I fixed all the line-to-long errors.

Need to take a pass to update the terminology and make it reflect what did and didn't get implemented, but there are more important holes to fill first. Tracking that at #787.

rjrjr:
Okay, this is stupid. I'm going to delete the second SHA and just turn off the line length check.

PR #789, opened 2022-05-05
Add in Latency Benchmark for Workflow

steve-the-edwards:


benchmarks/performance-poetry/complex-benchmark/src/main/java/com/squareup/benchmarks/performance/complex/poetry/benchmark/ComplexPoetryBenchmarks.kt
```diff
@@ -110,50 +112,121 @@ class ComplexPoetryBenchmarks {
...
+   * Frame-Latency-N: is the trace between passing the Rendering to the view layer and the
+   *  'post frame rendered callback' for the Nth frame in the scenario. In other words, this traces
+   *  the time it takes from a Rendering produced by Workflow to process through the Workflow UI
+   *  layer and then be rendered in the next frame.
+   *
+   * XScreen-onY-Z: is the time between the execution of event handler 'onY' and the production of
+   *   the next root Rendering by Workflow for the Zth instance of the 'onY' handler on X Screen.
+   *   In other words, this measures the time Workflow takes in processing a UI event into a new
+   *   Rendering. This will be similar to the render pass traced above, but more comprehensive to
+   *   include all of the event handling time.
```
RBusarow, line 153: These are great.  They make perfect sense to me.

RBusarow (approved):
I don't know very much about the libraries being used, but this LGTM!

benchmarks/performance-poetry/complex-poetry/src/main/java/com/squareup/benchmarks/performance/complex/poetry/instrumentation/WorkflowUiEventsTracer.kt
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  private const val EVENT_COOKIE = 12345
+}
+
+fun StanzaScreen.trace(): StanzaScreen {
+  return copy(
+    onGoUp = {
+      WorkflowUiEventsTracer.startTraceForEvent(WorkflowUiEvent("onGoUp", "StanzaScreen"))
+      onGoUp()
```
steve-the-edwards, line 69: Notes from talking with @pyricau - wrap just the callback as its on the main thread this will be in the same frame and we can measure exactly what WOrkflow is doing - then instrument the main thread and check the perfetto trace to ensure we got everything.
```suggestion
      trace {
           onGoUp()
       }
```

steve-the-edwards (commented):
Also from discussing with @pyricau - can find a way to wrap the response to Workers as well so we get exactly what's done by Workflow. Play around with the hooks and then confirm on perfetto.

PR #790, opened 2022-05-06
Support late/lazy configuration creation in dependency guard

RBusarow:
Wraps dependency guard configuration in `afterEvaluate { }`.  This ensures that configurations have been created before we look through the list for runtime classpaths.

Without this change, KMP modules get no dependency reports since they have no configurations when the convention plugin is applied.

### DIFF
--- a/buildSrc/src/main/java/dependency-guard.gradle.kts
+++ b/buildSrc/src/main/java/dependency-guard.gradle.kts
@@ -1,31 +1,38 @@
 import com.android.build.gradle.TestedExtension
 import com.dropbox.gradle.plugins.dependencyguard.DependencyGuardPluginExtension
 
-val configurationNames = when {
-  // record the root project's *build* classpath
-  project == rootProject -> listOf("classpath")
-  // Android variants and their configurations are added later in the configuration phase,
-  // so we can't look them up now using the `configurations` property.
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/java/dependency-guard.gradle.kts
```diff
@@ -1,31 +1,38 @@
...
-  // This is actually pretty robust, since if this configuration ever changes,
-  // dependency-guard will fail when trying to look it up.
-  extensions.findByType<TestedExtension>() != null -> listOf("releaseRuntimeClasspath")
-  // If we got here, we're either in an empty "parent" module without a build plugin
-  // (and no configurations), or we're in a vanilla Kotlin module.  In this case, we can just look
-  // at configuration names.
-  else -> configurations
-    .map { it.name }
-    // anything ending with 'runtimeClasspath' but not 'testRuntimeClasspath'
-    .filter { it.matches("^(?!test)\\w*[rR]untimeClasspath$".toRegex()) }
```
RBusarow, line 19: This regex works for the minimal cases, but incorrectly matches to `___TestRuntimeClasspath` configs for other source sets, like `jmhTestRuntimeClasspath` or `jvmTestRuntimeClasspath`.

Issue #370, opened 2021-03-12
Thoughts on decoupling render pass from view updates, and better support immediate state change / output


rjrjr:
Distilling [slack monologue thread](https://workflow-community.slack.com/archives/CHTFPR277/p1615574769018700):

 * It'd be nice to have a way to conflate renderings when they're coming "too fast". 
 * Thinking in particular of attractive techniques people resort to in order to have a workflow immediately emit output "without rendering", leading to cascades of dozens of render passes kicked off by a single user event.
 * Typical scheme is to fire off a Worker<Unit> on first render and have it update state / emit output.
 * Can we find a way to prevent emitting renderings until the action queue is empty?
 * Maybe couple this with the ability to emit output from initialState

0legg:
Both `initialState` and `onPropsChanged` are similar to the regular `action` minus emitting an output. Making both methods return `WorkflowAction` instead of `WorkflowState` will enable codebase unification (when initial state isn't a special case) and will reduce the described ugliness.

0legg:
One of usecases where it could be valuable is the case when the workflow is outputting some function of its state (extremely, workflow output is the same model as a state). It's a legit usecase, especially for dumb child workflows that delegate most of handling to the parent workflow. Right now consumers have to implement the ugly hack with one-shot worker, but it's not elegant and could be racy or error-prone.

rjrjr:
@zach-klippenstein I'm going to first play around with the simplest flavor of this I can think of: simply suppress emitting a rendering while the actionSink is not empty. I expect to find out that it's not simple at all, but at least it's a place to start digging.

I'm less excited about emitting output from `initialState`, if only because of the unspeakable migration nightmare it would be. I guess my hope is that I can find some low hanging fruit to reduce the penalty when people hack up their own mechanisms to achieve that effect.

I think it's well worth doing because I'm finding it's so hard to avoid an intial render storm as a deep workflow tree is set up, and all kinds of async workers are subscribed to at set up. It's hard to optimize things to avoid an initial redundant update from such workers (if you can even be confident they're redundant), and why should people have to fret about that in the first place?

zach-klippenstein:
Could you give a more specific example of a workflow that emits output as a function of state @0legg?

0legg:
@zach-klippenstein Please take a look at `AddressWorkflow` at internal repo, this is exact the case of state and output being the same.

steve-the-edwards:
[...]

PR #795, opened 2022-05-18
Optimize Action Handling by handling all possible actions within a reasonable frame rate.

steve-the-edwards:
We introduce this change via a `RuntimeConfig` sealed interface that we can optionally pass to `renderWorkflowIn` which is then forwarded to the `WorkflowRunner` which makes the actual decisions about how to run. Renaming `nextOutput` to `processActions` (a bit more general as its doing more than getting an Output in the OutputT sense), we also change its return type to a sealed interface which can handle `PropsUpdated` and `TimeoutForFrame`. This means we erase the `OutputT` from the internal implementation's signature but I have not seen any issues as a result of that as it is used internally or after we cast back to a `WorkflowOutput<T>`.

--- testing discussion:

With multiple runtime configurations we do the `RenderWorkflowInTest`s and `WorkflowRunnerTest`s once with both runtimes.

To do this we've built a simple ParameterizedTestRunner for kotlin common test code.

For all the other tests, we use a property (`-Pworkflow.runtime=timeout`) to override the baseline default runtime. For jvm unit tests, this will be pulled into the jvm fork running the tests and then the test runtime will use this runtime as specified. For Android tests, we have to build the app with this property and then run the integration tests against that app built with a particular runtime.

Note that because we introduce this frame timeout in order to process multiple actions, we provide a rendering only as fast as the frame rate times out. So when we run the tests with the modified runtime we need to account for this. This means if we expect a change in UI tests, we will need to wait for the frame timeout before verifying, which means we have to brute force Espresso a bit.

Things are more complicated for Compose because the Recomposer also has its own dispatcher, so we don't just nee to wait for the Workflow runtime's frame timeout, we also then need to wait for the Recomposer to finish well ... recomposing.  We don't want to use `rememberCoroutineScope` for our Workflow runtime because while that will tie it to the scope of the composition, we can't control the dispatcher in that scope easily in order to advance past the Workflow runtime's frame timeout.

Those changes remain so that the tests support both runtimes.

--- back to implementation discussion:

Note further, we are currently using a hard-coded frame timeout of 30ms because its somewhere between 60fps's 16ms and @pyricau's research for touch interaction at 100ms. In tuning this value it also produced the best reduction in render passes of the values I tried (10ms, 16ms, 30ms). Theoretically much better than this hard coded value will be to integrate a monotonic frame clock, which for Android can be tied to the Android Choreographer. I noticed this is what molecule wraps to time the emitted view model updates - we should take a look there and see if we can apply the same mechanism. See https://github.com/cashapp/molecule#frame-clock . Fine tuning this is covered by #810 .

We also introduce `currentTimeMillis()` expect so that we can calculate the timeout length on all platforms as we process actions.

Addresses #370 .


samples/compose-samples/src/androidTest/java/com/squareup/sample/compose/hellocompose/HelloComposeTest.kt
```diff
@@ -25,13 +26,19 @@ class HelloComposeTest {
...
-    composeRule.onNodeWithText("Goodbye")
-      .assertIsDisplayed()
-      .performClick()
-    composeRule.onNodeWithText("Hello")
-      .assertIsDisplayed()
+    runBlocking {
+      composeRule.onNodeWithText("Hello")
+        .assertIsDisplayed()
+        .performClick()
+      composeRule.awaitIdle()
```
steve-the-edwards, line 33: @RBusarow - after the change in the runtime to process multiple actions we introduce a suspending yield - for the onTimeout select clause even if there is only 1 action to process. This means that with the main immediate dispatcher tied to the IdlingResource watcher we don't get immediate synchronous results to assert on.

We have to wait and then advance the time in order to stabilize these tests - I need to talk with you about how to do this in a better general way.

steve-the-edwards:
Will still need a `currentTimeMillis` implementation for ios. cinterop or NSDate or kotlinx-datetime (this last one seems heavy?)

@bnvinay92 Do you have any good thoughts for this for `ios` target?

rjrjr:
> We should keep running _all_ these tests for both runtimes, but I haven't thought of a good way to to do that yet and it's a lot of work to set up the test matrix for runtime versions.

Maybe we should just go ahead and copy paste, set up two separate shards?

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowRunner.kt
```diff
@@ -53,18 +59,62 @@ internal class WorkflowRunner<PropsT, OutputT, RenderingT>(
...
-  // Tick _might_ return an output, but if it returns null, it means the state or a child
-  // probably changed, so we should re-render/snapshot and emit again.
-  suspend fun nextOutput(): WorkflowOutput<OutputT>? = select {
+  /**
+   * Stop processing and go to render on 1 of 3 conditions:
+   * 1. Props have changed.
+   * 2. tick produces an Output to be handled.
+   * 3. The frame 'times out'.
+   */
+  suspend fun processActions(): WorkflowOutput<OutputT>? {
```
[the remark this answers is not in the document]
rjrjr: There are still references to `nextOutput()` in the big comment above. Any others?

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowRunner.kt
```diff
@@ -53,18 +59,62 @@ internal class WorkflowRunner<PropsT, OutputT, RenderingT>(
...
+      while (processingResult == null && frameTimeLeft > 0) {
+        processingResult = select {
+          // N.B. that select clauses use declaration order to break ties, so the timeout goes first.
+          onTimeout(timeMillis = frameTimeLeft) frameTimeout@{
+            // return FrameTimeout so we know that it's high time to render now!
+            FrameTimeout
+          }
+          onPropsUpdated()
+          // Tick the workflow tree.
+          rootNode.tick(this)
```
rjrjr, line 100:
```suggestion
          // Have the workflow tree fill out the select to wait for the next event / output.
          rootNode.tick(this)
```

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
```diff
@@ -75,14 +109,28 @@ class RenderWorkflowInTest {
...
+      runtimeConfig = RuntimeConfig(processMultipleActions)
+    ) {}
     testScope.advanceUntilIdle()
 
     assertFalse(sideEffectWasRan)
   }
 
   @Test
-  fun `side effects from initial rendering in non-root workflow are never started when scope cancelled before start`() { // ktlint-disable max-line-length
+  fun `side effects from initial rendering in non-root workflow are never started when scope cancelled before start baseline`() =
```
rjrjr, line 124: Seems like this would be a good use of a parameterized test / `@Parameters`. (Fine to wait for follow up, if ever.)

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
```diff
@@ -46,27 +47,60 @@ class RenderWorkflowInTest {
...
+    val renderings = renderWorkflowIn(
+      workflow = workflow,
+      scope = pausedTestScope,
+      props = props,
+      runtimeConfig = RuntimeConfig(processMultipleActions)
+    ) {}
     assertEquals("props: foo", renderings.value.rendering)
   }
 
+  /* ktlint-disable max-line-length */
```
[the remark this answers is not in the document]
steve-the-edwards: You are right it's not needed!

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/internal/SubtreeManagerTest.kt
```diff
@@ -167,9 +168,9 @@ internal class SubtreeManagerTest {
       assertFalse(tickOutput.isCompleted)
 
       eventHandler("event!")
-      val update = tickOutput.await()!!.value
+      val update = tickOutput.await().value
 
-      val (_, output) = update.applyTo("props", "state")
+      val (_, output) = update!!.applyTo("props", "state")
```
rjrjr, line 173: Nit: seems like it'd be nicer to put the `!!` in the declaration, like it was originally. Same for the two `initialAction!!` bits below.

steve-the-edwards:
> 

Hmm. I'm loathe to copy-paste. But I think your point about multiple shards is good, because we could move the runtime choice from being in Test Util code to being in the executing environment. As in maybe we just use a system property for the sample.

workflow-runtime/src/iosMain/kotlin/com.squareup.workflow1.internal/SystemUtils.kt
```diff
@@ -0,0 +1,6 @@
+package com.squareup.workflow1.internal
+
+import platform.Foundation.NSDate
+import platform.Foundation.timeIntervalSince1970
+
+actual fun currentTimeMillis(): Long = (NSDate().timeIntervalSince1970 * 1000).toLong()
```
rjrjr, line 6: Should this be `CACurrentMediaTime()` instead? https://stackoverflow.com/questions/358207/iphone-how-to-get-current-milliseconds

> [timeIntervalSince1970] can return an earlier value when you call it after couple of miliseconds (e.g. call it consecutively and you might not end up with an increasing sequence)

Would require a dependency on `QuartzCore`, don't know if that's okay.

rjrjr (approved):
This is magnificent.

I can't wait to see what happens when we tie it to choreographer. Not clear to me that will actually be a win, at least for our apps. OTOH the option to configure it that way could be valuable for other, display intensive ones? Maybe we'll see a win from being able to keep our update schedule in phase with Choreographer: e.g. the timeout becomes based on the number of ticks from it, rather than the clock.

rjrjr:
> > 
> 
> Hmm. I'm loathe to copy-paste. But I think your point about multiple shards is good, because we could move the runtime choice from being in Test Util code to being in the executing environment. As in maybe we just use a system property for the sample tests.

Yeah, I just meant copy/paste of the github config -- I had assumed the tests were already configured from the environment and that GHA matrix itself was proving tricky. Certainly would not want to duplicate the test suites.

benchmarks/performance-poetry/complex-poetry/src/main/java/com/squareup/benchmarks/performance/complex/poetry/PerformancePoetryActivity.kt
```diff
@@ -129,12 +117,52 @@ class PerformancePoetryActivity : AppCompatActivity() {
...
+      (debuggable || profileable)
+    ) {
+      // Add main message tracing to see everything happening in Perfetto.
+      // See https://py.hashnode.dev/tracing-main-thread-messages for more background.
+      Looper.getMainLooper().setMessageLogging { log ->
+        if (!Trace.isEnabled()) {
+          return@setMessageLogging
+        }
+        if (log.startsWith('>')) {
+          val label = buildSectionLabel(log).take(127)
```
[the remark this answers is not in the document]
steve-the-edwards: max length for Perfetto label name.

samples/compose-samples/build.gradle.kts
```diff
@@ -3,6 +3,7 @@ plugins {
   `kotlin-android`
   `android-sample-app`
   `android-ui-tests`
+  `compose-ui-tests`
```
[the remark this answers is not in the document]
steve-the-edwards: I just added it.

samples/compose-samples/src/androidTest/java/com/squareup/sample/compose/hellocompose/HelloComposeTest.kt
```diff
@@ -25,13 +28,21 @@ class HelloComposeTest {
...
-    composeRule.onNodeWithText("Goodbye")
-      .assertIsDisplayed()
-      .performClick()
-    composeRule.onNodeWithText("Hello")
-      .assertIsDisplayed()
+    runBlocking {
+      composeRule.onNodeWithText("Hello")
+        .assertIsDisplayed()
+        .performClick()
+      composeRule.settleForNextRendering()
```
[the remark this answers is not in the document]
steve-the-edwards: I mean it's mostly a heuristic approach, not sure how rock solid it will be. But it was from brute force trial and error.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowRunner.kt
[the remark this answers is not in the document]
steve-the-edwards: Good point, I will take a look and update.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
```diff
@@ -133,7 +138,7 @@ public fun <PropsT, OutputT, RenderingT> renderWorkflowIn(
       // It might look weird to start by consuming the output before getting the rendering below,
       // but remember the first render pass already occurred above, before this coroutine was even
       // launched.
-      val output = runner.nextOutput()
+      val output: WorkflowOutput<OutputT>? = runner.processActions()
```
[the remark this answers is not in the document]
steve-the-edwards: internal to the `processActions` implementation, when we would otherwise be getting `null` we don't know the `OutputT` type of which we have a null instance. I don't think it will matter.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RuntimeConfig.kt
```diff
@@ -0,0 +1,23 @@
...
+ * @param frameTimeoutMs: How long before the next rendering should be provided if
+ * [processMultipleActions] is true.
+ * Some research has shown touch events need 100ms latency on mobile, we want to be sure to be
+ * less than that - what exactly this should be though is not clear.
+ * TODO (https://github.com/square/workflow-kotlin/issues/810): fine-tune this value, or
+ * hook up to a frame clock.
+ */
+public data class RuntimeConfig(
+  public val processMultipleActions: Boolean = false,
+  public val frameTimeoutMs: Long = 30L
```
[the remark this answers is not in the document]
steve-the-edwards: Oh that sealed class is very clear - yes I like that.

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
steve-the-edwards: - [ ] Use Junit @ParametrizedTest for the runtime tests.

workflow-runtime/src/iosMain/kotlin/com.squareup.workflow1.internal/SystemUtils.kt
steve-the-edwards: I'm not sure. We would need specific iOS focused validation of this runtime anyway.

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
steve-the-edwards: Welp, couldn't use JUnit because in common code - but made our own

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/internal/WorkflowRunnerTest.kt
```diff
@@ -34,45 +36,86 @@ internal class WorkflowRunnerTest {
     scope.cancel()
   }
 
-  @Test fun `initial nextRendering() returns initial rendering`() {
+  @Test fun `initial nextRendering returns initial rendering baseline`() =
```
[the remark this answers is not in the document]
steve-the-edwards: - [ ] Yes we can pass config in.

workflow-config/README.md
```diff
@@ -0,0 +1,17 @@
...
+property from the gradle build to choose a runtime for JVM tests.
+
+e.g. add "-Pworkflow.runtime=timeout".
+
+Note that this will only work for jvm based tests, use config-android for applications and
+application tests.
+
+## workflow-config:config-android
+
+Configuration for the Workflow Runtime when building an applications.
```
RBusarow, line 15:
```suggestion
Configuration for the Workflow Runtime when building an application.
```

workflow-config/config-android/gradle.properties
```diff
@@ -0,0 +1,3 @@
+POM_ARTIFACT_ID=workflow-config-android
+POM_NAME=Workflow Runtime Android Configuration
+POM_PACKAGING=jar
```
RBusarow, line 3:
```suggestion
POM_PACKAGING=aar
```

workflow-config/config-jvm/gradle.properties
```diff
@@ -0,0 +1,3 @@
+POM_ARTIFACT_ID=workflow-config-jvm
+POM_NAME=Workflow Runtime JVM Configuration
+POM_PACKAGING=aar
```
RBusarow, line 3: This should be a `jar` and the `artifacts.json` will need to be updated.

And I'm realizing that the artifacts plugin should probably have a check which compares the packaging to the module type.

rjrjr (approved):

workflow-config/config-jvm/gradle.properties
steve-the-edwards: Whoops. I think when I changed them I changed the jvm one to a jar rather than vice versa.

samples/compose-samples/src/main/java/com/squareup/sample/compose/hellocompose/App.kt
```diff
@@ -21,8 +22,13 @@ import com.squareup.workflow1.ui.plus
 private val viewEnvironment = ViewEnvironment.EMPTY + ViewRegistry(HelloBinding)
 
 @Composable fun App() {
+  println("%%%% In Hello Compose: ${AndroidRuntimeConfigTools.getAppWorkflowRuntimeConfig()}")
```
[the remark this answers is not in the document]
steve-the-edwards: Nope. Will remove. Sorry.

workflow-config/README.md
```diff
@@ -0,0 +1,17 @@
+# workflow-config
+
+## workflow-config:config-jvm
+
+Configuration for the Runtime while running on the JVM. This allows one to specify a project
```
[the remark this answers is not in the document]
steve-the-edwards: Good point. Will update.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RuntimeConfig.kt
```diff
@@ -0,0 +1,22 @@
+package com.squareup.workflow1
+
+/**
+ * A specification of the Workflow Runtime.
```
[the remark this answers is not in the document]
steve-the-edwards: Sounds smart.

workflow-config/README.md
```diff
@@ -0,0 +1,27 @@
...
+`val runtimeConfig = JvmTestRuntimeConfigTools.getTestRuntimeConfig()`
+
+## workflow-config:config-android
+
+Configuration for the Workflow Runtime when building an application.
+
+Add "-Pworkflow.runtime=timeout" to the gradle command for building the app.
+
+Your application will also need to fetch the configured [RuntimeConfig] and pass it to
+[renderWorkflowIn]. You can fetch it with the utility provided:
```
rjrjr, line 25:
```suggestion
[renderWorkflowIn] or [renderAsState]. You can fetch it with the utility provided:
```

Do the square braces here actually work?

PR #797, opened 2022-05-24
Convert :workflow-core to kmp

bnvinay92:


RBusarow (commented):
Thanks for this!  It looks great!

I ran this branch's snapshot through the CI on our main project and everything's green.  So, as soon as we've added some KDoc to `WorkflowIdentifierType`, I can approve.

bnvinay92:
> Thanks for this! It looks great!
> 
> I ran this branch's snapshot through the CI on our main project and everything's green. So, as soon as we've added some KDoc to `WorkflowIdentifierType`, I can approve.

Thanks. Added KDoc and removed the redundant dependency [here](https://github.com/square/workflow-kotlin/pull/797/commits/78bbd5fb04c9bc6d8b0b74ebe802466cadcb65f9).

RBusarow (approved):
Great, thanks!

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/Void.kt
```diff
@@ -0,0 +1,3 @@
+package com.squareup.workflow1
+
+internal class Void private constructor()
```
steve-the-edwards, line 3: kdoc on why this is here will also be helpful!

steve-the-edwards (approved):
🥳

PR #798, opened 2022-06-01
Add High Frequency Event Tests and Benchmarks and cleanup

steve-the-edwards:
This simulates an 'action storm' where we have to process a lot of actions at once.
Also, cleaning up benchmarks.

All baseline data now saved.


benchmarks/performance-poetry/complex-poetry/src/main/java/com/squareup/benchmarks/performance/complex/poetry/PerformancePoemWorkflow.kt
```diff
@@ -89,39 +93,73 @@ class PerformancePoemWorkflow(
...
-            TraceableWorker.from("PoemLoading") {
-              isLoading.value = true
-              delay(simulatedPerfConfig.complexityDelay)
-              // No Output for Worker is necessary because the selected index
-              // is already in the state.
+          if (repeat > 0) {
+            // Running a flow that emits 'repeat' number of times
+            context.runningWorker(
+              flow {
+                while (repeat > 0) {
```
RBusarow, line 125: Because of the check above, this is always true.

steve-the-edwards: good point, but I think this is clearer than `while (true)`. Maybe not.

The point is the flow needs to keep running and emitting as long as we are in this state

steve-the-edwards: Updated. 

```
 if (repeat > 0) {
            // Running a flow that emits 'repeat' number of times
            context.runningWorker(
              flow {
                while (true) {
                  // As long as this Worker is running we want to be emitting values.
                  delay(2)
                  emit(repeat)
                }
              }.asTraceableWorker("EventRepetition")
            ) {
```

WDYT @RBusarow ?

RBusarow (approved):

benchmarks/performance-poetry/complex-poetry/src/main/java/com/squareup/benchmarks/performance/complex/poetry/PerformancePoemWorkflow.kt
RBusarow: LGTM!

PR #799, opened 2022-06-02
Convert :workflow-runtime to KMM

bnvinay92:
- Replaces SendChannel#offer with SendChannel#trySend.
- org.jetbrains.kotlinx:kotlinx-coroutines-test:1.5.1 only ships a jvm artifact so bumped only this test dependency in commonTest to 1.6.1 (which ships artifacts for all platforms).


workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
```diff
@@ -34,40 +36,31 @@ import kotlin.test.assertTrue
...
+  private val testScope = TestScope(UnconfinedTestDispatcher())
 
   @Test fun `initial rendering is calculated synchronously`() {
     val props = MutableStateFlow("foo")
     val workflow = Workflow.stateless<String, Nothing, String> { "props: $it" }
     // Don't allow the workflow runtime to actually start.
-    expectedSuccessScope.pauseDispatcher()
-    val renderings = renderWorkflowIn(workflow, expectedSuccessScope, props) {}
+
+    val renderings = renderWorkflowIn(workflow, TestScope(), props) {}
```
steve-the-edwards, line 54: Can't we use `pausedTestScope` here?

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
```diff
@@ -188,6 +180,7 @@ class RenderWorkflowInTest {
     }
     sink.send("unchanging state")
     sink.send("unchanging state")
+    pausedTestScope.advanceUntilIdle()
```
steve-the-edwards, line 183: why is this here? I don't think `pausedTestScope` is used in this test is it?

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/internal/WorkflowRunnerTest.kt
```diff
@@ -133,19 +129,21 @@ internal class WorkflowRunnerTest {
...
-    dispatcher.resumeDispatcher()
-
     // The order in which props update and workflow update are processed is deterministic, based
     // on the order they appear in the select block in nextOutput.
-    val firstOutput = scope.async { runner.nextOutput() }
-      .getCompleted()
+    val firstOutput = with(scope.async { runner.nextOutput() }) {
+      scope.advanceUntilIdle()
+      getCompleted()
+    }
```
steve-the-edwards, line 137: We could throw this in a utility test extension function.

steve-the-edwards (approved):
Thanks @bnvinay92  - this is so exciting!

LGTM. @RBusarow also to take a look as he is tracking along with these close.y

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/internal/WorkflowRunnerTest.kt
RBusarow: +1

There's some subtlety to this, so giving it a function with a good name would probably make it easier to read.

RBusarow (approved):
I had the same nitpicks as @steve-the-edwards, but overall this is great!

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/internal/WorkflowRunnerTest.kt
bnvinay92: Tried this: 268daedd27bf086efe4f39847112db9a7a41810b

PR #802, opened 2022-06-09
Update Kotlin Check task for Core/Runtime KMP

steve-the-edwards:
This is a maintenance burden until we find a longer term solution.

Also note we will need to use a different Github Action runner for the ios tests.

### DIFF
--- a/.github/workflows/kotlin.yml
+++ b/.github/workflows/kotlin.yml
@@ -78,7 +78,7 @@ jobs :
           gradle-dependencies-cache-key : |
             gradle/libs.versions.toml
           arguments : |
-            test apiCheck artifactsCheck dependencyGuard lint ktlintCheck jvmWorkflowNodeBenchmarkJar --no-daemon --stacktrace --continue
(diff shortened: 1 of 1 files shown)


.github/workflows/kotlin.yml
```diff
@@ -78,7 +78,7 @@ jobs :
           gradle-dependencies-cache-key : |
             gradle/libs.versions.toml
           arguments : |
-            test apiCheck artifactsCheck dependencyGuard lint ktlintCheck jvmWorkflowNodeBenchmarkJar --no-daemon --stacktrace --continue
+            test apiCheck artifactsCheck dependencyGuard lint ktlintCheck jvmWorkflowNodeBenchmarkJar :workflow-runtime:jvmTest :workflow-core:jvmTest --no-daemon --stacktrace --continue
```
bnvinay92, line 81: Perhaps we could replace the `test` task with `check`?

bnvinay92: Managed to get a red-green on CI with `check` here: https://github.com/square/workflow-kotlin/pull/806

PR #804, opened 2022-06-09
Makes `OverlayDialogFactory` more consistent with `ScreenViewFactory`

rjrjr:
I'm sad about all the copy/paste API here, but every time I try to consolidate things I wind up with an opaque mess of parameter types that usually breaks things. A Dialog is not a View, both classes are dying anyway one hopes, and attempts to factor out a pretty shared wrapper API for theme seem like a bad use of time.

Also note:

* the `BackButtonScreen` wrapping hack in `ModalScreenOverlayDialogFactory` was redundant with the `onBackPressedHandler =` calls on the following lines, removed
* implementors of `ModalScreenOverlayDialogFactory` now have access to the entire `ScreenViewHolder` for the content, and can see its actual type
* `ModalScreenOverlayOnBackPressed` is now `ModalScreenOverlayBackButtonHelper`, and eliminates all direct dependencies on Jetpack's  `OnBackPressedDispatcherOwner` from `ModalScreenOverlayDialogFactory` -- necessary for we poor souls who remain incompatible with it.


workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewHolder.kt
```diff
@@ -50,6 +50,25 @@ public interface ScreenViewHolder<in ScreenT : Screen> {
...
+ * Each [ScreenViewRunner] instance is paired with a single [View] instance,
+ * its neighbor in a [ScreenViewHolder].
+ *
+ * This is the interface you'll implement directly to update Android view code
+ * from your [Screen] renderings. A [ScreenViewRunner] serves as the strategy
+ * object of a [ScreenViewHolder] instantiated by a [ScreenViewFactory] -- the
+ * runner provides the implementation for the holder's [ScreenViewHolder.show]
+ * method.
+ */
+@WorkflowUiExperimentalApi
```
rjrjr, line 64: Just moved this to a more appropriate home.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/AlertOverlayDialogFactory.kt
```diff
@@ -53,29 +52,26 @@ public open class AlertOverlayDialogFactory : OverlayDialogFactory<AlertOverlay>
...
         }
       }
-  }
-
-  open override fun updateDialog(
-    dialog: Dialog,
-    rendering: AlertOverlay,
-    environment: ViewEnvironment
-  ) {
-    (dialog as AlertDialog).apply {
```
rjrjr, line 63: See how we're able to get rid of this cast? That's a big benefit of this approach.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/DialogSession.kt
```diff
@@ -0,0 +1,159 @@
...
+import com.squareup.workflow1.ui.androidx.WorkflowSavedStateRegistryAggregator
+
+/**
+ * Used by [LayeredDialogSessions] to manage lifecycle and view persistence concerns for an
+ * [OverlayDialogHolder], as well as enforcing modal behavior.
+ *
+ * @param modal if true, ignore keyboard and touch events when [CoveredByModal] is also true
+ */
+@WorkflowUiExperimentalApi
+internal class DialogSession(
```
rjrjr, line 26: This is mostly just a rename of `DialogHolder`, don't know why git can't see that.

rjrjr: I'd suggest that reviewers view the raw file, copy it, and then use AS's _Compare to Clipboard_ feature against `DialogHolder` on `main`.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/ModalScreenOverlayBackButtonHelper.kt
```diff
@@ -16,31 +17,38 @@ import com.squareup.workflow1.ui.onBackPressedDispatcherOwnerOrNull
  * that predates `OnBackPressedDispatcher`.
  */
 @WorkflowUiExperimentalApi
-public fun interface ModalScreenOverlayOnBackPressed {
+public interface ModalScreenOverlayBackButtonHelper {
+  public fun initialize(contentView: View) {
+    // If the content view has no backPressedHandler, add a no-op one to
+    // ensure that the `onBackPressed` call below will not leak up to handlers
+    // that should be blocked by this modal session.
+    if (contentView.backPressedHandler == null) contentView.backPressedHandler = { }
```
rjrjr, line 27: This is inadequate, can be stomped updates of the content view.

- [ ] wrap the content view factory instead.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/ModalScreenOverlayDialogFactory.kt
```diff
@@ -81,35 +79,23 @@ public open class ModalScreenOverlayDialogFactory<O : ScreenOverlay<*>>(
...
-    val contentViewHolder = wrappedContentRendering.toViewFactory(initialEnvironment)
-      .startShowing(wrappedContentRendering, initialEnvironment, context).apply {
-        // If the content view has no backPressedHandler, add a no-op one to
-        // ensure that the `onBackPressed` call below will not leak up to handlers
-        // that should be blocked by this modal session.
-        if (view.backPressedHandler == null) view.backPressedHandler = { }
+  ): OverlayDialogHolder<O> {
+    val contentViewHolder = initialRendering.content.toViewFactory(initialEnvironment)
+      .startShowing(initialRendering.content, initialEnvironment, context).apply {
+        environment[ModalScreenOverlayBackButtonHelper].initialize(view)
```
rjrjr, line 85: Perhaps these should be methods on this interface, with default implementations?

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/OverlayDialogHolder.kt
```diff
@@ -0,0 +1,112 @@
...
+ */
+@WorkflowUiExperimentalApi
+public fun OverlayDialogHolder<*>.canShow(overlay: Overlay): Boolean {
+  // The ShowingNothing case covers bootstrapping, during the first call to show()
+  // from OverlayDialogFactory.start().
+  return showing.let { it is NoOverlay || compatible(it, overlay) }
+}
+
+/**
+ * Updates the [view] managed by the receiver to display [overlay], and
```
rjrjr: s/view/dialog/

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/ModalScreenOverlayBackButtonHelper.kt
rjrjr: And this realization is why #805 happened.

rjrjr:
This needs tweaks to the back button handling, and that's deeply related to modality, and I need to make the modal opt-in more flexible. So pausing this while I do that.

PR #806, opened 2022-06-10
Replace test with check

bnvinay92:
Seeing what happens on CI when we try to run all tests.


steve-the-edwards:
So how is the `iosX64Test` running in CI?

steve-the-edwards:
> So how is the `iosX64Test` running in CI?

nvm you answered on slack that it was because this is running on the linux machine and it is automatic?

steve-the-edwards (approved):

PR #808, opened 2022-06-11
More flexible modality.

rjrjr:
Introduces the `ModalOverlay` interface. Moves as much code as possible from `BodyAndModalsContainer` to `LayeredDialogs`, and allows them to turn on modal event blocking behavior per managed `Overlay` instance, rather than always enforcing it.

Very sad that I wasn't able to get rid of `BodyAndOverlaysContainer.onAttachedToWindow`, but the SavedStateRegistry timing is just too fussy to allow it.

Git was dumb about diff'ing across renames, so they aren't all done in the first commit for ease of review.

### DIFF
--- a/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/BodyAndOverlaysContainer.kt
+++ b/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/BodyAndOverlaysContainer.kt
@@ -20,14 +20,16 @@ import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
 import com.squareup.workflow1.ui.WorkflowViewStub
 
 @WorkflowUiExperimentalApi
-// TODO Rename this BodyAndOverlaysContainer
-internal class BodyAndModalsContainer @JvmOverloads constructor(
+internal class BodyAndOverlaysContainer @JvmOverloads constructor(
   context: Context,
   attributeSet: AttributeSet? = null,
   defStyle: Int = 0,
   defStyleRes: Int = 0
 ) : FrameLayout(context, attributeSet, defStyle, defStyleRes) {
   /**
+   * The unique `SavedStateRegistry` key passed to [LayeredDialogs.onAttachedToWindow],
+   * derived from the first rendering passed to [update]. See the doc on
+   * [LayeredDialogs.onAttachedToWindow] for details.
    */
   private lateinit var savedStateParentKey: String
 
@@ -41,7 +43,7 @@ internal class BodyAndModalsContainer @JvmOverloads constructor(
   )
 
   fun update(
-    newScreen: BodyAndModalsScreen<*, *>,
+    newScreen: BodyAndOverlaysScreen<*, *>,
     viewEnvironment: ViewEnvironment
   ) {
     savedStateParentKey = Compatible.keyFor(viewEnvironment[Showing])
@@ -126,10 +128,10 @@ internal class BodyAndModalsContainer @JvmOverloads constructor(
     }
   }
 
-  companion object : ScreenViewFactory<BodyAndModalsScreen<*, *>>
+  companion object : ScreenViewFactory<BodyAndOverlaysScreen<*, *>>
   by ScreenViewFactory.fromCode(
     buildView = { _, initialEnvironment, context, _ ->
-      BodyAndModalsContainer(context)
+      BodyAndOverlaysContainer(context)
         .let { view ->
           view.id = R.id.workflow_body_and_modals_container
           view.layoutParams = (LayoutParams(MATCH_PARENT, MATCH_PARENT))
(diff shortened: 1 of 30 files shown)


workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/container/ModalOverlay.kt
```diff
@@ -0,0 +1,10 @@
+package com.squareup.workflow1.ui.container
+
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+
+/**
+ * Marker interface identifying [Overlay] renderings whose presence
+ * indicates that events are blocked from lower layers.
+ */
+@WorkflowUiExperimentalApi
+public interface ModalOverlay : Overlay
```
rjrjr, line 10: Make `AlertOverlay` extend this.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/OverlayArea.kt
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Reports the the area of the screen whose events should be blocked by any modal [Overlay],
+ * in the style reported by [View.getGlobalVisibleRect][android.view.View.getGlobalVisibleRect].
+ * Expected to be supplied by containers that support [BodyAndOverlaysScreen].
+ */
+@WorkflowUiExperimentalApi
+internal class OverlayArea(
+  val bounds: StateFlow<Rect>
+) {
```
steve-the-edwards, line 18: SO this is how we specify whether to block events around overlays now?

steve-the-edwards (approved):

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/OverlayArea.kt
rjrjr: _Big response here, I'll update the commit message to include it, once you confirm it's intelligible._

This isn't new, I just renamed it from `ModalOverlayArea` b/c it's not specific to modal behavior. `OverlayArea` is still how we broadcast the bounds of the view under the dialogs, so that the dialogs' size and position can be restricted to that area. It's always been about positioning, not modality. And it's mainly about making sure that the dialog isn't too big, e.g. so that we can build status and tutorial bars that don't get eclipsed by our modal windows.

```
[ StatusBarContainer ]
[ BodyAndOverlaysContainer ]
[ TutorialBarContainer ]
```

`OverlayArea` here tracks the size and position of `BodyAndOverlaysContainer`, so any related dialog windows won't stray over the status or tutorial bars while they're visible. None of that changes with this PR.

The big change here is related to the tricky code we've built to make covered views start ignoring events the instant we decide to show a dialog, because there is a period after the call to Android's `Dialog.show()` where events that should be intercepted by the new dialog window aren't. It's fundamentally about when we set `allowEvents` fields in `DialogHolder` and `LayeredDialogs` to true or false.

While `allowEvents` is false, the covered view ignores all UI events, regardless of the size of the dialog windows above it. In our illustration above, even if the dialog windows over the body in `BodyAndOverlaysContainer` decide to be tiny, the body will ignore all events while any dialog is showing.

Previously, `LayeredDialogs.modal` was set at construction time, and `BodyAndModalsContainer` set it to true. That meant that we wanted all managed dialog windows to be considered modal, meaning `allowEvents` should be false while any of them are visible. So, if you wanted to use smaller dialog windows for something like a floating toast, which should not block events, you're hosed. You have to fork `BodyAndModalsContainer` to create `LayeredDialogs` with `modal` false, and have a separate wrapping container to host your non-modal toasts.

With this PR, `LayeredDialogs` looks for the `ModalOverlay` marker interface. `LayeredDialog.allowEvents` is set to false only if one of the managed `Overlay` renderings also implements `ModalOverlay`. `DialogHolder.allowEvents` is set to false only for dialogs whose z index is lower than a `ModalOverlay` sibling. 

Also note how we still look for the value of the  `CoveredByModal` `ViewEnvironment` value while deciding if we're in that situation, so that it's still possible to nest `BodyAndOverlayContainer`s.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/BodyAndOverlaysContainer.kt
```diff
@@ -0,0 +1,142 @@
...
+
+@WorkflowUiExperimentalApi
+internal class BodyAndOverlaysContainer @JvmOverloads constructor(
+  context: Context,
+  attributeSet: AttributeSet? = null,
+  defStyle: Int = 0,
+  defStyleRes: Int = 0
+) : FrameLayout(context, attributeSet, defStyle, defStyleRes) {
+  /**
+   */
```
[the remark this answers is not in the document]
rjrjr: Thanks.

```kotlin
  /**
   * The unique `SavedStateRegistry` key passed to [LayeredDialogs.onAttachedToWindow],
   * derived from the first rendering passed to [update]. See the doc on
   * [LayeredDialogs.onAttachedToWindow] for details.
   */
```

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/OverlayArea.kt
rjrjr: I'm adding this discussion to the `ScreenOverlayDialogFactory` in the next PR, gonna merge this as is.

PR #817, opened 2022-06-22
Upgrade coroutines to 1.6.4 (including coroutines.test changes)

steve-the-edwards:
For non-test use we had already migrated from `offer` to `trySend` for channels.

For tests (the larger change), see the test migration guide followed here: https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-test/MIGRATION.md 

This has been tested successfully internally in Square POS via a snapshot.

Regarding whether we should convert all `runBlocking` to `runTest` when used in tests, its a choice for each situation (we have converted many but not all). See Dmitry Khalanskiy's (coroutines maintainer) to that [question](https://kotlinlang.slack.com/archives/C1CFAFJSK/p1658222819199879?thread_ts=1658166870.419349&cid=C1CFAFJSK):

> If you're not relying on virtual time, you can stick with runBlocking, it's just a bit less convenient:
> There's no runBlocking on JS, so the test won't be usable in multiplatform code. This was the message behind the line: 1.6 introduced multiplatform coroutine testing.
> runBlocking returns the value that its block returns. JUnit, for example, expects the test functions to return Unit. So, something like @Test fun test() = runBlocking { assertFailsWith<IllegalArgumentException> { throw IllegalArgumentException() } } will, surprisingly, not be executed by JUnit, because test() returns an IllegalArgumentException.
> runTest is purely for tests, which means we can enhance it with other niceties that have no place in production code. For example, we recently added a way to cancel some of the coroutines in a test when the test body finishes: https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-test/kotlinx.coroutines.test/-test-scope/background-scope.html You won't have that with runBlocking.


vgonda (approved):
Sweet! Learn stuff we can apply to register? 😉

rjrjr (approved):

RBusarow (approved):

PR #819, opened 2022-06-23
add a `published` convention plugin and enforce `-SNAPSHOT` versions in CI

RBusarow:


.github/workflows/publish-snapshot.yml
```diff
@@ -22,6 +22,9 @@ jobs:
           distribution: 'zulu'
           java-version: 11
 
+      - name : Check for -SNAPSHOT version
```
rjrjr, line 25: Nice, thanks.

rjrjr (approved):

PR #820, opened 2022-06-23
Update Tutorial2.md to clarify an instruction.

tyler-higashi:
Made a clarification about what steps a tutorial follower should take to temporarily resolve RootWorkflow's renderChild call on a child Workflow with Nothing as its output type. The override of renderChild needed does not include a lambda as its last parameter.


afollestad (approved):

steve-the-edwards (approved):
LGTM.

PR #821, opened 2022-06-24
replace `burrunan/gradle-cache-action` with `gradle/gradle-build-action`

RBusarow:
The older action, `burrunan/gradle-cache-action`, turns GitHub's freebie storage into a remote Gradle build cache.  This causes tons of small requests, and as a result we sometimes get 429 errors or timeouts.  When this happens, the best case scenario is that Gradle just treats it as a cache miss and rebuilds something locally.  In the worst case, CI just fails.

This replacement action is Gradle's own.  It automatically handles keys and partitioning of the caches, and still uses build cache instead of `/build`.  But, it saves/restores those caches in large chunks instead of making a remote build cache.  As a result, the I/O is faster and there are far fewer requests.

### DIFF
--- a/.github/workflows/kotlin.yml
+++ b/.github/workflows/kotlin.yml
@@ -33,31 +33,23 @@ jobs :
           java-version : 11
 
       ## Actual task
-      - uses : burrunan/gradle-cache-action@v1
+      - uses: gradle/gradle-build-action@v2
         name : Assemble with gradle — make sure everything builds
         with :
-          gradle-dependencies-cache-key : |
-            gradle/libs.versions.toml
           arguments : |
-            assemble --no-daemon --stacktrace
-          concurrent : true
-          gradle-build-scan-report : false
-          gradle-distribution-sha-256-sum-warning : false
+            assemble --stacktrace
+          cache-read-only: false
 
       # This should ideally be done as a Check job below, but it needs to be done as a separate
       # step after running assemble. Heckin' ridikalus.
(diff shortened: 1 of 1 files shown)


.github/workflows/kotlin.yml
```diff
@@ -33,31 +33,23 @@ jobs :
...
         name : Assemble with gradle — make sure everything builds
         with :
-          gradle-dependencies-cache-key : |
-            gradle/libs.versions.toml
           arguments : |
-            assemble --no-daemon --stacktrace
-          concurrent : true
-          gradle-build-scan-report : false
-          gradle-distribution-sha-256-sum-warning : false
+            assemble --stacktrace
```
rjrjr, line 40: I take it `--no-daemon` is built in now?

rjrjr (approved):

.github/workflows/kotlin.yml
RBusarow: More or less.
https://github.com/gradle/gradle-build-action#stopping-the-gradle-daemon
> By default, the action will stop all running Gradle daemons in the post-action step, prior to saving the Gradle User Home state. This allows for any Gradle User Home cleanup to occur, and avoid file-locking issues on Windows.
>
> If caching is unavailable or the cache is in read-only mode, the daemon will not be stopped and will continue running after the job is completed.

steve-the-edwards (approved):
Love it!

PR #823, opened 2022-06-28
Compose Runtime Integration Take 2

steve-the-edwards:


benchmarks/performance-poetry/complex-poetry/src/main/java/com/squareup/benchmarks/performance/complex/poetry/MaybeLoadingGatekeeperWorkflow.kt
```diff
@@ -48,5 +49,36 @@ class MaybeLoadingGatekeeperWorkflow<T : Any>(
     )
   }
 
+  @Composable
+  override fun Rendering(
+    renderProps: Unit,
+    renderState: IsLoading,
+    context: RenderContext,
+    hoistRendering: @Composable (rendering: MayBeLoadingScreen) -> Unit
+  ) {
```
steve-the-edwards, line 58: I'm stumped here. These overrides are recognized by the compiler as it runs in the IDE but when tracing through the call always goes through to the base `StatefulWorkflow::Rendering` implementation?

@zach-klippenstein any idea why this override wouldn't be recognized at runtime?

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/StatefulWorkflow.kt
```diff
@@ -105,6 +106,34 @@ public abstract class StatefulWorkflow<
...
 
+  @Composable
+  public open fun Rendering(
+    renderProps: PropsT,
+    renderState: StateT,
+    context: RenderContext,
+    hoistRendering: @Composable (rendering: RenderingT) -> Unit
+  ): Unit {
+    // By default call render() as a fallback.
+    hoistRendering(render(renderProps, renderState, context))
```
steve-the-edwards, line 117: I did this so I don't have to modify *every* workflow and test to get this to compile. However, at runtime the Workflow.Rendering gets resolved here rather than to the child function. 😢

benchmarks/performance-poetry/complex-poetry/src/main/java/com/squareup/benchmarks/performance/complex/poetry/MaybeLoadingGatekeeperWorkflow.kt
steve-the-edwards: Reproduced in small project. I believe bug in Molecule - https://github.com/cashapp/molecule/issues/91

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/StatefulWorkflow.kt
steve-the-edwards: https://github.com/cashapp/molecule/issues/91

Issue #665, opened 2022-02-15
WorkflowRuntime re-write with Compose tree mechanisms?


steve-the-edwards:
Placeholder for this thought in the `optimization` backlog.

@zach-klippenstein and I once talked about re-writing the `WorkflowNode` tree using compose mechanisms (and optimizations?) but the problem there (IIRC) was that we don't want to have multiple compose-node types in the same app (e.g. UI compose-nodes and then Workflow nodes).

rjrjr:
I think that's supported now.

steve-the-edwards:
The work is being done in #824 some discussion in https://github.com/square/workflow-kotlin/discussions/272

PR #824, opened 2022-07-04
Compose in the Runtime Take 3

steve-the-edwards:
To address #665


workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
```diff
@@ -186,17 +211,63 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
...
+    return rendering
+  }
+
+  @Composable
+  private fun RenderingWithStateType(
+    workflow: StatefulWorkflow<PropsT, StateT, OutputT, RenderingT>,
+    props: PropsT,
+    setRendering: (RenderingT) -> Unit
+  ): Unit {
+    key(props) {
```
zach-klippenstein, line 228: If you're comparing props anyway, I don't see the reason for putting this in a key function (and if it is necessary, it should probably be an implementation detail of UpdatePropsAndState).

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
```diff
@@ -186,17 +211,63 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
...
+      UpdatePropsAndState(workflow, props)
```
zach-klippenstein, line 229: Nit: Compose conventions discourage naming unit-returning composable functions as verbs.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
```diff
@@ -205,19 +276,34 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
...
   }
 
+  @Composable
+  private fun UpdatePropsAndState(
+    workflow: StatefulWorkflow<PropsT, StateT, OutputT, RenderingT>,
+    newProps: PropsT
+  ) {
+    if (newProps != lastProps) {
+      val newState = interceptor.intercept(workflow, this)
+        .onPropsChanged(lastProps, newProps, state.value)
```
zach-klippenstein, line 292: I hope everyone heeded our advice to not perform side effects in this function 😅

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowRunner.kt
```diff
@@ -52,7 +56,7 @@ internal class WorkflowRunner<PropsT, OutputT, RenderingT>(
   private val rootNode = WorkflowNode(
     id = workflow.id(),
     workflow = workflow,
-    initialProps = currentProps,
+    initialProps = currentProps.value,
```
zach-klippenstein, line 59: Nit: I don't see you using the delegate syntax for `MutableState`s very much, it would make a lot of the code in this PR more idiomatic and a teeny bit cleaner.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/SubtreeManager.kt
```diff
@@ -114,6 +116,39 @@ internal class SubtreeManager<PropsT, StateT, OutputT>(
...
+
+  @Composable
+  override fun <ChildPropsT, ChildOutputT, ChildRenderingT> Rendering(
+    child: Workflow<ChildPropsT, ChildOutputT, ChildRenderingT>,
+    props: ChildPropsT,
+    key: String,
+    handler: (ChildOutputT) -> WorkflowAction<PropsT, StateT, OutputT>
+  ): ChildRenderingT {
+    val stagedChild = remember (child, props, key, handler) {
+      prepareStagedChild(
```
zach-klippenstein, line 136: I don't believe this is safe to call from composition since it modifies non-snapshot state in this class.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
```diff
@@ -215,25 +215,38 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
   private fun RenderingWithStateType(
     workflow: StatefulWorkflow<PropsT, StateT, OutputT, RenderingT>,
     props: PropsT,
-    hoistRendering: @Composable (RenderingT) -> Unit
-  ) {
+    state: StateT
+  ): RenderingT {
     val composedState = remember(state, props) {
```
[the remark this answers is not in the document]
steve-the-edwards: This was old so must have been from a previous review you did of this before pushing out the review. I am going to disregard :).

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowRunner.kt
steve-the-edwards: The delegate syntax continues to confuse me and I don't necessarily like the syntactic sugar obfuscating what is happening.

In this case `currentProps` is a class property. How would I use the delegate syntax here as I thought that's only when using `remember`? And `remember` is only for within a `@Composable`?

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/SubtreeManager.kt
steve-the-edwards: given that `SideEffect` only occurs after Composition though and we need this `stagedChild` to call the Rendering waht are the options here?

I thought `remember` would effectively avoid calling this multiple times across recomposition?

Do you mean that the StagingList needs to be `State`?

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowRunner.kt
zach-klippenstein: Remember has nothing to do with delegation. Property delegation is a [well-documented Kotlin language feature](https://kotlinlang.org/docs/delegated-properties.html). It's supported for `MutableState` because it has `getValue` and `setValue` extension methods.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/SubtreeManager.kt
zach-klippenstein: Remember will prevent calling it again on subsequent compositions, but it will still be called if the initial composition is discarded and never applied.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowRunner.kt
steve-the-edwards: Ah I was getting confused by all the examples including the `remember` in the delegate.

I see now that you can just use:
```
val currentProps: PropsT by mutableStateOf(props.value)
```

that makes a lot more sense.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/SubtreeManager.kt
steve-the-edwards: I think i'm just going to have to re-write or modify `ActiveStagingList` in a Snapshot-State-friendly way

steve-the-edwards:
```
  /**
   * Prepare the staged child while only modifying [children] in a SideEffect. This will ensure
   * that we do not inappropriately modify non-snapshot state.
   */
  @Composable
  private fun <ChildPropsT, ChildOutputT, ChildRenderingT> StagedChild(
    child: Workflow<ChildPropsT, ChildOutputT, ChildRenderingT>,
    props: ChildPropsT,
    key: String,
    handler: (ChildOutputT) -> WorkflowAction<PropsT, StateT, OutputT>
  ): WorkflowChildNode<*, *, *, *, *> {
    val childState = remember (child, key, props, handler) {
      children.forEachStaging {
        require(!(it.matches(child, key))) {
          "Expected keys to be unique for ${child.identifier}: key=\"$key\""
        }
      }
      mutableStateOf(children.firstActiveMatchingOrNull {
        it.matches(child, key)
      } ?: createChildNode(child, props, key, handler))
    }

    SideEffect {
      children.swapOrStage(
        predicate = {it.matches(child, key)},
        child = childState.value)
    }
    return childState.value
  }
```

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
steve-the-edwards: I think we can protect against that anyway like this:

```
    val newState = produceState(initialValue = state.value, newProps) {
      if (newProps != lastProps) {
        interceptor.intercept(workflow, this@WorkflowNode)
          .onPropsChanged(lastProps, newProps, state.value)
      }
    }
    state.value = newState.value
```

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
steve-the-edwards: Worthwhile to break convention here I think.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
steve-the-edwards: removed.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowChildNode.kt
```diff
@@ -58,6 +61,23 @@ internal class WorkflowChildNode<
...
+    props: Any?
+  ): R {
+    val renderingState = remember { mutableStateOf<R?>(null) }
+    @Suppress("UNCHECKED_CAST")
+    workflowNode.Rendering(
+      workflow as StatefulWorkflow<ChildPropsT, out Any?, ChildOutputT, Nothing>,
+      props as ChildPropsT,
+      setRendering = {
+        renderingState.value = it as R
+      }
```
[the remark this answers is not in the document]
steve-the-edwards: Done.

steve-the-edwards:
OK so I have updated this with a recent commit that isolates all of teh Compose optimizations as a plugin module

steve-the-edwards:
1. Trying to call into the Compose runtime directly (without it being a Composable) was a dead end for now as I worked on it for a bit with Zach and we got past his knowledge and idea of where to go next, so letting that go. The impact of this is we need a StatefulWorkflow/StatelessWorkflow extension to add the @Composable Rendering - but I think i did that relatively nicely.
2. This update is passing all the tests locally and is behind a warning annotation.
3. You would add the workflow-core-compose module and plugin the optimization to the runtime when calling renderWorkflowIn, as I do in performance poetry (and it shows the same rendering oprtimizations on the tests).
4. I like where this is at for pausing as its all isolated and I could likely rebase this without too much difficulty. (Obviously it doesn't have any compose runtime specific tests yet so I would want to add that when picking it back up).
5. Unfortunately I had to pull all of our runtime.internal classes into public land because we (still I think?) don't have good 'friend path' support for kotlin modules - and since I'm extending them all in the compose-core module I need access.
6. What I would like to do with this is turn off the publishing of these and the explicit API mode. So they can still be public for the other module but won't be part of the published API (haven't done this yet). Unless someone has a good idea for package-protected in kotlin?
7. Some of the polymorphism got a bit messier than at other times - would really appreciate ideas as to how it could be cleaner at points. WorkflowNode initialization is really the ugliest one as I had to add a startSession so that the initialization happens after the child class is instantiated and not just after the parent class.

steve-the-edwards:
Superseded by #886

PR #833, opened 2022-07-12
Do not change Workflow API for COmpose

steve-the-edwards:


steve-the-edwards:
This looks harder to do than originally thought. Not a rabbit hole that makes sense to go down now as we would have to understand more about the "Data Anchor" and why it is wrong when using the forced Compose call method.

@zach-klippenstein had recommendation which I will record for posterity:

> I would start by attaching a debugger, seeing what’s causing that anchor to be calculated as negative, and working backwards from there to see how it gets in a bad state

PR #836, opened 2022-07-13
Full backward compatibility for Screen.

rjrjr:
Extends the legacy `ViewRegistry.getFactoryForRendering` to handle any arbitrary `Screen` implementation, making every bit of code that works directly with `ViewRegistry.buildView` (including `DecorativeViewFactory`) future proof. Allows us to delete the legacy factory that had been created specifically for `BackStackScreen`.


rjrjr:
Actually, I'm still doing external testing on this, not sure it's worth reviewing yet.

rjrjr:
@RBusarow good to go!

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewShowRendering.kt
```diff
@@ -114,25 +114,35 @@ public fun <RenderingT : Any> View.showRendering(
...
+    // new Showing key and the old workflowViewState backing View.getRendering().
+    val updatedEnv = viewEnvironment + (Showing to asScreen(rendering))
+    workflowViewState = Started(rendering, updatedEnv, viewState.showRendering)
+
     viewState.showRendering.invoke(rendering, viewEnvironment)
   }
 }
 
 /**
- * **This will be deprecated in favor of [screenOrNull] very soon.**
```
RBusarow, line 124: 🔥

RBusarow (approved):

PR #837, opened 2022-07-15
Deprecates old view building API.

rjrjr:
I'm now confident that it's practical to migrate to Screen et al, so I've uncommented the related deprecation annotations. Note that this commit does not deprecate the old dialog building code, still kicking the tires on that.


RBusarow (approved):
Sweet!

PR #838, opened 2022-07-19
Simplifies ScreenOverlayDialogFactory.

rjrjr:
The old approach would force you to build an entire new `Dialog` if the content type changed, which is silly. Now we explicitly use `WorkflowLayout` as the content view, and eliminate the redundant code for dealing directly with `ScreenViewFactory`.

Also cleans up the unnecessary `.withEnvironment()` calls in `WorkflowLayout`, and makes its `show()` function more of a first class citizen.


rjrjr:
Non-starter, `ScreenOverlay.content` is not available to the dialog building code.

PR #839, opened 2022-07-20
Long lived branch with `Overlay` refinements

rjrjr:
Holding off on merging into `main` until I have more confidence in the changes, which I'm getting by applying them across Square's flagship Android apps; and until I'm ready to deal with the recent coroutines version update (#817).

### Introduces `OverlayDialogHolder.onUpdateBounds`

`OverlayDialogHolder.onUpdateBounds` replaces `ScreenOverlayDialogFactory.updateBounds`.

I finally tried to use `ScreenOverlayDialogFactory.updateBounds` and was reminded once again why it's bad to create interfaces that build something and then take it back later to be updated. If you want the renderings to be involved in setting the bounds policy, there was no way to smuggle that information along with the Dialog returned from `buildDialogWithContent`, so that `updateBounds` could honor it. Also, it was always pretty ugly that the bounds hook was special to the `ScreenOverlay` world.

The fix is make bounds maintenance part of the job of the `OverlayDialogHolder` interface. Square reviewers will notice that we're now an even more faithful rip off of Mosaic's DialogRunner -- hi @helios175.

### No more hard coded EnvironmentScreen in WorkflowLayout.

WorkflowLayout was wrapping things in EnvironmentScreen for no real reason. No longer.

### DIFF
--- a/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/ScreenOverlayDialogFactory.kt
+++ b/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/ScreenOverlayDialogFactory.kt
@@ -2,7 +2,6 @@ package com.squareup.workflow1.ui.container
 
 import android.app.Dialog
 import android.content.Context
-import android.graphics.Rect
 import android.graphics.drawable.ColorDrawable
 import android.util.TypedValue
 import android.view.KeyEvent
@@ -61,11 +60,6 @@ import kotlin.reflect.KClass
  * display those [Overlay] renderings look for [OverlayArea] value and restrict
  * themselves to the reported bounds.
  *
- * Dialogs created via [ScreenOverlayDialogFactory] implementations honor [OverlayArea]
- * automatically. [updateBounds] is called as the [OverlayArea] changes, and the
- * default implementation of that method sets created dialog windows to fill the given area --
- * not necessarily the entire display.
- *
  * Another [ViewEnvironment] value is maintained to support modality: [CoveredByModal].
  * When this value is true, it indicates that a dialog window driven by a [ModalOverlay]
  * is in play over the view, or is about to be, and so touch and click events should be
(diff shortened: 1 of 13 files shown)


rjrjr:
@steve-the-edwards Uh oh:

```
com.squareup.benchmarks.performance.complex.poetry.RenderPassTest > renderPassCounterFrameTimeoutComplexNoInitializingStateHighFrequencyEvents[test(AVD) - 10] FAILED 
	java.lang.AssertionError: Uh Oh! You have worsened the the number of Render Passes (lower better) by 2 (2.06%) for Runtime: FrameTimeout; the 'Raven navigation (no initializing state) scenario with high frequency events'! The value is now 99 (was 88..97). This likely results in worse performance. You can check with the timing benchmarks if you disagree.
	at org.junit.Assert.fail(Assert.java:89)
```

I'm surprised by this, and I don't see what I've done to cause it. Can I just bump the expected range, or do we need to dig in?

rjrjr:
Re-running that shard to see if this is flaky.

rjrjr:
Sure enough, it was a flake. :(

steve-the-edwards:
> Sure enough, it was a flake. :(

If that was just a flake/within reasonable bounds I am fine with bumping up the expected range.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/ScreenOverlayDialogFactory.kt
```diff
@@ -109,42 +103,33 @@ public open class ScreenOverlayDialogFactory<S : Screen, O : ScreenOverlay<S>>(
...
    */
-  public open fun buildDialogWithContent(content: ScreenViewHolder<S>): Dialog {
-    return Dialog(content.view.context).also { it.setContent(content) }
+  public open fun buildDialogWithContent(
+    initialRendering: O,
+    initialEnvironment: ViewEnvironment,
+    content: ScreenViewHolder<S>
+  ): OverlayDialogHolder<O> {
+    return OverlayDialogHolder(
+      initialEnvironment, Dialog(content.view.context).also { it.setContent(content) }
```
steve-the-edwards, line 122: nit:
```suggestion
      initialEnvironment, Dialog(content.view.context).apply { setContent(content) }
```
I think of `also` as side effects wrt to the receiver, whereas `apply` is explicitly about the receiver.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/ScreenOverlayDialogFactory.kt
```diff
@@ -159,8 +144,12 @@ public open class ScreenOverlayDialogFactory<S : Screen, O : ScreenOverlay<S>>(
     val contentViewHolder =
       buildContent(contentViewFactory, initialRendering.content, initialEnvironment, context)
 
-    return buildDialogWithContent(contentViewHolder).let { dialog ->
-      val window = requireNotNull(dialog.window) { "Dialog must be attached to a window." }
+    return buildDialogWithContent(
+      initialRendering,
+      initialEnvironment,
+      contentViewHolder
+    ).also { holder ->
```
steve-the-edwards, line 151: again nit: `apply` are we using other `this` references on `ScreenOverlayDialogFactory` that we need to keep `this` as that?

steve-the-edwards (approved):

vgonda (commented):
:stamp:

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/ScreenOverlayDialogFactory.kt
rjrjr: I generally find `apply` and `with` blocks pretty confusing, after a few lines I quickly lose track of that `this` is.

Like in this case, there is only a single use of `holder`, everything else is side effect. I find that a lot more apparent with this usage.

rjrjr: And even that one usage is strictly for getting the window, not for changing anything on the holder itself.

steve-the-edwards: OK ya that's valid.

rjrjr:
I'm going to merge #840 into this branch, and hold off on merging into main until I have a bit more confidence in this series of changes. I'm already working on a refinement of our back button handling in Overlay.

rjrjr:
> I'm going to merge #840 into this branch, and hold off on merging into main until I have a bit more confidence in this series of changes. I'm already working on a refinement of our back button handling in `Overlay`, and expect to float another PR targeting this branch today.

#840 is being weird in CI, though everything passes locally and all our internal tests like it. I'm going to go ahead and merge this branch and target #840 at `main`, hoping that makes CI sweet again.

PR #840, opened 2022-07-20
No more hard coded EnvironmentScreen in WorkflowLayout.

rjrjr:
WorkflowLayout was wrapping things in EnvironmentScreen for no real reason. No longer.


rjrjr:
@steve-the-edwards This seems consistent, and I don't understand how it's possible. These changes are entirely on the view side.

```
renderPassCounterFrameTimeoutComplexNoInitializingStateHighFrequencyEvents
```

```
java.lang.AssertionError: Congrats! You have improved the the number of Render Passes (lower better) by 33 (37.50%) for Runtime: FrameTimeout; the 'Raven navigation (no initializing state) scenario with high frequency events'! Please update the expected value for your config. The value is now 55 (was 88..97).
```

steve-the-edwards (approved):

PR #843, opened 2022-07-27
Simpler, better back button handling.

rjrjr:
`OverlayDialogHolder` now can provide an `onBackPressed: () -> Unit` function to be called in place of `Dialog.onBackPressed`, and we provide a default implementation that hooks in `OnBackPressedDispatcher`. This works for all kinds of dialogs, not just modal ones, and replaces `ModalScreenOverlayBackButtonHelper`. The other method that was on `ModalScreenOverlayBackButtonHelper`, which put the no-op `backPressedHandler` on the content view to ensure those on layer layers would not fire, is now built into the default implementation of `ScreenOverlayDialogFactory.buildDialogWithContent`, which is designed to be customized anyway.

Also locks down `AlertOverlayDialogFactory`, but moves all the interesting parts into a public function. Should be easier and safer for re-use.

So how is this simpler? Or at least less complex?

The old scheme had:

- a strategy object in the `ViewEnvironment`, `ModalScreenOverlayBackButtonHelper`, that was providing a one-size-fits-all-dialogs scheme for back button handling
- that scheme was only used by the `ScreenViewOverlay` subtype, not for `Overlay` in general
- and it was applied only for modals, which was very wrong

With this change:

- The individual `OverlayDialogHolder` that wraps a `Dialog` is directly responsible for defining its back button behavior, and can be changed from the default behavior directly by the feature developer
- The bizarre subtle `onBackPressed = {}` hack is moved to the factory class responsible for building the dialog, again in a spot that is naturally open for customization
- and we eliminate the boolean return value, meaning one less decision point for developers


workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/DialogSession.kt
```diff
@@ -46,14 +49,17 @@ internal class DialogSession(
...
+    holder.environment, holder.dialog, holder.onUpdateBounds, holder.onBackPressed
   ) { overlay, environment ->
     allowEvents = !environment[CoveredByModal]
     holder.show(overlay, environment)
   }
 
   val savedStateRegistryKey = Compatible.keyFor(holder.showing, index.toString())
 
+  private val KeyEvent.isBackPress: Boolean
+    get() = (keyCode == KEYCODE_BACK || keyCode == KEYCODE_ESCAPE) && action == ACTION_UP
```
steve-the-edwards, line 61: Hmm. Seems fragile. I guess if we have to check like this then better to have it centralized.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/DialogSession.kt
```diff
@@ -68,7 +74,18 @@ internal class DialogSession(
...
+          if (!allowEvents) return true
+
+          // If there is an onBackPressed handler invoke it instead of allowing
+          // the normal machinery to call Dialog.onBackPressed.
+          if (event.isBackPress) {
+            holder.onBackPressed?.invoke()
+            return true
+          }
+
+          // Allow the usual call to Dialog.onBackPressed.
```
steve-the-edwards, line 87: Confused. This isn't allowing the usual call to `onBackPressed` as that is handled above no? Do you mean route the regular `dispatchKeyEvent` handling for all other keys?

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/LayeredDialogSessions.kt
```diff
@@ -155,19 +155,13 @@ public class LayeredDialogSessions private constructor(
         // with the new Overlay at that index. Just update it.
         sessions[i].also { it.holder.show(overlay, dialogEnv) }
       } else {
-        // We need a new dialog for this overlay. Time to build it.
-        // We wrap our Dialog instances in DialogHolder to keep them
-        // paired with their current overlay rendering and environment.
-        // It would have been nice to keep those in tags on the Dialog's
-        // decor view, more consistent with what ScreenViewFactory does,
-        // but calling Window.getDecorView has side effects, and things
-        // break if we call it to early. Need to store them somewhere else.
```
steve-the-edwards, line 164: Why remove this comment now?

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/OverlayDialogFactoryFinder.kt
```diff
@@ -8,42 +8,8 @@ import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
 
 /**
  * [ViewEnvironment] service object used by [Overlay.toDialogFactory] to find the right
- * [OverlayDialogFactoryScreenViewFactory]. The default implementation makes [AndroidOverlay]
+ * [OverlayDialogFactory]. The default implementation makes [AndroidOverlay]
  * work, and provides default bindings for [AlertOverlay] and [FullScreenOverlay].
- *
- * Here is how this hook could be used to provide a custom dialog to handle [FullScreenOverlay]:
```
steve-the-edwards, line 14: This example is no longer applicable because of back press handling?

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/DialogSession.kt
```diff
@@ -68,7 +74,18 @@ internal class DialogSession(
...
         override fun dispatchKeyEvent(event: KeyEvent): Boolean {
-          return !allowEvents || realWindowCallback.dispatchKeyEvent(event)
+          // Consume all events if we've been told to do so.
+          if (!allowEvents) return true
+
+          // If there is an onBackPressed handler invoke it instead of allowing
+          // the normal machinery to call Dialog.onBackPressed.
+          if (event.isBackPress) {
+            holder.onBackPressed?.invoke()
+            return true
```
steve-the-edwards, line 84: I thought if the `onBackPressed` is not specified we don't want to block/"handle" back and _should_ actually let it fall through (as the comment below suggests).

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/DialogSession.kt
```diff
@@ -46,14 +49,17 @@ internal class DialogSession(
    * Wrap the given dialog holder to maintain [allowEvents] on each update.
    */
   val holder: OverlayDialogHolder<Overlay> = OverlayDialogHolder(
-    holder.environment, holder.dialog
+    holder.environment, holder.dialog, holder.onUpdateBounds, holder.onBackPressed
```
steve-the-edwards, line 52: For `onUpdateBounds` we weren't routing through the `onUpdateBounds` of the original holder before? So it was re-creating the default argument?

steve-the-edwards (commented):
I'm not sure I've gotten to coming around to how this is simpler, but I trust you and I think I need to do a more systematic learning of the modal/dialog/overlay handling anyway.

I do have a few questions about this particular PR though - noted in comments.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/DialogSession.kt
rjrjr: It's pretty well proven, we've been relying on this fragment for years now.

rjrjr:
> I'm not sure I've gotten to coming around to how this is simpler

Less complex, anyway.

The old scheme had a strategy object in the ViewEnvironment that was providing a one-size-fits-all-dialogs scheme for back button handling, and that scheme was only used by the ScreenViewOverlay subtype, not for Overlay in general. It also was applied only for modals, which was just wrong.

With this change, the OverlayDialogHolder that wraps the Dialog is directly responsible for defining its back button behavior; the bizarre subtle `onBackPressed = {}` hack is moved to the factory class responsible for building the dialog, again in a spot that is naturally open for customization; and we eliminate the boolean return value, meaning one less decision point for developers.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/LayeredDialogSessions.kt
rjrjr: It was already obsolete, and gets more obsolete with #845. Doesn't seem worth updating the editorial.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/OverlayDialogFactoryFinder.kt
rjrjr: It was already inaccurate, and I'm kind of exhausted, TBH.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/DialogSession.kt
rjrjr: Wow, good catch. You're absolutely right, and I don't understand how this didn't break `AlertOverlayDialogFactory` -- must be missing tests.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/DialogSession.kt
rjrjr: Yes, big oops.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/DialogSession.kt
rjrjr: Yup, totally broke it.

rjrjr: Fixed, updated the comment you mentioned, and added new `canGoBackFromAlert()` test.

```kotlin
          // If there is an onBackPressed handler invoke it instead of allowing
          // the normal machinery to call Dialog.onBackPressed.
          if (event.isBackPress) holder.onBackPressed?.let { onBackPressed ->
            onBackPressed.invoke()
            return true
          }

          // Allow the usual handling, including the usual call to Dialog.onBackPressed.
          return realWindowCallback.dispatchKeyEvent(event)
```

steve-the-edwards (approved):

Issue #849, opened 2022-07-29
Avoid recreating `RenderContext` so that the `actionSink` captured in lambdas can be memoized


steve-the-edwards:
Right now the Recomposer sees a different object each time and will force recomposition.

rjrjr:
Seems like we could still have them pass out the same / equal object from actionSink instead of this, rather than avoid the recreation entirely. Could be simpler.

steve-the-edwards:
Fixed by #850

PR #850, opened 2022-07-29
Reuse RenderContext; unfreeze before rendering

steve-the-edwards:
Addresses #849

### DIFF
--- a/.github/workflows/kotlin.yml
+++ b/.github/workflows/kotlin.yml
@@ -236,53 +236,54 @@ jobs :
           name : instrumentation-test-results-${{ matrix.api-level }}
           path : ./**/build/reports/androidTests/connected/**
 
-  frame-timeout-instrumentation-tests :
-    name : Frame Timeout Instrumentation tests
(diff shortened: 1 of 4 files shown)


rjrjr (approved):

steve-the-edwards:
The frame timeout tests are failing consistently. Which means there is somethign to investigate with re-using the RenderContext when we use the FrameTimeout optimization. This isn't used anywhere though yet that we know of. I would suggest we go ahead with merging this and I will turn off those tests for now.

@rjrjr @RBusarow WDYT?

RBusarow:
> The frame timeout tests are failing consistently. Which means there is somethign to investigate with re-using the RenderContext when we use the FrameTimeout optimization. This isn't used anywhere though yet that we know of. I would suggest we go ahead with merging this and I will turn off those tests for now.
> 
> @rjrjr @RBusarow WDYT?

Seems reasonable to me.

rjrjr:
+1, let's fix the real world problem ASAP.

PR #852, opened 2022-07-29
fix radiography packaging

RBusarow:
The upstream bug has been removed from kotlin/kotlinx libraries.  The exclusion allowed this module to build, but nothing would actually resolve when trying to consume it downstream.


rjrjr:
Thanks!

rjrjr (approved):

PR #853, opened 2022-07-30
capture KMP publications in artifacts-check

RBusarow:
This adds all publications to the `artifactsCheck` task, instead of just looking at the standard jvm ones.

It also pulls this data directly from the Gradle Maven publish plugin's `Publication` object, instead of just looking at what's defined `gradle.properties`.  That `Publication` object is what's used to create the actual pom, so it's a more robust test.

### DIFF
--- a/buildSrc/src/main/java/com/squareup/workflow1/buildsrc/artifacts/ArtifactsTask.kt
+++ b/buildSrc/src/main/java/com/squareup/workflow1/buildsrc/artifacts/ArtifactsTask.kt
@@ -8,6 +8,8 @@ import org.gradle.api.Project
 import org.gradle.api.file.ProjectLayout
 import org.gradle.api.file.RegularFile
 import org.gradle.api.plugins.JavaPluginExtension
+import org.gradle.api.publish.PublishingExtension
+import org.gradle.api.publish.maven.MavenPublication
 import org.gradle.api.tasks.Internal
 import org.gradle.api.tasks.OutputFile
 import org.gradle.kotlin.dsl.getByType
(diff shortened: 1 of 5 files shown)


buildSrc/src/main/java/com/squareup/workflow1/buildsrc/artifacts/ArtifactsTask.kt
```diff
@@ -52,27 +54,34 @@ abstract class ArtifactsTask(
...
+          .orEmpty()
+          .map { publication -> sub to publication }
+      }
+      .mapNotNull { (sub, publication) ->
 
-        val group = sub.properties["GROUP"] as? String
-        val artifactId = sub.properties["POM_ARTIFACT_ID"] as? String
-        val pomName = sub.properties["POM_NAME"] as? String
-        val packaging = sub.properties["POM_PACKAGING"] as? String
+        val group: String? = publication.groupId
```
RBusarow, line 66: Instead of reading Gradle properties, we can everything directly from the `Publication` object.  So, after everything done by the Vanniktech plugin.

RBusarow:
The updated `./gradlew artifactsCheck` task is passing locally but failing in CI (running on Ubuntu).  This is because of KMP's additional environment requirements.  The KMP plugin automatically skips the iOS targets, which means they aren't there for the artifacts plugin to read.

Here's the actual error:
```
Some Kotlin/Native targets cannot be built on this linux_x64 machine and are disabled:
    * In project ':workflow-core':
        * targets 'iosArm64', 'iosX64' (can be built with one of the hosts: macos_x64, macos_arm64)
    * In project ':workflow-runtime':
        * targets 'iosArm64', 'iosX64' (can be built with one of the hosts: macos_x64, macos_arm64)
To hide this message, add 'kotlin.native.ignoreDisabledTargets=true' to the Gradle properties.


> Task :artifactsCheck FAILED
	Artifact definitions don't match.  If this is intended, run `./gradlew artifactsDump` and commit changes.

	These artifacts are defined in `artifacts.json` but are missing from the project:

                     gradlePath  - :workflow-core
                publicationName  - iosArm64
                          group  - com.squareup.workflow1
                     artifactId  - workflow-core-iosarm64
                pom description  - Workflow Core
                      packaging  - klib

                     gradlePath  - :workflow-core
                publicationName  - iosX64
                          group  - com.squareup.workflow1
                     artifactId  - workflow-core-iosx64
                pom description  - Workflow Core
                      packaging  - klib

                     gradlePath  - :workflow-runtime
                publicationName  - iosArm64
                          group  - com.squareup.workflow1
                     artifactId  - workflow-runtime-iosarm64
                pom description  - Workflow Runtime
                      packaging  - klib

                     gradlePath  - :workflow-runtime
                publicationName  - iosX64
                          group  - com.squareup.workflow1
                     artifactId  - workflow-runtime-iosx64
                pom description  - Workflow Runtime
                      packaging  - klib
```

So as I see it, the options are:
1. Write something into `artifactsCheck` which checks the OS and skips any `"ios*64"` artifacts if it's not in macOS.  
  a. Which basically removes the checks from CI -- which isn't great.  It should at least print a warning.
2. Move this specific task into a macOS runner.
  a. The `check` task is currently being invoked on Ubuntu, in order to invoke the common tests.  We need to either call other, specific tasks, or option #1 needs to be done as well.

I think I need to do both.

rjrjr (approved):

PR #854, opened 2022-08-01
Add KMP target for iosSimulatorArm64

bnvinay92:
The coroutines bump to 1.6.4 lets us add the iosSimulatorArm64 target to run simulators on M1 macs.

### DIFF
--- a/buildSrc/src/main/java/kotlin-multiplatform.gradle.kts
+++ b/buildSrc/src/main/java/kotlin-multiplatform.gradle.kts
@@ -9,4 +9,14 @@ extensions.getByType(JavaPluginExtension::class).apply {
   targetCompatibility = JavaVersion.VERSION_1_8
 }
 
+kotlin {
+  sourceSets {
+    all {
+      languageSettings.apply {
+        optIn("kotlin.RequiresOptIn")
+      }
+    }
+  }
+}
+
 project.kotlinCommonSettings(bomConfigurationName = "commonMainImplementation")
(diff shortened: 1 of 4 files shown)


buildSrc/src/main/java/kotlin-multiplatform.gradle.kts
```diff
@@ -9,4 +9,25 @@ extensions.getByType(JavaPluginExtension::class).apply {
   targetCompatibility = JavaVersion.VERSION_1_8
 }
 
+kotlin {
```
RBusarow, line 12: glad you added this here. 👍

RBusarow (approved):

workflow-core/build.gradle.kts
```diff
@@ -5,28 +5,15 @@ plugins {
 
 kotlin {
   jvm { withJava() }
```
steve-the-edwards, line 7: Could you add a quick comment here to look at the `kotlin-multiplatform` include for the other KMP targets?

steve-the-edwards (approved):

workflow-core/build.gradle.kts
bnvinay92: Imported the function from buildSrc instead.

buildSrc/src/main/java/kotlin-multiplatform.gradle.kts
bnvinay92: I learnt about precompiled scripts from this project so thank you folks!

steve-the-edwards (approved):

PR #864, opened 2022-08-18
Add Rendering Per Frame; Rework Render Pass Per Frame

steve-the-edwards:


workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RuntimeConfig.kt
```diff
@@ -20,19 +20,44 @@ public annotation class WorkflowExperimentalRuntime
...
+   * This version of the runtime will process all actions queued within [actionWaitMs] AND call
+   * render() after processing them (i.e. cascade up and down the tree) before passing the
+   * updated rendering to the UI or until [frameTimeoutMs] has passed, at which point it will
+   * pass the new rendering to the UI (in other words it can't process forever if actions are
+   * queueing faster than [actionWaitMs]).
+   *
+   * i.e. this is one "Rendering" per 'frame'.
+   */
+  @WorkflowExperimentalRuntime
+  public data class RenderingPerFrame(
```
rjrjr, line 47: Sorry to bike shed, but I'm having a hard time distinguishing btw. the two names. Perhaps:

- `ExhaustActionsBeforeRenderPass`
- `DropExtraRenderings`

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowRunner.kt
```diff
@@ -72,58 +72,61 @@ internal class WorkflowRunner<PropsT, OutputT, RenderingT>(
...
-      // pass off something to the UI!
-      while (processingResult == null && frameTimeLeft > 0) {
+      // Then we can process any more actions queued up until the frame timeout when we should
+      // pass off something to the UI! If processingResult != null, that means we have an output
+      // for the external system, so we must go back and render then as well. If the result
+      // is ActionsExhausted, there are no actions ready to process.
+      while (
+        processingResult == null &&
+        processingResult !is ActionsExhausted &&
+        frameTimeLeft > 0
```
rjrjr, line 111: `processingResult == null && processingResult !is ActionsExhausted` is always true.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RuntimeConfig.kt
steve-the-edwards: Oh no, I thought these names were so smart. Maybe we should have a quick huddle to decide so it minimizes the thrash.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
```diff
@@ -74,11 +74,11 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
     Channel<WorkflowAction<PropsT, StateT, OutputT>>(capacity = UNLIMITED)
   private var state: StateT
 
-  private val context = RealRenderContext(
-    renderer = subtreeManager,
-    sideEffectRunner = this,
-    eventActionsChannel = eventActionsChannel
-  )
+  // private var context = RealRenderContext(
```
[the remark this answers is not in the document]
steve-the-edwards: This was temp testing I forgot to change back.

steve-the-edwards:
This work is obselete. Went with a slightly different implementation for Conflate Stale Renderings runtime.

PR #865, opened 2022-08-18
Update Poetry Cyborg for Perf Tests

steve-the-edwards:
The wait for interstitial was never working and was just timing out. This made these tests both slow and less reliable.

We do not need to go up to reset to the first screen unless the app was already running manually before the test (that is rare - so catch it with an exception).

Wait for poetry package was not working with depth()


RBusarow (approved):

rjrjr (approved):
Yikes. I guess UI Automater isn't a panacea.

PR #868, opened 2022-08-23
Add Conflate Stale Renderings Runtime

steve-the-edwards:
Performance tests for this new runtime are forthcoming, but not included in this PR.

### DIFF
--- a/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
+++ b/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
@@ -1,5 +1,6 @@
 package com.squareup.workflow1
 
+import com.squareup.workflow1.RuntimeConfig.ConflateStaleRenderings
 import com.squareup.workflow1.internal.WorkflowRunner
 import com.squareup.workflow1.internal.chained
 import kotlinx.coroutines.CancellationException
@@ -101,7 +102,7 @@ import kotlinx.coroutines.launch
  * A [StateFlow] of [RenderingAndSnapshot]s that will emit any time the root workflow creates a new
  * rendering.
  */
-@OptIn(ExperimentalCoroutinesApi::class)
+@OptIn(ExperimentalCoroutinesApi::class, WorkflowExperimentalRuntime::class)
 public fun <PropsT, OutputT, RenderingT> renderWorkflowIn(
   workflow: Workflow<PropsT, OutputT, RenderingT>,
   scope: CoroutineScope,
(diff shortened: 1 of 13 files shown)


workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
```diff
@@ -133,21 +134,60 @@ public fun <PropsT, OutputT, RenderingT> renderWorkflowIn(
     }
   )
 
+  suspend fun <PropsT, OutputT, RenderingT> renderAndEmitOutput(
+    runner: WorkflowRunner<PropsT, OutputT, RenderingT>,
+    actionResult: ActionProcessingResult?,
+    onOutput: suspend (OutputT) -> Unit
+  ): RenderingAndSnapshot<RenderingT> {
+    // After receiving an output from the actions that were processed,
```
rjrjr, line 142: Odd to have the same comment in two places (here and line 171). I think it's just confusing here.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
```diff
@@ -133,21 +134,60 @@ public fun <PropsT, OutputT, RenderingT> renderWorkflowIn(
...
+    val nextRenderAndSnapshot = runner.nextRendering()
+
+    when (actionResult) {
+      is WorkflowOutput<*> -> {
+        @Suppress("UNCHECKED_CAST")
+        val output = actionResult as? WorkflowOutput<OutputT>
+        output?.let { onOutput(it.value) }
+      }
+      else -> {} // no -op
+    }
```
rjrjr, line 154: Does a single cast not work?

```suggestion
   @Suppress("UNCHECKED_CAST")
   (actionResult as? WorkflowOutput<OutputT>)?.let {
     onOutput(it.value)
   }
```

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
```diff
@@ -133,21 +134,60 @@ public fun <PropsT, OutputT, RenderingT> renderWorkflowIn(
...
-      // don't surprise anyone with an unexpected rendering pass. Show's over, go home.
+      // After resuming from runner.processAction() our coroutine could now be cancelled, check so
+      // we don't surprise anyone with an unexpected rendering pass. Show's over, go home.
       if (!isActive) return@launch
 
       // After receiving an output, the next render pass must be done before emitting that output,
       // so that the workflow states appear consistent to observers of the outputs and renderings.
-      renderingsAndSnapshots.value = runner.nextRendering()
-      output?.let { onOutput(it.value) }
+      nextRenderAndSnapshot = renderAndEmitOutput(runner, actionResult, onOutput)
```
rjrjr, line 173: We're violating what this comment warns about -- under `ConflateStateRenderings`, we may emit several outputs before the next rendering. Question is, does that matter? @zach-klippenstein and I convinced each other that it does, but I don't believe that any more -- sanity check requested.

If we do think it's true, then this method needs to exit immediately after emitting output, right? Perhaps it becomes a `do {} while()`.

steve-the-edwards: IC what you mean. The wording needs more specificity. We *are* doing a Render Pass before emitting each output (they are tied together in the helper method).

What we are not doing is emitting the new *rendering* in the flow (and thus being able to update any UI).

You would likely have a better practical sense of use cases, but in theory the Output handling should be orthogonal to UI updates. Also we know that UI is already stale, so we wait to give you the fresh one. But Output by nature of being an event, can never be stale.

steve-the-edwards:
```

      // If the action did produce an Output, we send it immediately after the render pass.
      nextRenderAndSnapshot = renderAndEmitOutput(runner, actionResult, onOutput)

      if (runtimeConfig == ConflateStaleRenderings) {
        // With this runtime modification, we do not pass renderings we know to be stale. This
        // means that we may be calling onOutput out of sync with the update of the UI. Output
        // is an event though, and should always occur immediately - i.e. it cannot be stale.
```

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
steve-the-edwards: Removed and clarified.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
steve-the-edwards: Changed to this.

rjrjr (approved):

PR #870, opened 2022-08-30
adds new board to game over screen

### LINKED ISSUE #205: Add a "New Board" button to the Game Over dialog in the Dungeon Crawler sample.

LandonPatmore:
closes #205


samples/dungeon/app/src/main/java/com/squareup/sample/dungeon/GameSessionWorkflow.kt
```diff
@@ -46,6 +51,10 @@ class GameSessionWorkflow(
     data class GameOver(val board: Board) : State()
   }
 
+  sealed class Output {
+    object NewBoard : Output()
```
JoelBryceAnderson, line 55: Seems like this doesn't need to be a sealed class if we do the same thing any time any type of `Output` is emitted!

LandonPatmore: I went with this instead of just an `object Output` in the spirit of it being a sample that people can add to. I will foundations make the call here.

rjrjr: We actively discourage this kind of things as a speculative YAGNI (you ain't gonna need it) violation. Definitely wouldn't want to see it in a sample.

I won't be able to give the PR a proper review until next week. In the meantime, thanks for the contribution!

steve-the-edwards (approved):
LGTM with the change of `Output` to an `object NewBoard`

PR #872, opened 2022-09-10
It is not safe to touch decorView before Dialog.show()

rjrjr:
You'd think I'd remember that by now.
This should be the fix for #863, but I haven't managed to reproduce that so it's hard to say for sure. Hoping @bnvinay92 can confirm.

Fixes #863.

PR #873, opened 2022-09-12
update kotlin to 1.6.21

RBusarow:
- [x] run Square's internal CI against a snapshot (and pass)


rjrjr (approved):

steve-the-edwards (approved):
Nice!

rjrjr:
@RBusarow Presuming it passes the internal suite, I'd like to get https://github.com/square/workflow-kotlin/pull/872 on its own before this, to minimize the swirl around the AlertDialog fix. Should be able to do that today. Is that going to get in your way?

RBusarow:
> @RBusarow Presuming it passes the internal suite, I'd like to get #872 released on its own before this, to minimize the swirl around the AlertDialog fix. Should be able to do that today. Is that going to get in your way?

We have a handful of failing shards, but at first glance I don't think they're related to this update.  I won't be digging into that more until late today at the earliest, so I have no problem in holding back a bit. :)

RBusarow:
@rjrjr @steve-the-edwards I've added a second commit to drop the "-Xfriend-paths=___" compiler argument for `:workflow-runtime`.  `WorkflowNode` was no longer accessible to `WorkflowNodeBenchmark` after the Kotlin update.  Instead, I'm using the shiny newish `workflowNodeCompilation.associateWith(main)`.

https://github.com/square/workflow-kotlin/pull/873/commits/1256b0ada35f92410dffe80aa8ee50578c24e1b9

Issue #546, opened 2021-09-08
Wrapper renderings introduce an extra AndroidView unnecessarily when the wrapped rendering's factory is a ComposeScreenViewFactory.


zach-klippenstein:
I just realized this wrapping pattern (even before this change) might introduce an extra view unnecessarily when the wrapped rendering's factory is a `ComposeViewFactory`. It shouldn't affect correctness but something we should probably check/test and to keep in mind as yall adopt compose. 

__Originally posted by @zach-klippenstein in https://github.com/square/workflow-kotlin/pull/544#discussion_r704522274__

I think this could probably be addressed if you came up with a way to generalize the wrapping pattern (as @rjrjr suggested on [slack](https://workflow-community.slack.com/archives/CHTFPR277/p1631109753098600?thread_ts=1631109042.098100&channel=CHTFPR277&message_ts=1631109753.098600)), since `WorkflowRendering`'s `asComposeViewFactory` could use that to do the type check at the leaf.

rjrjr:
This is still a problem. I've tried and failed to find a general wrapping pattern a number of times, and I'm not confident a practical solution will emerge. A lot of wrapping is done for very view-system-specific purposes, like manipulating View instances and messing with Context. We'd need to come up with a two pass system that allows platform neutral code to manipulate just the structure and ViewEnvironment, and another that actually builds a View or provides a Composable.

In the meantime, I think we need to have the workflow compose module provide an alternative `ScreenViewFactoryFinder` that replaces the stock bindings for our out of the box wrappers (`NamedScreen`, `EnvironmentScreen`, `AsScreen`) with Compose variants. And provide tools for app developers to do the same.

rjrjr:
`VisualFactory` to be introduced with #874 should address this.

rjrjr:
`VisualFactory` is still far from production ready, and migrating to it will be yet another big lift. In the meantime, we might be able to give `asComposeViewFactory()` hooks to put Compose or Classic alternative bindings in place -- perhaps alternative `ScreenViewFactoryFinder` implementations?

@helios175 has done something similar to this in a one-off.

lucamtudor:
We're lucky enough to have almost 100% Compose UI so I'm planning to tackle this internally, and maybe just do this:

[...]

PR #874, opened 2022-09-13
VisualFactory replaces basically everything

rjrjr:
_This is nowhere near ready for review. It's not my fault that GitHub insists on bothering people about draft PRs._

@helios175 prototyped a drastically stripped down alternative to ScreenViewFactory and all its relatives. This branch applies that model to Workflow, particularly with an eye toward fixing #546.

* `interface VisualFactory<ContextT, RenderingT, VisibleT>` replaces all of ViewRegistry, ScreenViewFactoryFinder, ScreenViewFactory, ComposeScreenViewFactory OverlayDialogFactory, etc.
  * `VisibleT` might be `View`, `@Composable (Modifier) -> Unit`, `Dialog`...
  * `ContextT` is `ContextOrContainer` for `View`, `Context` for `Dialog` factories, `Unit` for Compose
  * There are no subtypes of `VisualFactory`, which makes it trivial to decorate

* `VisualFactory` builds `VisualHolder`, which replaces `ScreenViewHolder`, etc.

* `VisualFactoryConverter` simplifies converting a factory for one type of visual for another, e.g. to wrap `@Composable fun Content()` factories with `ComposeView`, or in the other direction to wrap a `View` builder with `@Composable fun AndroidView()`

* `VisualEnvironment` is just a new name for `ViewEnvironment`, moved to a new package. A `ViewEnvironment` wrapper still exists in the old location, but I don't know if this will really be practical for migration

* `Visualizer` provides the core support for `WorkflowViewStub`, `@Composable WorkflowRendering()`, other containers. 
  * Also replaces the static `ScreenViewFactory.startShowing` and `ScreenViewHolder.show` functions, I'm pretty sure.
  * The plan is to have one `Visualizer` able to turn any rendering into a `View`, and another able to turn any rendering into a Composable. Feature developers will bind their leaf `View` and composable factories to specific `VisualEnvironmentKey`s, and the `Visualizers` will wrap them with appropriate converters, standard factories, etc.

Most test are broken, but known passing tests are:

* `WorkflowViewStubLifecycleTest` exercising both `VisualFactory` implementations and a `VisualFactory` based wrapper of `ScreenViewFactoryFinder` for backward compatibility.

* `EnvironmentScreenAndroidIntegrationTest`

Lots and lots of `TODO` comments that should become github issues.

Biggest bit of remaining uncertainty is all the nooks and crannies of `OverlayDialogHolder`, in support of

* back button hooks
* bounds management
* modal behavior

Latest thinking there is to make `VisualT` into a struct that wraps the `Dialog` and its `updateBounds` method. For back button concerns, we may be able to get out of that business entirely with the recent introduction of `ViewTreeOnBackPressedDispatcherOwner` to JetPack. We should address that ASAP on the main branch.

Also still need to finish demonstrating the Compose integration, there is some WIP, but the prototype @helios175 built gives me a lot of confidence there.


workflow-ui/core-common/src/main/java/com/squareup/workflow1/visual/ExactTypeVisualFactory.kt
```diff
@@ -0,0 +1,55 @@
...
+        )?.let { delegateHolder ->
+          object : VisualHolder<RenderingT, VisualT> by delegateHolder {
+            override fun update(rendering: RenderingT): Boolean {
+              return predicate(rendering) && delegateHolder.update(rendering)
+            }
+          }
+        }
+      } else null
+    }
+  }
```
helios175, line 54: Can move to a different file.

workflow-ui/core-common/src/main/java/com/squareup/workflow1/visual/MultiRendering.kt
```diff
@@ -0,0 +1,50 @@
...
+ * different types or compatibilities, and so require different factories and holders.
+ */
+@WorkflowUiExperimentalApi
+public abstract class MultiRendering<UiContextT, VisualT> {
+
+  private var currentHolder: VisualHolder<Any, VisualT>? = null
+
+  public val currentVisual: VisualT get() = currentHolder!!.visual
+
+  public val currentVisualOrNull: VisualT? get() = currentHolder?.visual
```
helios175, line 16: I'd aim to name them without `current`. No strong opinion.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowViewStub.kt
```diff
@@ -210,36 +213,30 @@ public class WorkflowViewStub @JvmOverloads constructor(
...
+      WorkflowLifecycleOwner.installOn(newView)
+      doFirstUpdate()
+
+      if (inflatedId != NO_ID) newView.id = inflatedId
+      if (updatesVisibility) newView.visibility = visibility
+      background?.let { newView.background = it }
+      propagateSavedStateRegistryOwner(newView)
+
+      replaceOldViewInParent(parent, newView)
+    }
```
helios175, line 239: I'd eventually aim for making replaceWith receive a `doUpdate = true / false` and use false for this case.
Or have onNewVisualBeforeUpdate / onNewVisualAfterUpdate. 🤷 
Having MultiRendering do the update for us makes sense, but it shouldn't complicate us.

rjrjr: Not really following. I _love_ this, providing a lambda that gets called only if there's an update to do. Seems much simpler than what I was doing before.

rjrjr: Maybe give the lambda three arguments, build `oldViewOrNull` into it.

rjrjr:
This is a much cleaner API than the thing we evolved, but moving to it would require a herculean effort that we just can't make. It is also not the case that this is required to fix #546, I have something simpler in the works for that.

Still interested in landing this some day possibly at its own library, maybe Compose-first, but it's not part of the Workflow roadmap

PR #887, opened 2022-10-25
Add js target

hujim:
Rebased after the Kotlin 1.6.21 update.
Reviewing by commits should help.

- Add JS target and artifacts.
- Fix test names to non back-ticks for common tests can't have them for JS test support.
- Updated to use new coroutine testing library methods.
- Handle different results ErrorLoggingInterceptor for jsTargets in tests.
- Add JS WorkflowIdentifierTypeNamer to handle JS class comparison.

Will clean up commits once reviewed.

### DIFF
--- a/.github/workflows/kotlin.yml
+++ b/.github/workflows/kotlin.yml
@@ -164,6 +164,34 @@ jobs :
         with :
           report_paths : '**/build/test-results/test/TEST-*.xml'
 
+  js-tests :
+    name : JS Tests
+    runs-on : ubuntu-latest
+    timeout-minutes : 20
+    steps :
+      - uses : actions/checkout@v3
+      - uses : gradle/wrapper-validation-action@v1
+      - name : set up JDK 11
+        uses : actions/setup-java@v3
+        with :
+          distribution : 'zulu'
+          java-version : 11
+
+      ## JS Specific Tests (for KMP js actuals in core and runtime).
+      - uses : gradle/gradle-build-action@v2
+        name : Check with Gradle
+        with :
+          arguments : |
+            jsTest --stacktrace
+          cache-read-only : false
+
+      # Report as Github Pull Request Check.
+      - name : Publish Test Report
+        uses : mikepenz/action-junit-report@v3
+        if : always() # always run even if the previous step fails
+        with :
+          report_paths : '**/build/test-results/test/TEST-*.xml'
+
   performance-tests :
     name : Performance tests
     runs-on : macos-latest
--- a/gradle.properties
+++ b/gradle.properties
@@ -25,3 +25,8 @@ POM_DEVELOPER_ID=square
 POM_DEVELOPER_NAME=Square, Inc.
 POM_DEVELOPER_URL=https://github.com/square/
 SONATYPE_STAGING_PROFILE=com.squareup
+
+# As LEGACY JS compiler is deprecated and we only anticipate IR compatible consumers,
+# we're only supporting IR for now.
+# For details see https://kotlinlang.org/docs/js-ir-compiler.html
+kotlin.js.compiler=ir
--- a/workflow-core/src/jsMain/kotlin/com.squareup.workflow1/WorkflowIdentifierTypeNamer.kt
+++ b/workflow-core/src/jsMain/kotlin/com.squareup.workflow1/WorkflowIdentifierTypeNamer.kt
@@ -1,0 +1,31 @@
+package com.squareup.workflow1
+
+import kotlin.reflect.KClass
+
+internal actual object WorkflowIdentifierTypeNamer {
+  // Stores mappings between KClass instances and their assigned names.
+  val mappings = mutableMapOf<KClass<*>, String>()
+
+  // Note: This implementation does not differentiate between generic workflows.
+  // (ie. SomeGenericWorkflow<String> and SomeGenericWorkflow<Int> would both return back the same
+  // value.)
+  //
+  // Recommended workarounds:
+  // - Always provide a key for generic workflows
+  // - Create non-generic subclasses of generic workflows
+  public actual fun uniqueName(kClass: KClass<*>): String {
+    // Note: `kClass.qualifiedName` cannot be used here like other platforms as it's not supported
+    // for JS. Therefore, we construct a unique name of each static KClass based on its simple name
+    // and an index of when it was encountered.
+
+    val mapping = mappings[kClass]
+    if (mapping != null) {
+      return mapping
+    }
+
+    // `simpleName` does not differentiate between generic workflows.
+    val identifier = "${kClass.simpleName ?: kClass.hashCode()}(${mappings.size})"
+    mappings[kClass] = identifier
+    return identifier
+  }
+}
--- a/workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
+++ b/workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
@@ -54,7 +54,7 @@ class RenderWorkflowInTest {
     testScope = TestScope(UnconfinedTestDispatcher())
   }
 
-  @Test fun `initial rendering is calculated synchronously`() {
+  @Test fun initial_rendering_is_calculated_synchronously() {
     runtimeTestRunner.runParametrizedTest(
       paramSource = runtimeOptions,
       before = ::setup,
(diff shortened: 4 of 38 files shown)


workflow-core/src/jsMain/kotlin/com.squareup.workflow1/WorkflowIdentifierTypeNamer.kt
```diff
@@ -0,0 +1,23 @@
...
+    // Note: `kClass.qualifiedName` cannot be used here like other platforms as it's not supported
+    // for JS. Therefore, we construct a unique name of each static KClass based on its simple name
+    // and an index of when it was encountered.
+
+    val mapping = mappings[kClass]
+    if (mapping != null) {
+      return mapping
+    }
+
+    val identifier = "${kClass.simpleName ?: kClass.hashCode()}(${mappings.size})"
```
vng-sq, line 19: Note: This name doesn't differentiate against a genericized Workflow (although IIRC this same problem exists with `kClass.qualifiedName` in the native implementations too). Should I have put a comment here to mention this limitation?

workflow-runtime/src/jsTest/kotlin/com/squareup/workflow1/ErrorLoggingInterceptor.kt
```diff
@@ -0,0 +1,22 @@
+package com.squareup.workflow1
+
```
steve-the-edwards, line 2: I can't see what's different here that we are requiring to expect/actual these? Is it in a later commit?

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/WorkflowInterceptorTest.kt
```diff
@@ -198,7 +200,7 @@ internal class WorkflowInterceptorTest {
         key: String,
         sideEffect: suspend CoroutineScope.() -> Unit
       ) {
-        runBlocking { sideEffect() }
+        launch { sideEffect() }
```
steve-the-edwards, line 203: where do we ensure that this gets run? Here and elsewhere you could be using `runTest(UnconfinedTestDispatcher()` if we want everythign to go ahead eagerly.

workflow-core/src/jsMain/kotlin/com.squareup.workflow1/WorkflowIdentifierTypeNamer.kt
steve-the-edwards: Definitely put the comment about the limitations. Where were you seeing this problem for the kotlin native code? Do you have a link I can read more about this?
Was this a suggested solution for this with kotlin-js? Or something you invented?

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
```diff
@@ -93,7 +93,7 @@ class RenderWorkflowInTest {
   }
 
   @Test
-  fun side_effects_from_initial_rendering_in_root_workflow_are_never_started_when_scope_cancelled_before_start() {
+  fun `side_effects_from_initial_rendering_in_root_workflow_are_never_started_when_scope_cancelled_before_start`() {
```
steve-the-edwards, line 96: So we can use backticks just not spaces? was this becasue ktlint was failing these without the backticks?

hujim: yes precisely, we can use backticks but no special characters when compiling to JS.
ktlint marks this as a MAX_LENGTH violation without the backtick

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/WorkflowInterceptorTest.kt
hujim: Ah I might have missed this one, but I've added `advanceUntilIdle()` where these are required, I should add this here as well to prevent confusion.

But I've confirmed this isn't required in this particular test as though this could be ran in parallel, the render()+assertion still always happens at the end of the test in the single thread.

Does that make sense? Though it's possible I might be reading this wrong https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-test/kotlinx.coroutines.test/run-test.html

workflow-runtime/src/jsTest/kotlin/com/squareup/workflow1/ErrorLoggingInterceptor.kt
hujim: it's subtle but js uses `simpleName` instead of `qualifiedName` in the other targets

workflow-core/src/jsMain/kotlin/com.squareup.workflow1/WorkflowIdentifierTypeNamer.kt
vng-sq: If I run the following against the JVM, it doesn't include the generic type:

```
data class GenericType<T>(
  val t: T
)

val someType = GenericType<String>(t = "abc")
println(someType::class.qualifiedName) // prints "com.squareup.sandbox.GenericType"
```

As for this implementation, I invented this solution. This seems to work in practice since it appears Kotlin/JS uses a static class (probably stored on its prototype) to represent the KClass. Open to suggestions though!

.github/workflows/kotlin.yml
```diff
@@ -164,6 +164,34 @@ jobs :
         with :
           report_paths : '**/build/test-results/test/TEST-*.xml'
 
+  js-tests :
+    name : JS Tests
+    runs-on : macos-latest
```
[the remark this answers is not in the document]
RBusarow: The macOS runners are actually free now for public repos, but there's a limit of 5 concurrent jobs so it's still a good idea to use them sparingly.

https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits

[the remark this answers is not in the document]
hujim: good to know, I've changed to ubuntu cause these don't need mac at all.

workflow-core/src/jsMain/kotlin/com.squareup.workflow1/WorkflowIdentifierTypeNamer.kt
steve-the-edwards: Ah I C what you mean now.  Usuallly that is when the identifier would be overridden I believe.

workflow-runtime/src/iosTest/kotlin/com/squareup/workflow1/ErrorLoggingInterceptor.kt
```diff
@@ -0,0 +1,22 @@
...
+    errors += text
+  }
+
+  actual companion object {
+    actual val EXPECTED_ERRORS = listOf(
+      "ErrorLoggingInterceptor.logBeforeMethod threw exception:\n" +
+        IllegalArgumentException::class.qualifiedName.toString(),
+      "ErrorLoggingInterceptor.logAfterMethod threw exception:\n" +
+        IllegalArgumentException::class.qualifiedName.toString()
+    )
```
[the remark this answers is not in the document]
hujim: fixed!

hujim:
@steve-the-edwards ready for a final pass
I've merged the commits where it makes sense, but left some separate for better readability.

gradle.properties
```diff
@@ -25,3 +25,5 @@ POM_DEVELOPER_ID=square
 POM_DEVELOPER_NAME=Square, Inc.
 POM_DEVELOPER_URL=https://github.com/square/
 SONATYPE_STAGING_PROFILE=com.squareup
+
+kotlin.js.compiler=ir
```
steve-the-edwards, line 29: Do we have a link to a support article describing why this is the best?

steve-the-edwards (approved):

gradle.properties
vng-sq: https://kotlinlang.org/docs/js-ir-compiler.html

_Rather than directly generating JavaScript code from Kotlin source code, the Kotlin/JS IR compiler backend leverages a new approach. Kotlin source code is first transformed into a [Kotlin intermediate representation (IR)](https://kotlinlang.org/docs/whatsnew14.html#unified-backends-and-extensibility), which is subsequently compiled into JavaScript. For Kotlin/JS, this enables aggressive optimizations, and allows improvements on pain points that were present in the previous compiler, such as generated code size (through dead code elimination), and JavaScript and TypeScript ecosystem interoperability, to name some examples._

steve-the-edwards: Thanks! I should have been clearer though with the comment. I meant can you include that link as a comment in the .properties file so we have reference in the future

hujim: would just adding the link suffice? or do we prefer also a snippet of what this is?

hujim: added a bit of comment explaining the choice and the link

steve-the-edwards (approved):
LGTM!

PR #890, opened 2022-10-27
update the Vanniktech gradle publish plugin to 0.22.0

RBusarow:
TLDR - we don't need `--no-parallel` anymore.

https://github.com/vanniktech/gradle-maven-publish-plugin/blob/master/CHANGELOG.md#version-0220-2022-09-09

Version 0.22.0 *(2022-09-09)*
---------------------------------

- **NEW**: When publishing to maven central by setting `SONATYPE_HOST` or calling `publishToMavenCentral(...)` the plugin will now explicitly create a staging repository on Sonatype. This avoids issues where a single build would create multiple repositories
- The above change means that the plugin supports parallel builds and it is not neccessary anymore to use `--no-parallel` and `--no-daemon` together with `publish`
- **NEW**: When publishing with the `publish` or `publishAllPublicationsToMavenCentralRepository` tasks the plugin will automatically close the staging repository at the end of the build if it was successful.
- **NEW**: Option to also automatically release the staging repository after closing was susccessful
```
SONATYPE_HOST=DEFAULT # or S01
SONATYPE_AUTOMATIC_RELEASE=true
```
or
```
mavenPublishing {
  publishToMavenCentral("DEFAULT", true)
  // or publishToMavenCentral("S01", true)
}
```
- in case the option above is enabled, the `closeAndReleaseRepository` task is not needed anymore
- when closing the repository fails the plugin will fail the build immediately instead of timing out
- when closing the repository fails the plugin will try to print the error messages from Nexus
- increased timeouts for calls to the Sonatype Nexus APIs
- fixed incompatibility with the `com.gradle.plugin-publish` plugin
- added wokaround for Kotlin multiplatform builds reporting disabled build optimizations

### DIFF
--- a/.github/workflows/publish-snapshot.yml
+++ b/.github/workflows/publish-snapshot.yml
@@ -28,7 +28,7 @@ jobs:
       - name : Publish Snapshots
         run: |
           ./gradlew build --no-daemon
-          ./gradlew publish --no-parallel --no-daemon
+          ./gradlew publish --no-daemon
         env :
           ORG_GRADLE_PROJECT_mavenCentralUsername : ${{ secrets.SONATYPE_NEXUS_USERNAME }}
           ORG_GRADLE_PROJECT_mavenCentralPassword : ${{ secrets.SONATYPE_NEXUS_PASSWORD }}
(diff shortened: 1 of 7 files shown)


rjrjr (approved):
Nice! Thanks.

PR #896, opened 2022-11-09
Base case of WorkflowRendering with the new ComposeMultiRendering

wugeorgeq:
Commented out the rest of the `Samples.kt` just to make sure our changes to `WorkflowRendering` work so far


workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposeMultiRendering.kt
```diff
@@ -0,0 +1,22 @@
+package com.squareup.workflow1.ui.compose
+
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+import com.squareup.workflow1.visual.MultiRendering
+import com.squareup.workflow1.visual.VisualEnvironment
+import com.squareup.workflow1.visual.VisualHolder
+
+@WorkflowUiExperimentalApi
+public class ComposeMultiRendering : MultiRendering<Unit, ComposableLambda>() {
```
wugeorgeq, line 9: is `MultiRendering` the same as this "Aggregated visual factory" concept in [helios' writeup](https://github.com/squareup/helios-workflow-proposal)?

samples/compose-samples/src/main/java/com/squareup/sample/compose/hellocomposeworkflow/HelloComposeWorkflow.kt
```diff
@@ -57,5 +59,6 @@ fun HelloComposeWorkflowPreview() {
     onOutput = {},
     runtimeConfig = AndroidRuntimeConfigTools.getAppWorkflowRuntimeConfig()
   )
-  WorkflowRendering(rendering, ViewEnvironment.EMPTY)
+  // not sure if I should be adding the factory here
```
rjrjr, line 62: Yeah, I think this is the problem. Looking for the right place…

rjrjr: You were so close. Put this on line 57:

```kotlin
  val rendering by HelloComposeWorkflow
    .mapRendering {
      it.withEnvironment(
        ViewEnvironment.EMPTY.plus(Pair(ComposableFactoryKey, HelloFactory))
      )
    }
```

rjrjr: Oops, nope. That _should_ have fixed it, but we need to add support to `VisualFactory` support for `EnvironmentScreen`.

rjrjr: Haven't done that yet on the classic view spike either, instead I make `AndroidScreen` work and was careful to use that on anything I've tried to render so far.

rjrjr: You can see in `AndroidViewMultiRendering` how I added support for `Named` and `NamedScreen` for classic views, along with a lot of hand wringing about whether I did it in the right place or not.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposeMultiRendering.kt
rjrjr: It's more that `MultiRendering` subclasses are the things that use the aggregated factory. They're the shared code for the job that `WorkflowViewStub` and `WorkflowRendering()` both perform:

* holding on to a `var rendering`
* creating the visual for it (via the aggregated thing)
* updating the rendering and
  * updating the existing visual if it's compatible
  * or creating a new visual

The aggregated visual factory is the thing that includes all the leaf compose factories, all the leaf view factories, and the glue code that can convert one type to the other type.

wugeorgeq:
Still WIP, just pushing to cmd+s

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/WorkflowRendering.kt
```diff
@@ -84,14 +84,15 @@ public fun WorkflowRendering(
...
-      rendering.toViewFactory(viewEnvironment).asComposeViewFactory()
-    }
+    val multiRendering = remember { ComposeMultiRendering() }
+    multiRendering.replaceWith(
+      rendering = rendering,
+      context = Unit,
+      environment = viewEnvironment,
+      onNewVisual = { newView, doFirstUpdate ->
+        doFirstUpdate()
+      }
```
wugeorgeq, line 94: Oh shoot, I handwaved this, not sure if something more is supposed to go here

rjrjr: Are the lifecyle tests in `WorkflowRenderingTest` passing? If so…huh! If not, I expect the fix will require moving more of this method to happen inside that lambda. That's the thing that is giving us a new `@Composable content()` function to call.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposableFactory.kt
```diff
@@ -0,0 +1,52 @@
...
+public fun <RenderingT : Any> composableFactory(
+  block: @Composable (RenderingT, Modifier) -> Unit
+): ComposableFactory<RenderingT> =
+  object : ComposableFactory<RenderingT> {
+    override fun createOrNull(
+      rendering: RenderingT,
+      context: Unit,
+      environment: VisualEnvironment
+    ) = object : ComposableHolder<RenderingT> {
+      // By changing the mutable state, any "update"
```
helios175, line 30: nit: missing the rest of the phrase?

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposableFactory.kt
```diff
@@ -0,0 +1,52 @@
...
+      val currentRendering = mutableStateOf(rendering)
+      override val visual: ComposableLambda = { modifier ->
+        block(currentRendering.value, modifier)
+      }
+
+      override fun update(rendering: RenderingT): Boolean {
+        // By changing the mutable state, any composition using the "output" will be updated.
+        currentRendering.value = rendering
+        // TODO ?
```
helios175, line 39: I think that's it. It'll update it.

PR #897, opened 2022-11-10
Introduces `Wrapper`, `WithName`

rjrjr:
First use of `VisualFactory`'s ability to create general purpose factories that are agnostic about specific view and context types. And looky look, I think I've finally found a way to do that without sacrificing support for marker types like `Screen` and `Overlay`.

We introduce this class hierarchy:

```
abstract class Wrapper<W>
  abstract class WithName<W> : Wrapper<W>
    abstract class Named<W> : WithName<W>
    abstract class NamedScreen<W> : WithName<W>, Screen
```

Along with a new `VisualFactory.withName()` extension, a single function that can generate `VisualFactory` functions for simple subtypes of `WithName`.

Also moves `VisualHolder` out to its own file, otherwise unchanged.

`Wrapper` is justified mainly as a convenience to ensure that wrapper renderings include the support for `Compatible` that we almost always forget about. The standardized structure it will invite could also be a boon to test code.

`WithName<W> : Wrapper<W>` is base class for things like `NamedScreen`, and the forthcoming `NamedOverlay` peer that I already need to create for `Dialog`. Would have preferred to call it `Named`, but that name is taken for now.

`VisualFactory.withName()` provides general support for all concrete implementations of `WithName`. It can tranform a factory for renderings of type `R` into one for a particular subtype of `Wrapper<R>`. Note that we're able to delete `VisualFactory.named()` and `VisualFactory.namedScreen()` functions in this PR.

Still need to:
 - Build the same for `EnvironmentScreen`
 - Introduce `NamedOverlay` and `EnvironmentOverlay`.
 - Figure out exactly where these functions should live in the `MultiRendering` hierarchy, taking care that they can be reconfigured

`WorkflowViewStubLifecycleTest` is still green. It uses `WorkflowUiTestActivity.setRendering`, which relies on `NamedScreen`, so yes we're sure the new code is being exercised.

A little more context on why this is worth doing. It wouldn't need to be anywhere near so complicated if we were willing to abandon our `Screen` and `Overlay` marker interfaces, e.g. change `WorkflowViewStub.show(Screen)` back to `WorkflowViewStub.show(Any)`. But that compile-time distinction between view-like-things (`Screen`) and `window-like-things` (`Overlay`) is pretty valuable when building features, and it would be a shame to lose it -- especially if we are able to keep it from leaking into the fundamental machinery like `AndroidViewFactory` and `AndroidViewMultiRendering`, used mainly under the hood. What's exciting about the approach in this PR is that with a little bit of ceremony we get to keep the marker types without creating entire parallel hierarchies of copy / paste factory implementations.


workflow-ui/core-common/src/main/java/com/squareup/workflow1/visual/Wrapper.kt
```diff
@@ -0,0 +1,69 @@
...
+ * else.
+ */
+@WorkflowUiExperimentalApi
+public abstract class Wrapper<W : Any>(
+  public val wrapped: W,
+) : Compatible {
+  /** Read only once by [compatibilityKey], at construction time. */
+  protected open val name: String = this::class.qualifiedName ?: "Wrapper"
+
+  final override val compatibilityKey: String by lazy { Compatible.keyFor(wrapped, name) }
```
helios175, line 18: May be we should also add a qualified into the compatibility key?
That qualifier might be `""` by default but overridable?
For instance: WithName.qualifier should be name:
Named(slot1)-content should not be compatible with Named(slot2)-content.

workflow-ui/core-common/src/main/java/com/squareup/workflow1/visual/Wrapper.kt
```diff
@@ -0,0 +1,69 @@
...
+      environment: VisualEnvironment
+    ): VisualHolder<W, V> {
+      val delegateHolder = delegateFactory.create(rendering.wrapped, context, environment)
+
+      return object : VisualHolder<W, V> {
+        override val visual: V get() = delegateHolder.visual
+
+        override fun update(rendering: W): Boolean {
+          return delegateHolder.update(rendering.wrapped)
+        }
```
helios175, line 65: I dind't look at create vs createOrNull.
Is create an extension that forces the result not to be null?
Why not implement this createOrNull based on the delegate's createOrNull (and return null when appropriate)?

rjrjr: create is just a convenienc extension that calls createOrNull and throws with a useful message, replaces a bunch of redunant calls to requireNotNull. There is no delegate.

workflow-ui/core-common/src/main/java/com/squareup/workflow1/visual/Wrapper.kt
rjrjr: What would that provided that the name argument to `Compatible.keyFor()` doesn't provide already?

rjrjr:
Rebased to catch up with #895.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/visual/AndroidViewMultiRendering.kt
```diff
@@ -26,93 +22,47 @@ import com.squareup.workflow1.visual.ContextOrContainer.AndroidContext
  * to bind to the [AndroidViewMultiRendering] key.
  */
 @OptIn(WorkflowUiExperimentalApi::class)
-public class AndroidViewMultiRendering : MultiRendering<Context, View>() {
+public class AndroidViewMultiRendering : MultiRendering<ContextOrContainer, View>() {
 
   override fun create(
     rendering: Any,
-    context: Context,
+    context: ContextOrContainer,
```
rjrjr, line 29: @wugeorgeq Stumbled across this when catching up to your work, missed it when reviewing earlier. FYI.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowViewStub.kt
```diff
@@ -216,10 +217,12 @@ public class WorkflowViewStub @JvmOverloads constructor(
...
     val oldViewOrNull = showing.visualOrNull
 
-    showing.replaceWith(rendering, context, viewEnvironment) { newView, doFirstUpdate ->
+    val parent = actual.parent as? ViewGroup
+      ?: throw IllegalStateException("WorkflowViewStub must have a non-null ViewGroup parent")
 
-      val parent = actual.parent as? ViewGroup
-        ?: throw IllegalStateException("WorkflowViewStub must have a non-null ViewGroup parent")
+    showing.replaceWith(
+      rendering, AndroidContainer(parent), viewEnvironment
```
rjrjr, line 224: @wugeorgeq This was what drove the work in #895. If you diff this file again `origin/main`, you'll see that we had lost the bit where the factory is given access to `parent`.

wugeorgeq (approved):

PR #898, opened 2022-11-11
Eliminates overhead from ContextOrContainer

rjrjr:


wugeorgeq (approved):
NICE

rjrjr:
It's all a lie: https://mastodon.jakewharton.com/@jw/109322308449353917

> Unfortunately, use of polymorphism always forces boxing. See lines 113 and 119 which box the "instances" before invoking test(). Also see the signature of the test function on line 92 which only accepts the interface and therefore requires heap-allocated instances.

PR #899, opened 2022-11-14
`WithEnvironment` introduced, `MultiRendering` is now `Visualizer`

rjrjr:
- `MultiRendering` is now `Visualizer`, and it is no longer `abstact`.
- `WithEnvironment` provides `VisualT` agnostic support for `EnvironmentScreen`, etc. With updated unit test.
- Extracts `WithName` to its own file, otherwise unchanged.
- Adds `getFactory` function param to `VisualFactory.createOrNull`
  - needed because wrapping based approaches can't handle recursion
  - e.g, `fact.withNamedScreenSupport().withEnvScreenSupport()` must allow `NamedScreen(EnvScreen(foo))`
- `AndroidViewFactoryKey` renamed `LeafAndroidViewFactoryKey`. It is now private, behind better documented `VisualEnvironment` extensions
- Similar private key used by `Visualizer` is named `VisualizerAndroidViewFactoryKey`
- Fixes a few more spots where `Context` should have been `ContextOrContainer`
- `VisualFactory` is now a `fun interface`, allowing a lot of boilerplate to be deleted.


workflow-ui/core-android/src/main/java/com/squareup/workflow1/visual/AndroidViewFactory.kt
```diff
@@ -11,21 +11,24 @@ import com.squareup.workflow1.ui.ScreenViewFactory
...
+//  View.environmentOrNull working.
+
 @WorkflowUiExperimentalApi
 public typealias AndroidViewFactory<R> = VisualFactory<ContextOrContainer, R, View>
 
+@WorkflowUiExperimentalApi
 public sealed interface ContextOrContainer {
-  @JvmInline
-  public value class AndroidContext(public val context: Context) : ContextOrContainer
+  public class AndroidContext(public val context: Context) : ContextOrContainer
```
rjrjr, line 26: The `value class` thing wasn't actually giving any benefit, let's not lie to ourselves.

workflow-ui/core-android/src/test/java/com/squareup/workflow1/ui/container/LegacyEnvironmentScreenAndroidIntegrationTest.kt
```diff
@@ -0,0 +1,50 @@
...
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+import com.squareup.workflow1.ui.plus
+import com.squareup.workflow1.ui.show
+import com.squareup.workflow1.ui.startShowing
+import com.squareup.workflow1.ui.toViewFactory
+import com.squareup.workflow1.visual.VisualEnvironment.Companion.EMPTY
+import org.junit.Test
+import org.mockito.kotlin.mock
+
+internal class LegacyEnvironmentScreenAndroidIntegrationTest {
```
rjrjr, line 16: Nothing new here, just renamed the original test.

workflow-ui/core-common/src/main/java/com/squareup/workflow1/visual/WithName.kt
```diff
@@ -0,0 +1,41 @@
...
+@WorkflowUiExperimentalApi
+public abstract class WithName<W : Any>(
+  wrapped: W,
+  name: String
+) : Wrapper<W>(wrapped) {
+  public final override val name: String by lazy {
+    require(name.isNotBlank()) { "name must not be blank." }
+    name
+  }
+}
```
helios175, line 20: nit: Maybe add a `fun T.withName(name): WithName<T>` convenience?

rjrjr: I don't think that's possible, since `WithName` is an abstract class. We can't know what concrete subtype of `WithName` to instantiate.

workflow-ui/core-common/src/main/java/com/squareup/workflow1/visual/Visualizer.kt
```diff
@@ -29,13 +29,15 @@ import com.squareup.workflow1.ui.compatible
  * APIs for that.
  */
 @WorkflowUiExperimentalApi
-public abstract class MultiRendering<UiContextT, VisualT> {
+public class Visualizer<ContextT, VisualT>(
+  private val loadIntegrationFactory: (environment: VisualEnvironment) -> VisualFactory<ContextT, Any, VisualT>
```
helios175, line 33: <3
Strategy > template
(composition > inheritance)

workflow-ui/core-common/src/main/java/com/squareup/workflow1/visual/VisualFactory.kt
```diff
@@ -23,21 +23,25 @@ public interface VisualFactory<ContextT, in RenderingT, out VisualT> {
...
+   *
+   * @param getFactory can be used to make recursive calls to build VisualT
+   * instances for sub-parts of [rendering]
    */
   public fun createOrNull(
     rendering: RenderingT,
     context: ContextT,
-    environment: VisualEnvironment
+    environment: VisualEnvironment,
+    getFactory: (VisualEnvironment) -> VisualFactory<ContextT, Any, VisualT>
```
helios175, line 34: This looks redundant:
I know, one thing is the environment and another how to get the factory.
But you could also provide () -> VF...
Anyway I have another question that might affect this, so let's wait for that.

rjrjr: We need this in order for `WithEnvironment` to work. `WithEnvironment` needs to be able to redefine what might get returned by that function, that's its main purpose.

rjrjr: Hmm. I think? Not going to rathole on this until we discuss your main question.

helios175 (approved):

workflow-ui/core-common/src/main/java/com/squareup/workflow1/visual/WithEnvironment.kt
```diff
@@ -0,0 +1,33 @@
...
+    rendering: WithEnvironment<W>,
+    context: C,
+    environment: VisualEnvironment,
+    getFactory: (VisualEnvironment) -> VisualFactory<C, Any, V>
+  ): VisualHolder<WithEnvironment<W>, V> {
+    val mergedEnvironment = environment + rendering.environment
+
+    val delegateHolder = getFactory(mergedEnvironment).create(
+      rendering.wrapped, context, mergedEnvironment, getFactory
+    )
```
rjrjr, line 23: @helios175 This is why `getFactory` needs the `VisualEnvironment` argument -- so that `VisualFactory` implementations can mess with it. That's a primary use case.

PR #903, opened 2022-12-10
Add Configurable Recursive Rendering to Performance Poetry

steve-the-edwards:
Customizable via intent.

Depth is how many times to recursively render PerformancePoemsBrowserWorkflow in the tree.
Breadth is how many times to repeate rendering of PerformancePoemsBrowserWorkflow horizontally in the tree.

Only one path of the tree provides a rendering.

### DIFF
--- a/benchmarks/performance-poetry/complex-poetry/src/main/java/com/squareup/benchmarks/performance/complex/poetry/instrumentation/SimulatedPerfConfig.kt
+++ b/benchmarks/performance-poetry/complex-poetry/src/main/java/com/squareup/benchmarks/performance/complex/poetry/instrumentation/SimulatedPerfConfig.kt
@@ -1,6 +1,7 @@
 package com.squareup.benchmarks.performance.complex.poetry.instrumentation
 
 import android.os.Parcelable
+import com.squareup.sample.poetry.RecursionGraphConfig
 import kotlinx.parcelize.Parcelize
 
 /**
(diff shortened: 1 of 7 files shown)


rjrjr:
The commit message should be a comment in the code somewhere.

I don't understand this:

> Only one path of the tree provides a rendering.

It looked like you were rendering children for breadth as well as depth, just that most of them ignored output. What do you actually mean by "provides a rendering"?

steve-the-edwards:
> The commit message should be a comment in the code somewhere.
> 
> I don't understand this:
> 
> > Only one path of the tree provides a rendering.
> 
> It looked like you were rendering children for breadth as well as depth, just that most of them ignored output. What do you actually mean by "provides a rendering"?

All of them are 'rendered' from the POV of Workflow runtime, but only 1 path of the tree (the (n-1)th child of each depth) actually returns the rendering from `render()` and so only one path passes a rendering to the UI layer. I will make that clearer in the comment I add.

benchmarks/performance-poetry/complex-poetry/src/main/java/com/squareup/benchmarks/performance/complex/poetry/instrumentation/SimulatedPerfConfig.kt
```diff
@@ -8,14 +8,34 @@ import kotlinx.parcelize.Parcelize
...
+ *   (m-1)th rendering is returned from render() though which means that the other renders will be
+ *   work for the runtime to render but they won't be passed to the UI layer.
+ * [repeatOnNext] Is the number of times x that an action should be created when 'next' is pressed
+ *   in the poem before the action takes its effect - advances the state to the next stanza.
+ * [simultaneousActions] Determines the number of workers y that should be created for each
+ *   'complex' state transition. This will result in y actions that need to be handled
+ *   simultaneously and tries to represent a scenario of multiple Workflows listening to the same
+ *   action.
+ * [traceRenderingPasses], [traceFrameLatency], [traceEventLatency] are all flags to add
+ *   instrumentation for different performance measurements.
```
steve-the-edwards, line 31: @rjrjr  @RBusarow  - I added a description for `recursionGraph` here and while I was added I described each of the other config levers.

rjrjr:
> All of them are 'rendered' from the POV of Workflow runtime, but only 1 path of the tree (the (n-1)th child of each depth) actually returns the rendering from `render()` and so only one path passes a rendering to the UI layer. I will make that clearer in the comment I add.

Do the other ones render null? Is that mimicking a pattern in our prod code?

steve-the-edwards:
> > All of them are 'rendered' from the POV of Workflow runtime, but only 1 path of the tree (the (n-1)th child of each depth) actually returns the rendering from `render()` and so only one path passes a rendering to the UI layer. I will make that clearer in the comment I add.
> 
> Do the other ones render null? Is that mimicking a pattern in our prod code?

They don't render null, we just don't do anything with their renderings.

rjrjr (approved):

PR #904, opened 2022-12-12
Cache identifier

steve-the-edwards:
Avoid the class name lookup which happens on the `identifier` extension property getter each time currently.

Instead, when a Workflow is `IdCacheable` (as `StatelessWorkflow` and `StatefulWorkflow` are) we cache the `identifier` as `cachedIdentifier` and inline fetch this in the getter instead of computing it each time. We do this in `StatefulWorkflow` and `StatelessWorkflow` abstract classes using a lazy delegate that calls `computeIdentifier` (same impl as before). It's lazy in case we are an `ImpostorWorkflow` and we need our `realIdentifier` initialized before `computeIdentifier` is called.

### DIFF
--- a/workflow-core/src/commonMain/kotlin/com/squareup/workflow1/StatefulWorkflow.kt
+++ b/workflow-core/src/commonMain/kotlin/com/squareup/workflow1/StatefulWorkflow.kt
@@ -6,6 +6,7 @@ package com.squareup.workflow1
 
 import com.squareup.workflow1.StatefulWorkflow.RenderContext
 import com.squareup.workflow1.WorkflowAction.Companion.toString
+import kotlin.LazyThreadSafetyMode.NONE
 import kotlin.jvm.JvmMultifileClass
 import kotlin.jvm.JvmName
 
@@ -69,7 +70,7 @@ public abstract class StatefulWorkflow<
   StateT,
   out OutputT,
   out RenderingT
-  > : Workflow<PropsT, OutputT, RenderingT> {
+  > : Workflow<PropsT, OutputT, RenderingT>, IdCacheable {
 
   public inner class RenderContext internal constructor(
     baseContext: BaseRenderContext<PropsT, StateT, OutputT>
@@ -128,6 +129,18 @@ public abstract class StatefulWorkflow<
     renderState: StateT,
     context: RenderContext
   ): RenderingT
+
+  /**
+   * Use a lazy delegate so that any [ImpostorWorkflow.realIdentifier] will have been computed
+   * before this is initialized and cached.
+   *
+   * We use [LazyThreadSafetyMode.NONE] because access to these identifiers is thread-confined.
+   */
+  override val cachedIdentifier: WorkflowIdentifier by lazy(
+    mode = NONE
+  ) {
+    computeIdentifier()
+  }
 
   /**
    * Called whenever the state changes to generate a new [Snapshot] of the state.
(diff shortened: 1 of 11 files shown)


workflow-core/src/commonMain/kotlin/com/squareup/workflow1/WorkflowIdentifier.kt
```diff
@@ -146,15 +146,14 @@ public class WorkflowIdentifier internal constructor(
...
-      proxiedIdentifier = maybeImpostor?.realIdentifier,
-      description = maybeImpostor?.let { it::describeRealIdentifier }
-    )
-  }
+public fun Workflow<*, *, *>.computeIdentifier(): WorkflowIdentifier {
+  val maybeImpostor = this as? ImpostorWorkflow
+  return WorkflowIdentifier(
+    type = Snapshottable(this::class),
+    proxiedIdentifier = maybeImpostor?.realIdentifier,
+    description = maybeImpostor?.let { it::describeRealIdentifier }
```
steve-the-edwards, line 154: The reason we use a lazy delegate to initialize `identifier` in `StatelessWorkflow` and `StatefulWorkflow` abstract classes is because `computeIdentifier` does reference the `realIdentifier` of the impostor, which won't be initialized yet at the time that the abstract class initializer happens.

I guess I should probably document that..

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/Workflow.kt
```diff
@@ -104,6 +104,8 @@ public interface Workflow<in PropsT, out OutputT, out RenderingT> {
    */
   public fun asStatefulWorkflow(): StatefulWorkflow<PropsT, *, OutputT, RenderingT>
 
+  public val identifier: WorkflowIdentifier
```
rjrjr, line 107: kdoc

rjrjr:
If this pans out it will be a breaking change in a non-experimental interface, right? Wondering about proper SemVer treatment, wondering if we should provide a default implementation on the Workflow interface with a warning that it will be removed…

All of which is well out of scope for the PR, no reason to overthink things before we know it's a win.

rjrjr (approved):

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/StatefulWorkflow.kt
```diff
@@ -129,6 +129,10 @@ public abstract class StatefulWorkflow<
     context: RenderContext
   ): RenderingT
 
+  override val identifier: WorkflowIdentifier by lazy {
```
RBusarow, line 132: Since we're thread-confined while looking these up, would it make sense to use `LazyThreadSafetyMode.NONE` to avoid the locking behavior?

steve-the-edwards:
> If this pans out it will be a breaking change in a non-experimental interface, right? Wondering about proper SemVer treatment, wondering if we should provide a default implementation on the Workflow interface with a warning that it will be removed…
> 
> All of which is well out of scope for the PR, no reason to overthink things before we know it's a win.

We could get around this with a bit more syntactic sugar. We could make the caching property `internal` and keep the existing `.identifier` extension property as something which just pulled out the value of that backing property (rather than computed each time).

rjrjr (approved):

RBusarow (approved):

steve-the-edwards:
DNM: Note that we do not want to merge this until we determine its performance impact.

steve-the-edwards:
> > If this pans out it will be a breaking change in a non-experimental interface, right? Wondering about proper SemVer treatment, wondering if we should provide a default implementation on the Workflow interface with a warning that it will be removed…
> > All of which is well out of scope for the PR, no reason to overthink things before we know it's a win.
> 
> We could get around this with a bit more syntactic sugar. We could make the caching property `internal` and keep the existing `.identifier` extension property as something which just pulled out the value of that backing property (rather than computed each time).

Nvm this. We do it with an extension interface `IdCacheable` so that we don't need to modify the existing public interfaces.

steve-the-edwards:
Note that in our testing we did see a 28% improvement for the 'worst case' scenarios. No change to the median, but still a net positive.

steve-the-edwards:
Abandoned in favour of https://github.com/square/workflow-kotlin/pull/911

PR #911, opened 2023-01-12
Cache WorkflowIdentifier

steve-the-edwards:
Looking up the type can cost significantly when repeated. Instead of computing the identifier each time, we cache it.

We do it in a slightly funny way because we want to add this functionality on via a new interface - `IdCacheable` rather than modifying the core `Workflow` interface.

The logic that cares about this caching is all in the `identifier` extension.

### DIFF
--- a/.github/workflows/kotlin.yml
+++ b/.github/workflows/kotlin.yml
@@ -149,14 +149,6 @@ jobs :
             iosX64Test --stacktrace
           cache-read-only: false
 
-      ## iOS Specific Tests w/ strict memory model (for KMP ios actuals in core and runtime).
-      - uses: gradle/gradle-build-action@v2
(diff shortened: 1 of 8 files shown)


steve-the-edwards:
iOS Tests are failing due to old memory manager for Kotlin/Native not liking the modification once the class is capture in the closure related to the changes for caching `identifier`. We should update to Kotlin 1.7.20+ to use the new memory model for Kotlin/Native and see if this is still an issue.

steve-the-edwards:
- [ ] Decide whether removing the strict kotlin.native.memoryModel mode from a test run (as I have done here) is valid after Kotlin 1.7.20.

rjrjr (approved):

RBusarow (approved):

PR #912, opened 2023-01-26
update AGP to 7.4.1

RBusarow:
https://developer.android.com/studio/releases/gradle-plugin#7-4-0


steve-the-edwards (approved):
Thanks! Let's go!

PR #918, opened 2023-02-06
910: Integrate WorkerTester with TestScope

### LINKED ISSUE #910: Extend `WorkerTester` API to allow passing in a `TestDispatcher` or `TestScheduler`
This is to enable skipping of delays and controlling of virtual time.

Such as for:

```
worker.test {
  this.advanceTimeBy(..) <-----------
  assertThat(nextOutput()).isEqualTo(..)
}
```

steve-the-edwards:
Closes #910. We do not allow a `CoroutineContext`/`CoroutineDispatcher` to be passed in, instead favouring keeping the API for `WorkerTester.test` self contained. Within the test though you can now access the `TestCoroutineScheduler` to control virtual time.

We drive the flow backing the Worker with a Turbine rather than our own handle around a channel.


workflow-testing/src/main/java/com/squareup/workflow1/testing/WorkerTester.kt
```diff
@@ -1,25 +1,33 @@
...
 public interface WorkerTester<T> {
 
+  /**
+   * Access the [TestCoroutineScheduler] of the [kotlinx.coroutines.test.TestScope] running
+   * the [Worker]'s [test].
+   *
+   * This can be used to advance virtual time for the [CoroutineDispatcher] that the the Worker's
+   * flow is flowing on.
+   */
+  public val testCoroutineScheduler: TestCoroutineScheduler
```
steve-the-edwards, line 25: This is added to control virtual time on the scope in which the Turbine is launched and thus where the Worker's upstream flow is producing values.

workflow-testing/src/main/java/com/squareup/workflow1/testing/WorkerTester.kt
```diff
@@ -31,7 +39,7 @@ public interface WorkerTester<T> {
   /**
    * Suspends until the worker emits an output or finishes.
    *
-   * Throws an [AssertionError] if an output was emitted.
+   * Throws an [AssertionError] if an output was emitted or the Worker has an error.
```
steve-the-edwards, line 42: Prior to this change, we would not throw if there was an exception (we counted that as 'finished'). I changed that, in  favour of explicitly requiring the Worker to complete unexceptionally as a Turbine would.

This seemed the best choice to me. WDYT @RBusarow @rjrjr ?

workflow-testing/src/main/java/com/squareup/workflow1/testing/WorkerTester.kt
```diff
@@ -54,67 +62,78 @@ public interface WorkerTester<T> {
...
-          // isEmpty returns false if the channel is closed.
-          if (!channel.isEmpty && !channel.isClosedForReceive) {
+          try {
+            expectNoEvents()
+          } catch (e: AssertionError) {
             throw AssertionError("Expected no output to have been emitted.")
           }
         }
 
         override suspend fun assertFinished() {
```
steve-the-edwards, line 88: This mapping is the most complicated in order to get the same error messages in the AssertionError as before.

rjrjr: Seems worth saying so in a comment, e.g. "Complicated because these messages predate our Turbine integration and we wanted to keep them stable."

workflow-testing/src/test/java/com/squareup/workflow1/WorkerTest.kt
```diff
@@ -125,22 +124,19 @@ internal class WorkerTest {
...
-      // Run the timer on the test dispatcher so we can control time.
-      .transform { it.flowOn(testDispatcher) }
 
     worker.test {
       assertNoOutput()
       assertNotFinished()
 
-      testDispatcher.scheduler.advanceTimeBy(999)
-      testDispatcher.scheduler.runCurrent()
+      testCoroutineScheduler.advanceTimeBy(999)
```
rjrjr, line 133: Nice. Great argument in favor of the API change.

workflow-testing/src/main/java/com/squareup/workflow1/testing/WorkerTester.kt
RBusarow: Yep, I agree.

PR #919, opened 2023-02-08
`Container` and `Wrapper` interfaces, improved `forWrapper()`

### LINKED ISSUE #916: Yet another stab at a general wrapper / container interface
```kotlin
import com.squareup.workflow1.ui.Screen

/**
 * A model type comprised of a set of other models of the same base type.
 *
 * @param C the invariant base type of the contents of such a container,
 * usually [Screen] or [Overlay]. It is common for the [Container] itself to
 * implement [C], but that is not a requirement. E.g., an [Overlay] might be 
 * container of [Screen].
 * @param T the specific subtype of [C] collected by this [Container].
 */
interface Container<C, T : C> {
  fun asSequence(): Sequence<T>

  /**
   * Returns a [Container] with the [transform]ed contents of the receiver.
   * It is expected that an implementation will take advantage of covariance
   * to declare its own type as the return type, rather than plain old [Container].
   * This requirement is not enforced because recursive generics are a fussy nuisance.
   */
  fun <U : C> mapContent(transform: (T) -> U): Container<C, U>
}

/**
 * A singleton [Container].
 */
interface Wrapper<C, T : C> : Container<C, T> {
  val content: T

  override fun asSequence(): Sequence<T> = sequenceOf(content)

  override fun <U : C> mapContent(transform: (T) -> U): Wrapper<C, U>
}

class WrappyScreen<S : Screen>(
  override val content: S
) : Screen, Wrapper<Screen, S> {
  override fun <U : Screen> mapContent(transform: (S) -> U): WrappyScreen<U> {
    return WrappyScreen(transform(content))
  }
}

class BStack<T : Screen>(
  val top: T,
  val backstack: List<T> = emptyList()
) : Container<Screen, T> {
  override fun asSequence(): Sequence<T> {
    return backstack.asSequence() + top
  }

  override fun <U : Screen> mapContent(transform: (T) -> U): BStack<U> {
    return BStack(transform(top), backstack.map(transform))
  }
}
```

rjrjr:
Introduces the `Container` and `Wrapper` types, giving Workflow UI its first general notion of structure. Their integration with `Compatibile` reduces the likelihood of making the most common mistake with wrapper types (namely, forgetting to do that). And they standardize the `map()` function that gets implemented by wrappers more often than not.

Also updates `forWrapper` and `toUnwrappingViewFactory` to be defined in terms of `Wrapper`, allowing us to simplify their APIs by relying on `Wrapper.content` in default lambda implementations.

And while we're in the neighborhood, adds long needed `prepEnvironment` and `prepContext` function arguments that simplify transforming a `ScreenViewFactory` to pre-process its `ViewEnvironment` and `Context`. We use this new capability to simplify the default `ScreenViewFactory` implementation for `EnvironmentScreen`.

Closes #916


rjrjr:
Oops, shouldn't have PR'd this yet. Need to think harder about `prepEnvironment`.

PR #920, opened 2023-02-08
`Container` and `Wrapper` interfaces, improved `forWrapper()`

### LINKED ISSUE #916: Yet another stab at a general wrapper / container interface

rjrjr:
Introduces the `Container` and `Wrapper` types, giving Workflow UI its first general notion of structure. Their integration with `Compatibile` reduces the likelihood of making the most common mistake with wrapper types (namely, forgetting to do that). And they standardize the `map()` function that gets implemented by wrappers more often than not.

Also updates `forWrapper` and `toUnwrappingViewFactory` to be defined in terms of `Wrapper`, allowing us to simplify their APIs by relying on `Wrapper.content` in default lambda implementations.

And while we're in the neighborhood, adds long needed `prepEnvironment` and `prepContext` function arguments that simplify transforming a `ScreenViewFactory` to pre-process its `ViewEnvironment` and `Context`. We use this new capability to simplify the default `ScreenViewFactory` implementation for `EnvironmentScreen`.

Closes #916


workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/AsScreen.kt
```diff
@@ -7,16 +7,19 @@ package com.squareup.workflow1.ui
...
-      "AsScreen is for converting non-Screen renderings, it should not wrap Screen $rendering."
+    check(content !is Screen) {
+      "AsScreen is for converting non-Screen renderings, it should not wrap Screen $content."
     }
   }
 
-  override val compatibilityKey: String
-    get() = Compatible.keyFor(rendering, "AsScreen")
+  @Deprecated("Use wrapped", ReplaceWith("wrapped"))
+  public val rendering: W = content
```
rjrjr, line 19: - [ ] Either this is silly, in which case delete it; or useful, in which case do the same to `NamedScreen`, `EnvironmentScreen`, `BackButtonScreen`.

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/Container.kt
```diff
@@ -0,0 +1,46 @@
+package com.squareup.workflow1.ui
+
+import com.squareup.workflow1.ui.Compatible.Companion.keyFor
+
+/**
+ * A model type comprised of a set of other models.
```
steve-the-edwards, line 6: I sometimes get thrown off when you call these `model`s. Why model when we use `Rendering` elsewhere for the general type and `Screen` for the marker interface?

I get that this concept (and its generics) are generalized enough to be used for anything but this is opinionated and designed for use with Workflow UI Renderings, no?

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/Container.kt
```diff
@@ -0,0 +1,46 @@
...
+public interface Container<BaseT, out ContentT : BaseT> {
+  public fun asSequence(): Sequence<ContentT>
+
+  /**
+   * Returns a [Container] with the [transform]ed contents of the receiver.
+   * It is expected that an implementation will take advantage of covariance
+   * to declare its own type as the return type, rather than plain old [Container].
+   * This requirement is not enforced because recursive generics are a fussy nuisance.
+   */
+  public fun <U : BaseT> map(transform: (ContentT) -> U): Container<BaseT, U>
```
steve-the-edwards, line 30: I _think_ I'm _beginning_ to understand what you are trying to abstract with this `Container` and `map`. It could really use some examples in this kdoc though as its hard to picture.

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/Container.kt
```diff
@@ -0,0 +1,46 @@
...
+}
+
+/**
+ * A singleton [Container].
```
steve-the-edwards, line 34:
```suggestion
 * A singleton [Container] that wraps a rendering of type [T] as a rendering of type [U] and reports that it is
 [Compatible] with its [content].
```

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/AsScreen.kt
steve-the-edwards: Don't you mean use `content`?

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewFactory.kt
```diff
@@ -353,44 +327,32 @@ public fun interface ViewStarter {
...
     wrapperScreen: WrapperT,
     environment: ViewEnvironment,
     showUnwrappedScreen: (WrappedT, ViewEnvironment) -> Unit
-  ) -> Unit
-): ScreenViewFactory<WrapperT> {
+  ) -> Unit = { _, wrapperScreen, environment, showUnwrappedScreen ->
+    showUnwrappedScreen(wrapperScreen.content, environment)
+  }
+): ScreenViewFactory<WrapperT>
+  where WrapperT : Screen, WrapperT : Wrapper<Screen, WrappedT>, WrappedT : Screen {
```
steve-the-edwards, line 346: Whoa. This is cool. Has this always been valid syntax?

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewFactory.kt
```diff
@@ -139,107 +139,81 @@ public interface ScreenViewFactory<in ScreenT : Screen> : ViewRegistry.Entry<Scr
...
-    /**
-     * Creates a [ScreenViewFactory] for [WrapperT] that finds and delegates to the one for
-     * [WrappedT]. Allows [WrapperT] to wrap instances of [WrappedT] to add information or behavior,
-     * without requiring wasteful wrapping in the view system.
+     *    class WithTutorialTips<W : Screen>(
+     *      override val content: W
+     *    ) : AndroidScreen<WithTutorialTips<W>>, Wrapper<Screen, W> {
+     *      override val viewFactory = forWrapper<WithTutorialTips<W>, W>(
+     *        beforeShowing = { TutorialTipRunner.initialize(it.view) }
+     *      )
```
steve-the-edwards, line 166: 🔥

steve-the-edwards (approved):
😵 . I hope I managed to understand things.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewFactory.kt
rjrjr: Yup, ancient.

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/AsScreen.kt
rjrjr: Yup, thanks.

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/Container.kt
rjrjr: It's my recent forays into Compose leaking in. What we call "renderings" they generally refer to as "models." I like it better as a general term, but you're right: this is Workflow UI land, and "rendering" is entrenched.

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/Container.kt
rjrjr:
```kotlin
   *
   * For example, suppose we want to create `LoggingScreen`, one that wraps any
   * other screen to add some logging calls. Its implementation of this method
   * would be expected to have a return type of `LoggingScreen` rather than `Container`:
   *
   *    override fun <D : Screen> map(transform: (C) -> D): LoggingScreen<D> =
   *      LoggingScreen(transform(content))
```

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/Container.kt
rjrjr:
```
/**
 * A singleton [Container] -- that is, a rendering which wraps a single rendering of type
 * [ContentT].
 *
 * Usually a [Wrapper] is [Compatible] only with another of the same type, and with a
 * [Compatible] [content] instance. In aid of that, this interface extends [Compatible] and
 * provides a convenient default implementation of [compatibilityKey].
 */
```

rjrjr:
Update:
* fixes stale name in `AsScreen` deprecation
* adds missing deprecation conveniences to `NamedScreen`, `EnvironmentScreen`, `BackButtonScreen`
* Fixes `PanelOvelay` sample (it was missing its `map` implementation).
* Improves kdoc in `Container.kt`

rjrjr:
It occurs to me that a base `ScreenWrapper` interface would put `BaseT` a bit less in people's faces, going to try adding that.

rjrjr:
Update:
- Rebases to `main` again
- Fixes Tic Tac Toe sample
- Fixes `NamedScreen`

God bless CI and unit tests.

samples/containers/common/src/main/java/com/squareup/sample/container/panel/PanelOverlay.kt
```diff
@@ -6,6 +6,9 @@ import com.squareup.workflow1.ui.container.ModalOverlay
...
 
 @OptIn(WorkflowUiExperimentalApi::class)
-class PanelOverlay<T : Screen>(
-  override val content: T
-) : ScreenOverlay<T>, ModalOverlay
+class PanelOverlay<C : Screen>(
+  override val content: C
+) : ScreenOverlay<C>, ModalOverlay {
+  override fun <D : Screen> map(transform: (C) -> D): ScreenOverlay<D> =
+    PanelOverlay(transform(content))
```
steve-the-edwards, line 13:
```suggestion
  override fun <D : Screen> map(transform: (C) -> D): PanelOverlay<D> =
    PanelOverlay(transform(content))
```
?

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/Container.kt
```diff
@@ -0,0 +1,72 @@
...
+   * contents can be repackaged in interesting ways, e.g.:
+   *
+   *    val childBackStackScreen = renderChild(childWorkflow) { ... }
+   *    val loggingBackStackScreen = childBackStackScreen.map { LoggingScreen(it) }
+   */
+  public fun <ContentU : BaseT> map(transform: (ContentT) -> ContentU): Container<BaseT, ContentU>
+}
+
+/**
+ * A singleton [Container] -- that is, a rendering which wraps a single rendering of type
```
steve-the-edwards, line 49: nit: "singleton" Container still throws me off as I think, oh - it makes it so that this Rendering is a Singleton; which, i get, is absurd but I need the explanation sentence anyway so why not change what we are summarizing this as?

"A frame [Container]"?

steve-the-edwards (approved):

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/Container.kt
rjrjr:
```
 * A [Container] rendering that wraps exactly one other rendering, its [content]. These are
 * typically used to "add value" to the [content], e.g. an
 * [EnvironmentScreen][com.squareup.workflow1.ui.container.EnvironmentScreen] that allows
 * changes to be made to the the [ViewEnvironment].
```

rjrjr:
Addressed the last couple of issues that @steve-the-edwards raised, thanks Steve. Holding off on merging this until #944 has been vetted, maybe until it's published.

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/Container.kt
rjrjr: "the the" :/

rjrjr:
I've merged #967 down into this branch, merged master into it, and have set this for auto-merge.

PR #941, opened 2023-02-21
Close dialogs even if the managing view is never attached.

rjrjr:


workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/WorkflowAndroidXSupport.kt
```diff
@@ -15,6 +15,17 @@ import kotlin.reflect.cast
...
  */
 public object WorkflowAndroidXSupport {
+  /**
+   * Returns the [LifecycleOwner] managing [context].
+   *
+   * @throws IllegalArgumentException if [context] is unmanaged
+   */
+  @WorkflowUiExperimentalApi
+  public fun lifecycleOwnerFromContext(context: Context): LifecycleOwner =
+    requireNotNull(context.ownerOrNull(LifecycleOwner::class)) {
```
pyricau, line 25: I'm guessing this isn't your API but funny that we have "ownerOrNull"  instead of "requireOwner"

pyricau: Actually, this makes me think, should this turn into `Context.requireLifecycleOwner` ?

rjrjr: Very consistent convention we switched to ages ago. `orNull` is the weird case. If you don't specify it, non-null is implied.

rjrjr: `require*` is so Java.

rjrjr: `should this turn into Context.requireLifecycleOwner`

We've left these methods a bit inconvenient on purpose. Feature devs shouldn't need them, having them autocompolete off of `Context` would be more advertising than we really want.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/LayeredDialogSessions.kt
```diff
@@ -298,16 +301,36 @@ public class LayeredDialogSessions private constructor(
           "Expected a ViewTreeLifecycleOwner on $view"
         }
       }.also { dialogs ->
+        fun closeAll() {
+          dialogs.update(emptyList(), ViewEnvironment.EMPTY) {}
```
pyricau, line 305: an API that requires a lambda and yes we don't want to do anything in that lambda?

maybe that API needs tweaking? I have no idea what this is though.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/LayeredDialogSessions.kt
```diff
@@ -298,16 +301,36 @@ public class LayeredDialogSessions private constructor(
...
+        // But the Activity might finish before the hosting view is ever attached. And we have
+        // lots of time to show Dialogs before then. They will leak.
+        //
+        // To guard against that we hang a default observer directly off of the Activity that
+        // will close all Dialogs when it is destroyed; and we remove it as soon as the hosting
+        // view is attached for the first time.
+        val failsafe = object : DefaultLifecycleObserver {
+          override fun onDestroy(owner: LifecycleOwner) = closeAll()
+        }
+        lifecycleOwnerFromContext(view.context).lifecycle.addObserver(failsafe)
```
pyricau, line 319: Almost want to make this `observeUntilViewAttached` to show how the add and remove go together

pyricau (approved):
NICE FIX!

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/LayeredDialogSessions.kt
rjrjr: Non-normal call. I'm telling the `DialogManager` to show nothing, meaning close all Dialogs. The empty lambda is normally the one that updates the view underneat the dialogs. Nothing to update, we're closing up shop.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/LayeredDialogSessions.kt
rjrjr: I don't like to wrap up a few a one-off lines like this, even in the same file. I like to be able to see exactly what's happening.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/WorkflowAndroidXSupport.kt
steve-the-edwards: `<x>OrNull` is the Kotlin style - see e.g., Result - https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-result/get-or-null.html

steve-the-edwards (approved):
Nice!

rjrjr:
Expecting to publish this via branch `v1.9.0-beta02-release`. Running internal test suite against it first.

PR #944, opened 2023-02-22
compileSdkVersion 33; androidx.activity/appcompat upgrades; transitives

steve-the-edwards:
With upgrading of Compose, we revealed a tricky bug in `renderAsState` caught by the `SampleLauncherTest` while scrolling a `LazyColumn` to bring a `Composable` into view whose state was driven by `renderAsState`. The fix was to launch a coroutine to transfer the state from Workflow into Compose state via an `UNDISPATCHED` launched coroutine rather than an `onEach` on the flow.


samples/compose-samples/src/main/AndroidManifest.xml
```diff
@@ -2,14 +2,16 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools">
 
+  <uses-permission android:name="android.permission.POST_NOTIFICATIONS"/>
```
rjrjr, line 5: How come?

workflow-ui/compose/README.md
```diff
@@ -78,7 +78,7 @@ Note the difference in when the restoration happens relative to the lifecycle st
...
-| All views participate in this mechanism, and can’t opt-out.                                                                                                       | Views must opt-in by getting access to a `SavedStateRegistry` either via a `ViewTreeSavedStateRegistry` or some other way.                                                                                                                          |
+| All views participate in this mechanism, and can’t opt-out.                                                                                                       | Views must opt-in by getting access to a `SavedStateRegistry` either via a `findViewTreeSavedStateRegistryOwner` or some other way.                                                                                                                 |
```
rjrjr, line 81: Should drop "a".

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/ViewStateFrame.kt
```diff
@@ -29,7 +31,15 @@ internal data class ViewStateFrame(
   companion object CREATOR : Creator<ViewStateFrame> {
     override fun createFromParcel(parcel: Parcel): ViewStateFrame {
       val key = parcel.readString()!!
-      val viewState = parcel.readSparseArray<Parcelable>(ViewStateFrame::class.java.classLoader)!!
+      val viewState = if (VERSION.SDK_INT >= VERSION_CODES.TIRAMISU) {
```
rjrjr, line 34: I'm kind of amazed they didn't give us compat wrappers to do all this version checking

rjrjr: Should we make our own? We already have a module of Androidx support methods.

rjrjr (approved):

samples/compose-samples/src/main/AndroidManifest.xml
steve-the-edwards: New requirement to request this permission if your app posts any notifications. Good point that I didn't really look into why this was being linted and if there was a way of specifying that we _don't_ post notifications. I kind of thought that, hey, might as well just include it. Too messy?

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/ViewStateFrame.kt
steve-the-edwards: Could consider it for sure.

samples/compose-samples/src/main/AndroidManifest.xml
steve-the-edwards: So this is a branch new permission that only applies to API 33+ installs. On those you need to request the permission at runtime. There will be follow-on work for this. Alternative would be to leave it out and disable the Android Lint setting for this? I think this is likely the least intrusive way forward?

What will catch us is when we start running on API 33 devices. I will create an issue to track that.

gradle/libs.versions.toml
```diff
@@ -2,31 +2,35 @@
...
+androidx-activity = "1.6.1"
+androidx-appcompat = "1.6.1"
 androidx-benchmark = "1.1.1"
 androidx-cardview = "1.0.0"
-androidx-compose = "1.1.0-rc01"
 androidx-compose-compiler = "1.3.2"
+androidx-compose-foundation = "1.3.1"
+androidx-compose-material = "1.3.1"
+androidx-compose-runtime = "1.3.3"
+androidx-compose-ui = "1.3.3"
```
RBusarow, line 17: holy fragmentation

RBusarow (approved):

gradle/libs.versions.toml
vgonda: https://developer.android.com/jetpack/compose/bom/bom

Needs BOM

samples/compose-samples/build.gradle.kts
```diff
@@ -20,7 +20,10 @@ android {
 }
 
 dependencies {
+  val composeBom = platform(libs.androidx.compose.bom)
```
RBusarow, line 23: You should be able to lose this and just have `implementation(platform(libs.androidx.compose.bom))`.  The `androidTestImplementation` configuration extends `debugImplementation`, which in turn extends `implementation`.

gradle/libs.versions.toml
steve-the-edwards: Done.

samples/compose-samples/build.gradle.kts
steve-the-edwards: oh weird, the Android docs specified to do it this way :S.

steve-the-edwards: oh weird, the Android docs specified to do it this way :S.

samples/compose-samples/src/main/AndroidManifest.xml
steve-the-edwards: It looks for code that uses Notifications, but this can be in classes from dependent libraries - classes which aren't necessarily used. This is only a warning, but since we have warning=error it fails the build.

I'm going to take these out and take a lint warnings baseline, then use that going forward.

buildSrc/src/main/java/android-sample-app.gradle.kts
```diff
@@ -11,6 +12,12 @@ configure<TestedExtension> {
   buildFeatures.viewBinding = true
 }
 
+configure<BaseAppModuleExtension> {
+  lint {
+    baseline = file("lint-baseline.xml")
+  }
+}
```
steve-the-edwards, line 19: Added this to baseline android lint warnings at this snapshot.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/RenderAsState.kt
```diff
@@ -206,18 +206,23 @@ private class WorkflowRuntimeState<PropsT, OutputT : Any, RenderingT>(
...
 
-    renderings
-      .onEach {
-        renderingState.value = it.rendering
-        snapshotState.value = it.snapshot
-      }
+    workflowScope.launch(
+      start = UNDISPATCHED,
+      context = Dispatchers.Unconfined
+    ) {
```
steve-the-edwards, line 212: moved from shuttling these renderings and snapshot into compose state in an `onEach` started by a `launchIn` on `Dispatchers.Unconfined` to launching for the collection with an `CoroutineStart.UNDISPATCHED` coroutine and this fixes the timing issue we saw come up in the Compose sample test while scrolling the LazyColumn.

cc @rjrjr and @RBusarow .

I discussed this with @zach-klippenstein . He thinks there may still be some Compose bug here but this is making the behavior more reliable for us for now and while it makes this PR messy I'm inclined to merge it.

samples/compose-samples/src/main/AndroidManifest.xml
steve-the-edwards: Turns out I don't think we need `POST_NOTIFICATIONS` once i finished updating some transitive dependency libraries (this isn't in the lint baseline). Obviously one of the libraries had to update for Android 13 as well.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/ViewStateFrame.kt
steve-the-edwards: Did not do this as I think this type of branching is pretty standard practice and I'm ok with it as it makes it clear what can happen down the road rather than hiding that behind some syntactic sugar.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/RenderAsState.kt
RBusarow: It seems fine to me 👍

rjrjr: This could be a big win. The problem your describing sounds very similar to one that has been plaguing a lot of our Compose tests.

rjrjr: Oh, but we don't use `renderAsState` :(

PR #947, opened 2023-02-23
add Renovate groups for Kotlin and androidx-lifecycle updates

RBusarow:
With [package grouping](https://docs.renovatebot.com/noise-reduction/#package-grouping), Renovate will combine multiple update PRs into a single PR.  This is really nice for dependencies which must be updated together, like Kotlin and KSP.

If only some packages in the group have updates, then Renovate will make a PR for those packages by themselves.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -113,6 +113,8 @@ androidx-appcompat = { module = "androidx.appcompat:appcompat", version.ref = "a
 
 androidx-cardview = { module = "androidx.cardview:cardview", version.ref = "androidx-cardview" }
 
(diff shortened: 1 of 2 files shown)


gradle/libs.versions.toml
```diff
@@ -113,6 +113,8 @@ androidx-appcompat = { module = "androidx.appcompat:appcompat", version.ref = "a
 
 androidx-cardview = { module = "androidx.cardview:cardview", version.ref = "androidx-cardview" }
 
+androidx-compose-compiler = { module = "androidx.compose.compiler:compiler", version.ref = "androidx-compose-compiler" }
```
RBusarow, line 116: We use the version reference in the Android convention plugin, but that isn't enough for Renovate to pick it up.  So by defining the actual maven artifact here, Renovate will see it and include it in its update checks.

PR #952, opened 2023-02-28
Add interceptor methods: onNextRendering, onSnapshotTree

pyricau:
Also removing noise from some of the toString()


workflow-tracing/src/main/java/com/squareup/workflow1/diagnostic/tracing/TimelineWorkflowInterceptor.kt
```diff
@@ -0,0 +1,308 @@
...
+    object None : RenderingState
+    data class Rendering(val root: WorkflowSession): RenderingState
+    data class Snapshotting(val root: WorkflowSession): RenderingState
+  }
+
+  private val stack = ArrayDeque<String>()
+  private val causalChain = mutableListOf<Edge>()
+  private var nextRenderingCause: String? = null
+  private var renderingState: RenderingState = None
+  private var enqueuedActions = 0
```
steve-the-edwards, line 44: When you get a chance can you document what each of these are keeping track of?

pyricau: if I don't throw the whole thing away, yes :)

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/WorkerWorkflow.kt
```diff
@@ -89,8 +89,9 @@ private class EmitWorkerOutputAction<P, S, O>(
   private val output: O,
 ) : WorkflowAction<P, S, O>() {
   override fun toString(): String =
-    WorkflowIdentifierTypeNamer.uniqueName(EmitWorkerOutputAction::class) +
-      "(worker=$worker, key=\"$renderKey\")"
+    WorkflowIdentifierTypeNamer.uniqueName(EmitWorkerOutputAction::class)
+  // TODO tmp Removed because this creates noise.
+  // + "(worker=$worker, key=\"$renderKey\")"
```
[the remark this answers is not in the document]
pyricau: output would look like this:

```
sendAndAwaitApplication(com.squareup.workflow1.EmitWorkerOutputAction(worker=TypedWorker(java.lang.String (Kotlin reflection is not available)), key="1"))
```

key was useful, `worker=TypedWorker(java.lang.String (Kotlin reflection is not available))` definitely not useful. Really wish we'd get rid of these `Kotlin reflection is not available` in technical names btw.

The real reason I removed this is because `key="1"` has `"` and I needed to escape those in the output for the graph and was lazy so I just removed them.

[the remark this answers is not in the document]
pyricau: We should probably remove the `" "` I don't think we use those anywhere else.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/Worker.kt
```diff
@@ -273,7 +262,7 @@ public interface Worker<out OutputT> {
  */
 @OptIn(ExperimentalStdlibApi::class)
 public inline fun <reified OutputT> Flow<OutputT>.asWorker(): Worker<OutputT> =
-  TypedWorker(typeOf<OutputT>(), this)
+  TypedWorker({ OutputT::class.qualifiedName!! }, this)
```
pyricau, line 265: @rjrjr I'm getting rid of those gnarly "kotlin reflection not available"! At least for this one spot I found.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/WorkerWorkflow.kt
```diff
@@ -90,7 +90,7 @@ private class EmitWorkerOutputAction<P, S, O>(
 ) : WorkflowAction<P, S, O>() {
   override fun toString(): String =
     WorkflowIdentifierTypeNamer.uniqueName(EmitWorkerOutputAction::class) +
-      "(worker=$worker, key=\"$renderKey\")"
+      "(worker=$worker, key=$renderKey)"
```
pyricau, line 93: We typically don't use `"` in logged names.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/WorkerWorkflow.kt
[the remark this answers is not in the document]
rjrjr: > Really wish we'd get rid of these Kotlin reflection is not available in technical names btw.

Yes! They're horrible, but I've lost track of where they're sneaking in. If you find the source, please let me know and I'll jump on it.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt
```diff
@@ -87,6 +87,16 @@ public interface WorkflowInterceptor {
     session: WorkflowSession
   ): S = proceed(old, new, state)
 
+  /**
+   * Intercept calls to [com.squareup.workflow1.internal.WorkflowRunner.nextRendering].
```
rjrjr, line 91:
```suggestion
   * Intercept calls to [WorkflowRunner.nextRendering][com.squareup.workflow1.internal.WorkflowRunner.nextRendering].
```

rjrjr (approved):
Could you add a commit message explaining why this is needed, what use case it unlocks?

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt
steve-the-edwards: This definitely needs kdoc here as we are intercepting an internal method in the Runner.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt
```diff
@@ -98,6 +108,14 @@ public interface WorkflowInterceptor {
     session: WorkflowSession
   ): R = proceed(renderProps, renderState, null)
 
+  /**
+   * Intercept calls to [StatefulWorkflow.snapshotState] including the children calls.
+   */
+  public fun onSnapshotTree(
```
steve-the-edwards, line 114: This is a bit confusing. Why not `onSnapshotStateWithChildren`? Also could use kdoc explaining use case.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt
pyricau: done

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt
pyricau: done

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/WorkerWorkflow.kt
```diff
@@ -35,7 +35,7 @@ internal class WorkerWorkflow<OutputT>(
   ImpostorWorkflow {
 
   override val realIdentifier: WorkflowIdentifier = unsnapshottableIdentifier(workerType)
-  override fun describeRealIdentifier(): String = "worker $workerType"
+  override fun describeRealIdentifier(): String = workerType.toString()
```
pyricau, line 38: the workers have names that make it clear they're workers, no need for this.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNodeId.kt
```diff
@@ -48,6 +57,8 @@ internal data class WorkflowNodeId(
       return WorkflowNodeId(identifier, name)
     }
   }
+
+  override fun toString() = debugString
```
pyricau, line 61: Removing `WorkflowNodeId(identifier=foo, name=bar)` to just `foo` and `foo named bar`

rjrjr:
Still would like to see a commit message explaining what use case this unlocked.

steve-the-edwards (approved):

steve-the-edwards:
@pyricau you can use `./gradlew -p . apiDump` to update the API check.

steve-the-edwards (approved):
LGTM

PR #953, opened 2023-02-28
Add Property for Target Selection

steve-the-edwards:
Gives us the ability to specify targets for gradle commands with the `workflow.targets` property as any of `kmp`, `jvm`, `ios`, or `js`

### DIFF
--- a/workflow-core/README.md
+++ b/workflow-core/README.md
@@ -9,6 +9,14 @@ are `jvm`, `ios`, and `iosSimulatorSimulatorArm64`. If you are having issues wit
 ensure you have the correct version of XCode installed and can launch a simulator as it's specified
 in the gradle build file (Currently iPhone 14).
 
+You can also choose to specify your targets for build and test with the property `workflow.targets`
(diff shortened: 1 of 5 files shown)


workflow-core/README.md
```diff
@@ -9,6 +9,14 @@ are `jvm`, `ios`, and `iosSimulatorSimulatorArm64`. If you are having issues wit
 ensure you have the correct version of XCode installed and can launch a simulator as it's specified
 in the gradle build file (Currently iPhone 14).
 
+You can also choose to specify your targets for build and test with the property `workflow.targets`
+as either `kmp`, `android`, `ios`, `js`. The default is `kmp` (all the targets). Set this in your
+gradle.properties or specify the property in your gradle command, e.g.:
```
RBusarow, line 14: > Set this in your gradle.properties

It's probably worth calling out that this would be `~/.gradle/gradle.properties`, and not the project properties.

RBusarow (approved):
This is great!  Thanks!

workflow-core/README.md
steve-the-edwards: Done.

PR #970, opened 2023-03-16
Fixes bad readSparseArray call in WorkflowLayout

rjrjr:
Crash on restore on Tiramisu / 33 or later


steve-the-edwards (approved):
Thanks!

PR #975, opened 2023-03-21
Extracts `ScreenViewFactory.map` from `ScreenViewFactory.forWrapper`

### LINKED ISSUE #973: Alias renderings are too hard to deal with
The recent introduction of the `Wrapper` interface (#920) made it easier to create things like `NamedScreen`, where one rendering is used to modify how a wrapped one is displayed; but made it harder to implement "alias" types: custom renderings that are most easily displayed by transforming them into existing types.

This is a problem because we encourgage apps to use `BodyAndOverlaysScreen` as a building block to craft their own, stricter rules. 

e.g., it should be trivial to register a `ScreenViewFactory` that handles this:

```kotlin
class MyAppRootUI(
  val body: Screen,
  val maybeExactlyOnePopupScreen: MySpecificOverlayType?
  val maybeAnAlertOnTopOfThat: AlertOverlay?
)
```

by transforming it into this:

```kotlin
BodyAndOverlaysScreen(
  body = myRoot.body,
  overlays = listOfNotNull(
    myRoot.maybeExactlyOnePopupScreen,
    myRoot.maybeAnAlertOnTopOfThat
)
```

By coupling the `ScreenViewFactory.forWrapper` to the new `Wrapper` type, we have actually made that much harder.

So, at least two outcomes needed here:

* Convenient API for this use case
* Introduce a sample demonstrating this pattern, both to document it and to make it painfully obvious when we work against it.

rjrjr:
Coupling `Wrapper` to `ScreenViewFactory.forWrapper` was a mistake, made simple transformations that weren't in the strict `Wrapper` shape more difficult.

Also some compulsive tidying of sample code.

Fixes #973


workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewHolder.kt
```diff
@@ -26,6 +26,16 @@ public interface ScreenViewHolder<in ScreenT : Screen> {
    * maintained correctly, and [showing] keeps working.
    */
   public val runner: ScreenViewRunner<ScreenT>
+
+  public companion object {
```
rjrjr, line 30: Call sites aren't changed by this, it still looks like a constructor invocation of the `ScreenViewHolder` interface. Just seems more discoverable to me.

steve-the-edwards (approved):
LGTM. I think I'm beginning to understand this a bit more. Certainly the terminology is getting easier for me to grok

rjrjr:
> Certainly the terminology is getting easier for me to grok

You and me both.

PR #986, opened 2023-04-08
Readability tweaks to runtime

rjrjr:
1. Rename tick to onNextAction
2. Less confusing and slightly less hacky action handling return types


RBusarow (approved):

steve-the-edwards (approved):
Thanks for doing this! Every time I look at this code I think we need to do this but then I don't.

Issue #985, opened 2023-04-07
Offer config option to skip rendering when new state == old


rjrjr:
We have been afraid to simply make this change because it will inevitably break so many existing workflows. But everyone writes code assuming this is how things work, and that causes our worst performance problems. Let's at least offer the option to turn that behavior on, and consider gradually moving to make that the default.

Also not that if an output is emitted but no state is changed, there is still no reason to re-render. It seems to me we can consider posting output and changing state as orthogonal things.

steve-the-edwards:
So, with #992 we will add exactly this. See that PR for explanation and pertinent points. Tests are passing and results looking good re: the # of render passes.

A further step beyond this could actually be to record a path from the the root of the Workflow tree (R) to the node where the state changed started from the action (lets call this N).

I believe we only need to re-render the path of nodes from R-N as they are the only ones that can be affected by this state change, *within the next render pass*. That state change of course could have knock on effects for any and all of the other nodes but those would need to be applied in separate actions.

The path R-N could start as naive (render all nodes in the path and all the children of all those nodes), then we could optimize it even further where in the path R-N we only render a node A's child if that child is on the path R-N or if the state change to A has modify the props passed to the child.

I'm not sure how we would manage knitting together the cached renderings w/ new renderings in order to build the full rendering.

steve-the-edwards:
Fixed by #992

PR #992, opened 2023-04-20
WIP: 985: Add ability to propagate state comparison up tree

steve-the-edwards:
aiming to satisfy #985 . 

First commit propagates up the tree whether or not the state was changed. Deciding whether or not to change runtime behaviour with that information will be down to a runtime config.

For now, see the tests pass.


gradle.properties
```diff
@@ -8,7 +8,7 @@ android.useAndroidX=true
 systemProp.org.gradle.internal.publish.checksums.insecure=true
 
 GROUP=com.squareup.workflow1
-VERSION_NAME=1.10.0-beta06-SNAPSHOT
+VERSION_NAME=1.10.0-beta06-state3-SNAPSHOT
```
rjrjr, line 11: This looks like an accident.

samples/dungeon/timemachine/src/test/java/com/squareup/sample/timemachine/TimeMachineWorkflowTest.kt
```diff
@@ -25,7 +25,12 @@ class TimeMachineWorkflowTest {
     val delegateWorkflow = Workflow.stateful<String, Nothing, DelegateRendering>(
       initialState = "initial",
       render = { renderState ->
-        DelegateRendering(renderState, setState = eventHandler { s -> state = s })
+        DelegateRendering(
```
rjrjr, line 28: Please to revert accidental whitespace change.

workflow-config/config-android/src/main/java/com/squareup/workflow1/config/AndroidRuntimeConfigTools.kt
```diff
@@ -26,6 +29,7 @@ public class AndroidRuntimeConfigTools {
     @WorkflowExperimentalRuntime
     public fun getAppWorkflowRuntimeConfig(): RuntimeConfig {
       return when (BuildConfig.WORKFLOW_RUNTIME) {
+        "stateChange" -> RenderOnStateChangeOnly
         "conflate" -> ConflateStaleRenderings
```
rjrjr, line 33: So we can't turn on both `RenderOnStateChangeOnly` and `ConflateStaleRenderings`? Isn't that going to bite us pretty quickly?

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/WorkflowAction.kt
```diff
@@ -22,15 +22,25 @@ public abstract class WorkflowAction<in PropsT, StateT, out OutputT> {
...
-      this.maybeOutput = WorkflowOutput(output)
+      this.outputOrNull = WorkflowOutput(output)
+    }
+
+    /**
+     * The action will cause the Workflow to re-render regardless of whether or not state has
+     * changed, for all [RuntimeConfig]s.
+     */
+    public fun forceRerender() {
+      forcedRerender = true
```
rjrjr, line 43: Surprised this exists, and surprised that it's public. Hard to imagine feature devs putting this to anything but bad uses.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/WorkflowAction.kt
```diff
@@ -133,3 +143,28 @@ public class WorkflowOutput<out OutputT>(public val value: OutputT) : ActionProc
...
   override fun hashCode(): Int = value.hashCode()
 }
+
+/**
+ * An [ActionProcessingResult] is any possible outcome after the runtime does a loop of processing.
+ *
+ * Only [ActionApplied] needs the generic OutputT so we do not include it in the root
+ * interface here.
+ */
+public sealed interface ActionProcessingResult
```
rjrjr, line 153: It's interesting that `ActionProcessingResult` was already public, never really thought about that. Wondering if it's actually necessary.

gradle.properties
steve-the-edwards: yes, just keeping track of the Snapshot version I'm using for internal canary testing.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/WorkflowAction.kt
steve-the-edwards: I originally did not have this. I've now added this as in internal testing we have a number of scenarios where we drive 'other' (non-Workflow) UI via side effects that occur in actions, but then that require a re-render() to pick up the changes in that other UI. So either we add some kind of increasing counter to the state for those cases or we add a hook to force re-render() without state change like this. I wanted to just test with this first. Sorry I'm realizing now that this PR shoudl still be in draft. I'm not sure why it wasn't. Probably I was thinking at some point it was more ready than it is.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/WorkflowAction.kt
steve-the-edwards: It's used in runtime so yes.

workflow-config/config-android/src/main/java/com/squareup/workflow1/config/AndroidRuntimeConfigTools.kt
steve-the-edwards: Combined now to make both usable at the same time (and to test that).

.github/workflows/kotlin.yml
```diff
@@ -188,6 +188,62 @@ jobs :
         with :
           report_paths : '**/build/test-results/test/TEST-*.xml'
 
+  jvm-stateChange-runtime-test :
```
rjrjr, line 191: Woot! @RBusarow should own reviewing this for correctness, though.

workflow-config/config-android/src/main/java/com/squareup/workflow1/config/AndroidRuntimeConfigTools.kt
```diff
@@ -22,12 +22,21 @@ public class AndroidRuntimeConfigTools {
...
      */
     @WorkflowExperimentalRuntime
     public fun getAppWorkflowRuntimeConfig(): RuntimeConfig {
       return when (BuildConfig.WORKFLOW_RUNTIME) {
-        "conflate" -> ConflateStaleRenderings
-        else -> RenderPerAction
+        "conflate" -> ConflateStaleRenderings(renderOnStateChangeOnly = false)
+        "conflate-stateChange" -> ConflateStaleRenderings(renderOnStateChangeOnly = true)
+        "baseline-stateChange" -> RenderPerAction(renderOnStateChangeOnly = true)
+        else -> RuntimeConfig.DEFAULT_CONFIG
```
rjrjr, line 39: Seems like we should be a bit more strict so that we catch typos for people.

```suggestion
        "conflate" -> ConflateStaleRenderings(renderOnStateChangeOnly = false)
        "conflate-stateChange" -> ConflateStaleRenderings(renderOnStateChangeOnly = true)
        "baseline-stateChange" -> RenderPerAction(renderOnStateChangeOnly = true)
        "", "baseline" -> RuntimeConfig.DEFAULT_CONFIG
        else -> throw IllegalArgumentException("Unrecognized config \"${BuildConfig.WORKFLOW_RUNTIME}\"")
```

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/WorkflowAction.kt
```diff
@@ -133,3 +134,28 @@ public class WorkflowOutput<out OutputT>(public val value: OutputT) : ActionProc
...
+
+/**
+ * Result of applying an action.
+ *
+ * @param output: the potentially null [WorkflowOutput]. If null, then no output was set by the
+ *   action. Otherwise it is a [WorkflowOutput] around the output value of type [OutputT],
+ *   which could be null.
+ * @param stateChanged: whether or not the action changed the state.
+ */
+public data class ActionApplied<out OutputT>(
```
rjrjr, line 158: Do we actually rely on this being a `data class`? I remember that being a landmine when used in public API, though I forget exactly why.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
```diff
@@ -135,39 +134,57 @@ public fun <PropsT, OutputT, RenderingT> renderWorkflowIn(
...
       }
+
       else -> {} // no -op
     }
   }
 
+  /**
+   * We only check for this when [RuntimeConfig.renderOnStateChangeOnly] is true.
+   */
+  suspend fun maybeCheckNoStateChange(actionResult: ActionProcessingResult): Boolean {
```
rjrjr, line 157: How about `shortCircuitForUnchangedState`? And have the kdoc explain what should happen when true is returned.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RuntimeConfig.kt
```diff
@@ -20,20 +20,33 @@ public annotation class WorkflowExperimentalRuntime
...
  * A specification of the Workflow Runtime.
  */
-public sealed interface RuntimeConfig {
+public sealed class RuntimeConfig {
+  /**
+   * If state has not changed from an action cascade, do not re-render. This has been mostly
+   * proven out. However, be careful if you have any non-Workflow code you integrate with that
+   * depends on the Workflow tree re-rendering to pick up changes from its equivalent 'view model.'
+   * You should change some kind of Workflow state when updating that external code if you want
+   * Workflow to pick up the change and render again.
```
rjrjr, line 29:
```suggestion
   * If state has not changed from an action cascade (as determined via `equals()`),
   * do not re-render. For example, when this is true and `noAction()` is enqueued,
   * the current `render()` pass will short circuit and no rendering will be posted
   * through the `StateFlow` returned from `renderWorkflowIn()`.
   *
   * This has been mostly proven out. However, be careful if you have any non-Workflow 
   * code you integrate with that depends on the Workflow tree re-rendering to pick up
   * changes from its equivalent 'view model.' You should change some kind of Workflow
   * state when updating that external code if you want Workflow to pick up the change
   * and render again.
```

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
```diff
@@ -226,15 +227,26 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
...
+   * not this action has changed the current node's state.
    */
   private fun applyAction(
-    action: WorkflowAction<PropsT, StateT, OutputT>
-  ): ActionProcessingResult? {
-    val (newState, outputOrNull) = action.applyTo(lastProps, state)
+    action: WorkflowAction<PropsT, StateT, OutputT>,
+    childResult: ActionApplied<*>? = null
+  ): ActionProcessingResult {
+    val (newState, actionApplied) = action.applyTo(lastProps, state)
```
rjrjr, line 238:
```suggestion
    val (newState: StateT, actionApplied: ActionApplied<StateT>) = action.applyTo(lastProps, state)
```

Not positive about the `ActionApplied` type, but you get the idea.

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
```diff
@@ -172,10 +174,34 @@ class RenderWorkflowInTest {
   }
 
   private val runtimeMatrix = arrayOf(
-    Pair(RenderPerAction, RenderPerAction),
-    Pair(RenderPerAction, ConflateStaleRenderings),
-    Pair(ConflateStaleRenderings, RenderPerAction),
-    Pair(ConflateStaleRenderings, ConflateStaleRenderings),
+    Pair(RenderPerAction(), RenderPerAction()),
```
rjrjr, line 177: Holy combinatorial explosion! Would this get any simpler of the config was an enum set and `RuntimeConfig` was a three value enum? Baseline config would be an empty set.

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/internal/SubtreeManagerTest.kt
```diff
@@ -172,10 +174,34 @@ internal class SubtreeManagerTest {
...
+    val update = tickOutput.await().output!!.value!!
 
-    val (_, output) = update.applyTo("props", "state")
-    assertEquals("case output:workflow output:event!", output?.value)
+    val (_, result) = update.applyTo("props", "state")
+    assertEquals("case output:workflow output:event!", result.output!!.value)
+    assertFalse(result.stateChanged)
+  }
+
+  @Test fun tick_children_handles_no_child_output() = runTest {
```
rjrjr, line 184: nit / out of scope: we don't have a `tick()` method any more, should we update test and var names here?

rjrjr (approved):
Nothing but nits. So excited!

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/WorkflowAction.kt
rjrjr: Remember that you can implement positional params manually, it's pretty trivial.

steve-the-edwards: Like to support destructuring you mean?

Hmm, now I'm curious - what the was the data class issue in the public API?

.github/workflows/kotlin.yml
steve-the-edwards: i did do some cleanup after this - https://github.com/square/workflow-kotlin/pull/1011

workflow-config/config-android/src/main/java/com/squareup/workflow1/config/AndroidRuntimeConfigTools.kt
steve-the-edwards: I like the Set<Enum> idea!

I'm going to leave it as strings for the options though as I don't want to mess with the buildConfig stuff I set up before (the command line options are easy - but the buildconfig for the instrumentation tests is harder IIRC)

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RuntimeConfig.kt
steve-the-edwards: Used in the updated location in the enum.

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/internal/SubtreeManagerTest.kt
steve-the-edwards: Good point. #1012

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
rjrjr: "and process another action" seems to imply that we will render, just that we're going to process more actions first. Should this say "and process the next action (if any)"?

steve-the-edwards: Just missed me merging this :). But I will follow up with this change.

steve-the-edwards: Technically 'processing an action' can involve suspending until there is one but that's more than a little bit confusing.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/WorkflowAction.kt
steve-the-edwards: For posterity. I am going to do this change after all. Canonical resource describing backwards compatibility concerns: https://jakewharton.com/public-api-challenges-in-kotlin/

PR #993, opened 2023-04-24
Use RenderOnlyOnStateChange for RenderPassCounting

steve-the-edwards:
hard-code in this runtime for now.


steve-the-edwards:
This is obselete now. See #992

Issue #994, opened 2023-04-26
Deprecate `runningWorker` and add `@ReplaceWith` for `runningSideEffect` alternative


crocsandcoffee:
There is a variant of `runningWorker` for when your Worker doesn't output anything. The kdoc mentions to use `runningSideEffect` as the alternative. We should consider a formal `@Deprecation` of this `runningWorker` API and can use `@ReplaceWith` for in-place fix/migration. 

Not sure how this type of thing is handled but could then add warning in the deprecation message that this API will be removed in future version

steve-the-edwards:
Completely removed by https://github.com/square/workflow-kotlin/pull/1237

PR #996, opened 2023-05-03
Mark runningWorker with no output as @Deprecated

crocsandcoffee:
This addresses https://github.com/square/workflow-kotlin/issues/994


steve-the-edwards (approved):
Thanks!

crocsandcoffee:
I totally forgot about this.. I think I'm failing some tests because of lint errors. I'll address this ASAP so we can get it merged in and it stops collecting dust.

PR #1000, opened 2023-05-17
Introduces `ScreenTransitionLogger`

rjrjr:
`ScreenTransitionLogger` provides extremely minimal hooks to allow apps to implement whatever it is that "navigation logging" means to them. It is a `ViewEnvironment` service with a default no-op implementation.

`ScreenTransitionLogger.onTransition` is called whenever one of our stock stub views (`WorkflowViewStub` and `@Composable fun WorkflowRendering()`) is updated to:
- show its first `Screen` rendering, or to
- show one that is incompatible with whatever was last shown

Both the incoming and outgoing `Screen` are provided to `onTransition`. The latter may be null when the function is called with the first `Screen` a stub is showing.

For this first stab, we are very intentionally providing a very unambitious mechanism.
- We make no attempt to report the entire AndroidX lifecycle -- e.g., we do not report when stubs are set up or torn down, only when they show something new.
- We focus only on `Screen`, not `Overlay`. In practice we have found that all of our interesting `Overlay` implementations are based on `ScreenOverlay` and stubs anyway.
- We do not attempt to report that view code is "done" updating, only that we have determined that a new view is required.

Note that this approach _will_ log updates under `BackStackScreen`, because `BackStackContainer` is built around `WorkflowViewStub`. Updates to the UI root will also be reported, because `WorkflowLayout` wraps `WorkflowViewStub` as well.

It is worth noting what _is_ supported: `onTransition` is provided with the `ViewEnvironment` in addition to the incoming and outgoing `Screen`s. This should allow more complex apps to set up logging systems that reflect more complex structures. For example, a custom container tiling or layering multiple stubs could provide `ViewEnvironment` values naming these separate navigation contexts. Or if a particular custom container is very fine grained in its use of stubs, it can seed its children's environments with a no-op `ScreenTransitionLogger` to eliminate noise.

### DIFF
--- a/samples/tictactoe/app/src/main/java/com/squareup/sample/mainactivity/TicTacToeActivity.kt
+++ b/samples/tictactoe/app/src/main/java/com/squareup/sample/mainactivity/TicTacToeActivity.kt
@@ -10,18 +10,20 @@ import androidx.test.espresso.IdlingResource
 import com.squareup.sample.authworkflow.AuthViewFactories
 import com.squareup.sample.container.SampleContainers
 import com.squareup.sample.gameworkflow.TicTacToeViewFactories
+import com.squareup.workflow1.ui.ScreenTransitionLogger
+import com.squareup.workflow1.ui.ViewEnvironment
 import com.squareup.workflow1.ui.WorkflowLayout
 import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
-import com.squareup.workflow1.ui.container.withRegistry
+import com.squareup.workflow1.ui.container.withEnvironment
 import com.squareup.workflow1.ui.modal.AlertContainer
 import com.squareup.workflow1.ui.plus
-import kotlinx.coroutines.flow.collect
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.launch
 import timber.log.Timber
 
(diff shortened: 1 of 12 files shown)


rjrjr:
Offline from @zach-klippenstein 

> That kind of code is usually wrong, and very brittle, since it relies on recomposition happening only once. In this case, you avoid that with the compatible check, but it’s still a bad pattern and prone to breaking. A better option would be something like:

```kotlin
val lastScreen by rememberUpdatedState(screen)
val logger = viewEnv[TL]
LaunchedEffect(logger) {
  snapshotFlow { lastScreen }
    .scan { prev, curr -> 
      if (!compatible(prev, curr)) {
        logger.onNav(prev, curr)
      }
    }
    .collect()
}
```

The `SideEffect` approach triggers the event on recomposition, but the `snapshotFlow` approach explicitly triggers only on state change. The `compatible` calls filter out the difference, but the latter is a better expression of intent.

rjrjr:
@zach-klippenstein I actually had thought about `LaunchedEffect` but couldn't figure out what `key` to use. How did you pick `logger`?

rjrjr:
This is just not a good idea. Much much too noisy, would be very difficult to tame

PR #1003, opened 2023-05-31
Do not show `Dialog`s before `Activity` content view is attached.

### LINKED ISSUE #1001: `ComposeView` and `IllegalStateException: You can consumeRestoredStateForKey only after super.onCreate of corresponding component`
Update: we've found other, simpler recipes. `AndroidView` is not in the chain, dialog windows always are. More deets in comment.

We have a family of crashes that pass through the following points. (I'll post a complete stack trace in a follow on comment.)

```
IllegalStateException: You can consumeRestoredStateForKey only after super.onCreate of corresponding component
  SavedStateRegistry.consumeRestoredStateForKey
  AndroidComposeView.setOnViewTreeOwnersAvailable
  LifecycleRegistry.moveToState
  LifecycleRegistry.setCurrentState
  RealWorkflowLifecycleOwner.updateLifecycle
```

Very hard to reproduce. 

- There's a variant in production that looks like it's related to config change, can't repro
- There's one I can make happen on demand via an espresso test that fires a deep link intent from a notification -- relaunches the activity with the UI pointed to a specific screen. (Can't repro by firing the same intent from adb.)

In both cases, the common elements seem to be that we're building the UI for the first time in a new activity, and that `AndroidView {}` is involved. Debugger reveals a missing call to `SavedStateRegistry.consumeRestoredStateForKey`.

Version info:

```
  lifecycleVersion = '2.5.1'

  composeRuntimeVersion = '1.4.3'
  composeFoundationVersion = '1.4.3'
  composeUiVersion = '1.4.3'
  composeAnimationVersion = '1.4.3'
  composeMaterialVersion = '1.1.0-rc01'

  kotlinVersion = getGradleProperty('square.kotlinVersion', '1.8.10')
  composeCompilerVersion = getGradleProperty('square.composeCompilerVersion', '1.4.4')
```

rjrjr:
If the very first rendering would instantiate `Dialog` windows, their content views were getting attached before the `Activity`'s content view. This would cause us to restore saved view state in the wrong order. If a `ComposeView` is in a `Dialog` in this situation, we would crash with an `IllegalStateException` from `SavedStateRegistry.consumeRestoredStateForKey`.

The fix is for `LayeredDialogSessions` and `DialogCollator` to conspire to decouple created `Dialog` instances from showing them the first time. Normally we go ahead and do both at once. But if `LayeredDialogSessions.update` is called before the corresponding `BodyAndOverlaysContainer` view is attached to the `Activity` window, we defer showing until it is.

Fixes #1001

### DIFF
--- a/samples/nested-overlays/src/androidTest/java/com/squareup/sample/nestedoverlays/NestedOverlaysAppTest.kt
+++ b/samples/nested-overlays/src/androidTest/java/com/squareup/sample/nestedoverlays/NestedOverlaysAppTest.kt
@@ -3,6 +3,7 @@ package com.squareup.sample.nestedoverlays
 import androidx.test.espresso.Espresso.onView
 import androidx.test.espresso.ViewInteraction
 import androidx.test.espresso.action.ViewActions.click
+import androidx.test.espresso.action.ViewActions.pressBack
 import androidx.test.espresso.action.ViewActions.typeText
 import androidx.test.espresso.assertion.ViewAssertions.doesNotExist
 import androidx.test.espresso.assertion.ViewAssertions.matches
@@ -54,6 +55,28 @@ class NestedOverlaysAppTest {
     onTopCoverEverything().assertNotDisplayed()
     onBottomCoverBody().assertNotDisplayed()
     onBottomCoverEverything().assertNotDisplayed()
+  }
+
+  @Test fun backButtonWorks() {
+    onTopCoverEverything().perform(click())
+    onView(withText("Cover Body")).inRoot(isDialog()).perform(click())
+
+    onView(withText("Reveal Body")).inRoot(isDialog()).perform(pressBack())
+    onView(withId(R.id.button_bar_text)).inRoot(isDialog()).perform(pressBack())
+
+    onTopCoverBody().assertDisplayed()
+  }
+
+  @Test fun backButtonWorksAfterConfigChange() {
+    onTopCoverEverything().perform(click())
+    onView(withText("Cover Body")).inRoot(isDialog()).perform(click())
+
+    scenarioRule.scenario.recreate()
+
+    onView(withText("Reveal Body")).inRoot(isDialog()).perform(pressBack())
+    onView(withId(R.id.button_bar_text)).inRoot(isDialog()).perform(pressBack())
+
+    onTopCoverBody().assertDisplayed()
   }
 
   // https://github.com/square/workflow-kotlin/issues/966
--- a/samples/nested-overlays/src/androidTest/java/com/squareup/sample/nestedoverlays/NestedOverlaysAppTest.kt
+++ b/samples/nested-overlays/src/androidTest/java/com/squareup/sample/nestedoverlays/NestedOverlaysAppTest.kt
@@ -107,6 +107,20 @@ class NestedOverlaysAppTest {
       .check(matches(withText("banana")))
   }
 
+  @Test fun orderPreservedOnConfigChange() {
+    // Show the outer dialog
+    onTopCoverEverything().perform(click())
+
+    // Click the outer dialog's button to show the inner dialog.
+    onView(withText("Cover Body")).inRoot(isDialog()).perform(click())
+
+    // "Config change"
+    scenarioRule.scenario.recreate()
+
+    // The green "Cover Everything" dialog is on top.
+    onView(withText("Reveal Body")).inRoot(isDialog()).check(matches(isDisplayed()))
+  }
+
   // https://github.com/square/workflow-kotlin/issues/314
   @Test fun whenBodyAndOverlaysStopsBeingRenderedDialogsAreDismissed() {
     onBottomCoverBody().perform(click())
--- a/samples/nested-overlays/src/main/java/com/squareup/sample/nestedoverlays/NestedOverlaysWorkflow.kt
+++ b/samples/nested-overlays/src/main/java/com/squareup/sample/nestedoverlays/NestedOverlaysWorkflow.kt
@@ -7,9 +7,11 @@ import com.squareup.workflow1.Snapshot
 import com.squareup.workflow1.StatefulWorkflow
 import com.squareup.workflow1.ui.Screen
 import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+import com.squareup.workflow1.ui.container.BackButtonScreen
 import com.squareup.workflow1.ui.container.BodyAndOverlaysScreen
 import com.squareup.workflow1.ui.container.FullScreenModal
 
+@OptIn(WorkflowUiExperimentalApi::class)
 object NestedOverlaysWorkflow : StatefulWorkflow<Unit, State, Nothing, Screen>() {
   data class State(
     val showTopBar: Boolean = true,
@@ -46,15 +48,19 @@ object NestedOverlaysWorkflow : StatefulWorkflow<Unit, State, Nothing, Screen>()
     val outerSheet = if (!renderState.showOuterSheet) {
       null
     } else {
+      val closeOuter = context.eventHandler { state = state.copy(showOuterSheet = false) }
       FullScreenModal(
-        ButtonBar(
-          Button(
-            name = R.string.close,
-            onClick = context.eventHandler { state = state.copy(showOuterSheet = false) }
+        BackButtonScreen(
+          ButtonBar(
+            Button(
+              name = R.string.close,
+              onClick = closeOuter
+            ),
+            context.toggleInnerSheetButton(renderState),
+            color = android.R.color.holo_green_light,
+            showEditText = true,
           ),
-          context.toggleInnerSheetButton(renderState),
-          color = android.R.color.holo_green_light,
-          showEditText = true
+          onBackPressed = closeOuter
         )
       )
     }
@@ -62,20 +68,24 @@ object NestedOverlaysWorkflow : StatefulWorkflow<Unit, State, Nothing, Screen>()
     val innerSheet = if (!renderState.showInnerSheet) {
       null
     } else {
+      val closeInner = context.eventHandler { state = state.copy(showInnerSheet = false) }
       FullScreenModal(
-        ButtonBar(
-          Button(
-            name = R.string.close,
-            onClick = context.eventHandler { state = state.copy(showInnerSheet = false) }
+        BackButtonScreen(
+          ButtonBar(
+            Button(
+              name = R.string.close,
+              onClick = closeInner
+            ),
+            toggleTopBarButton,
+            toggleBottomBarButton,
+            Button(
+              name = R.string.nuke,
+              onClick = context.eventHandler { state = State(nuked = true) }
+            ),
+            color = android.R.color.holo_red_light,
+            showEditText = true
           ),
-          toggleTopBarButton,
-          toggleBottomBarButton,
-          Button(
-            name = R.string.nuke,
-            onClick = context.eventHandler { state = State(nuked = true) }
-          ),
-          color = android.R.color.holo_red_light,
-          showEditText = true
+          onBackPressed = closeInner
         )
       )
     }
--- a/workflow-ui/core-android/src/androidTest/java/com/squareup/workflow1/ui/container/DialogIntegrationTest.kt
+++ b/workflow-ui/core-android/src/androidTest/java/com/squareup/workflow1/ui/container/DialogIntegrationTest.kt
@@ -1,10 +1,18 @@
 package com.squareup.workflow1.ui.container
 
 import android.app.Dialog
+import android.text.SpannableStringBuilder
 import android.view.View
+import android.view.ViewGroup.LayoutParams
+import android.view.ViewGroup.LayoutParams.MATCH_PARENT
 import android.widget.EditText
 import androidx.activity.ComponentActivity
 import androidx.lifecycle.Lifecycle.State.DESTROYED
+import androidx.test.espresso.Espresso.onView
+import androidx.test.espresso.assertion.ViewAssertions.matches
+import androidx.test.espresso.matcher.RootMatchers.isDialog
+import androidx.test.espresso.matcher.ViewMatchers.isDisplayed
+import androidx.test.espresso.matcher.ViewMatchers.withText
 import androidx.test.ext.junit.rules.ActivityScenarioRule
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import com.google.common.truth.Truth.assertThat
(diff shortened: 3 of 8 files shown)


rjrjr:
This probably won't play nicely with `DialogCollator` as is.

rjrjr:
> This probably won't play nicely with `DialogCollator` as is.

Confirmed, dialogs can re-open in the wrong order on config change. Shouldn't be a hard fix though. Just have the outermost container view put something in the view environment telling children to do nothing, and then make it the only one to update when the attach event comes.

Special bonus: this breaks every method in `DialogIntegrationTest`, even though the _Nested Overlays_ sample works fine manually (outside of config change).

workflow-ui/core-android/src/androidTest/java/com/squareup/workflow1/ui/container/DialogIntegrationTest.kt
```diff
@@ -79,12 +91,14 @@ internal class DialogIntegrationTest {
 
     scenario.onActivity { activity ->
       val root = WorkflowLayout(activity)
+      activity.setContentView(root)
```
rjrjr, line 94: Oops, these tests were never actually attaching anything to the activity window. Now they do.

rjrjr:
> Special bonus: this breaks every method in `DialogIntegrationTest`, even though the _Nested Overlays_ sample works fine manually (outside of config change).

Derp. That test was never actually callint `Activity.setContentView()`. (It doesn't run against the sample app, that's a different test.)

rjrjr:
Realizing that in the recursive case, nested instances will always defer and replay for no reason. That seems bad. Could be why the performance shard "broke" (actually reported an unexpected improvement in render count).

rjrjr:
This works, but under the hood it's a pretty drastic change to the order in which a UI hierarchy is updated -- a parent view can no longer assume that children will update synchronously, at least once an `Overlay` gets into the path. E.g. `BackButtonScreen` relies on this convention to ensure that children will stack their handlers before or after the parent's own, based on `BackButtonScreen.shadow`.

Of course this is a pretty janky, brittle mechanism, only in use in a few spots. Maybe this is the moment to improve those spots? Maybe we can find a fix to this attachment order problem that covers the whole tree, instead of just `LayeredDialogSessions`?

rjrjr:
Ideally we would build the `Dialog`s first and then show them all at once, but that's simply not an option -- there are calls that it is unsafe to make on a `Dialog`'s `DecorView` before `show()`.

rjrjr:
Could try to make `WorkflowLifecycleOwner` refuse to move to `RESUMED` before it is attached? Hard to imagine that won't break 500 other things. And maybe this fantasy of "children will update exactly and only when their parents say to" is already doomed by Compose.

rjrjr:
> Could try to make `WorkflowLifecycleOwner` refuse to move to `RESUMED` before it is attached? Hard to imagine that won't break 500 other things. And maybe this fantasy of "children will update exactly and only when their parents say to" is already doomed by Compose.

Doesn't change anything, they're basically always attached.

rjrjr:
> Realizing that in the recursive case, nested instances will always defer and replay for no reason. That seems bad. Could be why the performance shard "broke" (actually reported an unexpected improvement in render count).

Untrue, the new code is only firing in the problem case -- bootstrapping an entire view tree that includes dialog windows.

rjrjr:
Well now.

That "always builds or updates all the things at once and in the expected order" invariant is a lot harder to live without than I realized.

Trying a new approach, which is maybe simpler and less fragile anyway. Basically, if we're not attached to a window when building dialogs, the only thing we defer until attach time is showing them -- we build them all immediately, just like today.

`DialogCollator` actually makes that very simple. Now to see if this approach still prevents the `IllegalStateException` from `consumeRestoredStateForKey`… 🤞

rjrjr:
OMFG it works.

rjrjr:
It works it works it works it works. For the problems that I know about. Can't wait to see what it breaks instead!

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/BodyAndOverlaysContainer.kt
```diff
@@ -84,11 +84,11 @@ internal class BodyAndOverlaysContainer @JvmOverloads constructor(
   }
 
   override fun dispatchTouchEvent(event: MotionEvent): Boolean {
-    return !dialogs.allowEvents || super.dispatchTouchEvent(event)
+    return !dialogs.allowContentEvents || super.dispatchTouchEvent(event)
```
rjrjr, line 87: Unrelated method rename to try to make it clearer that this is strictly about event handling in the activity window, not the dialogs.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/DialogCollator.kt
```diff
@@ -60,19 +60,29 @@ internal class DialogSessionUpdate(
  * and [DialogCollator.scheduleUpdates]
  *
  * @param existingSessions [DialogSession] instances that were running before the pending update
+ *
+ * @param onRootUpdateFinished honored only for the outermost [LayeredDialogSessions] caller.
+ * If provided, called with the complete set of [DialogSession] created by
+ * [DialogCollator.scheduleUpdates], not just those of the client identified by [id].
+ * https://github.com/square/workflow-kotlin/issues/1001
```
rjrjr, line 67: This URL isn't really useful here. It's already in the comments that are at the heart of the full fix.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/DialogSession.kt
```diff
@@ -125,7 +127,18 @@ internal class DialogSession(
         }
       }
     }
+  }
+
+  /**
+   * One time call to shows the managed [Dialog][OverlayDialogHolder.dialog] for the first time.
+   * Called after [initNewDialog] and [destroyDialog]
+   *
+   * See also [setVisible], used to dismiss and re-show and existing one.
```
rjrjr, line 136:
```suggestion
   * One time call to show the managed [Dialog][OverlayDialogHolder.dialog] for the first time.
   * Called between [initNewDialog] and [destroyDialog].
   *
   * See also [setVisible], used to dismiss and re-show an existing one.
```

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/LayeredDialogSessions.kt
```diff
@@ -113,7 +113,34 @@ public class LayeredDialogSessions private constructor(
 
   private var sessions: List<DialogSession> = emptyList()
 
-  public var allowEvents: Boolean = true
+  /**
+   * Reflects whether [onAttachedToWindow] has been called yet, so that
+   * we can avoid creating `Dialog` windows before the `Activity` window
```
rjrjr, line 118:
```suggestion
   * we can avoid [showing][showNewDialog] `Dialog` windows before the `Activity` window
```

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/LayeredDialogSessions.kt
```diff
@@ -113,7 +113,34 @@ public class LayeredDialogSessions private constructor(
...
+   * is true, holds on to the set of all [DialogSession] to be shown at
+   * attach time, instead of immediately. Necessary to ensure that view state
+   * is restored in the correct order -- the Activity contentView must be attached
+   * to its window before those of any Dialog instance.
+   *
+   * Note that this is a superset of [sessions], includes built by
+   * any nested [LayeredDialogSessions] instances during recursive [update]
+   * calls.
+   *
+   * https://github.com/square/workflow-kotlin/issues/1001
```
rjrjr, line 134:
```suggestion
   * Set by an outermost instance only. If we are in a root [update] call before 
   * [attached] is true, holds on to the set of all [DialogSession] to be 
   * [shown][showNewDialog] at attach time, instead of immediately. Necessary to
   * ensure that view state is restored in the correct order -- the `Activity`
   * `contentView` must be attached to its window before that of any `Dialog`.
   *
   * Note that this is a superset of [sessions], includes built by
   * any nested [LayeredDialogSessions] instances during recursive [update]
   * calls.
   *
   * https://github.com/square/workflow-kotlin/issues/1001
```

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/BodyAndOverlaysContainer.kt
steve-the-edwards: Why `allowContentEvents` and not `allowBodyEvents`? I think Body would help me understand what you mean.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/LayeredDialogSessions.kt
```diff
@@ -113,7 +113,35 @@ public class LayeredDialogSessions private constructor(
...
+   * Note that this is a superset of [sessions], also includes sessions built by
+   * any nested [LayeredDialogSessions] instances during recursive [update]
+   * calls.
+   *
+   * https://github.com/square/workflow-kotlin/issues/1001
+   */
+  private var deferredShow: List<DialogSession>? = null
+
+  /**
+   * Clients should ignore UI events targeting their [content][ScreenOverlay.content]
```
steve-the-edwards, line 140: so this is where the `content` naming comes from? Hmm. How do I map body to content? are they equivalent? 'Body' is the container for 'content'?

steve-the-edwards (commented):
Nice! A few naming questions as usual 😄

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/BodyAndOverlaysContainer.kt
rjrjr: I wrestled with that. I went with `content` because it matches `ScreenOverlay.content`, which in turn comes from the `Wrapper` interface.  But really they're both still ambiguous, aren't they -- dialog content? dialog body?

How about `allowBaseEvents`? That would match the `updateBase` param to `LayeredDialogSessions.update`.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/DialogSession.kt
```diff
@@ -87,10 +88,11 @@ internal class DialogSession(
   private val KeyEvent.isBackPress: Boolean
     get() = (keyCode == KEYCODE_BACK || keyCode == KEYCODE_ESCAPE) && action == ACTION_UP
 
-  fun initAndShowDialog(
-    parentLifecycleOwner: LifecycleOwner,
-    initialEnvironment: ViewEnvironment
-  ) {
+  /**
+   * One time call to set up our brand new [OverlayDialogHolder] instance.
```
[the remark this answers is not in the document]
rjrjr: That gets into another level of API inconsistency. `holder.show` is public, `OverlayDialogHolder.show`, and it parallel `ScreenViewHolder.show`. The is definitely an awkward seam, but it's all in non-public code. 

Basically, `Dialog.show()` conflicts with our `show(someRendering, environment)` convention. But it's only us on the Workflow team who take the cognitive hit, b/c feature code never gets anywhere near `Dialog.show()` or `Dialog.dismiss()`.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/LayeredDialogSessions.kt
rjrjr: Oh, derp: and that kdoc is TOTALLY WRONG! Link should be to `BodyAndOverlays.content`!

rjrjr: Tell you what, let's decouple this naming conversation from the ordering fix. Maybe even move it upstream.

rjrjr:
```kotlin
  /**
   * While this is true, clients should ignore UI events targeting the
   * [body][BodyAndOverlaysScreen.body] view beneath the
   * [Dialog][android.app.Dialog] windows they manage. It reflects
   * that a [ModalOverlay] is (or soon will be) showing over the body screen.
   */
  public var allowBodyEvents: Boolean = true
```

samples/nested-overlays/src/main/java/com/squareup/sample/nestedoverlays/NestedOverlaysWorkflow.kt
```diff
@@ -46,36 +48,44 @@ object NestedOverlaysWorkflow : StatefulWorkflow<Unit, State, Nothing, Screen>()
...
-          Button(
-            name = R.string.close,
-            onClick = context.eventHandler { state = state.copy(showOuterSheet = false) }
+        BackButtonScreen(
+          ButtonBar(
+            Button(
+              name = R.string.close,
+              onClick = close
+            ),
+            context.toggleInnerSheetButton(renderState),
```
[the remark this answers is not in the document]
rjrjr: Oops, nope, can't do that either.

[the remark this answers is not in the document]
rjrjr: Combination of `vararg` and default values is pretty hard to work with, I don't think I can make this better.

rjrjr:
@steve-the-edwards PTAL, adopted most of your naming fixes.

samples/nested-overlays/src/androidTest/java/com/squareup/sample/nestedoverlays/NestedOverlaysAppTest.kt
```diff
@@ -56,6 +57,28 @@ class NestedOverlaysAppTest {
     onBottomCoverEverything().assertNotDisplayed()
   }
 
+  @Test fun backButtonWorks() {
+    onTopCoverEverything().perform(click())
+    onView(withText("Cover Body")).inRoot(isDialog()).perform(click())
```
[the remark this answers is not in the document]
rjrjr: I don't know why, but I always find that makes tests harder for me to read.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/BodyAndOverlaysContainer.kt
rjrjr: See insane stream of consciousness below, `allowBodyEvents` and `updateBody()` FTW.

ijwhelan (approved):
Still a novice at Workflow internals but I'm not seeing any obvious issues, changes make sense to me.

PR #1008, opened 2023-06-02
adding custom composite actions for CI

RBusarow:
This adds some custom internal actions, which removes some boilerplate from job definitions and ensures that everything's consistent.

For each step which invokes a Gradle task, we will:

- set up JDK 11 (zulu)
- set Gradle properties and jvm args, where the memory settings are OS-specific
  - `-Xmx10g` for MacOS
  - `-Xmx5g` for Ubuntu
  - `-Xmx3g` for Windows
- use `gradle-build-action@v2` for caching
- invoke whatever Gradle task(s)
- upload a heap dump in the event of a memory error

This also adds a `gradle-task-with-commit` action which also does all the above.  In the future, it will automagically run format/baseline tasks before a "check" task if it has permissions to push to the PR branch.  **For the moment, this is not actually set up.**  I need to get us a service token with the appropriate permissions first.  So for now, the jobs which use `gradle-task-with-comment` will just execute the "check" version for everyone.

### DIFF
--- a/.github/actions/gradle-task-with-commit/action.yml
+++ b/.github/actions/gradle-task-with-commit/action.yml
@@ -1,0 +1,89 @@
+name : Set Up Gradle and Run Task
+description : This action performs a "fix"-type Gradle task and commits/pushes changes if possible.
+
+inputs :
+  java-version :
+    description : 'The Java version to set up'
+    default : '11'
+  distribution :
+    description : 'The JDK distribution to use'
+    default : 'zulu'
+  fix-task :
+    description: 'The task to be executed if the user has a PAT and the branch is not a fork'
+    required: true
+  check-task:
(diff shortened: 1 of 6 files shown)


rjrjr:
> This also adds a `gradle-task-with-commit` action which does all the above, except it will automagically run format/baseline tasks before a "check" task if it has permissions to push to the PR branch. So it will automatically fix KtLint issues.

How automatically are we talking about here?

.github/actions/gradle-task-with-commit/action.yml
```diff
@@ -0,0 +1,74 @@
+name : Set Up Gradle and Run Task
```
RBusarow, line 1: - [ ] rename this

steve-the-edwards (approved):
Nice!

rjrjr (approved):

Issue #1010, opened 2023-06-08
Remove mobile.dev integration


steve-the-edwards:
They have let us know they are focusing on UI Testing with Maestro and the performance testing and app size analysis will be shutting down. We can remove uploading our sample APKs to their servers.

PR #1011, opened 2023-06-08
Clean Up Actions and Remove JMH Benchmarks

### LINKED ISSUE #998: Remove JMH Benchmarks
Was used for pure-Java benchmarks of some core runtime functions to ensure they weren't getting out of hand.
We are not validating against their results in CI though (and not in the practice of running them locally to check).

They caused headaches for KMP migration, and now possibly Kotlin 1.8 migration.

They were introduced prior to androidx.microbenchmark. While having the pure JVM benchmark is nice, androidx.microbenchmark  is more amenable to our toolchain and getting the validation in the context of Android is likely more useful anyway.

Follow up work would be to reproduce something similar with microbenchmark on android targets _and_ run that in CI.

steve-the-edwards:
Closes #998 , and #1010 .

Renames shards that were previously ambiguous regarding runtimes.


workflow-runtime/build.gradle.kts
```diff
@@ -14,28 +11,7 @@ kotlin {
...
-          defaultSourceSet {
-            dependencies {
-              implementation(libs.kotlinx.benchmark.runtime)
-
-              implementation(main.compileDependencyFiles + main.output.classesDirs)
-            }
-          }
-        }
-      }
-    }
```
RBusarow, line 38: 🔥

workflow-runtime/src/jvmWorkflowNode/kotlin/com/squareup/workflow1/RenderBenchmarkResults.txt
RBusarow, line 1: 🔥

RBusarow (approved):

PR #1025, opened 2023-06-14
use delegating shard tasks for `connectedCheck`

RBusarow:
These shard tasks collectively depend upon all Android `connectedCheck` tasks in the entire project.

Each shard depends upon the `connectedCheck` tasks of some subset of Android projects. Projects are assigned to a shard by counting the number of `@Test` annotations within their `androidTest` directory, then associating those projects to a shard in a round-robin fashion.

These shards are invoked in CI using a matrix. If the number of shards changes, the `connectedCheckShardMatrixYamlUpdate` task can automatically update the workflow file so that they're all invoked.

The shard tasks are invoked as:

```shell
# roughly 1/3 of the tests
./gradlew connectedCheckShard1
# the second third
./gradlew connectedCheckShard2
# the last third
./gradlew connectedCheckShard3
```

The task filtering we currently use in CI (`./gradlew fooShard1 -x :my-project:foo`) will still work here, however the "cost" of the excluded task's tests is still accounted for when the sharding is performed.

### DIFF
--- a/build-logic/src/main/java/com/squareup/workflow1/buildsrc/shardConnectedChecks.kt
+++ b/build-logic/src/main/java/com/squareup/workflow1/buildsrc/shardConnectedChecks.kt
@@ -63,18 +63,20 @@ fun shardConnectedCheckTasks(target: Project) {
 
   shardAssignments.forEach { shard ->
 
+    val projects by shard.projectsLazy
+
+    val paths by lazy {
+      projects.joinToString(prefix = "[ ", postfix = " ]") { it.path }
+    }
+
     target.tasks.register("connectedCheckShard${shard.number}") {
 
       group = "Verification"
-
-      val projects = shard.projects
 
       validateSharding(
         projectsWithTestCount = projectsWithTestCount.value,
         shardAssignments = shardAssignments
       )
-
-      val paths = projects.joinToString(prefix = "[ ", postfix = " ]") { it.path }
 
       description = "Runs $connectedTestName in projects: $paths"
 
@@ -83,6 +85,24 @@ fun shardConnectedCheckTasks(target: Project) {
       }
 
       dependsOn(assignedTests)
+    }
+
+    target.tasks.register("prepareConnectedCheckShard${shard.number}") {
+
+      validateSharding(
+        projectsWithTestCount = projectsWithTestCount.value,
+        shardAssignments = shardAssignments
+      )
+
+      description = "Builds all artifacts for running connected tests in projects: $paths"
+
+      val regex = Regex("""prepare[A-Z]\w*AndroidTestArtifacts""")
+
+      val prepareTasks = projects.map { project ->
+        project.tasks.matching { it.name.matches(regex) }
+      }
+
+      dependsOn(prepareTasks)
     }
   }
 }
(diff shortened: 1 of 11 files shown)


build-logic/src/main/java/com/squareup/workflow1/buildsrc/shardConnectedChecks.kt
```diff
@@ -0,0 +1,111 @@
...
+ * # the last third
+ * ./gradlew connectedCheckShard3
+ * ```
+ *
+ * @param target the root project which gets the shard tasks
+ * @param shardCount the number of shard tasks to create.
+ */
+fun shardConnectedCheckTasks(
+  target: Project,
+  shardCount: Int = 3
```
rjrjr, line 34: Why provide a default, seems like that just invites mistakes where you forget to pass the matrix value and don't notice.

build-logic/src/main/java/com/squareup/workflow1/buildsrc/shardConnectedChecks.kt
```diff
@@ -0,0 +1,111 @@
...
+          { it.path }
+        )
+      )
+
+    var shardIndex = 0
+
+    sortedProjects.groupBy { (shardIndex++ % shardCount) + 1 }
+  }
+
+  (1..shardCount).map { shardIndex ->
```
rjrjr, line 92: If we have an off by one error and some boundary tests fall between the cracks, how will we notice that shard1 + shard2 + shard3 < allShards? Wondering if there's an invariant check we can put somewhere. 🤔

rjrjr: Maybe extract a core of this routine so that you can unit test it?

rjrjr (approved):

build-logic/src/main/java/com/squareup/workflow1/buildsrc/shardConnectedChecks.kt
RBusarow: I wavered.  I didn't want the configuration option to be out of place in the root `build.gradle.kts`, but I also didn't want it to be hard-coded.  So I found a third bad option and went with that. ;)

Fixed.  It's now a `private const val`.

build-logic/src/main/java/com/squareup/workflow1/buildsrc/shardConnectedChecks.kt
RBusarow: I didn't (and still don't) want to introduce unit tests to `build-logic`, since the tasks for an included build are not hooked into Gradle's implicit task delegation by default.

For example, if we invoke `./gradlew test` today, it'll invoke all unit tests in kotlin-jvm modules.  We just expect that to work.  But that hook (implicit task delegation) only applies to sub-projects. The `build-logic` build is not affected by this, so its `test` task would never be invoked.

We can manually stitch things together, such as with this in the root `build.gradle.kts`:

```kotlin
tasks.named("test") {
  dependsOn(gradle.includedBuild("build-logic").task(":test"))
}
```

...but if anything were to happen to that, then the sharding could break in the way you described and no one would be the wiser.

So, instead I refactored it to be as off-by-one-proof as I could.  As the tasks are configuring, they assert that the shard allocations total up to the exact number of Android projects and there are no duplicates.  So they test themselves every time they're used.

PR #1027, opened 2023-06-14
2/4 Better support for `View.findViewTreeOnBackPressedDispatcherOwner`.

rjrjr:
`fun View.findViewTreeOnBackPressedDispatcherOwner()` is AndroidX's preferred entry point to the exciting new world of `OnBackPressedDispatcherOwner`. With this PR we support it explicitly, in particular taking care to ensure that it can be called by newly constructed views before they have been attached to a parent. That is, we take care to make eager calls `View.setViewTreeSavedStateRegistryOwner` on every view we build.

We accomplish this mainly by riding the rails previously laid down via `WorkflowLifecycleOwner.installOn`, which now requires an `OnBackPressedDispatcherOwner` parameter. (This is the method used by `WorkflowViewStub` _et al_ to ensure that we're managing the JetPack Lifecycle correctly, and `OnBackPressedDispatcherOwner` is just another piece of that puzzle.)

In aid of that, we introduce a new `ViewEnvironmentKey`, `OnBackPressedDispatcherOwnerKey`. It is initialized by `WorkflowLayout`, our new `ComponentDialog.setContent` extension, and `@Composable fun WorkflowRendering()`. This key is not intended for use by feature code, it's more of an implementation detail that has to stay public to allow custom containers to be built. It ensures that `WorkflowViewStub` and friends will have access to the correct `OnBackPressedDispatcherOwner` before they have access to a parent view.

- [ ] TODO: add tests of `@Composable fun BackHandler()`, in both activity and dialog windows.

### DIFF
--- a/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/ComponentDialogSetContent.kt
+++ b/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/ComponentDialogSetContent.kt
@@ -5,7 +5,7 @@ import android.util.TypedValue
 import android.view.WindowManager.LayoutParams.FLAG_DIM_BEHIND
 import android.view.WindowManager.LayoutParams.FLAG_NOT_TOUCH_MODAL
 import androidx.activity.ComponentDialog
-import androidx.activity.setViewTreeOnBackPressedDispatcherOwner
+import com.squareup.workflow1.ui.OnBackPressedDispatcherOwnerKey
 import com.squareup.workflow1.ui.Screen
 import com.squareup.workflow1.ui.ViewEnvironment
 import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
@@ -19,7 +19,7 @@ import com.squareup.workflow1.ui.toViewFactory
  * the Dialog's content via instances of a particular type of [ScreenOverlay].
  *
  * Dialogs managed this way are compatible with
- * [View.setBackHandler][com.squareup.workflow1.ui.setBackHandler],
+ * [View.backPressedHandler][com.squareup.workflow1.ui.backPressedHandler],
  * and honor the [OverlayArea] and [CoveredByModal] values placed in
  * the [ViewEnvironment] by the standard [BodyAndOverlaysScreen] container.
  */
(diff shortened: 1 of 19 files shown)


workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/ComponentDialogSetContent.kt
```diff
@@ -62,10 +62,17 @@ public fun <C : Screen, O : ScreenOverlay<C>> ComponentDialog.setContent(
...
+  // imitation of that one, and is going to be removed soon.
   return OverlayDialogHolder(
     initialEnvironment = environment,
     dialog = this,
+    onBackPressed = null
   ) { newOverlay, newEnvironment ->
-    contentHolder.show(newOverlay.content, newEnvironment)
+    contentHolder.show(
+      newOverlay.content,
+      newEnvironment + (OnBackPressedDispatcherOwnerKey to this@setContent)
```
steve-the-edwards, line 75:
```suggestion
      // Note we need to add back the OnBackPressedDispatcherOwner each time we receive a new ViewEnvironment.
      newEnvironment + (OnBackPressedDispatcherOwnerKey to this@setContent)
```

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/DialogSession.kt
```diff
@@ -150,12 +155,29 @@ internal class DialogSession(
...
+      // so 99% chance we'll hit the Activity before the stub.
+      val onBack = (dialog as? OnBackPressedDispatcherOwner)
+        ?: holder.environment.map[OnBackPressedDispatcherOwnerKey] as? OnBackPressedDispatcherOwner
+        ?: decorView.onBackPressedDispatcherOwnerOrNull()
+        ?: object : OnBackPressedDispatcherOwner {
+          override fun getLifecycle(): Lifecycle =
+            error("To support back press handling extend ComponentDialog: $dialog")
+
+          override fun getOnBackPressedDispatcher(): OnBackPressedDispatcher =
+            error("To support back press handling extend ComponentDialog: $dialog")
```
steve-the-edwards, line 171: Nice!

steve-the-edwards (approved):

PR #1028, opened 2023-06-14
3/4 `fun View.setBackHandler()` replaces `val View.backPressedHandler`

rjrjr:
`val View.backPressedHandler` is overly complicated and kind of naive. The replacement, `fun View.setBackHandler()`, carefully echoes the API and behavior of [`@Composable fun BackHandler()`](https://developer.android.com/reference/kotlin/androidx/activity/compose/package-summary#BackHandler(kotlin.Boolean,kotlin.Function0)). We also provide a single argument overload that take a nullable handler function, and sets `enabled` to true for non-`null` handlers, false for `null`.

The biggest change in implementation is that we now use the preferred `OnBackPressedDispatcher.addCallback(LifecycleOwner, OnBackPressedCallback)` overload, which should allow AndroidX to do all the lifecycle bookkeeping for us -- taking advantage of all the hard work we've done to make `WorkflowLifecycleOwner` behave correctly.

This work revealed a problem where there was no `WorkflowLifecycleOwner` in place in time for the first update of the content view in `ScreenOverlayDialogFactory`, which prevented us from repeating that mistake in the new `ComponentDialog.setContent()` extension function introduced two PRs up. So that's nice.

### DIFF
--- a/samples/dungeon/timemachine-shakeable/src/main/java/com/squareup/sample/timemachine/shakeable/ShakeableTimeMachineLayoutRunner.kt
+++ b/samples/dungeon/timemachine-shakeable/src/main/java/com/squareup/sample/timemachine/shakeable/ShakeableTimeMachineLayoutRunner.kt
@@ -13,7 +13,7 @@ import com.squareup.workflow1.ui.ScreenViewRunner
 import com.squareup.workflow1.ui.ViewEnvironment
 import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
 import com.squareup.workflow1.ui.WorkflowViewStub
-import com.squareup.workflow1.ui.backPressedHandler
+import com.squareup.workflow1.ui.setBackHandler
 import kotlin.time.Duration
 import kotlin.time.Duration.Companion.milliseconds
 import kotlin.time.ExperimentalTime
@@ -44,7 +44,7 @@ class ShakeableTimeMachineLayoutRunner(
     environment: ViewEnvironment
   ) {
     // Only handle back presses explicitly if in playback mode.
-    view.backPressedHandler = rendering.onResumeRecording.takeUnless { rendering.recording }
+    view.setBackHandler(rendering.onResumeRecording.takeUnless { rendering.recording })
 
     seek.max = rendering.totalDuration.toProgressInt()
     seek.progress = rendering.playbackPosition.toProgressInt()
--- a/samples/tictactoe/app/src/main/java/com/squareup/sample/authworkflow/LoginViewFactory.kt
+++ b/samples/tictactoe/app/src/main/java/com/squareup/sample/authworkflow/LoginViewFactory.kt
@@ -4,7 +4,7 @@ import com.squareup.sample.tictactoe.databinding.LoginLayoutBinding
 import com.squareup.workflow1.ui.ScreenViewFactory
 import com.squareup.workflow1.ui.ScreenViewFactory.Companion.fromViewBinding
 import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
-import com.squareup.workflow1.ui.backPressedHandler
+import com.squareup.workflow1.ui.setBackHandler
 
 @OptIn(WorkflowUiExperimentalApi::class)
 internal val LoginViewFactory: ScreenViewFactory<LoginScreen> =
@@ -15,5 +15,5 @@ internal val LoginViewFactory: ScreenViewFactory<LoginScreen> =
       rendering.onLogin(loginEmail.text.toString(), loginPassword.text.toString())
     }
 
-    root.backPressedHandler = { rendering.onCancel() }
+    root.setBackHandler(rendering.onCancel)
   }
--- a/workflow-ui/core-android/src/androidTest/java/com/squareup/workflow1/ui/ViewBackHandlerTest.kt
+++ b/workflow-ui/core-android/src/androidTest/java/com/squareup/workflow1/ui/ViewBackHandlerTest.kt
@@ -1,0 +1,54 @@
+package com.squareup.workflow1.ui
+
+import android.view.View
+import androidx.activity.ComponentActivity
+import androidx.test.ext.junit.rules.ActivityScenarioRule
+import com.google.common.truth.Truth.assertThat
+import com.squareup.workflow1.ui.androidx.WorkflowLifecycleOwner
+import com.squareup.workflow1.ui.internal.test.IdlingDispatcherRule
+import leakcanary.DetectLeaksAfterTestSuccess
+import org.junit.Rule
+import org.junit.Test
+import org.junit.rules.RuleChain
+
+@OptIn(WorkflowUiExperimentalApi::class)
+internal class ViewBackHandlerTest {
+  private val scenarioRule = ActivityScenarioRule(ComponentActivity::class.java)
+  private val scenario get() = scenarioRule.scenario
+
+  @get:Rule val rules: RuleChain = RuleChain.outerRule(DetectLeaksAfterTestSuccess())
+    .around(scenarioRule)
+    .around(IdlingDispatcherRule)
+
+  private var viewHandlerCount = 0
+  private fun viewBackHandler() {
+    viewHandlerCount++
+  }
+
+  @Test fun itWorksWhenHandlerIsAddedBeforeAttach() {
+    scenario.onActivity { activity ->
+      val view = View(activity)
+      WorkflowLifecycleOwner.installOn(view, activity)
+      view.setBackHandler { viewBackHandler() }
+
+      activity.setContentView(view)
+      assertThat(viewHandlerCount).isEqualTo(0)
+
+      activity.onBackPressedDispatcher.onBackPressed()
+      assertThat(viewHandlerCount).isEqualTo(1)
+    }
+  }
+
+  @Test fun itWorksWhenHandlerIsAddedAfterAttach() {
+    scenario.onActivity { activity ->
+      val view = View(activity)
+      activity.setContentView(view)
+
+      view.setBackHandler { viewBackHandler() }
+      assertThat(viewHandlerCount).isEqualTo(0)
+
+      activity.onBackPressedDispatcher.onBackPressed()
+      assertThat(viewHandlerCount).isEqualTo(1)
+    }
+  }
+}
--- a/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/BackPressHandler.kt
+++ b/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/BackPressHandler.kt
@@ -13,6 +13,7 @@ import com.squareup.workflow1.ui.androidx.WorkflowAndroidXSupport.onBackPressedD
  * A function passed to [View.backPressedHandler], to be called if the back
  * button is pressed while that view is attached to a window.
  */
+@Deprecated("Use View.backHandler()")
 @WorkflowUiExperimentalApi
 public typealias BackPressHandler = () -> Unit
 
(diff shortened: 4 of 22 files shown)


samples/dungeon/timemachine-shakeable/src/main/java/com/squareup/sample/timemachine/shakeable/ShakeableTimeMachineLayoutRunner.kt
```diff
@@ -44,7 +44,9 @@ class ShakeableTimeMachineLayoutRunner(
     environment: ViewEnvironment
   ) {
     // Only handle back presses explicitly if in playback mode.
-    view.backPressedHandler = rendering.onResumeRecording.takeUnless { rendering.recording }
+    rendering.onResumeRecording.takeUnless { rendering.recording }
+      ?.let { view.setBackHandler(onBack = it) }
+      ?: view.setBackHandler(enabled = false) { }
```
rjrjr, line 49:
```suggestion
    view.setOrClearBackhandler(rendering.onResumeRecording.takeUnless { rendering.recording })
```

rjrjr: A lot of these.

samples/tictactoe/app/src/main/java/com/squareup/sample/authworkflow/LoginViewFactory.kt
```diff
@@ -15,5 +15,5 @@ internal val LoginViewFactory: ScreenViewFactory<LoginScreen> =
       rendering.onLogin(loginEmail.text.toString(), loginPassword.text.toString())
     }
 
-    root.backPressedHandler = { rendering.onCancel() }
+    root.setBackHandler(onBack = rendering.onCancel)
```
rjrjr, line 18: If you use `setOrClearBackHandler` then we don't need the `onBack =`. 🤔

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/ContentDialogSetContent.kt
```diff
@@ -63,8 +65,9 @@ public fun <C : Screen, O : ScreenOverlay<C>> ComponentDialog.setContent(
   // into ComponentDialog will be used. Our default implementation basically
   // duplicates that one.
   return OverlayDialogHolder(
-    environment,
-    this
+    initialEnvironment = environment,
+    dialog = this,
+    onBackPressed = null
```
rjrjr, line 70: Should pull `onBackPressed = null` into the upstream PR, important.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/ScreenOverlayDialogFactory.kt
```diff
@@ -59,7 +61,17 @@ public open class ScreenOverlayDialogFactory<S : Screen, O : ScreenOverlay<S>>(
...
     val contentViewHolder = initialRendering.content.toViewFactory(initialEnvironment)
-      .startShowing(initialRendering.content, initialEnvironment, context)
+      .startShowing(initialRendering.content, initialEnvironment, context) { view, doStart ->
+        // Note that we never call destroyOnDetach for this owner. That's okay because
+        // DialogSession.showNewDialog puts one in place above us on the decor view,
+        // and cleans it up. It's in place by the time we attach to the window, and
+        // so becomes our parent.
+        WorkflowLifecycleOwner.installOn(
+          view,
+          initialEnvironment.onBackPressedDispatcherOwner(view)
```
rjrjr, line 71: This will pull in the wrong OnBackPressedDispatcherOwner, but that was always the case. This problem is why we've deprecated `ScreenOverlayDialogFactory` and replaced it with `fun ComponentDialog.setContent()` -- it was fundamentally broken, requiring us to create the content view before the `Dialog` that would host it.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/BackButtonScreen.kt
```diff
@@ -28,24 +29,27 @@ public class BackButtonScreen<C : Screen>(
   public val shadow: Boolean = false,
   public val onBackPressed: (() -> Unit)? = null
 ) : Wrapper<Screen, C>, AndroidScreen<BackButtonScreen<C>> {
+  // If they change the shadow value, we need to build a new view to reorder the handlers.
+  override val compatibilityKey: String = keyFor(content, "BackButtonScreen+shadow:$shadow")
+
```
helios175, line 34: nice

rjrjr: I suspect you are the only other person on the planet able to appreciate that.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewBackHandler.kt
rjrjr, line 1: Should name this file `SetBackHandler.kt`

workflow-ui/core-android/src/androidTest/java/com/squareup/workflow1/ui/ViewBackHandlerTest.kt
```diff
@@ -0,0 +1,54 @@
...
+@OptIn(WorkflowUiExperimentalApi::class)
+internal class ViewBackHandlerTest {
+  private val scenarioRule = ActivityScenarioRule(ComponentActivity::class.java)
+  private val scenario get() = scenarioRule.scenario
+
+  @get:Rule val rules: RuleChain = RuleChain.outerRule(DetectLeaksAfterTestSuccess())
+    .around(scenarioRule)
+    .around(IdlingDispatcherRule)
+
+  private var viewHandlerCount = 0
```
steve-the-edwards, line 23: nit:

```suggestion
  private var viewHandlerBackPressedCount = 0
```

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/BackPressHandler.kt
```diff
@@ -13,6 +13,7 @@ import com.squareup.workflow1.ui.androidx.WorkflowAndroidXSupport.onBackPressedD
  * A function passed to [View.backPressedHandler], to be called if the back
  * button is pressed while that view is attached to a window.
  */
+@Deprecated("Use View.backHandler()")
```
steve-the-edwards, line 16:
```suggestion
@Deprecated("Use View.setBackHandler()")
```

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/BackPressHandler.kt
```diff
@@ -24,7 +25,9 @@ public typealias BackPressHandler = () -> Unit
  * That means that this is a last-registered-first-served mechanism, and that it is
  * compatible with Compose back button handling.
  */
+@Suppress("DEPRECATION")
 @WorkflowUiExperimentalApi
+@Deprecated("Use setOrClearBackHandler")
```
steve-the-edwards, line 30:
```suggestion
@Deprecated("Use setBackHandler")
```

I see you went through some naming iterations 😄

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewBackHandler.kt
```diff
@@ -0,0 +1,72 @@
+package com.squareup.workflow1.ui
+
+import android.view.View
+import androidx.activity.OnBackPressedCallback
+import androidx.lifecycle.ViewTreeLifecycleOwner
+import com.squareup.workflow1.ui.androidx.WorkflowAndroidXSupport.onBackPressedDispatcherOwnerOrNull
+
+/**
+ * A function to be called if the device back button is pressed while this
```
steve-the-edwards, line 9:
```suggestion
 * Sets a function to be called if the device back button is pressed while this
```

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewBackHandler.kt
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * @param onBack the function to run from [OnBackPressedCallback.handleOnBackPressed]
+ */
+@WorkflowUiExperimentalApi
+public fun View.setBackHandler(
+  enabled: Boolean = true,
+  onBack: () -> Unit
+) {
+  val callback = onBackPressedCallbackOrNull ?: MutableOnBackPressedCallback().apply {
+    onBackPressedCallbackOrNull = this
```
steve-the-edwards, line 26: well this is an interesting pattern. 🤔 .

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewBackHandler.kt
```diff
@@ -0,0 +1,72 @@
...
+ * back gesture is made. If null, the relevant [OnBackPressedCallback] will be disabled,
+ * but it will still exist -- this [View]'s priority in the
+ * [OnBackPressedDispatcher][androidx.activity.OnBackPressedDispatcher] queue
+ * will not change, should a non-null handler be provided by a later call.
+ */
+@WorkflowUiExperimentalApi
+public fun View.setBackHandler(onBack: (() -> Unit)?) {
+  onBack?.let { setBackHandler(enabled = true, it) }
+    ?: setBackHandler(enabled = false) {}
+}
```
steve-the-edwards, line 56: Do you think `setBackHandler(null)` is the most understandable way of clearing/disabling this?

I personally like `setBackHandler(enabled = false)` and then making `{}` the default value of `onBack` in the version of the function above (not needing this overload then). I could see how you could end up with `setBackHandler(false)` but even that I don't find confusing.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ViewBackHandler.kt
```diff
@@ -0,0 +1,72 @@
...
+
+@WorkflowUiExperimentalApi
+private var View.onBackPressedCallbackOrNull: MutableOnBackPressedCallback?
+  get() = getTag(R.id.view_back_handler) as MutableOnBackPressedCallback?
+  set(value) {
+    setTag(R.id.view_back_handler, value)
+  }
```
steve-the-edwards, line 63: You truly are an Android view maestro how you nicely abstract this. 👨🏻‍🍳 💋

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/ScreenOverlayDialogFactory.kt
```diff
@@ -39,12 +41,12 @@ public open class ScreenOverlayDialogFactory<S : Screen, O : ScreenOverlay<S>>(
...
 
     return OverlayDialogHolder(initialEnvironment, dialog) { overlayRendering, environment ->
-      // For a modal, on each update put a no-op backPressedHandler in place on the
+      // For a modal, on each update put a no-op backHandler in place on the
       // decorView before updating, to ensure that the global androidx
       // OnBackPressedDispatcher doesn't fire any set by lower layers. We put this
       // in place before each call to show(), so the real content view will be able
       // to clobber it.
-      if (modal) content.view.backPressedHandler = {}
+      if (modal) content.view.setBackHandler {}
```
steve-the-edwards, line 49: Given all the no-op handlers we use, having a default of `{}` might be nice. Although maybe that's just in these internal use cases?

steve-the-edwards (approved):

PR #1031, opened 2023-06-15
4/4 `asDialogHolderWithContent`, `ScreenViewHolder.startShowing`

rjrjr:
Hide whitespace while reviewing this, especially in `ScreenViewFactory.kt`.

1. `setContent` as a method name gave no clue that we were returning a new manager object, so now it's called `asDialogHolderWithContent`

2. Hard coding the call to `Dialog.setContent` interferes with code that needs to make that call itself. We now accept a `setContent: (ScreenViewHolder<C>) -> Unit` lambda to do that work, which defaults to the current behavior.

3. Decouple building and starting views. This allows us to ensure that a Dialog's content view is attached to its window before its first rendering -- before the first call to `ScreenViewHolder.show`. Necessary to ensure that the strictly view-tree-based AndroidX `findViewTreeOnBackPressedDispatcherOwner()` call works immediately. This was always the original intention, but never actually found a use case until now.

4. Moves most of that default behavior to a public `Dialog.fixBackgroundAndDimming` extension function to keep it available and more self documenting (hard learned lessons in there). Pretty similar to how `ScreenOverlayDialogFactory` acts.

5. Big kdoc update


steve-the-edwards (approved):

steve-the-edwards:
Btw - I don't remember seeing naming update in the deprecation warning strings. Need to add that.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/AsDialogHolderWithContent.kt
```diff
@@ -22,15 +24,22 @@ import com.squareup.workflow1.ui.toViewFactory
...
   overlay: O,
-  environment: ViewEnvironment
+  environment: ViewEnvironment,
+  turnOffChrome: Boolean = true,
+  setContent: (ComponentDialog, ScreenViewHolder<C>) -> Unit = { d, v -> d.setContentView(v.view) }
 ): OverlayDialogHolder<O> {
   // Note that we always tell Android to make the window non-modal, regardless of our own
   // notion of its modality. Even a modal dialog should only block events within
-  // the appropriate bounds, but Android makes them block everywhere.
+  // the appropriate bounds, but Android makes them block everywhere by default.
```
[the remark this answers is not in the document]
rjrjr: It's not impossible, just a bit inconvenient. e.g. they can always turn the flag back on in their custom `setContentView` method. 

I definitely want to minimize the chance of clobbering this bit by accident. Without it we lose the main promise of `BodyAndOverlaysScreen`, that it will restrict the windows to the area of the body view.

It occurs to me that `toggleChrome` and `setContent` are redundant. I'm inclined to get rid of parameter, make the function public, and have the default `setContent` implementation call that. That's more like how the deprecated `abstract class ScreenOverlayDialogFactory` works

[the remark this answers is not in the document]
rjrjr: Renamed `Dialog.turnOffChrome()` function to `Dialog.fixBackgroundAndDimming().

rjrjr:
@steve-the-edwards Good morning! I've smoked and canaried the heck out of this, and it's ready to go. The three upstream PRs haven't changed since you reviewed them, but this one could definitely use a new review.

Be sure to check _ignore whitespace_!

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/ScreenViewFactory.kt
```diff
@@ -270,17 +282,9 @@ public fun <ScreenT : Screen> ScreenT.toViewFactory(
...
- * instances to [WorkflowViewStub.show], which will apply the [ScreenViewFactory] machinery for you.
- *
- * Creates a [ScreenViewHolder] wrapping a [View] able to display a stream of [ScreenT] renderings,
- * starting with [initialRendering].
- *
- * To add more initialization behavior (typically a call to
- * [WorkflowLifecycleOwner.installOn][com.squareup.workflow1.ui.androidx.WorkflowLifecycleOwner.installOn]),
- * provide a [viewStarter].
+ * Convenience that combines [ScreenViewFactory.buildView] and [ScreenViewHolder.startShowing],
+ * since we rarely need to do work between those two calls.
```
steve-the-edwards, line 286: Nit: Probably could move this below the main function below.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/container/OverlayDialogFactory.kt
```diff
@@ -7,40 +7,41 @@ import com.squareup.workflow1.ui.ViewRegistry
...
+ * Factory for [Dialog] instances that can show renderings of type [OverlayT].
+ * Most easily implemented using the [ScreenOverlay] interface to define the
+ * [Dialog] `contentView`, and calling the [asDialogHolderWithContent]
+ * extension function to set it.
  *
- * To minimize boilerplate, have your rendering classes implement [AndroidOverlay] to associate
- * them with appropriate an appropriate [OverlayDialogFactory]. For more flexibility, and to
- * avoid coupling your workflow directly to the Android runtime, see [ViewRegistry].
+ * To really minimize boilerplate, have your rendering classes implement [AndroidOverlay] to
+ * directly associate them with appropriate an appropriate [OverlayDialogFactory].
```
steve-the-edwards, line 16:
```suggestion
 * directly associate them with an appropriate [OverlayDialogFactory].
```

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/container/BodyAndOverlaysScreen.kt
```diff
@@ -13,6 +13,61 @@ import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
...
+ * And we want to ensure that any modal windows do not obscure the tutorial, if
+ * it's showing:
+ *
+ *    +----+=============+------+
+ *    |  My|             |      |
+ *    |    | MyEditModal |      |
+ *    |    |             |      |
+ *    +----+=============+------+
+ *    | MyTutorialScreen        |
+ *    +-------------------------+
```
steve-the-edwards, line 37: 😮 . ASCII diagram skills. 🔥

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/container/BodyAndOverlaysScreen.kt
```diff
@@ -13,6 +13,61 @@ import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
...
+ *         bar = tutorialScreenOrNull,
+ *       )
+ *     )
+ *
+ * Whatever structure you settle on for your root rendering, it is important
+ * to render the same structure every time. If your app will ever want to show
+ * an [Overlay], it should always render [BodyAndOverlaysScreen], even when
+ * there is no [Overlay] to show. Otherwise your entire view tree will be rebuilt,
+ * since the view built for a `MyBodyAndBottomBarScreen` cannot be updated to show
+ * a [BodyAndOverlaysScreen] rendering.
```
steve-the-edwards, line 69: Nice recent addition :)

steve-the-edwards (approved):
LGTM! 2 nits/typos.

PR #1035, opened 2023-06-20
Update Turbine to 0.13.0

steve-the-edwards:
Update Turbine to 0.13.0; Update WorkerTester

assertNoOutput now fails if there is a completion of the WOrker's flow.

### DIFF
--- a/workflow-testing/src/test/java/com/squareup/workflow1/testing/WorkerTesterTest.kt
+++ b/workflow-testing/src/test/java/com/squareup/workflow1/testing/WorkerTesterTest.kt
@@ -15,11 +15,26 @@ import kotlin.test.assertFalse
 @OptIn(ExperimentalCoroutinesApi::class)
 class WorkerTesterTest {
 
-  @Test fun `assertNoOutput passes after worker finishes without emitting`() {
-    val worker = Worker.finished<Unit>()
+  @Test fun `assertNoOutput succeeds in live flow without output`() {
+    val worker = Worker.create<Unit> {
+      suspendCancellableCoroutine { }
+    }
     worker.test {
       assertNoOutput()
     }
+  }
+
+  @Test fun `assertNoOutput fails after worker finishes without emitting`() {
+    val worker = Worker.finished<Unit>()
+    val error = assertFailsWith<AssertionError> {
+      worker.test {
+        assertNoOutput()
+      }
+    }
+    assertEquals(
+      expected = "Expected no output, completion, or error to have been emitted.",
+      actual = error.message
+    )
   }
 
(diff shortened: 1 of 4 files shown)


RBusarow (approved):

workflow-testing/src/test/java/com/squareup/workflow1/testing/WorkerTesterTest.kt
```diff
@@ -15,13 +15,28 @@ import kotlin.test.assertFalse
...
+  @Test fun `assertNoOutput succeeds in live flow without output`() {
+    val worker = Worker.create<Unit> {
+      suspendCancellableCoroutine { }
+    }
     worker.test {
       assertNoOutput()
     }
   }
 
+  @Test fun `assertNoOutput fails after worker finishes without emitting`() {
```
RotBolt, line 27: if worker finishes without emitting anything, then `assertNoOutput` should succeed right? as the name suggest it should check for no output

rjrjr: @steve-the-edwards I'm confused by this too.

rjrjr: I'm confused by two things: I think @RotBolt is right that the method name is wrong; and I'm unclear why we want this semantic.

Why is it an error for a worker to complete when we `assertNoOutput()`? This change broke some existing tests that were building workers via `Observable.just()`, for example.

steve-the-edwards: I think you guys are right I should have made more noise about the change in this semantic. I just took the easy way out and inherited the change from `expectNoEvents()` when it started throwing on Completion after Turbine 0.3.0.

I've done a quick look and I'm not sure there is good way to reproduce the behavior we had before. @RotBolt could you create an issue to update this or rename the assertion to something like `assertNoOutputOrCompletion()`?

I think I am ok with having places where we use flowOf { 1 } or Observable.just(false) to start using `assertFinished()` instead of `assertNoOutput()`. But maybe that's a cop out?

rjrjr: It's easy to replace `Observable.just(false)` with `BehaviorRelay.createDefault(false)`. Is there a similar idiom for `Flow`?

steve-the-edwards:
```
flow { emit(false); suspendCancellableCoroutine {} }
```

so tidy right? 

hmm. must be something better

PR #1090, opened 2023-07-12
Update AndroidX Dependencies

steve-the-edwards:
And make corresponding changes due to API updates.


RBusarow (approved):

rjrjr (approved):
Thanks Steve! I was kind of dreading this, thought it would be more work than this.

PR #1091, opened 2023-07-13
Cache the AVD in CI before executing any tests

RBusarow:
background: https://workflow-community.slack.com/archives/CHTFPR277/p1689193974005269

I was using `actions/cache@v3` to restore and possibly save the AVD in instrumented test shards.

That basic "cache" action handles both reading and writing to the cache.  The writing is done during a post-action, meaning that if the AVD was cached, it was cached *after* it had been used for the instrumented tests.

Now, we use the newer `actions/cache/restore@v3` and `actions/cache/save@v3` actions, like we use elsewhere.  Now, if there was a cache miss and we have to make a new AVD, it is cached before the tests are executed so that it doesn't have any of our test apps installed.

### DIFF
--- a/.github/actions/gradle-tasks-with-emulator/action.yml
+++ b/.github/actions/gradle-tasks-with-emulator/action.yml
@@ -1,13 +1,13 @@
 name : Run Android Instrumentation Tests with Artifact and AVD Caching
-description: This action sets up Gradle, runs a preparatory task, runs Android tests on an emulator, and uploads test results.
+description : This action sets up Gradle, runs a preparatory task, runs Android tests on an emulator, and uploads test results.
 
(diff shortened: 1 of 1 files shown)


rjrjr:
Be sure to hide whitespace when reviewing this!

rjrjr (approved):

RBusarow:
> Be sure to hide whitespace when reviewing this!

Yeah, sorry about that.  This is the style set by `.editorconfig`.

rjrjr:
> > Be sure to hide whitespace when reviewing this!
> 
> Yeah, sorry about that. This is the style set by `.editorconfig`.

Well worth it.

PR #1092, opened 2023-07-18
Introduces `fun Screen.deepMap(Screen): Screen`

rjrjr:
Introduces and uses `Screen`-specific subtypes of the `Container` and `Wrapper` interfaces. This allows the inroduction of `Screen.deepMap()`, which allows us to apply a transformation to the "real" `Screen`s collected in a `Container`, no matter how deeply wrapped they are.


workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/VisualHolder.kt
```diff
@@ -0,0 +1,24 @@
+package com.squareup.workflow1.ui
+
+@WorkflowUiExperimentalApi
+public interface VisualHolder<in RenderingT, out VisualT> {
```
RBusarow, line 4: It seems like some of the kdoc for `VisualFactory.createOrNull` should be here as well.

rjrjr: Uh oh. I think this and `VisualHolder` leaked in from an unrelated PR.

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/ScreenContainer.kt
```diff
@@ -0,0 +1,34 @@
...
+ *    val backStack = BackStackScreen(SomeWrapper(theRealScreen))
+ *    val loggingBackStack = backStack.deepMap { WithLogging(it) }
+ *
+ * `loggingBackStack` will have a structure like so:
+ *
+ *    BackStackScreen(SomeWrapper(WithLogging(theRealScreen)))
+ */
+@WorkflowUiExperimentalApi
+public fun Screen.deepMap(transform: (Screen) -> Screen): Screen {
+  return if (this is ScreenContainer<*>) {
```
rjrjr, line 29: You know what? I don't know that `ScreenContainer` is actually necessary for this. I'm going to put this PR back into draft.

- [ ] `ScreenContainer` probably isn't necessary.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/UiFactory.kt
```diff
@@ -0,0 +1,24 @@
...
+   *
+   * @param getFactory can be used to make recursive calls to build VisualT
+   * instances for sub-parts of [rendering]
+   */
+  public fun createOrNull(
+    rendering: RenderingT,
+    context: ContextT,
+    environment: ViewEnvironment,
+    getFactory: (ViewEnvironment) -> VisualFactory<ContextT, Any, VisualT>
+  ): VisualHolder<RenderingT, VisualT>?
```
rjrjr, line 23: - [ ] Unused, did this leak in from the work for #546?

PR #1096, opened 2023-08-03
1093: Yield in runtime to run side effects that were launched lazily

### LINKED ISSUE #1093: Side effects / Workers / LifecycleWorker methods not called if rendered/not rendered without ever yielding dispatch
since worker is built on `runningSideEffect`, the easier way to reproduce this is:

```
 runningSideEffect("test") {
   try {
   } finally {
     error()
   }
}
```

If this is rendered and not rendered within the same thread frame (e.g. using the Main.Immediate dispatcher and cycling back and forth), `error()` will never get called.

At the very least this needs to be updated in the documentation.

This is because the Job's for side effects are started using `CoroutineStart.LAZY` so that we don't end up sending anything to the `actionSink` until after the `render()` method is complete and the `RenderContext` frozen (then we call `start()` on all the side effect jobs - see [here](https://github.com/square/workflow-kotlin/blob/551ee52b3153667113ba1cadc504b710813fd207/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt#L217).

```
    baseRenderContext.unfreeze()
    val rendering = interceptor.intercept(workflow, this)
      .render(props, state, context)
    baseRenderContext.freeze()

    // Side effect jobs are launched lazily, since they can send actions to the sink, and can only
    // be started after context is frozen.
    sideEffects.forEachStaging { it.job.start() }
    sideEffects.commitStaging { it.job.cancel() }
```

What would be better if we provided a hook to the side effect's Job's `invokeOnCompletion` which will be invoked even if cancelled before the coroutine gets dispatched.  E.g. see this test:

```
fun main() = runBlocking {

    val job = launch(start = CoroutineStart.LAZY) {
        try {
            repeat(1000) { i ->
                println("job: I'm sleeping $i ...")
                delay(500L)
            }
        } finally {
            println("I'm still run in the finally block.")
        }
    }
    job.invokeOnCompletion { cause ->
        println("I'm complete! $cause")
    }
    job.start()
    //yield()
    job.cancel()
    println("done test.")
}
```

Which prints "I'm complete" but not what is in the finally block (without the `yield()`).

I propose this could be called.

`RenderContext.onNoLongerRendered()`. It has to be outside of `runningSideEffect`'s lambda because that won't ever get invoked. We want to attach this right on the `Job` we create for the side effect.

steve-the-edwards:
Add in some `yield()` in the runtime so that we can start any running side effect after each render() before processing the next action, which might tear down that side effect's Job.

Side Effect Job's are launched lazily in case they send to teh actionSink (so we can complete the render before they get dispatched).

Also brings in `headlessIntegrationTest` which was being used internally for some time.

Brings this in in order to right some tests for whether or not these side effects are run under certain circumstances.

Closes #1093


workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
```diff
@@ -204,6 +207,8 @@ public fun <PropsT, OutputT, RenderingT> renderWorkflowIn(
       renderingsAndSnapshots.value = nextRenderAndSnapshot
       // And emit the Output.
       sendOutput(actionResult, onOutput)
+      // Yield if there are any side effects (they were launched lazily) that need starting.
+      yield()
```
rjrjr, line 211: Are you sure we shouldn't `yield()` first, before we allow the UI to update?

steve-the-edwards: in order to get the side effects started before then you mean? Or because we don't want the collectors of the StateFlow of renderings to be in contention for dispatch at the same time?

rjrjr: The former was what I was thinking of -- seems like a side effect should happen, or start happening, before / while the UI is updating.

rjrjr: Thought exercise: what happens if something on the receiving end of `sendOutput` kills off the workflow runtime before that `yield()` happens (if it that's possible). Would we be surprised that the last declared side effects were never run? I think we would.

steve-the-edwards: all legitimate points. I adjusted the ordering in the updated PR.

steve-the-edwards:
This isn't desirable as it creates unnecessary thrash after the `render()` pass.

steve-the-edwards:
Closing and this is not needed. It doesn't work and we went with #1106 instead.

PR #1099, opened 2023-08-07
Add PetFinder Module for database example

### LINKED ISSUE #769: Write database-backed example for kotlin
We need example of driving long, random access list from a database —  RecyclerView pulling from Rooms or SqlDelight

RotBolt:
Will close #769


RotBolt:
Currently its in DRAFT mode. I will make incremental updates and commits and squash the unnecessary ones if required

PR #1102, opened 2023-08-09
1093: Launch Side Effects atomically so they are always started

steve-the-edwards:
Add headlessIntegrationTest to renderWorkflowIn with a nice Turbine attached.

Check that we do not use the Unconfined dispatcher within the Workflow runtime as it does not provide ordering guarantees. Coroutines are resumed on whatever thread was last suspended. This means we could start sending actions within a side effect into a unfrozen RenderContext as we cannot guarantee the ordering of handling side effects.

Update documentation and tests to note the need to use a Dispatcher other than Dispatchers.Unconfined.

### DIFF
--- a/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
+++ b/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
@@ -19,7 +19,7 @@ import com.squareup.workflow1.internal.RealRenderContext.SideEffectRunner
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineName
 import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.CoroutineStart.LAZY
+import kotlinx.coroutines.CoroutineStart.ATOMIC
 import kotlinx.coroutines.DelicateCoroutinesApi
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.Job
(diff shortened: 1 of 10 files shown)


workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
```diff
@@ -260,7 +261,13 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
     sideEffect: suspend CoroutineScope.() -> Unit
   ): SideEffectNode {
     val scope = this + CoroutineName("sideEffect[$key] for $id")
-    val job = scope.launch(start = LAZY, block = sideEffect)
+    // Side effect jobs are automatic because even if the side effect is run, then not run
```
rjrjr, line 264:
```suggestion
    // Side effect jobs are ATOMIC because even if the side effect is run, then not run
```

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
```diff
@@ -543,35 +485,6 @@ class RenderWorkflowInTest {
     }
   }
 
-  @Test
-  fun `side_effects_from_initial_rendering_in_non_root_workflow_are_never_started_when_initial_render_of_non_root_workflow_fails`() {
```
rjrjr, line 547: Should we repurpose these tests to confirm that they _are_ run?

steve-the-edwards:
This doesn't work because while the wrapped coroutine will be started, the block declared and passed to `runningSideEffect` will never be called.

Issue #990, opened 2023-04-13
Consider version of `initialState` that provides a `CoroutineScope` identical to the `WorkflowNode`'s


steve-the-edwards:
Then for example, a StateFlow operator could be launched from onInitialState under a scope identical to the workflow's scope.

steve-the-edwards:
Thinking about this more...

we have had 2 main use cases for a StateFlow :

1. So that we could get `.value` out of the `StateFlow` in `initialState` to avoid one extra bootstrap re-rendering (for performance reasons). If we are going to do that `stateIn` in `initialState` then it either won't be synchronous (the suspending version) or we already have the correct `initialValue` synchronously (as props or otherwise), in which case we don't really need a `StateFlow` from `initialState` onwards. We can just populate the Workflow's state with the initial value and use the `flow` form a `Worker` in `render()`.
1. The 2nd case is if we have operators that we want to do on one or more upstream StateFlows. Given any operator has work to do and brings this back to a `flow`, we need to `stateIn` after the operator in order to get it back to a `StateFlow` to access `.value` synchronously in `initialState`. The operator combination is likely a concern of the Workflow usage only, so we want to be able to model the code for it there. It would be better to be able to do the operators and then a `stateIn` attached to the Workflow's lifecycle itself (thus going back to providing the `CoroutineScope` in `initialState`). However, as I think through this now I'm not sure this case doesn't just collapse into the first. If we have all upstream `StateFlow` signals we can `.value` them all already in `initialState` and then run the operator/combination logic synchronously (it better be fast) in `initialState` to get a value to populate our state without extra re-renders. Then we're happy enough if the operators result in a `flow` to collect from in a `Worker`.

So now I've convinced myself that I don't see a case where we _need_ `CoroutineScope` in `initialState` to accomplish something we couldn't before. It might be a useful convenience, but I need to think about whether or not we want to enable that convenience vs. encouraging doing the above steps. 🤔

rjrjr:
Case one is less interesting now that #992 has landed, but this still seems like it would solve a lot of other problems.

steve-the-edwards:
> still seems like it would solve a lot of other problems.

In terms of discoverability and having good tools 'at hand' you mean?

I feel like I've ruled out what I thought before were the main use cases in terms of being blocked without this.

rjrjr:
You have a much clearer picture of the situation than I do. Let's definitely not build this until we have real people with real use cases in hand.

blakelee:
[...]

PR #1104, opened 2023-08-31
nodeCreated Method and Piping

steve-the-edwards:
For #1096 and #990


steve-the-edwards:
Not needed as this goes too far with `WorkflowLocal` for now. Instead see #1106

PR #1105, opened 2023-09-01
Add HeadlessIntegrationTest to Workflow

steve-the-edwards:
This has been used internally at Square for some time now.

### DIFF
--- a/workflow-testing/src/main/java/com/squareup/workflow1/testing/HeadlessIntegrationTest.kt
+++ b/workflow-testing/src/main/java/com/squareup/workflow1/testing/HeadlessIntegrationTest.kt
@@ -1,0 +1,189 @@
+package com.squareup.workflow1.testing
+
+import app.cash.turbine.ReceiveTurbine
+import app.cash.turbine.test
+import com.squareup.workflow1.RuntimeConfig
+import com.squareup.workflow1.RuntimeConfigOptions
+import com.squareup.workflow1.Workflow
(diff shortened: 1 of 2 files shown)


workflow-testing/src/main/java/com/squareup/workflow1/testing/HeadlessIntegrationTest.kt
```diff
@@ -0,0 +1,189 @@
...
+import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.flow.drop
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.test.UnconfinedTestDispatcher
+import kotlinx.coroutines.test.runTest
+import kotlin.coroutines.CoroutineContext
+import kotlin.time.Duration.Companion.milliseconds
+
+/**
+ * This is a test harness to run integration tests for a Workflow tree. The parameters passed here are
```
rjrjr, line 25: If we had had this built alreay, do you think we would have bothered with the `launchForTestingFrom*` machinery?

rjrjr (approved):

workflow-testing/src/main/java/com/squareup/workflow1/testing/HeadlessIntegrationTest.kt
steve-the-edwards: No.

steve-the-edwards:
Not sure why CI keeps failing. Abandoning this commit.

PR #1106, opened 2023-09-01
Add CoroutineScope to initialState; SessionWorkflow to aid rollout

### LINKED ISSUE #990: Consider version of `initialState` that provides a `CoroutineScope` identical to the `WorkflowNode`'s

### LINKED ISSUE #1093: Side effects / Workers / LifecycleWorker methods not called if rendered/not rendered without ever yielding dispatch

steve-the-edwards:
Adds 'Lifecycle' related tests to prove out SessionWorkflow's ability to effectively track lifecycle in the failing scenario. Adds comments describing the failing scenario of issue #1093, and how this can workaround it to fix it.

Also address #990 which has been separately requested.

Also adds `HeadlessIntegrationTest` to the OSS Library, which we have used internally for some time.

Closes #990
Closes #1093

### DIFF
--- a/workflow-core/src/commonMain/kotlin/com/squareup/workflow1/BaseRenderContext.kt
+++ b/workflow-core/src/commonMain/kotlin/com/squareup/workflow1/BaseRenderContext.kt
@@ -95,6 +95,23 @@ public interface BaseRenderContext<out PropsT, StateT, in OutputT> {
    * that the workflow runtime is running in. The side effect coroutine will not be started until
    * _after_ the first render call than runs it returns.
    *
+   * Note that there is currently an [issue](https://github.com/square/workflow-kotlin/issues/1093)
+   * when a [runningSideEffect] (and thus also [runningWorker], or the parent Workflow of either
+   * via [renderChild]) is declared as running (or rendering) in one render pass and
+   * then not declared in the next render pass and both those consecutive render passes happen
+   * synchronously - i.e. without the [CoroutineDispatcher][kotlinx.coroutines.CoroutineDispatcher]
+   * for the Workflow runtime being able to dispatch asynchronously. This is because the jobs for
+   * side effects are launched lazily in order to ensure they happen after the render pass, but if
+   * the [CoroutineScope]'s job (the parent for all these jobs) is cancelled before these lazy
+   * coroutines have a chance to dispatch, then they will never run at all. For more details, and
+   * to report problems with this, see the [issue](https://github.com/square/workflow-kotlin/issues/1093).
+   * If you need guaranteed execution for some code in this scenario (like for cleanup),
+   * please use a [SessionWorkflow] and the [SessionWorkflow.initialState] that provides the
+   * [CoroutineScope] which is equivalent to the lifetime of the Workflow node in the tree. The
+   * [Job][kotlinx.coroutines.Job] can be extracted from that and used to get guaranteed to be
+   * executed lifecycle hooks, e.g. via [Job.invokeOnCompletion][kotlinx.coroutines.Job.invokeOnCompletion].
+   *
+   *
    * @param key The string key that is used to distinguish between side effects.
    * @param sideEffect The suspend function that will be launched in a coroutine to perform the
    * side effect.
@@ -280,6 +297,10 @@ public inline fun <reified W : Worker<Nothing>, PropsT, StateT, OutputT>
  * workers are compared by their _declared_ type, not their actual type. This means that if you
  * pass a worker stored in a variable to this function, the type that will be used to compare the
  * worker will be the type of the variable, not the type of the object the variable refers to.
+ *
+ * Note that there is currently an [issue](https://github.com/square/workflow-kotlin/issues/1093)
+ * which can effect whether a [Worker] is ever executed.
+ * See more details at [BaseRenderContext.runningSideEffect].
  *
  * @param key An optional string key that is used to distinguish between identical [Worker]s.
  */
--- a/workflow-core/src/commonMain/kotlin/com/squareup/workflow1/LifecycleWorker.kt
+++ b/workflow-core/src/commonMain/kotlin/com/squareup/workflow1/LifecycleWorker.kt
@@ -14,6 +14,10 @@ import kotlin.jvm.JvmName
  *
  * A [Worker] is stopped when its parent [Workflow] finishes a render pass without running the
  * worker, or when the parent workflow is itself torn down.
+ *
+ * Note that there is currently an [issue](https://github.com/square/workflow-kotlin/issues/1093)
+ * which can effect whether a [LifecycleWorker] is ever executed.
+ * See more details at [BaseRenderContext.runningSideEffect].
  */
 public abstract class LifecycleWorker : Worker<Nothing> {
 
--- a/workflow-core/src/commonMain/kotlin/com/squareup/workflow1/SessionWorkflow.kt
+++ b/workflow-core/src/commonMain/kotlin/com/squareup/workflow1/SessionWorkflow.kt
@@ -1,0 +1,175 @@
+package com.squareup.workflow1
+
+import kotlinx.coroutines.CoroutineScope
+
+/**
+ * An extension of [StatefulWorkflow] that gives [initialState] a [CoroutineScope]
+ * that corresponds with the lifetime of _session_ driven by this Workflow.
+ *
+ * A session begins the first time a parent passes a child [Workflow] of a particular type to
+ * [renderChild] with a particular [key] parameter. It ends when the parent executes [render]
+ * without making a matching [renderChild] call. The [CoroutineScope] that is passed to
+ * [initialState] is created when a session starts (when [renderChild] is first called), and
+ * [cancelled][kotlinx.coroutines.Job.cancel] when the session ends.
+ *
+ * This API extension exists on [StatefulWorkflow] as well, but it is confusing because the version
+ * of [initialState] that does not have the [CoroutineScope] must also be implemented as it is
+ * an abstract fun, even though it would never be used.
+ * With this version, that confusion is removed and only the version of [initialState] with the
+ * [CoroutineScope] must be implemented.
+ */
+@WorkflowExperimentalApi
+public abstract class SessionWorkflow<
+  in PropsT,
+  StateT,
+  out OutputT,
+  out RenderingT
+  > : StatefulWorkflow<PropsT, StateT, OutputT, RenderingT>() {
+
+  /**
+   * @see [StatefulWorkflow.initialState] for kdoc on the base function of this method.
+   *
+   * This version adds the following:
+   * @param workflowScope A [CoroutineScope] that has been created when this Workflow is first
+   * rendered and canceled when it is no longer rendered.
+   *
+   * This [CoroutineScope] can be used to:
+   *
+   *  - set reliable teardown hooks, e.g. via [Job.invokeOnCompletion][kotlinx.coroutines.Job.invokeOnCompletion].
+   *
+   *  -  own the transforms on a [StateFlow][kotlinx.coroutines.flow.StateFlow],
+   *     linking them to the lifetime of a Workflow session. For example,
+   *     here is how you might safely combine two `StateFlow`s:
+   *
+   *     data class MyState(
+   *       val derivedValue: String,
+   *       val derivedWorker: Worker<String>
+   *     )
+   *
+   *     override fun initialState(
+   *       props: Unit,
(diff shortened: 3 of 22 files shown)


rjrjr:
I love this.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/BaseRenderContext.kt
```diff
@@ -95,6 +95,22 @@ public interface BaseRenderContext<out PropsT, StateT, in OutputT> {
...
+   * when a [runningSideEffect] (and thus also [runningWorker], or the parent Workflow of either
+   * via [renderChild]) is declared as running (or rendering) in one render pass and
+   * then not declared in the next render pass and both those consecutive render passes happen
+   * synchronously - i.e. without the [CoroutineDispatcher][kotlinx.coroutines.CoroutineDispatcher]
+   * for the Workflow runtime being able to dispatch asynchronously. This is because the jobs for
+   * side effects are launched lazily in order to ensure they happen after the render pass, but if
+   * the [CoroutineScope]'s job (the parent for all these jobs) is cancelled before these lazy
+   * coroutines have a chance to dispatch, then they will never run at all. For more details, and
+   * to report problems with this, see the [issue](https://github.com/square/workflow-kotlin/issues/1093).
+   * If you need guaranteed execution for some code in this scenario (like for cleanup), please
```
rjrjr, line 108:
```suggestion
   *
   * If you need guaranteed execution for some code in this scenario (like for cleanup), please
```

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/BaseRenderContext.kt
```diff
@@ -95,6 +95,22 @@ public interface BaseRenderContext<out PropsT, StateT, in OutputT> {
...
+   * use a [SessionWorkflow] and the [SessionWorkflow.initialState] that provides the
+   * [CoroutineScope] which is equivalent to the lifetime of the Workflow node in the tree. The
+   * [Job][kotlinx.coroutines.Job] can be extracted from that and used to get guaranteed to be
+   * executed lifecycle hooks.
```
rjrjr, line 112:
```suggestion
   * executed lifecycle hooks, e.g. via
   * [Job.disposeOnCompletion][kotlinx.coroutines.disposeOnCompletion].
```

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/BaseRenderContext.kt
```diff
@@ -281,6 +297,9 @@ public inline fun <reified W : Worker<Nothing>, PropsT, StateT, OutputT>
  * pass a worker stored in a variable to this function, the type that will be used to compare the
  * worker will be the type of the variable, not the type of the object the variable refers to.
  *
+ * Note that there is currently an [issue](https://github.com/square/workflow-kotlin/issues/1093)
+ * which can effect [runningWorker]. See more details at [BaseRenderContext.runningSideEffect].
```
rjrjr, line 301:
```suggestion
 * which can whether a Worker is ever executed. See more details at [BaseRenderContext.runningSideEffect].
```

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/LifecycleWorker.kt
```diff
@@ -14,6 +14,9 @@ import kotlin.jvm.JvmName
  *
  * A [Worker] is stopped when its parent [Workflow] finishes a render pass without running the
  * worker, or when the parent workflow is itself torn down.
+ *
+ * Note that there is currently an [issue](https://github.com/square/workflow-kotlin/issues/1093)
+ * which can effect a [LifecycleWorker]. See more details at [BaseRenderContext.runningSideEffect].
```
rjrjr, line 19:
```suggestion
 * which can effect whether a [LifecycleWorker] is ever executed.
 * See more details at [BaseRenderContext.runningSideEffect].
```

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/SessionWorkflow.kt
```diff
@@ -0,0 +1,154 @@
...
+
+import kotlinx.coroutines.CoroutineScope
+
+/**
+ * @see [StatefulWorkflow] for the main description of this class.
+ *
+ * This is an extension of the API that allows for [initialState] to also receive a [CoroutineScope]
+ * that corresponds with the lifetime of this Workflow. In other words, the [CoroutineScope] is
+ * created when an session of this Workflow is active. A session is active from the first time one
+ * node of it is rendered, until it is no longer rendered.
```
rjrjr, line 11:
```suggestion
 * An extension of [StatefulWorkflow] that gives [initialState] to a [CoroutineScope]
 * that corresponds with the lifetime of a _session_ driven by this [Workflow].
 * 
 * A session begins the first time a parent passes a child [Workflow] of a particular
 * type to [renderChild] with a particular [key] parameter. It ends when the parent executes
 * [render] without making a matching [renderChild] call. The [CoroutineScope]
 * that is passed to [initialState] is created when a session starts (when [renderChild]
 * is first called), and [cancelled][Job.cancel] when the session ends. 
```

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/StatefulWorkflow.kt
```diff
@@ -93,6 +92,18 @@ public abstract class StatefulWorkflow<
     snapshot: Snapshot?
   ): StateT
 
+  /**
+   * @see [SessionWorkflow.initialState].
```
rjrjr, line 96: Do we really need to add this? Can the caller instead do an `is SessionWorkflow` check? If so, remember to delete the last paragraph of the `SessionWorkflow` class kdoc.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/SessionWorkflow.kt
```diff
@@ -0,0 +1,154 @@
...
+   * @param workflowScope A [CoroutineScope] that has been created when this Workflow is first
+   * rendered and canceled when it is no longer rendered.
+   *
+   * This [CoroutineScope] be used to own the transforms on, e.g. a [StateFlow][kotlinx.coroutines.flow.StateFlow],
+   * in order to link them to the lifetime of this Workflow's session - from first rendering until
+   * no longer rendered. e.g.,
+   * ```
+   * val transformedStateFlow = myStateFlow1.combine(myStateFlow2).
+   *   stateIn(workflowScope, SharingStarted.Eagerly, myStateFlow1.value + myStateFlow2.value)
+   * ```
```
rjrjr, line 40: We usually indent code blocks instead of relying on triple backtick. Same formatting. 

```suggestion
   * This [CoroutineScope] be used to:
   *
   *  - build reliable set up and teardown hooks, e.g. via
   *    [Job.disposeOnCompletion][kotlinx.coroutines.disposeOnCompletion]
   *
   *  -  own the transforms on a [StateFlow][kotlinx.coroutines.flow.StateFlow],
   *     linking them to the lifetime of a Workflow's session. For example,
   *     here is how you might safely combine two `StateFlow`s:
   *
   *     data class MyState(
   *       val derivedValue: String,
   *       val derivedWorker: Worker<String>
   *     )
   *
   *     override fun initialState(
   *       props: Unit,
   *       snapshot: Snapshot?,
   *       workflowScope: CoroutineScope
   *     ): MyState {
   *       val transformedStateFlow = stateFlow1.combine(stateFlow2).
   *         stateIn(workflowScope, SharingStarted.Eagerly, "${stateFlow1.value}-${stateFlow2.value}")
   *
   *       return MyState(
   *         transformedStateFlow.value,
   *         transformedStateFlow.asWorker()
   *       )
   *     }
```

workflow-testing/src/test/java/com/squareup/workflow1/ParameterizedTestRunner.kt
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * This file is copied from workflow-runtime:commonTest so our tests that test across the runtime
+ * look consistent. We could have used a JUnit library like Jupiter, but didn't.
+ *
+ * This file is copied so as to avoid creating a workflow-core-testing module (for now).
+ *
+ * We do our best to tell you what the parameter was when the failure occured by wrapping
+ * assertions from kotlin.test and injecting our own message.
+ */
+class ParameterizedTestRunner<P : Any> {
```
rjrjr, line 20: Should we slap the experimental annotation here?

rjrjr (approved):

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/BaseRenderContext.kt
rjrjr: (Just suggesting adding a linebreak.)

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/StatefulWorkflow.kt
steve-the-edwards: We call this from the runtime in `WorkflowNode` - which is all based on `StatefulWorkflow`, so yes, I think we need this unless we want to switch the order of the hierarchy.

workflow-testing/src/test/java/com/squareup/workflow1/ParameterizedTestRunner.kt
steve-the-edwards: Sure. not too concerned either way.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/BaseRenderContext.kt
steve-the-edwards: Changed manually.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/SessionWorkflow.kt
steve-the-edwards: Changed manually.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/SessionWorkflow.kt
steve-the-edwards: Changed manually.

PR #1108, opened 2023-09-10
Correcting sample code snippet in Tutorial 5

sssurvey:
In `Tutorial5.md`, there was an oversight in the code example for `WelcomeWorkflow#onLogin`. The example incorrectly included the `isNotEmpty()` check before it was supposed to be added by the developer. Additionally, it had an incorrect variable name, `state.name`, which has been corrected to `state.username`.

This pull request removes the `isNotEmpty()` check and makes the necessary variable name correction. 

Thank you!

### DIFF
--- a/samples/tutorial/Tutorial5.md
+++ b/samples/tutorial/Tutorial5.md
@@ -75,9 +75,7 @@ object WelcomeWorkflow : StatefulWorkflow<Unit, State, LoggedIn, WelcomeScreen>(
   }
 
   internal fun onLogin() = action {
-    if (state.name.isNotEmpty()) {
-      setOutput(LoggedIn(state.name))
-    }
+      setOutput(LoggedIn(state.username))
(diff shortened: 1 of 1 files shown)


samples/tutorial/Tutorial5.md
```diff
@@ -75,9 +75,7 @@ object WelcomeWorkflow : StatefulWorkflow<Unit, State, LoggedIn, WelcomeScreen>(
   }
```
steve-the-edwards, line 75: looks like more needs to be changed here. This is still using `name` rather than `username`

sssurvey: Yes @steve-the-edwards you are right, updated with `username` just now. I also forget to update line `323`, it is now updated as well. Thanks!

steve-the-edwards (approved):

PR #1110, opened 2023-09-21
Friendlier construction API for `BackStackScreen`

rjrjr:
* Deprecates the confusing `bottom, rest: List` constructor with an eye toward making it private
* Adds `fromList` and `fromListOrNull` factory functions to `BackStackScreen.Companion`, since no one ever finds the `List` extensions

We'll preserve the `bottom, rest: List` constructor as the core one because it makes it impossible to build an invalid instance.


rjrjr:
Actually…let me update this to have a fundamental private constructor that just takes a list. It'll be more efficient and the public API will still prevent mistakes.

PR #1116, opened 2023-10-12
Modernize tutorial final code

rjrjr:
_There are a lot of indentation changes here, good idea to hide whitespace while reviewing._

It's time to get the Tutorial caught up with undeprecated API. This PR modifies only the final tutorial module. Once we're happy with the code, I'll back port the changes and update the prose in follow ups.

- Use `AndroidScreen` and drop `ViewRegistry`
- Use `View.setBackHandler`
- Use `TextController`
- More consistent naming, code style for actions and event handlers in `Screen` renderings

I did not introduce `RenderContext.eventHandler {}`, that seems like it would just be confusing to a newcomer.

Also not introducing big new blocks of material, in particular not introducing `Overlay`. I do think we should do that, but for this release I just want to focus on getting the deprecated code deleted.


samples/tutorial/tutorial-final/src/main/java/workflow/tutorial/TodoListWorkflow.kt
```diff
@@ -29,14 +29,15 @@ object TodoListWorkflow : StatelessWorkflow<ListProps, Output, TodoListScreen>()
...
+      username = renderProps.username,
+      todoTitles = titles.map { it.textValue },
+      onTodoSelected = { context.actionSink.send(selectTodo(it)) },
+      onBackClick = { context.actionSink.send(postGoBack) },
+      onAddClick = { context.actionSink.send(postNewTodo) }
     )
   }
 
-  private fun onBack() = action {
+  private val postGoBack = action {
```
steve-the-edwards, line 40: same comments.

samples/tutorial/tutorial-final/src/main/java/workflow/tutorial/TodoModel.kt
```diff
@@ -1,6 +1,15 @@
...
-  val note: String
-)
+  val title: TextController,
+  val note: TextController
+) {
+  constructor(
+    title: String,
+    note: String
+  ) : this(TextController(title), TextController(note))
+}
```
steve-the-edwards, line 15: is this used directly in the rendering?

workflow-ui/container-common/src/main/java/com/squareup/workflow1/ui/modal/AlertContainerScreen.kt
```diff
@@ -16,10 +16,10 @@ import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
...
 // @Deprecated(
-//   "Use BodyAndModalsScreen and AlertOverlay",
+//   "Use BodyAndOverlaysScreen and AlertOverlay",
 //   ReplaceWith(
-//     "BodyAndModalsScreen<B>(beneathModals, modals)",
-//     "com.squareup.workflow1.ui.container.BodyAndModalsScreen"
+//     "BodyAndOverlaysScreen<B>(beneathModals, modals)",
+//     "com.squareup.workflow1.ui.container.BodyAndOverlaysScreen"
 //   )
 // )
```
steve-the-edwards, line 24: will need rebase I take it?

steve-the-edwards (approved):

samples/tutorial/tutorial-final/src/main/java/workflow/tutorial/TodoEditWorkflow.kt
```diff
@@ -54,33 +56,20 @@ object TodoEditWorkflow : StatefulWorkflow<EditProps, State, Output, TodoEditScr
...
-  internal fun onTitleChanged(title: String) = action {
-    state = state.withTitle(title)
-  }
-
-  internal fun onNoteChanged(note: String) = action {
-    state = state.withNote(note)
-  }
-
-  private fun onDiscard() = action {
+  private val postDiscard = action {
```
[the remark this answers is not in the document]
rjrjr: > LIke is the action about what it does (posting a save) or what it responds to (on "Save" clicked)? I feel like we've often used the latter, so maybe that's what is tripping me up. Maybe the former is preferrable?

I like the `*Action` notion, will adopt.

samples/tutorial/tutorial-final/src/main/java/workflow/tutorial/TodoListScreen.kt
```diff
@@ -1,14 +1,24 @@
...
  * `val onButtonTapped: () -> Unit`.
  */
+@OptIn(WorkflowUiExperimentalApi::class)
 data class TodoListScreen(
   val username: String,
   val todoTitles: List<String>,
   val onTodoSelected: (Int) -> Unit,
-  val onBack: () -> Unit
-)
+  val onBackClick: () -> Unit,
```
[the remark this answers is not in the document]
rjrjr: 🤔

samples/tutorial/tutorial-final/src/main/java/workflow/tutorial/TodoModel.kt
rjrjr: Yes. That's the normal pattern for `TextController`, which is why I've pushed back when anyone talks about generalizing it.

workflow-ui/container-common/src/main/java/com/squareup/workflow1/ui/modal/AlertContainerScreen.kt
rjrjr: Yup!

samples/tutorial/tutorial-final/src/main/java/workflow/tutorial/TodoEditWorkflow.kt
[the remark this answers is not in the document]
rjrjr: Actually…

What I like about the awkward `post*` names was that they communicated what was going to happen in response to the UI event. I don't think that's an implementation detail, it helps make visible what the role of this workflow is in the overall structure. And once I've put the names in that very active format, the `*Action` suffix starts to sound noisy. 

WDYT:

```kotlin
      onTodoSelected = { context.actionSink.send(requestSelection(it)) },
      onBackPressed = { context.actionSink.send(requestGoBack) },
      onAddPressed = { context.actionSink.send(requestNew) }
```

PR #1120, opened 2023-10-13
Update Kotlin to 1.9 and friends

steve-the-edwards:


steve-the-edwards:
Not needed. Done elsewhere.

PR #1122, opened 2023-10-20
Updates: Gradle 8.4; AGP 8.1.2; Java 17; Remove Kotlin DSL Plugin Use

steve-the-edwards:
A bit too coupled but it builds 🤷🏻


build-logic/build.gradle.kts
```diff
@@ -1,7 +1,8 @@
 plugins {
```
rjrjr, line 1: So if we've removed this plugin, how is it that the `*.kts` files are still working?

rjrjr (approved):

build-logic/build.gradle.kts
rjrjr: Is the idea that `*.kts` support is intrinsic, and the plugin is required just to extend it?

PR #1133, opened 2023-10-25
revert jdk target to 1.8

RBusarow:


steve-the-edwards (approved):

benchmarks/dungeon-benchmark/build.gradle.kts
```diff
@@ -13,12 +15,12 @@ android {
   compileSdk = libsCatalog.version("compileSdk").toInt()
 
   compileOptions {
-    sourceCompatibility = JavaVersion.VERSION_17
-    targetCompatibility = JavaVersion.VERSION_17
+    sourceCompatibility = javaTargetVersion
```
rjrjr, line 18: Do we really need to change `sourceCompatilibity` too? Will that, say, prevent us using language features that aren't actually supported?

rjrjr (approved):

benchmarks/dungeon-benchmark/build.gradle.kts
RBusarow: That's correct keeping them in sync will prevent us from inadvertently using Java language features that don't exist in the target environment.

build-logic/src/main/java/com/squareup/workflow1/buildsrc/internal/catalogs.kt
```diff
@@ -44,3 +27,48 @@ fun DependencyHandler.kotlin(
   module: String,
   version: String? = null
 ): Any = "org.jetbrains.kotlin:kotlin-$module${version?.let { ":$version" } ?: ""}"
+
+/**
+ * the jdk used in packaging
```
rjrjr, line 32: Isn't this redundant w/kgx?

RBusarow: I'd thought it was, but sadly no.  The catalog entry/property names are unique to individual projects.

PR #1135, opened 2023-11-14
BREAKING: Stop stomping `View.layoutParams`, new parameter for `replaceViewInParent`

rjrjr:
`WorkflowViewStub.replaceViewInParent` now can be configured not to propagate `layoutParams`, by setting new property `propagatesLayoutParams` to `false`. When it's `true` (the default), the stub's `layoutParams` are passed as a new third argument to the (customizable) `replaceViewInParent` function, to make it clearer that this function bears responsibility for applying them.

Motivated because `BodyAndOverlaysContainer` and `WorkflowLayout` were being dictatorial about the body view's layout params for no clear reason, leading to layout surprises. They're the first to opt out of `propagatesLayoutParams`.

### DIFF
--- a/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowViewStub.kt
+++ b/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowViewStub.kt
@@ -5,6 +5,7 @@ import android.graphics.drawable.Drawable
 import android.util.AttributeSet
 import android.view.View
 import android.view.ViewGroup
+import android.view.ViewGroup.LayoutParams
 import androidx.annotation.IdRes
 import androidx.savedstate.SavedStateRegistryOwner
 import androidx.savedstate.findViewTreeSavedStateRegistryOwner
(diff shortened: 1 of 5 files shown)


traviswestbrook (approved):

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowViewStub.kt
```diff
@@ -85,6 +86,8 @@ public class WorkflowViewStub @JvmOverloads constructor(
    */
   public var updatesVisibility: Boolean = true
 
+  public var pushesLayoutParams: Boolean = true
```
ijwhelan, line 89: kdoc here?

Also wondering if `propagatesLayoutParams` might be more clear.

ijwhelan (approved):

rjrjr:
Renamed to `propagatesLayoutParams`, added missing kdoc, and punched up kdoc on `replaceOldViewInParent`. Will push again to beef up the commit message.

PR #1137, opened 2023-11-15
Deprecates `LifecycleWorker` in favor of `SessionWorkflow.initialState`

### LINKED ISSUE #1136: Un-deprecate runningWorker for LifecycleWorker
Getting a deprecation warning for:

```kotlin
context.runningWorker(teardownWorker)
```

Which looks like:

```
private val teardownWorker = object : LifecycleWorker() {
  override fun onStopped() {
    mediaPlayer.shutDown()
  }
}
```

The alternative says to use `runningSideEffect`, but it doesn't seem to work and we would need to rewrite `LifecycleWorker`.

rjrjr:
Fixes #1136


LandonPatmore (approved):

rjrjr:
Had to suppress a few deprecation warnings in tests.

steve-the-edwards (approved):
Thanks!

steve-the-edwards:
#1136 already fixed with other docs. See more added here - https://github.com/square/workflow-kotlin/pull/1140

PR #1140, opened 2023-12-11
Add More SessionWorkflow lifecycle documentation/suggestions

steve-the-edwards:


workflow-core/src/commonMain/kotlin/com/squareup/workflow1/SessionWorkflow.kt
```diff
@@ -70,7 +70,29 @@ public abstract class SessionWorkflow<
    *     or [Flow.onCompletion][kotlinx.coroutines.flow.onCompletion] to handle that.
    *
    *     If you have a general cleanup operation that is fast and thread-safe then you could use
-   *     [Job.invokeOnCompletion][kotlinx.coroutines.Job.invokeOnCompletion].
+   *     [Job.invokeOnCompletion][kotlinx.coroutines.Job.invokeOnCompletion]. Otherwise use
+   *     [awaitCancellation][kotlinx.coroutines.awaitCancellation] with a finally block. Examples:
+   *
+   *     override fun initialState(
```
rjrjr, line 76: You need to indent this more to get it to format as a code block. Use Ctrl-J to preview.

rjrjr: Note that the existing codeblocks are a mess

steve-the-edwards: Ok, fixed both.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/SessionWorkflow.kt
```diff
@@ -70,7 +70,29 @@ public abstract class SessionWorkflow<
...
+   *       props: Unit,
+   *       snapshot: Snapshot?,
+   *       workflowScope: CoroutineScope
+   *     ): MyState {
+   *       someService.start()
+   *       workflowScope.coroutineContext.job.invokeOnCompletion {
+   *         // If quick and does not need to on Workflow runtime thread.
```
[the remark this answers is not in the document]
steve-the-edwards: "If quick and does not need to execute on the Workflow runtime's thread."

rjrjr (approved):

PR #1141, opened 2023-12-11
Add testRender with CoroutineScope for SessionWorkflow

### LINKED ISSUE #1138: SessionWorkflow throws when trying to unit test it with RenderTester#testRender
When trying to unit test a Workflow that inherits from `SessionWorkflow` using `RenderTester#testRender` I get this error:

> SessionWorkflow should never call initialState without the CoroutineScope.
java.lang.IllegalStateException: SessionWorkflow should never call initialState without the CoroutineScope.
	at com.squareup.workflow1.SessionWorkflow.initialState(SessionWorkflow.kt:82)
	at com.squareup.workflow1.testing.RenderTesterKt.testRender(RenderTester.kt:29)
	

It looks like `RenderTester#testRender` is calling an overloaded `initialState` method that `SessionWorkflow` doesn't support, see https://github.com/square/workflow-kotlin/blob/b7e37bc04fff285fbb346c28c1008817a5a1578e/workflow-core/src/commonMain/kotlin/com/squareup/workflow1/SessionWorkflow.kt#L78-L84.

steve-the-edwards:
Fixes #1138

### DIFF
--- a/workflow-testing/src/main/java/com/squareup/workflow1/testing/RenderTester.kt
+++ b/workflow-testing/src/main/java/com/squareup/workflow1/testing/RenderTester.kt
@@ -1,14 +1,17 @@
 package com.squareup.workflow1.testing
 
 import com.squareup.workflow1.ActionApplied
+import com.squareup.workflow1.SessionWorkflow
 import com.squareup.workflow1.StatefulWorkflow
 import com.squareup.workflow1.Workflow
 import com.squareup.workflow1.WorkflowAction
+import com.squareup.workflow1.WorkflowExperimentalApi
 import com.squareup.workflow1.WorkflowIdentifier
 import com.squareup.workflow1.WorkflowOutput
 import com.squareup.workflow1.identifier
 import com.squareup.workflow1.testing.RenderTester.ChildWorkflowMatch
 import com.squareup.workflow1.workflowIdentifier
+import kotlinx.coroutines.CoroutineScope
 import kotlin.reflect.KClass
 import kotlin.reflect.KType
 import kotlin.reflect.KTypeProjection
@@ -19,6 +22,7 @@ import kotlin.reflect.KTypeProjection
  *
  * See [RenderTester] for usage documentation.
  */
+@OptIn(WorkflowExperimentalApi::class) // Opt-in is only for the argument check.
 @Suppress("UNCHECKED_CAST")
 public fun <PropsT, OutputT, RenderingT> Workflow<PropsT, OutputT, RenderingT>.testRender(
   props: PropsT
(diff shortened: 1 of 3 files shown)


workflow-testing/src/main/java/com/squareup/workflow1/testing/RenderTester.kt
```diff
@@ -19,14 +22,43 @@ import kotlin.reflect.KTypeProjection
...
 @Suppress("UNCHECKED_CAST")
 public fun <PropsT, OutputT, RenderingT> Workflow<PropsT, OutputT, RenderingT>.testRender(
   props: PropsT
 ): RenderTester<PropsT, *, OutputT, RenderingT> {
   val statefulWorkflow = asStatefulWorkflow() as StatefulWorkflow<PropsT, Any?, OutputT, RenderingT>
   return statefulWorkflow.testRender(
     props = props,
-    initialState = statefulWorkflow.initialState(props, null)
+    initialState = run {
+      require(this !is SessionWorkflow<PropsT, *, OutputT, RenderingT>) {
```
rjrjr, line 34: nice

workflow-testing/src/main/java/com/squareup/workflow1/testing/RenderTester.kt
```diff
@@ -19,14 +22,43 @@ import kotlin.reflect.KTypeProjection
...
+
+/**
+ * Create a [RenderTester] to unit test an individual render pass of this [SessionWorkflow],
+ * using the workflow's [initial state][StatefulWorkflow.initialState], in the [workflowScope].
+ *
+ * See [RenderTester] for usage documentation.
+ */
+@OptIn(WorkflowExperimentalApi::class)
+@Suppress("UNCHECKED_CAST")
+public fun <PropsT, OutputT, RenderingT> Workflow<PropsT, OutputT, RenderingT>.testRender(
```
rjrjr, line 50: Can't you make this version an extension on `SessionWorkflow` and avoid the runtime check?

rjrjr (approved):

workflow-testing/src/main/java/com/squareup/workflow1/testing/RenderTester.kt
steve-the-edwards: ya I thought about that but then I thought about how in your tests you might not have the reified `SessionWorkflow` type? Maybe you were passing it around or dealign with it just as a `Workflow`? Maybe I was overthinking it though. Perhaps we could just make it an extension on `SessionWorkflow` 🤔

rjrjr: I think so. I'm going to write a unit test against `class RealFooWorkflow`, not `interface FooWorkflow`.

steve-the-edwards: ya, I think you are right. I will change it.

- [ ] testRender + CoroutineScope extension extends `SessionWorkflow` not `Workflow`

PR #1144, opened 2023-12-12
Modernize tutorial final code

rjrjr:
It's time to get the Tutorial caught up with undeprecated API. This PR modifies only the final tutorial module. Once we're happy with the code, I'll back port the changes and update the prose in follow ups.

- Use `AndroidScreen` and drop `ViewRegistry`
- Better, more consistent use of `BackStackScreen`
- Use `View.setBackHandler`
- Use `TextController`
- Delete a lot of `// Exactly what the function name and the code say` comments
- More consistent naming, code style for actions and event handlers in `Screen` renderings
  - Event handler fields are named `onVerbPhrase`, like `onBackPressed`
  - Action names are verb phrases describing the action that is being taken, not the event that is being handled
  - Output names are generally in terms of the semantic event being reported to the parent, rather than describing what the parent will do

Thus:

```kotlin
return TodoListScreen(
  onRowPressed = { context.actionSink.send(reportSelection(it)) }
```

```kotlin
  private fun reportSelection(index: Int) = action {
    // Tell our parent that a todo item was selected.
    setOutput(TodoSelected(index))
  }
```

I did not introduce `RenderContext.eventHandler {}`, that seems like it would just be confusing to a newcomer.

Also not introducing big new blocks of material, in particular not introducing `Overlay`. I do think we should do that, but for this release I just want to focus on getting the deprecated code deleted.


workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/container/BackStackScreen.kt
```diff
@@ -115,6 +115,13 @@ public operator fun <T : Screen> BackStackScreen<T>.plus(
   return other?.let { BackStackScreen(frames + it.frames) } ?: this
 }
 
+@WorkflowUiExperimentalApi
+public operator fun <T : Screen> BackStackScreen<T>.plus(
+  other: BackStackScreen<T>?
+): BackStackScreen<T> {
+  return other?.let { BackStackScreen(frames + it.frames) } ?: this
+}
```
steve-the-edwards, line 123: Could separate this out to a different PR that's not tutorial related. but also 🤷🏻

steve-the-edwards (approved):
LGTM. Still don't feel like the way we patternize `action` creation is ideal - but I'm not sure what I offered is better.

Also, if you have a backlog of what else to add to the tutorial can I request:
1. A SessionWorkflow
2. Other testing types - like `headlessIntegrationTest` or `launchForTestingFrom...`?

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/container/BackStackScreen.kt
rjrjr: Yeah, this seems like it leaked in from another PR (that I thought I had already merged).

rjrjr:
> Also, if you have a backlog of what else to add to the tutorial can I request:
> 
> 1. A SessionWorkflow
> 2. Other testing types - like `headlessIntegrationTest` or `launchForTestingFrom...`?

We need proper code labs. And more people. :/

rjrjr:
Merging to the staging `ray/tutorial-update` branch. Red shards are current noise plus a compile error in tutorial 2, which hasn't been updated yet.

Issue #669, opened 2022-02-16
BackStackScreen causes ComposeRenderings to disappear when transition animations kick-in


lucamtudor:
I've changed ":samples:compose-samples" & ":samples:tutorial" to play with BackStackScreen, ComposeRendering & LayoutRunners. The code producing this bug isn't really noteworthy. The whole Activity is driven with a WorkflowLayout and the render method looks like:

```kotlin
  override fun render(
    renderProps: Unit,
    renderState: State,
    context: RenderContext
  ): Any {
    val backstack: List<Any> = List(renderState.children) { index ->
      when (index % 3) {
        0 -> {
          LegacyRendering(
            id = "$index Compose-GoForwardRendering",
            goBack = { context.actionSink.send(goBack()) },
            goForward = { context.actionSink.send(goForward()) },
            rendering = GoForwardComposeRendering(
              step = index,
              content = context.renderChild(TextInputWorkflow, key = "text-input-$index"),
              onNextClicked = { context.actionSink.send(goForward()) },
            ),
          )
        }

        1 -> {
          val welcomeScreen = context.renderChild(WelcomeWorkflow, key = "welcome-$index") {
            WorkflowAction.noAction()
          }
          LegacyRendering(
            id = "$index Legacy-Welcome Screen",
            goBack = { context.actionSink.send(goBack()) },
            goForward = { context.actionSink.send(goForward()) },
            rendering = welcomeScreen,
          )
        }

        else -> {
          val hello = context.renderChild(HelloWorkflow)
          LegacyRendering(
            id = "$index Compose-GoBackRendering",
            goBack = { context.actionSink.send(goBack()) },
            goForward = { context.actionSink.send(goForward()) },
            rendering = GoBackComposeRendering(
              step = index,
              content = hello,
              onGoBack = { context.actionSink.send(goBack()) },
            )
          )
        }
      }
    }

    return backstack.toBackStackScreen()
  }
```

Here's a short video showcasing the problem:

https://user-images.githubusercontent.com/3036785/154375879-ac8b07da-2e1d-4b3e-8150-cef1dafcdf46.mp4

rjrjr:
I believe this was fixed by the introduction of `ScreenComposableFactory` and the new compose binding for `NamedScreen`. Leaving the ticket open as a reminder to verify that.

PR #1146, opened 2023-12-18
BREAKING: Replaces `ComposeScreenViewFactory` with `ScreenComposableFactory`

### LINKED ISSUE #546: Wrapper renderings introduce an extra AndroidView unnecessarily when the wrapped rendering's factory is a ComposeScreenViewFactory.

rjrjr:
**_Please be sure to review `workflow-ui/compose/README.md`, which GitHub helpfully hides by default because it's big._**

_There are two commits in this PR, and it'd be wise to review them in sequence:_

### _`ViewRegistry.Key` allows multiple factory types per rendering type_

Paves the way for parallel Classic and Compose implementations of wrappers like `NamedScreen`, to fix #546. Ironically, breaks Compose support in the process, but that is fixed by the next commit.

### _Replaces `ComposeScreenViewFactory` with `ScreenComposableFactory`_

This commit takes advantage of the new `ViewRegistry.Key` (which allows renderings to be bound to multiple UI factories) to fix a long standing problem where wrapper screens -- things like `NamedScreen` and `EnvironmentScreen` -- used in a Compose context would cause needless calls to `@Composeable fun AndroidView()` and `ComposeView()`.

For example, consider this rendering:

```
BodyAndOverlaysScreen(
  body = SomeComposeScreen(
    EnvironmentScreen(
      SomeOtherComposeScreen
    )
  )
)
```

Before this change, that would create a View hierarchy something like this:

```
BodyAndOverlaysContainer : FrameLayout {
  mChildren[0] = ComposeView {
    // compose land
    SomeComposeScreen.Content {
      AndroidView {
        ComposeView {
          // nested compose land
            SomeOtherComposeScreen.Content()
```

Now it will look this way:

```
BodyAndOverlaysContainer : FrameLayout {
  mChildren[0] = ComposeView {
    // compose land
    SomeComposeScreen.Content {
      SomeOtherComposeScreen.Content()
```

`ScreenComposableFactory` replaces `ComposeScreenViewFactory`, and `ComposeScreen` no longer extends `AndroidScreen`. Compose support is now a first class citizen, instead of a hack bolted on to View support.

Unfortunately, `ViewEnvironment.withComposeInteropSupport()` (see below) now must be called near the root of a Workflow app to enable the seamless Compose > Classic > Compose handling that used to be built in to `WorkflowRendering` and `ComposeScreenViewFactory`. This means that call is required for Compose support for built in rendering types like `BodyAndOverlaysScreen` and `BackStackScreen`, which so far are backed only by classic View implementations.

Other introductions, changes:

- `Screen.toComposableFactory()`, used by `WorkflowRendering()` in the same way that `WorkflowViewStub` uses `Screen.toViewFactory()`

- `ScreenComposableFactoryFinder`, a `ViewEnvironment`-based strategy object used by `Screen.toComposableFactory()` the same way that `Screen.toViewFactory()` uses `ScreenViewFactoryFinder`. The default implementation provides Compose bindings for `NamedScreen` and `EnvironmentScreen`, fixing #546.

- `ScreenViewFactoryFinder.getViewFactoryForRendering()` can now return `null`. A `requireViewFactoryForRendering()` extension is introduced for use when `null` is not acceptable.

- `ViewEnvironment.withComposeInteropSupport()`, which wraps the found `ScreenComposableFactoryFinder` and `ScreenViewFactoryFinder` with implementations that allow Compose contexts to handle renderings bound only to `ScreenViewFactory`, and classic contexts to handle renderings bound only to `ScreenComposableFactory`. Replaces the logic that used to be in the private `ScreenViewFactory.asComposeViewFactory()` extension in `WorkflowRendering()`.

- `Screen.Preview()` is introduced. The existing `Preview()` extension functions were tied to `ScreenViewFactory`, making them much less useful. It is still the case that previews work for non-Compose UI code just fine. Which is pretty cool, really.

Fixes #546


rjrjr:
- [ ] Use `ViewRegistry.getFactoryFor()` or get rid of it
- [ ] Update kdoc to note that `ScreenViewFactoryFinder` method now can return `null`, and mention new `require...` method

rjrjr:
This is looking really solid. I don't quite have our internal test suite passing yet, but so far everything I've found has been newly exposed bugs in our workflow integration, especially the hellish duct tape supporiting our remaining pre-workflow code. I'm pretty optimistic about shipping this month.

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/ViewRegistry.kt
```diff
@@ -61,24 +63,60 @@ import kotlin.reflect.KClass
...
+      return "Key(renderingType=$renderingType, factoryType=$factoryType)"
+    }
+  }
+
+  /**
+   * Implemented by a factory that can build some kind of UI for view models
+   * of type [RenderingT], and which can be listed in a [ViewRegistry]. The
+   * [Key.factoryType] field of [key] must be the type of this [Entry].
+   */
   public interface Entry<in RenderingT : Any> {
```
rjrjr, line 102: It is very tempting to rename `Entry` to `UiFactory`. I avoided that in this PR to minimize surface area and migration pain. But maybe we should allow ourselves one more round of renaming before locking everything down?

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ScreenComposableFactory.kt
```diff
@@ -0,0 +1,280 @@
...
+   * The composable content of this [ScreenComposableFactory]. This method will be called
+   * any time [rendering] or [environment] change. It is the Compose-based analogue of
+   * [ScreenViewRunner.showRendering][com.squareup.workflow1.ui.ScreenViewRunner.showRendering].
+   */
+  @Composable public fun Content(
+    rendering: ScreenT,
+    environment: ViewEnvironment
+  )
+
+  public companion object {
```
rjrjr, line 131: Isn't this redundant with `screenComposableFactory(` above?

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/ViewRegistry.kt
rjrjr: Nah.

workflow-ui/compose/src/androidTest/java/com/squareup/workflow1/ui/compose/WorkflowRenderingTest.kt
```diff
@@ -178,18 +158,63 @@ internal class WorkflowRenderingTest {
...
-      WorkflowRendering(rendering, ViewEnvironment.EMPTY)
+      WorkflowRendering(rendering, env)
     }
 
     onView(withText("two")).check(matches(isDisplayed()))
     wrapperText.value = "OWT"
     onView(withText("OWT")).check(matches(isDisplayed()))
   }
 
+  @Test fun namedScreenStaysInTheSameComposeView() {
```
rjrjr, line 169: Ta da!

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/WorkflowRendering.kt
```diff
@@ -146,90 +134,3 @@ public fun WorkflowRendering(
...
   return lifecycleOwner
 }
-
-/**
- * Returns a [ComposeScreenViewFactory] that makes it convenient to display this [ScreenViewFactory]
- * as a composable. If this is a [ComposeScreenViewFactory] already it just returns `this`,
- * otherwise it wraps the factory in one that manages a classic Android view.
- */
-@OptIn(WorkflowUiExperimentalApi::class)
-private fun <ScreenT : Screen> ScreenViewFactory<ScreenT>.asComposeViewFactory() =
```
rjrjr, line 156: This became `ScreenComposableFactoryFinder`

workflow-ui/compose/README.md
```diff
@@ -105,85 +111,108 @@ We have looked at a few ways of implementing this:
...
 
-We will also provide a construction analogous to `WorkflowViewStub` to allow Compose-based factories to idiomatically display child renderings.
+We also provide [`@Composable fun WorkflowRendering()`](https://github.com/square/workflow-kotlin/blob/9bfd5119fabd0a3dfbc25bf7d93e52c7b31bb4cd/workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/WorkflowRendering.kt), a construction function analogous to `WorkflowViewStub` to allow Compose-based factories to idiomatically display child renderings.
 
-The above two concepts coordinate, and when a Compose-based factory is delegating to a child rendering that is also bound to a composable factory, we can skip the detour out into the Android view world and simply call the child composable directly from the parent.
+> You'll note that there is no `OverlayComposableFactory` family of interfaces. So far, all of our window management is strictly via classic Android `Dialog` calls. There is nothing stopping us (or you) from adding Compose-based `Overlay` support in the future as a replacement, but we're definitely not making any promises on that front.
```
rjrjr, line 123: @ekeitho 👀

workflow-ui/compose-tooling/src/main/java/com/squareup/workflow1/ui/compose/tooling/Previews.kt
```diff
@@ -0,0 +1,113 @@
...
+  viewEnvironmentUpdater: ((ViewEnvironment) -> ViewEnvironment)? = null
+) {
+  val previewEnvironment =
+    rememberPreviewViewEnvironment(placeholderModifier, viewEnvironmentUpdater, mainFactory = this)
+  WorkflowRendering(rendering, previewEnvironment, modifier)
+}
+
+/**
+ * Like [ScreenComposableFactory.Preview], but for non-Compose [ScreenViewFactory] instances.
+ * Yes, you can preview classic [View][android.view.View] code this way.
```
ekeitho, line 90: whoa, neat.

workflow-ui/compose/README.md
```diff
@@ -1,11 +1,9 @@
...
+It was originally published as a [blog post](https://developer.squareup.com/blog/jetpack-compose-support-in-workflow).
+Since then, our approach has been overhauled. With that overhaul this document has been updated to reflect how the new system works, but it's still very deisgn-doc like. To skip past all the big picture and implementation verbiage and get right to how to use this stuff, jump down to [API Design][#api-design] below.
```
ekeitho, line 6: design-doc*

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ScreenComposableFactoryFinder.kt
```diff
@@ -0,0 +1,69 @@
...
+    rendering: ScreenT
+  ): ScreenComposableFactory<ScreenT>? {
+    val factoryOrNull: ScreenComposableFactory<ScreenT>? =
+      environment[ViewRegistry].getFactoryFor(rendering)
+
+    @Suppress("UNCHECKED_CAST")
+    return factoryOrNull
+      ?: (rendering as? ComposeScreen)?.let {
+        ScreenComposableFactory<ComposeScreen> { rendering, environment ->
+          // Do we need to key on either, both params?
```
vgonda, line 26: Do we need to answer this before merging?

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ScreenComposableFactoryFinder.kt
```diff
@@ -0,0 +1,69 @@
...
+          rendering.Content(environment)
+        } as ScreenComposableFactory<ScreenT>
+      }
+      // Support for Compose BackStackScreen, BodyAndOverlaysScreen treatments would go here
+      //   See similar blocks in ScreenViewFactoryFinder
```
vgonda, line 31: Should we have a TODO here to add them?

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ViewEnvironmentWithComposeSupport.kt
```diff
@@ -0,0 +1,57 @@
...
+      rendering: ScreenT
+    ): ScreenComposableFactory<ScreenT>? {
+      return rawComposableFactoryFinder.getComposableFactoryForRendering(environment, rendering)
+        ?: rawViewFactoryFinder.getViewFactoryForRendering(environment, rendering)
+          ?.asComposableFactory()
+    }
+  }
+
+  // TODO consider something like LocalHasViewFactoryRootBeenApplied to enforce
+  // no duplicate calls
```
vgonda, line 53: Drawing attention to this lingering TODO in case it needs attention before merging

vgonda (approved):

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ViewEnvironmentWithComposeSupport.kt
rjrjr: Oh, thanks!

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ScreenComposableFactoryFinder.kt
rjrjr: That work isn't planned, so I'm reluctant to imply it. I'll tighten up the language a bit.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ScreenComposableFactoryFinder.kt
rjrjr: Answer appears to be no, deleting.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ViewEnvironmentWithComposeSupport.kt
rjrjr: Going to hold off on this until the need makes itself clear.

workflow-ui/compose-tooling/src/main/java/com/squareup/workflow1/ui/compose/tooling/Previews.kt
```diff
@@ -0,0 +1,113 @@
...
+ *       ).Preview()
+ *     }
+ */
+@WorkflowUiExperimentalApi
+@Composable
+public fun Screen.Preview(
+  modifier: Modifier = Modifier,
+  placeholderModifier: Modifier = Modifier,
+  viewEnvironmentUpdater: ((ViewEnvironment) -> ViewEnvironment)? = null
+) {
```
ijwhelan, line 39: 😮

ijwhelan (approved):

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ScreenComposableFactory.kt
```diff
@@ -0,0 +1,233 @@
...
+import com.squareup.workflow1.ui.ViewRegistry
+import com.squareup.workflow1.ui.ViewRegistry.Key
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+import com.squareup.workflow1.ui.androidx.OnBackPressedDispatcherOwnerKey
+import com.squareup.workflow1.ui.show
+import com.squareup.workflow1.ui.startShowing
+import kotlin.reflect.KClass
+
+@WorkflowUiExperimentalApi
+public inline fun <reified ScreenT : Screen> ScreenComposableFactory(
```
ekeitho, line 26: trying to understand this better, i was thinking this would of been of type `ComposeScreen` to make it more assertive to only be used with compose; but after reading the comment below `Use [ScreenComposableFactory] directly if you need to prevent your [Screen] rendering classes from depending on Compose at compile time.` and the being able to inject things like `NamedScreen`, `EnvironmentScreen` into the compose content - is the reasoning why?

maybe having that sort of comment near the top why it isn't `ComposeScreen` would be helpful for readability.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ScreenComposableFactory.kt
```diff
@@ -0,0 +1,233 @@
...
+ *     val viewRegistry = ViewRegistry(fooComposableFactory, …)
+ *     val viewEnvironment = ViewEnvironment.EMPTY + viewRegistry
+ *
+ *     renderWorkflowIn(
+ *       workflow = MyWorkflow.mapRendering { it.withEnvironment(viewEnvironment) }
+ *     )
+ *
+ */
+@WorkflowUiExperimentalApi
+public interface ScreenComposableFactory<in ScreenT : Screen> : ViewRegistry.Entry<ScreenT> {
```
ekeitho, line 84: outside of scope of this, but out of curiosity if we were wanting to add a compose overlay factory.

would that entail creating a new `ComposeOverlayDialogFactory` which implements `ScreenComposableFactory` & then adding to the [checks in ScreenComposableFactoryFinder](https://github.com/square/workflow-kotlin/pull/1146/files#diff-488dc3940ee9cf848df3a29160328f6b811dcca43c0f130d52d10d66fad40cfdR30) searching for `ComposeOverlayDialogFactory`?

or would that entail creating a new factory like this one, with it's relative factory finder?

ekeitho (approved):

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ScreenComposableFactory.kt
rjrjr: `ComposeScreen` is a possible implementation, one of many possible concrete implementations of `Screen`. Tying this to `ComposeScreen` would be like tying it to `ArrayList` instead of `List`. Really, `ComposeScreen` is just a convenience so that you don't have to deal with `ScreenComposableFactory` directly.

rjrjr: I'll find a way to get that across better in the docs.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ScreenComposableFactory.kt
rjrjr: That would require:

- A Compose binding (`ScreenComposableFactory` implementation) for `BodyAndOverlaysScreen` -- an alternative to the existing `BodyAndOverlaysContainer`

- A new `OverlayComposableFactory` / `OverlayComposableFactoryFinder` pairing

That's assuming you weren't going to use any non-Compose dialog code at all.

If you were going to try to make it coexist with the `DialogComposableFactory` you would also need to enhance the logic in `withComposeInteropSupport()`. But that would only be feasible if your Compose version was still going to create `Dialog` instances, so that Compose-managed ones could interleave with legacy instances. Even then, you would somehow have to make `DialogCollator` et al agnostic between `Dialog` intances created via `Compose` v. those created the other way. No idea if that's practical.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ScreenComposableFactory.kt
rjrjr: @ekeitho I just pushed up a separate SHA that tweaks some kdoc, PTAL.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ScreenComposableFactory.kt
ekeitho: yeah exactly not going to use any non-compose dialog code at all - thanks for explaining!

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ScreenComposableFactory.kt
ekeitho: this makes a lot more sense to me, thank you!

rjrjr:
As soon as this merges (still trying to get our #*&$(& benchmarks to run), we should follow up with a new Compose-based BackStackScreen treatment that fixes #669.

rjrjr:
> As soon as this merges (still trying to get our #*&$(& benchmarks to run), we should follow up with a new Compose-based BackStackScreen treatment that fixes #669.

Benchmarks look good. No real improvement in the instrumented portion of our real world app, but that's to be expected since our Compose code is still almost entirely at the leaf level. But the reassurance that the latency around those leaves didn't climb is meaningful.

workflow-ui/compose/build.gradle.kts
```diff
@@ -13,6 +13,10 @@ android {
   composeOptions {
     kotlinCompilerExtensionVersion = libs.versions.androidx.compose.compiler.get()
   }
+  defaultConfig {
```
rjrjr, line 16: Oops, didn't mean to merge this.

rjrjr: No, not accident, required by a crummy old AVD version. But this ripples to break other things (how did I get a green PR?). Need to consider spreading the blast radius or else upgrading the AVDs.

@RBusarow @steve-the-edwards, what's your kneejerk?

PR #1150, opened 2023-12-21
Cleanup publish-snapshot

steve-the-edwards:
No running from ray/ui-update branch

No artifactsCheck as we use a LInux Runner


steve-the-edwards:
Putting this on hold for in the new year.

steve-the-edwards:
Manually triggered - finished successfully here - https://github.com/square/workflow-kotlin/actions/runs/7389619523

RBusarow (approved):

PR #1154, opened 2024-01-17
BREAKING: container > navigation

rjrjr:
In preparation for (finally!) locking down the workflow-ui API, renames the `com.squareup.workflow1.ui.container` package to `navigation`, since that's the common theme of nearly everything in there. I think this simple change should help make our navigation story a bit more self documentating. Yes, I'll still write navigation docs.

Other changes:

 * `View.setBackHandler` moves from `ui` to `ui.navigation`
 * `EnvironmentScreen`, `Screen.withEnvironment()`, `Screen.withRegistry()` which this PR moves up to the `ui` package. References to these had bled out into other spots already, so this feels like a move in the right direction.
 * `OnBackPressedDispatcherOwnerKey` moves from `ui` to `ui.androidx`

Also, note that the packages of the deprecated classes in the `container-android` and `container-common` modules were not changed, since they're about to be deleted.


samples/containers/poetry/src/main/java/com/squareup/sample/poetry/StanzaListScreen.kt
```diff
@@ -16,9 +16,9 @@ import com.squareup.workflow1.ui.ScreenViewFactory
 import com.squareup.workflow1.ui.ScreenViewRunner
 import com.squareup.workflow1.ui.ViewEnvironment
 import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
-import com.squareup.workflow1.ui.container.BackStackConfig
-import com.squareup.workflow1.ui.container.BackStackConfig.Other
-import com.squareup.workflow1.ui.setBackHandler
+import com.squareup.workflow1.ui.navigation.BackStackConfig
+import com.squareup.workflow1.ui.navigation.BackStackConfig.Other
+import com.squareup.workflow1.ui.container.setBackHandler
```
rjrjr, line 21: Missed some spots :/

ekeitho (approved):
LGTM

rjrjr:
Force push fixes file name of `ViewSetBackHandler.kt` and improves commit message.

samples/nested-overlays/src/main/java/com/squareup/sample/nestedoverlays/NestedOverlaysWorkflow.kt
```diff
@@ -53,11 +55,11 @@ object NestedOverlaysWorkflow : StatefulWorkflow<Unit, State, Nothing, Screen>()
         BackButtonScreen(
           ButtonBar(
             Button(
-              name = R.string.close,
+              name = string.close,
```
ijwhelan, line 58: Android Studio strikes again

ijwhelan: Looks like it's only this file

ijwhelan (approved):
Overall LGTM, just one place where Android Studio tried to be too helpful

samples/nested-overlays/src/main/java/com/squareup/sample/nestedoverlays/NestedOverlaysWorkflow.kt
rjrjr: Thanks for catching that. It's just crazy making.

PR #1156, opened 2024-01-18
Document how WorkflowAction can delegate

rjrjr:
It's an obscure use case, but it comes up every now and then

### DIFF
--- a/workflow-core/src/commonMain/kotlin/com/squareup/workflow1/WorkflowAction.kt
+++ b/workflow-core/src/commonMain/kotlin/com/squareup/workflow1/WorkflowAction.kt
@@ -13,7 +13,9 @@ import kotlin.jvm.JvmOverloads
  *
  * A [WorkflowAction]'s [apply] method is executed in the context of an [Updater][WorkflowAction.Updater],
  * which provides access to the current [props][WorkflowAction.Updater.props] and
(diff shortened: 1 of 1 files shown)


workflow-core/src/commonMain/kotlin/com/squareup/workflow1/WorkflowAction.kt
```diff
@@ -10,12 +10,28 @@ import kotlin.jvm.JvmOverloads
 
 /**
  * An atomic operation that updates the state of a [Workflow], and also optionally emits an output.
+ *
+ * A [WorkflowAction]'s [apply] method is executed in the context of an [Updater][WorkflowAction.Updater],
+ * which provides access to the current [props][WorkflowAction.Updater.props] and
+ * [state][WorkflowAction.Updater.props]. The latter can be updated with a new [StateT] instance
+ * that will become the current one after the [apply] function finishes.
```
rjrjr, line 17:
```suggestion
 * A [WorkflowAction]'s [apply] method is executed in the context of an [Updater][WorkflowAction.Updater],
 * which provides access to the current [props][WorkflowAction.Updater.props] and
 * [state][WorkflowAction.Updater.state],
 * along with a [setOutput][WorkflowAction.Updater.setOutput] function.
 * The [state][WorkflowAction.Updater.state] can be updated with a new [StateT] instance
 * that will become the current one after the [apply] function finishes.
```

PR #1158, opened 2024-01-19
Fix tutorial fileTemplates

rjrjr:
Our templates were out of date, and our `install.sh` script doesn't work any more.
Replaced with an IDE-produced zip file that the IDE is willing to import.

Replaces `Layout Runner (ViewBinding)` with `Android Screen (ViewBinding)`:

```kotlin
package ${PACKAGE_NAME}

import com.squareup.workflow1.ui.AndroidScreen
import com.squareup.workflow1.ui.ScreenViewFactory
import com.squareup.workflow1.ui.ScreenViewRunner

data class $Name(
  // TODO: add properties needed to update $VIEW_BINDING_TYPE
) : AndroidScreen<$Name> {

  override val viewFactory =
    ScreenViewFactory.fromViewBinding($VIEW_BINDING_TYPE::inflate, ::${Name}Runner)
}

private class ${Name}Runner(
  private val viewBinding: $VIEW_BINDING_TYPE
) : ScreenViewRunner<$Name> {

  override fun showRendering(
    rendering: $Name,
    environment: ViewEnvironment
  ) {
    TODO("Update viewBinding from rendering")
  }
}
```


ijwhelan:
Any reason not to encourage usage of the other flavor of `fromViewBinding()` that allows for a trailing lambda containing the `ScreenViewRunner`'s logic instead?

ijwhelan (approved):

rjrjr:
> Any reason not to encourage usage of the other flavor of `fromViewBinding()` that allows for a trailing lambda containing the `ScreenViewRunner`'s logic instead?

Good catch. I actually considered that but backed away from it, because that makes it so easy to accidentally capture the first rendering instead of using the one passed to the lambda. I figure for a tutorial it's good to have all the moving parts visible, even at the expense of more boilerplate. And I really doubt these templates will get much use outside of the tutorial.

PR #1159, opened 2024-01-20
Tutorial update

rjrjr:
Long lived branch that is collecting commits that bring the tutorial up to date with undeprecated workflow UI API. Expect rebases. So far the source code in `samples/tutorial/tutorial-final` is in the proper form. No prose has been updated yet.


rjrjr:
The branch is still there and I'm working on it actively (again, finally) but it's wasteful to keep this PR open and kick off CI runs that won't pass and won't be looked at.

PR #1165, opened 2024-02-02
set GHA runner max heap size dynamically based upon runner hardware

RBusarow:
Before this change, we set the max heap size based upon runner OS. That worked well enough, since we only used the free GitHub-hosted runners with well-known hardware. 

Now that we have two different Ubuntu runner sizes, that old strategy is leaving free memory on the table.

### DIFF
--- a/.github/actions/gradle-args/action.yml
+++ b/.github/actions/gradle-args/action.yml
@@ -21,24 +21,39 @@ runs:
       run: |
         runnerOS=$RUNNER_OS
 
-        # Set common JVM arguments
-        jvmArgs="-XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8"
+        totalMemory=0
+        # How much memory does the OS require? 
+        memoryOverhead=0
 
         case $runnerOS in
           macOS)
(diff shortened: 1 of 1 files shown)


.github/actions/gradle-args/action.yml
```diff
@@ -21,25 +21,40 @@ runs :
...
+            memoryOverhead=3
             ;;
           Linux)
-            jvmArgs="-Xmx4g -XX:MaxMetaspaceSize=3g $jvmArgs"
+            totalMemory=$(awk '/MemTotal/ {print int($2/1024/1024+0.5)}' /proc/meminfo)
+            memoryOverhead=3
             ;;
           Windows)
-            jvmArgs="-Xmx3g -XX:MaxMetaspaceSize=756m $jvmArgs"
+            # Fetch and parse memory in MB, then convert to GB
```
rjrjr, line 38: Wow, are you actually set up to test this?

RBusarow: ChatGPT was kind enough to show me how to do this.  It does seem to work, but I do not take credit. :)

PR #1181, opened 2024-02-08
publish snapshots and releases using macos-latest runners

RBusarow:


.github/workflows/publish-release.yml
```diff
@@ -1,38 +1,38 @@
...
+  publish-release:
+    runs-on: macos-latest
+    if: github.repository == 'square/workflow-kotlin'
+    timeout-minutes: 35
 
-    steps :
-      - uses : actions/checkout@v3
-      - uses : gradle/wrapper-validation-action@v1
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
```
rjrjr, line 13: Not released yet?

rjrjr: `kotlin.yml` is still using `v3`. Should we be consistent?

rjrjr (approved):

.github/workflows/publish-release.yml
rjrjr: @RBusarow ping, enquiring minds want to know.

PR #1183, opened 2024-02-13
ScreenViewFactoryFinder main error message mentions Compose.

rjrjr:
Fixes the error message you see when missing a call to `ViewEnvironment.withComposeInteropSupport()`. A little odd since this module has no dependency on Compose, but the requirement is otherwise much too hard to discover.


crocsandcoffee (approved):
🎉

PR #1185, opened 2024-02-17
Create workflow-ui/core module.

MilanJovic92:
Create workflow-ui-core KMP module.
Copied implementation from workflow-ui-core-common to the new module.

### DIFF
--- a/artifacts.json
+++ b/artifacts.json
@@ -189,6 +189,60 @@
     "publicationName": "maven"
   },
   {
+    "gradlePath": ":workflow-ui:core",
+    "group": "com.squareup.workflow1",
+    "artifactId": "workflow-ui-core-iosarm64",
+    "description": "Workflow UI Core",
+    "packaging": "klib",
+    "javaVersion": 8,
+    "publicationName": "iosArm64"
+  },
+  {
+    "gradlePath": ":workflow-ui:core",
+    "group": "com.squareup.workflow1",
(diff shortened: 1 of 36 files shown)


steve-the-edwards:
@MilanJovic92  can you give more context on why we should do this or what it accomplishes? Can you link this to an issue?

MilanJovic92:
> @MilanJovic92 can you give more context on why we should do this or what it accomplishes? Can you link this to an issue?

@steve-the-edwards Of course. This PR copies everything from `workflow-ui-core-common` module which publishes the JVM library to the new module `workflow-ui-core` that will publish the KMP package.

The idea is to enable the creation of `Screen` and other helpful UI elements inside KMP projects that use `workflow-core`.

I haven't found the Issue for this, but I can create one if it's necessary. 

After and if this PR gets merged, the old `workflow-ui-core-common` should be deleted.

blakelee:
I was able to get Workflow working in my KMP project by making these changes as well as pulling in all of the classes from `workflow-ui:compose` and deleting `ViewEnvironmentWithComposeSupport.kt` along with a few other Android only functions. I'm sure there's more support needed per platform but Workflow would be great as a multiplatform option since the alternatives aren't nearly as good.

MilanJovic92:
@blakelee That's exactly what I tried to achieve with this PR. I would love to see Workflow with full-fledged support of KMP.

PR #1186, opened 2024-02-23
always check out the `github.head_ref` in CI

RBusarow:
The default behavior for the `actions/checkout` is to check out the `github.ref` commit.

In the case of `pull_request_target` triggers, the `github.ref` commit is the last commit on the target branch (typically `main`), instead of the PR branch:

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

> This event runs in the context of the base of the pull request, rather than in the context of the merge commit, as the pull_request event does.

Unfortunately, that means that unless we specify the correct PR branch, the `kotlin.yml` workflow jobs will just execute against the `main`.

### DIFF
--- a/.github/workflows/kotlin.yml
+++ b/.github/workflows/kotlin.yml
@@ -26,7 +26,10 @@ jobs :
     name : Build all
     runs-on : workflow-kotlin-test-runner-ubuntu-4core
     steps :
-      - uses : actions/checkout@v3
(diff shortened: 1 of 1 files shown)


rjrjr (approved):
I continue to be amazed at how hard it is to do this correctly. Are we really that weird? Or setup seems so clean.

PR #1201, opened 2024-04-09
Makes `CoveredByModal` public.

rjrjr:
Held off on this originally because we didn't have a use case. Now we do -- trying to keep a non-workflow system that does its own window management well behaved.


darshanparajuli (approved):
thank you!

RBusarow (approved):

PR #1202, opened 2024-04-22
Stop pretending everyone implements view state right.

rjrjr:
Makes the two sites where we call `View.restoreHierarchyState` defensive, catching and logging thrown exceptions.

We normally prefer to let exceptions fly (https://en.wikipedia.org/wiki/Fail-fast_system), but `View.onRestoreInstanceState` is a particularly unforgiving method. Naive implementations throw if a view of one type attempts to restore state written by one of another type, which can happen if view ids are mismanaged. When these errors cause crashes in large apps, they are unlikely to be discovered before release and can be hellish to track down.

Crashing in such a case is an overreaction when the alternative, typically losing scroller positions or partially entered text values on rotation or when popping a back stack, is such a non-event.


pforhan (approved):
LGTM

ijwhelan (approved):

PR #1207, opened 2024-06-25
BREAKING: Compose overhaul makes ViewEnvironment less in your face.

rjrjr:
* Introduces `LocalWorkflowEnvironment` and eliminates explicit `ViewEnvironment` parameters from `ScreenComposableFactory.Content`, `ComposeScreen.Content`, etc. It is put in place automatically by the default implementation of `ScreenComposableFactoryFinder`.

* Adds optional `CompositionRoot` argument to `ViewEnvironment.withComposeInteropSupport()`. `CompositionRoot` is our existing hook to ensure `CompositionLocal`s (e.g. for UI themes) are put in play above the `@Composable Content()` function invoked for any `Screen`.

* Replaces `ViewEnvironment.withCompositionRoot` with `@Composable fun ViewEnvironment.RootScreen`, as our preferred Compose-friendly alternative to `WorkflowLayout`

An `Activity` that uses `setContent {}` instead of `setContentView()` can now kick things off like so:

```kotlin
  override fun onCreate(savedInstanceState: Bundle?) {
    super.onCreate(savedInstanceState)

    val environment : ViewEnvironment = ViewEnvironment.EMPTY +
      // ...
      .withComposeInteropSupport()

    val rootScreen by RootWorkflow.renderAsState(
      props = Unit,
      onOutput = {},
    )

    setContent {
      environment.RootScreen(rootScreen)
    }
  }
```

### DIFF
--- a/workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/LocalWorkflowEnvironment.kt
+++ b/workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/LocalWorkflowEnvironment.kt
@@ -1,0 +1,10 @@
+package com.squareup.workflow1.ui.compose
+
+import androidx.compose.runtime.ProvidableCompositionLocal
+import androidx.compose.runtime.compositionLocalOf
+import com.squareup.workflow1.ui.ViewEnvironment
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+
+@WorkflowUiExperimentalApi
+public val LocalWorkflowEnvironment: ProvidableCompositionLocal<ViewEnvironment> =
+  compositionLocalOf { ViewEnvironment.EMPTY }
--- a/workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ScreenComposableFactoryFinder.kt
+++ b/workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ScreenComposableFactoryFinder.kt
@@ -1,5 +1,7 @@
 package com.squareup.workflow1.ui.compose
 
+import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.runtime.remember
 import com.squareup.workflow1.ui.EnvironmentScreen
 import com.squareup.workflow1.ui.NamedScreen
 import com.squareup.workflow1.ui.Screen
@@ -22,8 +24,8 @@ public interface ScreenComposableFactoryFinder {
     @Suppress("UNCHECKED_CAST")
     return factoryOrNull
       ?: (rendering as? ComposeScreen)?.let {
-        ScreenComposableFactory<ComposeScreen> { rendering, environment ->
-          rendering.Content(environment)
+        ScreenComposableFactory<ComposeScreen> { rendering ->
+          rendering.Content()
         } as ScreenComposableFactory<ScreenT>
       }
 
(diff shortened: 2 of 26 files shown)


workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/LocalWorkflowEnvironment.kt
```diff
@@ -0,0 +1,10 @@
+package com.squareup.workflow1.ui.compose
+
+import androidx.compose.runtime.ProvidableCompositionLocal
+import androidx.compose.runtime.compositionLocalOf
+import com.squareup.workflow1.ui.ViewEnvironment
+import com.squareup.workflow1.ui.WorkflowUiExperimentalApi
+
+@WorkflowUiExperimentalApi
+public val LocalWorkflowEnvironment: ProvidableCompositionLocal<ViewEnvironment> =
+  compositionLocalOf { ViewEnvironment.EMPTY }
```
rjrjr, line 10: @zach-klippenstein @vgonda @ekeitho I was torn here between `compositionLocalOf` or `staticCompositionLocalOf`. Seemed like the former gave us a better hope of having bookkeeping in place that would reduce the impact of updates to the environment?

ekeitho: yeah `staticCompositionLocalOf` should be used for things that are unlikely to change often, because it recomposes the entire scope that it's provided in. so things like `LocalDensity` is a good example usage of `staticCompositionLocalOf` - however `compositionLocalOf` would be a good use if it's value changes more frequently and so Compose does more tracking of where the value of the composition local is read. so if `ViewEnviornment` is changed frequently i'd say `compositionLocalOf` is the better option here.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ScreenComposableFactoryFinder.kt
```diff
@@ -22,27 +24,32 @@ public interface ScreenComposableFactoryFinder {
...
-          val innerFactory = rendering.content.toComposableFactory(comboEnv)
-          innerFactory.Content(rendering.content, comboEnv)
-          // WorkflowRendering(rendering.content, comboEnv)
+        ScreenComposableFactory<EnvironmentScreen<*>> { rendering ->
+          val currentEnv = LocalWorkflowEnvironment.current
+          val innerFactory = rendering.content.toComposableFactory(
+            currentEnv + rendering.environment
+          )
+
+          val comboEnv = remember(currentEnv) { currentEnv + rendering.environment }
```
ekeitho, line 49: Should rendering.environment be included as a key in the remember function to ensure comboEnv is recomputed when it changes? Currently, it's only keyed on currentEnv, which may not cover all use cases.

ekeitho (approved):
really awesome stuff!

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ViewEnvironmentWithComposeSupport.kt
```diff
@@ -1,11 +1,34 @@
...
+ * for a pure Compose application. Makes the receiver available via [LocalWorkflowEnvironment]
+ * and runs the composition bound to [screen].
+ *
+ * Note that any app relying on stock navigation classes like
+ * [BackStackScreen][com.squareup.workflow1.ui.navigation.BackStackScreen] or
+ * [BodyAndOverlaysScreen][com.squareup.workflow1.ui.navigation.BodyAndOverlaysScreen]
+ * are not pure Compose, and must call [ViewEnvironment.withComposeInteropSupport] first.
+ */
+@WorkflowUiExperimentalApi
+@Composable public fun ViewEnvironment.RootScreen(
```
wardellbagby, line 23: Nit: I'm bikeshedding but I think we should stick with either `Screen` or `Rendering` as a naming convention for these functions.

wardellbagby (approved):
Everything tracks and makes sense to me at a high-level but I'm realizing that I need to take more of a deep-dive into the Workflow Compose integration. I've been taking it for granted.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ScreenComposableFactoryFinder.kt
rjrjr: Yes, it can change.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ViewEnvironmentWithComposeSupport.kt
rjrjr: Couldn't agree more. Should be `screen`.

rjrjr: I should make a separate pass to catch old `rendering: Screen` instances :/

rjrjr:
@RBusarow any guesses as to why the checks aren't running on this PR?

rjrjr:
> @RBusarow any guesses as to why the checks aren't running on this PR?

Maybe lingering effects from the earlier outage. Closed and opened to try again.

PR #1219, opened 2024-08-06
Return the `Job` from `WorkflowLayout.take` so it can be canceled.

rjrjr:


wardellbagby (approved):

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLayout.kt
```diff
@@ -91,19 +92,23 @@ public class WorkflowLayout(
    * @param [collectionContext] additional [CoroutineContext] we want for the coroutine that is
    * launched to collect the renderings. This should not override the [CoroutineDispatcher][kotlinx.coroutines.CoroutineDispatcher]
    * but may include some other instrumentation elements.
+   *
+   * @return the [Job] started to collect [renderings], to allow callers to
+   * [cancel][Job.cancel] collection -- e.g., before calling [take] again with a new
+   * [renderings] flow
```
zach-klippenstein, line 98: Nit: Might be worth documenting that it's not _required_ that callers cancel the returned `Job`, and that the job will always be a child of any `Job` present in `collectionContext`.

zach-klippenstein (approved):

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLayout.kt
rjrjr: That's a good point, thanks.

```
   * @return the [Job] started to collect [renderings], to give callers the option to
   * [cancel][Job.cancel] collection -- e.g., before calling [take] again with a new
   * [renderings] flow. In most cases the caller can ignore this, interacting with
   * the [Job] is very unusual.
```

PR #1221, opened 2024-09-24
`safeAction`, `safeEventHandler`

rjrjr:
People are confused by the fact that a `WorkflowAction` can't assume that a sealed class / interface `StateT` is the same subtype that it was at render time when the action fires. And those that do understand it resent this boilerplate:

```kotlin
action {
  (state as? SpecificState)?.let { currentState ->
    // whatever
  }
}
```

So we introduce `StatefulWorkflow.safeAction` and `StatefulWorkflow.RenderContext.safeEventHandler` as conveniences to do that cast for you.

### DIFF
--- a/workflow-core/api/workflow-core.api
+++ b/workflow-core/api/workflow-core.api
@@ -155,6 +155,7 @@ public final class com/squareup/workflow1/Snapshots {
 public abstract class com/squareup/workflow1/StatefulWorkflow : com/squareup/workflow1/IdCacheable, com/squareup/workflow1/Workflow {
 	public fun <init> ()V
 	public final fun asStatefulWorkflow ()Lcom/squareup/workflow1/StatefulWorkflow;
+	public final fun defaultOnFailedCast (Ljava/lang/String;Lkotlin/reflect/KClass;Ljava/lang/Object;)V
 	public fun getCachedIdentifier ()Lcom/squareup/workflow1/WorkflowIdentifier;
 	public abstract fun initialState (Ljava/lang/Object;Lcom/squareup/workflow1/Snapshot;)Ljava/lang/Object;
 	public fun initialState (Ljava/lang/Object;Lcom/squareup/workflow1/Snapshot;Lkotlinx/coroutines/CoroutineScope;)Ljava/lang/Object;
--- a/workflow-core/src/commonMain/kotlin/com/squareup/workflow1/BaseRenderContext.kt
+++ b/workflow-core/src/commonMain/kotlin/com/squareup/workflow1/BaseRenderContext.kt
@@ -303,7 +303,7 @@ public interface BaseRenderContext<out PropsT, StateT, in OutputT> {
   public fun <E1, E2, E3, E4, E5, E6, E7, E8, E9> eventHandler(
     name: () -> String = { "eventHandler" },
     update: WorkflowAction<@UnsafeVariance PropsT, StateT, @UnsafeVariance OutputT>
-    .Updater.(E1, E2, E3, E4, E5, E6, E7, E8, E9) -> Unit
+      .Updater.(E1, E2, E3, E4, E5, E6, E7, E8, E9) -> Unit
   ): (E1, E2, E3, E4, E5, E6, E7, E8, E9) -> Unit {
     return { e1, e2, e3, e4, e5, e6, e7, e8, e9 ->
       actionSink.send(action(name) { update(e1, e2, e3, e4, e5, e6, e7, e8, e9) })
@@ -313,7 +313,7 @@ public interface BaseRenderContext<out PropsT, StateT, in OutputT> {
   public fun <E1, E2, E3, E4, E5, E6, E7, E8, E9, E10> eventHandler(
     name: () -> String = { "eventHandler" },
     update: WorkflowAction<@UnsafeVariance PropsT, StateT, @UnsafeVariance OutputT>
-    .Updater.(E1, E2, E3, E4, E5, E6, E7, E8, E9, E10) -> Unit
+      .Updater.(E1, E2, E3, E4, E5, E6, E7, E8, E9, E10) -> Unit
   ): (E1, E2, E3, E4, E5, E6, E7, E8, E9, E10) -> Unit {
     return { e1, e2, e3, e4, e5, e6, e7, e8, e9, e10 ->
       actionSink.send(action(name) { update(e1, e2, e3, e4, e5, e6, e7, e8, e9, e10) })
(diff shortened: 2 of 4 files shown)


workflow-core/src/commonMain/kotlin/com/squareup/workflow1/StatefulWorkflow.kt
```diff
@@ -73,7 +75,20 @@ public abstract class StatefulWorkflow<
 
   public inner class RenderContext internal constructor(
     baseContext: BaseRenderContext<PropsT, StateT, OutputT>
-  ) : BaseRenderContext<@UnsafeVariance PropsT, StateT, @UnsafeVariance OutputT> by baseContext
+  ) : BaseRenderContext<@UnsafeVariance PropsT, StateT, @UnsafeVariance OutputT> by baseContext {
+
+    public inline fun <reified CurrentStateT : Any> safeEventHandler(
```
rjrjr, line 80: Note the dirty trick here: we can't make `CurrentStateT: StateT`, because `StateT` is not required to extend `Any`, but `KClass` (provider of the `safeCast()` call) requires it. So we accept any old type for `CurrentStateT` and…why not?

zach-klippenstein:
```suggestion
    public inline fun <reified CurrentStateT : StateT & Any> safeEventHandler(
```

rjrjr: That works! 👯

zach-klippenstein (approved):

workflow-core/api/workflow-core.api
```diff
@@ -155,6 +155,7 @@ public final class com/squareup/workflow1/Snapshots {
 public abstract class com/squareup/workflow1/StatefulWorkflow : com/squareup/workflow1/IdCacheable, com/squareup/workflow1/Workflow {
 	public fun <init> ()V
 	public final fun asStatefulWorkflow ()Lcom/squareup/workflow1/StatefulWorkflow;
+	public final fun defaultOnFailedCast (Ljava/lang/String;Lkotlin/reflect/KClass;Ljava/lang/Object;)V
```
rjrjr, line 158: Weird that this is the only change from `apiDump`. Maybe its blind to deeper nesting?

rjrjr, line 158: Why is this the only change caught by `apiDump`?

afollestad (approved):
Thanks for doing this!

When `WorkflowActionFactory` is deprecated, we'll want to call out that `currentState` (in the receiver) is replaced with the `oldState` lambda parameter

rjrjr:
> Thanks for doing this!
> 
> When `WorkflowActionFactory` is deprecated, we'll want to call out that `currentState` (in the receiver) is replaced with the `oldState` lambda parameter

@afollestad I actually called it `currentState` in the function definitions, so the IDE autocomplete will provide that name for us. And yes, we will need to call that out.

rjrjr:
@afollestad obsessively tweaked the kdoc samples to say `currentState` instead of `oldState`.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/StatefulWorkflow.kt
```diff
@@ -73,7 +75,377 @@ public abstract class StatefulWorkflow<
...
+     *            setOutput(CanceledStart)
+     *          },
+     *          onStartGame =
+     *            context.safeEventHandler<NewGame, String, String> { currentState, x, o ->
+     *              state = Playing(currentState.gameType, PlayerInfo(x, o))
+     *            }
+     *        )
+     *      }
+     *
+     * This is not an uncommon case. Consider accidental rapid taps on
```
jeffhashfield, line 97: Do you think it would be helpful to mention an event handler does not synchronously execute its update lambda? I find a lot of people don't understand that it creates and queues a WorkflowAction to be executed later. (At least I think that's what happens. Hopefully I have that correct!) Understanding that can help make it clearer why there's time for a second button tap to queue a second WorkflowAction before the first WorkflowAction is executed and changes state/produces a new rendering without the event handler, and then those WorkflowActions are pulled off the queue and executed in sequence.

rjrjr: That's a really good callout Jeff, I'll add that.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/BaseRenderContext.kt
```diff
@@ -130,6 +130,45 @@ public interface BaseRenderContext<out PropsT, StateT, in OutputT> {
    * given [update] function, and immediately passes it to [actionSink]. Handy for
    * attaching event handlers to renderings.
    *
+   * It is important to understand that the [update] lambda you provide here
```
rjrjr, line 133: @jeffhashfield PTAL

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/StatefulWorkflow.kt
```diff
@@ -73,7 +75,381 @@ public abstract class StatefulWorkflow<
...
   public inner class RenderContext internal constructor(
     baseContext: BaseRenderContext<PropsT, StateT, OutputT>
-  ) : BaseRenderContext<@UnsafeVariance PropsT, StateT, @UnsafeVariance OutputT> by baseContext
+  ) : BaseRenderContext<@UnsafeVariance PropsT, StateT, @UnsafeVariance OutputT> by baseContext {
+
+    /**
+     * Like [eventHandler], but no-ops if [state][WorkflowAction.Updater.state] has
+     * changed to a different type than [CurrentStateT] by the time [update] fires.
+     *
+     * It is also important to understand that **even if [update] is called, there is
```
rjrjr, line 84: @jeffhashfield Also added a callout here to that doc.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/StatefulWorkflow.kt
```diff
@@ -73,7 +75,381 @@ public abstract class StatefulWorkflow<
...
+   *      state = when (game.ending) {
+   *        Quitting -> MaybeQuitting(currentState.playerInfo, game)
+   *        else -> GameOver(currentState.playerInfo, game)
+   *      }
+   *    }
+   *
+   * This is not an uncommon case. Consider accidental rapid taps on
+   * a button, where the first tap event moves the receiving [StatefulWorkflow]
+   * to a new state. There is no reason to expect that the later taps will not
+   * fire the (now stale) event handler a few more times. No promise can be
```
jeffhashfield, line 421: > the (now stale) event handler

Is it confusing to use the term `event handler` here? The WorkflowAction created by `safeAction` is probably being used to handle an event, but it might be better to use a different term that isn't easily confused with the `eventHandler` method. How about `the (now stale) [update] lambda`? That change would also work well in the `safeEventHandler` documentation.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/BaseRenderContext.kt
jeffhashfield: I like it!

PR #1224, opened 2024-10-21
Makes `TextController` support equals

jkwiz:
Allows comparing `TextController` by the current `textValue`.


rjrjr:
LGTM, but could you add a unit test?

PR #1226, opened 2024-11-02
`action {}`, `eventHandler {}` with no debugging name is deprecated

rjrjr:
Logs of actions are nearly useless if they have no name. Let's stop making things so convenient that we go blind.

### DIFF
--- a/samples/dungeon/timemachine/src/test/java/com/squareup/sample/timemachine/TimeMachineWorkflowTest.kt
+++ b/samples/dungeon/timemachine/src/test/java/com/squareup/sample/timemachine/TimeMachineWorkflowTest.kt
@@ -25,7 +25,7 @@ class TimeMachineWorkflowTest {
     val delegateWorkflow = Workflow.stateful<String, Nothing, DelegateRendering>(
       initialState = "initial",
       render = { renderState ->
-        DelegateRendering(renderState, setState = eventHandler { s -> state = s })
+        DelegateRendering(renderState, setState = eventHandler("") { s -> state = s })
       }
     )
     val clock = TestTimeSource()
--- a/workflow-core/src/commonMain/kotlin/com/squareup/workflow1/BaseRenderContext.kt
+++ b/workflow-core/src/commonMain/kotlin/com/squareup/workflow1/BaseRenderContext.kt
@@ -137,7 +137,7 @@ public interface BaseRenderContext<out PropsT, StateT, in OutputT> {
    *    return SomeScreen(
    *      onClick = {
    *        context.actionSink.send(
-   *          action { state = SomeNewState }
+   *          action("onClick") { state = SomeNewState }
    *        }
    *      }
    *    )
@@ -145,7 +145,7 @@ public interface BaseRenderContext<out PropsT, StateT, in OutputT> {
    *  with this:
    *
    *    return SomeScreen(
-   *      onClick = context.eventHandler { state = SomeNewState }
+   *      onClick = context.eventHandler("onClick") { state = SomeNewState }
    *    )
    *
    * Notice how your [update] function is passed to the [actionSink][BaseRenderContext.actionSink]
@@ -161,7 +161,7 @@ public interface BaseRenderContext<out PropsT, StateT, in OutputT> {
    *        MyAnalytics.log("SomeScreen was clicked")
    *
    *        context.actionSink.send(
-   *          action {
+   *          action("onClick") {
    *            // This happens eventually.
    *            state = SomeNewState
    *          }
@@ -173,6 +173,31 @@ public interface BaseRenderContext<out PropsT, StateT, in OutputT> {
    * as a debugging aid
    * @param update Function that defines the workflow update.
    */
+  @Suppress("DEPRECATION")
+  public fun eventHandler(
+    name: String,
+    // Type variance issue: https://github.com/square/workflow-kotlin/issues/891
+    update: WorkflowAction<
+      @UnsafeVariance PropsT,
+      StateT,
+      @UnsafeVariance OutputT
+      >.Updater.() -> Unit
+  ): () -> Unit = eventHandler({ name }, update)
+
+  @Deprecated(
+    "Always provide a debugging name",
+    ReplaceWith("eventHandler(\"TODO: debugging name\", onFailedCast, update)")
+  )
+  public fun eventHandler(
+    // Type variance issue: https://github.com/square/workflow-kotlin/issues/891
+    update:
+    WorkflowAction<@UnsafeVariance PropsT, StateT, @UnsafeVariance OutputT>.Updater.() -> Unit
+  ): () -> Unit = eventHandler("eventHandler", update)
+
+  @Deprecated(
+    "Use a static string.",
+    ReplaceWith("eventHandler(\"TODO: debugging name\", onFailedCast, update)")
+  )
   public fun eventHandler(
     name: () -> String = { "eventHandler" },
     // Type variance issue: https://github.com/square/workflow-kotlin/issues/891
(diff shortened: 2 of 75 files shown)


samples/dungeon/timemachine/src/test/java/com/squareup/sample/timemachine/TimeMachineWorkflowTest.kt
```diff
@@ -25,7 +25,7 @@ class TimeMachineWorkflowTest {
     val delegateWorkflow = Workflow.stateful<String, Nothing, DelegateRendering>(
       initialState = "initial",
       render = { renderState ->
-        DelegateRendering(renderState, setState = eventHandler { s -> state = s })
+        DelegateRendering(renderState, setState = eventHandler("") { s -> state = s })
```
steve-the-edwards, line 28: Should we have a `require(!name.isEmpty())` or is that just mean/unecessary?

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/BaseRenderContext.kt
```diff
@@ -207,6 +278,32 @@ public interface BaseRenderContext<out PropsT, StateT, in OutputT> {
...
+    name: String,
+    update: WorkflowAction<@UnsafeVariance PropsT, StateT, @UnsafeVariance OutputT>.Updater.(
+      E1,
+      E2,
+      E3
+    ) -> Unit
+  ): (E1, E2, E3) -> Unit = eventHandler({ name }, update)
+
+  @Deprecated(
+    "Requiring a function to provide a debugging name was absurd, we're sorry.",
```
steve-the-edwards, line 304: Here (and elsewhere with this pattern) you mean the fact that we made `name` a function? It reads confusing, because we now are requiring you to provide a debugging name, but not via a function.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/BaseRenderContext.kt
```diff
@@ -173,6 +173,31 @@ public interface BaseRenderContext<out PropsT, StateT, in OutputT> {
...
+  @Suppress("DEPRECATION")
+  public fun eventHandler(
+    name: String,
+    // Type variance issue: https://github.com/square/workflow-kotlin/issues/891
+    update: WorkflowAction<
+      @UnsafeVariance PropsT,
+      StateT,
+      @UnsafeVariance OutputT
+      >.Updater.() -> Unit
+  ): () -> Unit = eventHandler({ name }, update)
```
steve-the-edwards, line 185: I'm confused why this one delegates to the original `name: () -> String = { "eventHandler" }` one that we have deprecated? 

Why isn't this one like this:

```suggestion
  ): () -> Unit {
    return {
      actionSink.send(action(name, update))
    }
  }
```

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/BaseRenderContext.kt
```diff
@@ -173,6 +173,31 @@ public interface BaseRenderContext<out PropsT, StateT, in OutputT> {
...
+
+  @Deprecated(
+    "Always provide a debugging name",
+    ReplaceWith("eventHandler(\"TODO: debugging name\", onFailedCast, update)")
+  )
+  public fun eventHandler(
+    // Type variance issue: https://github.com/square/workflow-kotlin/issues/891
+    update:
+    WorkflowAction<@UnsafeVariance PropsT, StateT, @UnsafeVariance OutputT>.Updater.() -> Unit
+  ): () -> Unit = eventHandler("eventHandler", update)
```
steve-the-edwards, line 195: why do we have to introduce these versions without the `name` parameter just to deprecate it? Is this about the way in which the deprecation detection works with the optional parameter?

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/StatefulWorkflow.kt
```diff
@@ -76,6 +76,16 @@ public abstract class StatefulWorkflow<
   public inner class RenderContext internal constructor(
     baseContext: BaseRenderContext<PropsT, StateT, OutputT>
   ) : BaseRenderContext<@UnsafeVariance PropsT, StateT, @UnsafeVariance OutputT> by baseContext {
+    public inline fun <reified CurrentStateT : StateT & Any> safeEventHandler(
```
steve-the-edwards, line 79: not sure I understand why this one isn't deprecated? (Or why its introduced?)

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
```diff
@@ -230,7 +230,7 @@ class RenderWorkflowInTest {
         render = { _, renderState ->
           Pair(
             renderState,
-            { newState -> actionSink.send(action { state = newState }) }
+            { newState -> actionSink.send(action("") { state = newState }) }
```
steve-the-edwards, line 233: we could add a `testAction` in the testing package to avoid this kind of boilerplate/frustration?

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/BaseRenderContext.kt
rjrjr: Yes, exactly.

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
rjrjr: That's a good point. Or really, the tests in question could whip that up themselves, it's not like this is a common problem for feature developers.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/BaseRenderContext.kt
rjrjr: D'oh.

rjrjr: Oh, because that would make it recursive. I'd have to make this PR even bigger.

Thinking about it, _why_ deprecate the lazy ones? The hard part was getting all these new entry points defined; can definitely see a benefit to having the lazy option; and you'll note I didn't deprecate the fundamental lazy `action {}` variants.

I guess I was…being lazy.

steve-the-edwards: I'm ok with keeping the lazy ones, but we can't keep them as is because they have the default `= { "eventHandler" }` which is what we're trying to avoid, no?

samples/dungeon/timemachine/src/test/java/com/squareup/sample/timemachine/TimeMachineWorkflowTest.kt
rjrjr: I'm still torn on this. We definitely shouldn't be throwing exceptions for empty output from the `name: () -> String` variants, that seems like it would be a major landmine. And really, even the static string version seems ripe to blow up only after release -- you enter a a state that is less common, you forgot to unit test it, and suddenly an obscure feature is blowing up. I don't think we should do it.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/BaseRenderContext.kt
rjrjr: Ah of course. Will fix.

rjrjr: Bah. I don't have lazy ones for `safeEventHandler` because it seemed ridiculous. 

Flip flopping again. Let's deprecate them, and I'll fix the deprecation message.

The 0.01% of places where laziness might possibly matter can just use `action({}) {}` directly.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/StatefulWorkflow.kt
rjrjr: Oops, lost the deprecation annotation. Thanks.

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
rjrjr: Now that I've convinced myself we shouldn't police for empty strings, inclined to leave the tests as is.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/BaseRenderContext.kt
rjrjr:
```kotlin
  @Deprecated(
    "Use a static string.",
```

steve-the-edwards (approved):
Got it. Makes sense now!

PR #1230, opened 2024-11-11
make sure events are not assumed and always update to lifecycle obser…

ekeitho:
…ver event changes

### DIFF
--- a/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/WorkflowLifecycleOwner.kt
+++ b/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/WorkflowLifecycleOwner.kt
@@ -8,6 +8,8 @@ import androidx.annotation.VisibleForTesting
 import androidx.annotation.VisibleForTesting.Companion.PRIVATE
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.Lifecycle.Event
+import androidx.lifecycle.Lifecycle.Event.ON_DESTROY
+import androidx.lifecycle.Lifecycle.Event.ON_PAUSE
 import androidx.lifecycle.Lifecycle.State.DESTROYED
 import androidx.lifecycle.Lifecycle.State.INITIALIZED
 import androidx.lifecycle.Lifecycle.State.RESUMED
@@ -190,12 +192,12 @@ internal class RealWorkflowLifecycleOwner(
       oldLifecycle?.removeObserver(this)
       parentLifecycle?.addObserver(this)
     }
-    updateLifecycle()
+    updateLifecycle(Event.ON_START)
   }
 
   override fun onViewDetachedFromWindow(v: View) {
     viewIsAttachedToWindow = false
-    updateLifecycle()
+    updateLifecycle(ON_PAUSE)
   }
 
   /** Called when the [parentLifecycle] changes state. */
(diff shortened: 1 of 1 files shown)


workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/WorkflowLifecycleOwner.kt
```diff
@@ -190,66 +192,44 @@ internal class RealWorkflowLifecycleOwner(
...
       parentLifecycle?.addObserver(this)
     }
-    updateLifecycle()
+    updateLifecycle(Event.ON_START)
   }
 
   override fun onViewDetachedFromWindow(v: View) {
     viewIsAttachedToWindow = false
-    updateLifecycle()
+    updateLifecycle(ON_PAUSE)
```
rjrjr, line 200: This feels really good.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/WorkflowLifecycleOwner.kt
```diff
@@ -190,66 +192,44 @@ internal class RealWorkflowLifecycleOwner(
...
 
   /** Called when the [parentLifecycle] changes state. */
   override fun onStateChanged(
     source: LifecycleOwner,
     event: Event
   ) {
-    updateLifecycle()
+    // we should always consume what the new event is from onStateChanged. if we try to
+    // assume the event like we did in the past, we can get into scenarios where we try to put
+    // a view in created, when it's parent hasn't completed fully it's super.onCreate yet.
```
rjrjr, line 210:
```suggestion
    // a view in created, when its parent hasn't completed fully its super.onCreate yet.
```

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/WorkflowLifecycleOwner.kt
```diff
@@ -190,66 +192,44 @@ internal class RealWorkflowLifecycleOwner(
...
+    updateLifecycle(event)
   }
 
   override fun destroyOnDetach() {
     if (!destroyOnDetach) {
       destroyOnDetach = true
-      updateLifecycle()
+      updateLifecycle(Event.ON_DESTROY)
```
rjrjr, line 217: This subverts the purpose of `destroyOnDetach()`, which is to make sure things are still alive while the view is visible. I think instead you want to change `onViewDetachedFromWindow` to look at `destroyOnDetach` and use that to choose between `ON_PAUSE` and `ON_DESTROY`. Maybe also need to change the guard clause at the top of `onViewAttachedToWindow` to look directly at `destroyOnDetach`.

rjrjr (approved):
Seems like the right direction except it breaks `destroyOnDetach`.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/WorkflowLifecycleOwner.kt
```diff
@@ -225,6 +225,9 @@ internal class RealWorkflowLifecycleOwner(
     }
 
     localLifecycle.currentState = when {
+      event != null -> {
+        event.targetState
```
zach-klippenstein, line 229: Don't we need to make sure the local lifecycle isn't greater than the parent? E.g. if the parent hasn't STARTED yet, we shouldn't either.

ekeitho: I think this should be renamed to `parentEvent`  since the events should originate from the parent lifecycle. For example, onStateChanged ensures that the local lifecycle mirrors the parent's state and doesn't exceed it.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/WorkflowLifecycleOwner.kt
```diff
@@ -190,32 +191,38 @@ internal class RealWorkflowLifecycleOwner(
...
   ) {
-    updateLifecycle()
+    updateLifecycle(event.targetState)
   }
 
   override fun destroyOnDetach() {
     if (!destroyOnDetach) {
       destroyOnDetach = true
-      updateLifecycle()
+      updateLifecycle(Event.ON_DESTROY.targetState)
```
rjrjr, line 219: Just to make sure the comment doesn't get lost across updates: we can't ship this way. 

- [ ] `destroyOnDetach()` has to mean "move to destroyed the next time `onViewDetachedFromWindow()` is called.

PR #1232, opened 2024-11-14
Add public `debuggingName` to `WorkflowAction`, use it in default `toString()`

steve-the-edwards:
This is a more useful default implementation and allows clearer overriding via `debuggingName`.


rjrjr:
> This will also block data class WorkflowActions.

Block them, or just prevent them from generating horrible `toString()` implementations?

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/WorkflowAction.kt
```diff
@@ -31,6 +31,9 @@ import kotlin.jvm.JvmOverloads
  */
 public abstract class WorkflowAction<in PropsT, StateT, out OutputT> {
 
+  // Debugging name, will be returned by `toString()`
+  public open val name: String = CommonKClassTypeNamer.uniqueName(this::class)
```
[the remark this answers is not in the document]
steve-the-edwards: we could, but we don't have control over that. if someone makes a massive `data class MyAction: WorkflowAction` with a tonne of members (as we have seen) then `toString()` will still print all those out.

`final toString()` prevents that and make sure that `WorkflowAction` logging is always based on the `name` which can only be so big (in reality)

steve-the-edwards:
> > This will also block data class WorkflowActions.
> 
> Block them, or just prevent them from generating horrible `toString()` implementations?

I don't think there is a way that I've seen to subclass with a `data class` if `toString()` is `final` as you can't opt-out of the `toString()` implementation in the `data class`.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/WorkflowAction.kt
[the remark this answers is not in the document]
rjrjr: Right, but I'm saying the real problem is that we're using `toString()` in logging at all, which we do because we had no alternative. But now `name` exists and we can log that — avoiding any call to `toString()` in production code as nature intended. That seems like a complete solution, I don't see why we need to mess with `toString()` as well.

rjrjr:
> > > This will also block data class WorkflowActions.
> > 
> > 
> > Block them, or just prevent them from generating horrible `toString()` implementations?
> 
> I don't think there is a way that I've seen to subclass with a `data class` if `toString()` is `final` as you can't opt-out of the `toString()` implementation in the `data class`.

This compiles:

```kotlin

abstract class Foo {
  final override fun toString(): String {
    return super.toString()
  }
}

data class Bar(val s: String) : Foo() {

}
```

But I still think taking over toString() is a bridge too far.

steve-the-edwards:
> > > > This will also block data class WorkflowActions.
> > > 
> > > 
> > > Block them, or just prevent them from generating horrible `toString()` implementations?
> > 
> > 
> > I don't think there is a way that I've seen to subclass with a `data class` if `toString()` is `final` as you can't opt-out of the `toString()` implementation in the `data class`.
> 
> This compiles:
> 
> ```kotlin
> abstract class Foo {
>   final override fun toString(): String {
>     return super.toString()
>   }
> }
> 
> data class Bar(val s: String) : Foo() {
> 
> }
> ```
> 
> But I still think taking over toString() is a bridge too far.

Nice! My assumption was wrong then. This is even better. let's talk offline about why you don't like `final toString()`

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/WorkflowAction.kt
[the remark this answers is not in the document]
steve-the-edwards: > Right, but I'm saying the real problem is that we're using toString() in logging at all, which we do because we had no alternative. But now name exists and we can log that — avoiding any call to toString() in production code as nature intended. That seems like a complete solution, I don't see why we need to mess with toString() as well.


Ahhh IC. that makes a tonne of sense

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/StatefulWorkflow.kt
```diff
@@ -1007,6 +1007,8 @@ public fun <PropsT, StateT, OutputT, RenderingT>
     name: () -> String,
     update: WorkflowAction<PropsT, StateT, OutputT>.Updater.() -> Unit
   ): WorkflowAction<PropsT, StateT, OutputT> = object : WorkflowAction<PropsT, StateT, OutputT>() {
+  override val debuggingName: String
```
rjrjr, line 1010: Should update the `@param` for `name` to mention `debuggingName` instead of `toString()`. Just to draw people's attention to the better thing.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/WorkflowAction.kt
```diff
@@ -31,6 +31,9 @@ import kotlin.jvm.JvmOverloads
  */
 public abstract class WorkflowAction<in PropsT, StateT, out OutputT> {
 
+  // Debugging name, will be returned by `toString()`
```
rjrjr, line 34:
```suggestion
  /** Debugging name. Handy for logging and returned by `toString()`. */
```

workflow-testing/src/test/java/com/squareup/workflow1/testing/RealRenderTesterTest.kt
```diff
@@ -48,6 +48,9 @@ internal class RealRenderTesterTest {
   private interface OutputWhateverChild : Workflow<Unit, Unit, Unit>
   private interface OutputNothingChild : Workflow<Unit, Nothing, Unit>
 
+  // [WorkflowAction::toString] includes "@-${hashCode()}", so strip it.
+  private fun String.removeActionHashCodes(): String = replace(Regex("-@([0-9]*)"), "")
```
rjrjr, line 52: Not sure where `toString()` is leaking in, but can you instead change this test to rely on `debugginName` directly and avoid the need for this function? Same for the other instances.

rjrjr (approved):

workflow-testing/src/test/java/com/squareup/workflow1/testing/RealRenderTesterTest.kt
steve-the-edwards: This one is a good candidate for that, but not the `TracingWorkflowInterceptorTest` as we want that to use the actual `toString()` or the traces won't be as useful.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/StatefulWorkflow.kt
steve-the-edwards: Done.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/WorkflowAction.kt
steve-the-edwards: Done.

PR #1234, opened 2024-11-19
Refactor to ComposeLifecycleOwner for Better Lifecycle Synchronization

ekeitho:
Refactor LifecycleOwner to ComposeLifecycleOwner for better lifecycle management

Extracted anonymous `LifecycleOwner` & introduced `RememberObserver` implementation into a reusable `ComposeLifecycleOwner` class. This now synchronizes its lifecycle with the parent Lifecycle & immediately synchornizes on creation of `ComposeLifecycleOwner` instead of where before we were synchronized to the parent **AFTER COMPOSITION**.


workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposeLifecycleOwner.kt
```diff
@@ -0,0 +1,75 @@
...
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.LifecycleEventObserver
+import androidx.lifecycle.LifecycleOwner
+import androidx.lifecycle.LifecycleRegistry
+
+/**
+ * A custom [LifecycleOwner] that synchronizes its lifecycle with a parent [Lifecycle] and
+ * integrates with Jetpack Compose's lifecycle through [RememberObserver].
+ *
+ * **Purpose:**
```
zach-klippenstein, line 13: Nit: You can use markdown headers (e.g. `## Purpose:`)

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposeLifecycleOwner.kt
```diff
@@ -0,0 +1,75 @@
...
+ */
+internal class ComposeLifecycleOwner(
+  private val parentLifecycle: Lifecycle
+) : LifecycleOwner, RememberObserver {
+
+  private val registry = LifecycleRegistry(this)
+  override val lifecycle: Lifecycle
+    get() = registry
+
+  private val parentObserver = LifecycleEventObserver { _, event ->
```
zach-klippenstein, line 48: Can we just implement this directly in this class? Saves one lambda allocation.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposeLifecycleOwner.kt
```diff
@@ -0,0 +1,75 @@
...
+    // Any time the parent lifecycle changes state, perform the same change on our lifecycle.
+    registry.handleLifecycleEvent(event)
+  }
+
+  init {
+    // We do this right away to ensure that the parent lifecycle is observed as soon as possible
+    // to prevent any inconsistencies.
+    parentLifecycle.addObserver(parentObserver)
```
zach-klippenstein, line 56: Constructors should not have side effects. Please pull this code into the `remember` lambda.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/WorkflowRendering.kt
```diff
@@ -101,34 +95,9 @@ public fun WorkflowRendering(
...
-      // If we're leaving the composition it means the WorkflowRendering is either going away itself
-      // or about to switch to an incompatible rendering – either way, this lifecycle is dead. Note
-      // that we can't transition from INITIALIZED to DESTROYED – the LifecycleRegistry will throw.
-      // WorkflowLifecycleOwner has this same check.
-      if (lifecycleOwner.registry.currentState != INITIALIZED) {
-        lifecycleOwner.registry.currentState = DESTROYED
-      }
-    }
+  val lifecycleOwner = remember(parentLifecycle) {
+    ComposeLifecycleOwner(parentLifecycle)
```
zach-klippenstein, line 100: (See comment on other file)
```suggestion
    ComposeLifecycleOwner(parentLifecycle).also(parentLifecycle::addObserver)
```

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposeLifecycleOwner.kt
```diff
@@ -0,0 +1,75 @@
...
+    onForgotten()
+  }
+
+  override fun onForgotten() {
+    parentLifecycle.removeObserver(parentObserver)
+
+    // If we're leaving the composition, ensure the lifecycle is cleaned up
+    if (registry.currentState != Lifecycle.State.INITIALIZED) {
+      registry.currentState = Lifecycle.State.DESTROYED
+    }
```
zach-klippenstein, line 72: One behavior change this introduces that we might not want is that it will destroy the child lifecycle before sending a new one, if the parent lifecycle instance changes. I don't think the parent instance should ever change though, unless `movableContentOf` is used or something, so it's probably not gonna be an issue immediately.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposeLifecycleOwner.kt
```diff
@@ -0,0 +1,75 @@
...
+    get() = registry
+
+  private val parentObserver = LifecycleEventObserver { _, event ->
+    // Any time the parent lifecycle changes state, perform the same change on our lifecycle.
+    registry.handleLifecycleEvent(event)
+  }
+
+  init {
+    // We do this right away to ensure that the parent lifecycle is observed as soon as possible
+    // to prevent any inconsistencies.
```
zach-klippenstein, line 55: We can be more specific: "We need to synchronize the lifecycles before the child ever even sees the lifecycle because compose's contract tries to guarantee that the lifecycle is in at least the `CREATED` state by the time composition is actually running. If we don't synchronize the lifecycles right away, then we break that invariant. One concrete case of this is that `SavedStateRegistry` requires its lifecycle to be `CREATED` before reading values from it, and consuming values from an SSR is a valid thing to do from composition directly, and in fact `AndroidComposeView` itself does this."

zach-klippenstein (approved):

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposeLifecycleOwner.kt
```diff
@@ -0,0 +1,74 @@
...
+    parentLifecycle.addObserver(parentObserver)
+  }
```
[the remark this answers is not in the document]
zach-klippenstein: Nah we don't need another composable, although I might move rememberChildLifecycle to this file and make this class private.

zach-klippenstein:
Even if we can't figure out how to test the original bug, we could at least have a test that ensures the a child composable sees the correct lifecycle state immediately.

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposeLifecycleOwner.kt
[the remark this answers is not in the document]
ekeitho: i can move over rememberChildLifecycle now & make that class private

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposeLifecycleOwner.kt
ekeitho: Interesting, i guess that means we can just not key off parent lifecycle on the remember call

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposeLifecycleOwner.kt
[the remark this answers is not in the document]
ekeitho: I've moved everything into `ComposeLifecycleOwner`

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposeLifecycleOwner.kt
ekeitho: I made the update to ensure this behavior didn't change since i'd hope it would work with `movableContentOf` & wrote a test for it too: https://github.com/square/workflow-kotlin/pull/1234/files#diff-be84e80fcb3d58b96b21bf4311c0bcdc15c6f2d0d90b6628ce1f220bd012301eR85

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposeLifecycleOwner.kt
```diff
@@ -0,0 +1,103 @@
...
+    source: LifecycleOwner,
+    event: Event
+  ) {
+    registry.handleLifecycleEvent(event)
+  }
+
+  companion object {
+    fun installOn(parentLifecycle: Lifecycle): ComposeLifecycleOwner {
+      return ComposeLifecycleOwner(parentLifecycle).also {
+        // We need to synchronize the lifecycles before the child ever even sees the lifecycle
```
ekeitho, line 92: i want to inline this comment appropriately deferring against use method reference for this one

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposeLifecycleOwner.kt
zach-klippenstein: I'm guessing you didn't push that yet? It doesn't look changed to me. This should push this code to be even more similar to the View one.

ekeitho: yeah i removed `parentLifecycle` from the remember key. check out: https://github.com/square/workflow-kotlin/pull/1234/files#diff-be84e80fcb3d58b96b21bf4311c0bcdc15c6f2d0d90b6628ce1f220bd012301eR76

it will retain the same  instance of `LifecycleOwner` regardless of parent lifecycle changing.

rjrjr (approved):

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposeLifecycleOwner.kt
```diff
@@ -0,0 +1,103 @@
...
+import androidx.lifecycle.LifecycleEventObserver
+import androidx.lifecycle.LifecycleOwner
+import androidx.lifecycle.LifecycleRegistry
+
+/**
+ * Returns a [LifecycleOwner] that is a mirror of the current [LocalLifecycleOwner] until this
+ * function leaves the composition. Similar to [WorkflowLifecycleOwner] for views, but a
+ * bit simpler since we don't need to worry about attachment state.
+ */
+@Composable internal fun rememberChildLifecycleOwner(
```
rjrjr, line 18: I don't see this being called anywhere outside of the test.

rjrjr: Oh, derp, you're just fixing the existing call.

rjrjr (approved):
:shipit:

workflow-ui/compose/src/main/java/com/squareup/workflow1/ui/compose/ComposeLifecycleOwner.kt
zach-klippenstein: Oh, that's also incorrect because if the parent changes then we'll be syncing to the wrong one. You need to add some code so the remember function can update the `parentLifecycle` on this new class, and make sure we remove/add observers correctly when that happens (without necessarily destroying the child).

ekeitho: i think i got what you mean now: https://github.com/square/workflow-kotlin/pull/1234/commits/953e2ab89b67999574ba5953255318a4d3bf8636

PR #1237, opened 2024-12-04
Remove Deprecated Worker, eventHandler, and action APIs

steve-the-edwards:
Some have been long deprecated, others more recently deprecated (the anonymous actions and eventHandlers - but we have not received any concerns regarding removing them - https://github.com/square/workflow-kotlin/discussions/1236, https://workflow-community.slack.com/archives/CHTFPR277/p1732635078961279?thread_ts=1730740915.096009&cid=CHTFPR277)

Also, prefix `eventHandler` `debuggingName`s with `eH:`

### DIFF
--- a/workflow-core/src/commonMain/kotlin/com/squareup/workflow1/Worker.kt
+++ b/workflow-core/src/commonMain/kotlin/com/squareup/workflow1/Worker.kt
@@ -6,7 +6,6 @@ package com.squareup.workflow1
 import com.squareup.workflow1.Worker.Companion.create
 import com.squareup.workflow1.Worker.Companion.from
 import com.squareup.workflow1.Worker.Companion.fromNullable
-import kotlinx.coroutines.FlowPreview
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.FlowCollector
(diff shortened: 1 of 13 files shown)


workflow-core/src/commonMain/kotlin/com/squareup/workflow1/Worker.kt
```diff
@@ -247,10 +202,7 @@ public interface Worker<out OutputT> {
      * builder functions that have the same output type.
      */
     public inline fun <reified OutputT> fromNullable(
-      // This could be crossinline, but there's a coroutines bug that will cause the coroutine
-      // to immediately resume on suspension inside block when it is crossinline.
-      // See https://youtrack.jetbrains.com/issue/KT-31197.
-      noinline block: suspend () -> OutputT?
+      crossinline block: suspend () -> OutputT?
```
steve-the-edwards, line 205: Also changed this to `crossinline` as that issue was fixed a number of years ago.

workflow-rx2/src/main/java/com/squareup/workflow1/rx2/RxWorkers.kt
```diff
@@ -81,17 +80,3 @@ public inline fun <reified T : Any> Single<out T?>.asWorker(): Worker<T> =
...
- * The [Completable] will be subscribed to when the [Worker] is started, and disposed when it is
- * cancelled.
- *
- * The key is required for this operator because there is no type information available to
- * distinguish workers.
- *
- * TODO: https://github.com/square/workflow-kotlin/issues/526 once this is removed.
- */
-@Suppress("DEPRECATION")
-public fun Completable.asWorker(): Worker<Nothing> = Worker.createSideEffect { await() }
```
steve-the-edwards, line 97: One would just use

```
runningSideEffect { myCompletable.await() }
```

zach-klippenstein (approved):
I love deleting code! This extremely violates semver but discussed offline and since we don't think this library has many users other than us, we don't care.

PR #1239, opened 2024-12-04
Add smol caching & a utility method

pyricau:


workflow-core/src/commonMain/kotlin/com/squareup/workflow1/WorkflowIdentifier.kt
```diff
@@ -58,6 +58,21 @@ public class WorkflowIdentifier internal constructor(
 
   private val proxiedIdentifiers = generateSequence(this) { it.proxiedIdentifier }
 
+  private val cachedToString by lazy {
```
zach-klippenstein, line 61: FYI: Each `by lazy` allocates an additional private object to store the property delegate. It's cheaper to just use a nullable backing property and initialize it yourself. Usually matters more for framework code, so we might want to avoid `by lazy` here.

E.g.
```kotlin
private var _cachedToString: String?
private val cachedToString: String
  get() = _cachedToString ?: calculateToString().also { _cachedToString = it }
```

zach-klippenstein (approved):

steve-the-edwards:
See build failure - we run in `explicit API` mode, so you need to add return types explicitly.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/WorkflowIdentifier.kt
```diff
@@ -58,6 +58,21 @@ public class WorkflowIdentifier internal constructor(
...
+      ?: proxiedIdentifiers
+        .joinToString { it.typeName }
+        .let { "WorkflowIdentifier($it)" }
+  }
+
+  /**
+   * Either a [KClass] or [KType] representing the "real" type that this identifier
+   * identifies – i.e. which is not an [ImpostorWorkflow].
+   */
+  public val realIdentifierType by lazy {
```
steve-the-edwards, line 72: This is what needs an explicit type declared.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/WorkflowIdentifier.kt
pyricau: Did that but made cachedToString volatile because I'm not 100% this is actually single threaded. volatile is still better than using synchronization primitives as there's no locking, and the downside is that we might sometimes compute it twice then set it, which is ok.

pyricau:
```
e: file:///home/runner/work/workflow-kotlin/workflow-kotlin/workflow-core/src/commonMain/kotlin/com/squareup/workflow1/WorkflowIdentifier.kt:69:57 Platform declaration clash: The following declarations have the same JVM signature (getRealIdentifierType()Lcom/squareup/workflow1/WorkflowIdentifierType;):
    fun `<get-realIdentifierType>`(): WorkflowIdentifierType defined in com.squareup.workflow1.WorkflowIdentifier
    fun getRealIdentifierType(): WorkflowIdentifierType defined in com.squareup.workflow1.WorkflowIdentifier
    ```
    
 I was surprised that the IDE was fine with it.

pyricau:
To avoid the conflict, renamed the field to `realType` which is better anyway

steve-the-edwards (approved):

PR #1242, opened 2024-12-16
Update to Kotlin 1.9.24, w/ corresponding Compose, KSP

steve-the-edwards:
Kotlin and KSP 1.9.24
compose compiler - 1.5.14
compose bom - 2024.09.02
androidx.lifecycle to 2.8.1


samples/compose-samples/src/main/java/com/squareup/sample/compose/launcher/SampleLauncherApp.kt
```diff
@@ -148,14 +148,15 @@ private fun launchSample(
...
 }
 
-private fun Modifier.disableTouchInput(): Modifier = pointerInput(Unit) {
-  forEachGesture {
-    awaitPointerEventScope {
-      awaitPointerEvent(Initial).let { event ->
-        event.changes.forEach { change ->
-          if (change.pressed != change.previousPressed) change.consume()
+private fun Modifier.disableTouchInput(): Modifier =
+  then(
```
zach-klippenstein, line 152: You don't need `then` when calling other modifier factories that already properly chain. This actually ends up calling `then` twice which i'm surprised doesn't have negtive side effects.

workflow-ui/compose/src/androidTest/java/com/squareup/workflow1/ui/compose/ComposeLifecycleOwnerTest.kt
```diff
@@ -32,7 +31,9 @@ class ComposeLifecycleOwnerTest {
     composeTestRule.setContent {
       parentLifecycle.currentState = RESUMED
       childLifecycleOwner = rememberChildLifecycleOwner(parentLifecycle)
-      CompositionLocalProvider(LocalLifecycleOwner provides childLifecycleOwner) {
+      CompositionLocalProvider(
+        androidx.lifecycle.compose.LocalLifecycleOwner provides childLifecycleOwner
```
zach-klippenstein, line 35: Why remove this import?

zach-klippenstein (approved):

workflow-ui/compose/src/androidTest/java/com/squareup/workflow1/ui/compose/ComposeLifecycleOwnerTest.kt
steve-the-edwards: just didn't get the import in after the change to the androidx.lifecycle version. will add.

PR #1244, opened 2024-12-16
Update androidx library versions (and a few others)

steve-the-edwards:


gradle/libs.versions.toml
```diff
@@ -1,6 +1,6 @@
 [versions]
 
-androidTools = "8.1.2"
+agpVersion = "8.1.2"
```
steve-the-edwards, line 3: this was just misnamed and so I changed it here to be clear.

gradle/libs.versions.toml
```diff
@@ -9,36 +9,37 @@ targetSdk = "33"
...
+androidx-test = "1.6.1"
 androidx-test-espresso = "3.5.1"
 androidx-test-junit-ext = "1.1.5"
 androidx-test-runner = "1.5.2"
 androidx-test-truth-ext = "1.5.0"
-androidx-tracing = "1.1.0"
-androidx-transition = "1.4.1"
+androidx-tracing = "1.2.0"
+androidx-transition = "1.5.1"
+androidx-viewbinding = "8.1.2"
```
steve-the-edwards, line 38: we did not have a version for viewbinding as we used the `androidTools` version, though we were only using that version for the AGP version, we need to separate their tracking.

afollestad (approved):

afollestad:
Sorry,I guess my review doesn't count 😂

PR #1248, opened 2025-01-20
1247: Partial Tree Rerendering

### LINKED ISSUE #1247: Partial Tree Re-rendering
Performance optimization to track the path of changed workflow node state after an action. Only re-render the nodes that have changed state, or who have children with changed state.

steve-the-edwards:
Re-render the node only if:
1. The state (including props) of this node or props has changed.
2. The state of one of this node's descendants has changed.

In order to do this, we have to propagate the action cascade all the way up to the root to record a single path of changed nodes.

Note we currently do *not* re-render if the workflow *instance* (the behavior definition instance, not the node) changed (unless the state also changed!), assuming the new workflow is still compatible with the old rendering because its identifier did not change.

Closes #1247.

### DIFF
--- a/workflow-core/src/commonMain/kotlin/com/squareup/workflow1/NullableInitBox.kt
+++ b/workflow-core/src/commonMain/kotlin/com/squareup/workflow1/NullableInitBox.kt
@@ -1,0 +1,28 @@
+package com.squareup.workflow1
+
+import kotlin.jvm.JvmInline
+
+/**
+ * Used to wrap immutable nullable values whose holder may not yet be initialized.
+ * Check [isInitialized] to see if the value has been assigned.
+ */
+@JvmInline
+public value class NullableInitBox<T>(private val _value: Any? = Uninitialized) {
+  /**
+   * Whether or not a value has been set for this [NullableInitBox]
+   */
+  public val isInitialized: Boolean get() = _value !== Uninitialized
+
+  /**
+   * Get the value this has been initialized with.
+   *
+   * @throws [IllegalStateException] if the value in the box has not been initialized.
+   */
+  @Suppress("UNCHECKED_CAST")
+  public fun getOrThrow(): T {
+    check(isInitialized) { "NullableInitBox was fetched before it was initialized with a value." }
+    return _value as T
+  }
+
+  public object Uninitialized
+}
--- a/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RuntimeConfig.kt
+++ b/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RuntimeConfig.kt
@@ -46,10 +46,10 @@ public enum class RuntimeConfigOptions {
    * Otherwise return the cached rendering (as there is no way it could have changed).
    *
    * Note however that you must be careful using this because there may be external
-   * state that your Workflow's draw in and re-render and if that is not explicitly
-   * tracked within that Workflow's state, then it will not re-render. In this case,
-   * make sure that the state is tracked within the Workflow's state (even through
-   * an artificial token) in some way.
+   * state that your Workflow draws in and re-renders, and if that is not explicitly
+   * tracked within that Workflow's state then the Workflow will not re-render.
+   * In this case  make sure that the implicit state is tracked within the Workflow's
+   * `StateT` in some way, even if only via a hash token.
    */
   @WorkflowExperimentalRuntime
   PARTIAL_TREE_RENDERING,
--- a/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RuntimeConfig.kt
+++ b/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RuntimeConfig.kt
@@ -39,6 +39,22 @@ public enum class RuntimeConfigOptions {
   RENDER_ONLY_WHEN_STATE_CHANGES,
 
   /**
+   * Only re-render each active Workflow node if:
+   * 1. It's own state changed, OR
+   * 2. One of it's descendant's state has changed.
+   *
+   * Otherwise return the cached rendering (as there is no way it could have changed).
+   *
+   * Note however that you must be careful using this because there may be external
+   * state that your Workflow's draw in and re-render and if that is not explicitly
+   * tracked within that Workflow's state, then it will not re-render. In this case,
+   * make sure that the state is tracked within the Workflow's state (even through
+   * an artificial token) in some way.
+   */
+  @WorkflowExperimentalRuntime
+  PARTIAL_TREE_RENDERING,
+
+  /**
    * If we have more actions to process, do so before passing the rendering to the UI layer.
    */
   @WorkflowExperimentalRuntime
--- a/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RuntimeConfig.kt
+++ b/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RuntimeConfig.kt
@@ -40,8 +40,8 @@ public enum class RuntimeConfigOptions {
 
   /**
    * Only re-render each active Workflow node if:
-   * 1. It's own state changed, OR
-   * 2. One of it's descendant's state has changed.
+   * 1. Its own state changed, OR
+   * 2. One of its descendant's state has changed.
    *
    * Otherwise return the cached rendering (as there is no way it could have changed).
    *
--- a/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
+++ b/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
@@ -3,14 +3,17 @@ package com.squareup.workflow1.internal
 import com.squareup.workflow1.ActionApplied
 import com.squareup.workflow1.ActionProcessingResult
 import com.squareup.workflow1.NoopWorkflowInterceptor
+import com.squareup.workflow1.NullableInitBox
 import com.squareup.workflow1.RenderContext
 import com.squareup.workflow1.RuntimeConfig
 import com.squareup.workflow1.RuntimeConfigOptions
+import com.squareup.workflow1.RuntimeConfigOptions.PARTIAL_TREE_RENDERING
 import com.squareup.workflow1.StatefulWorkflow
 import com.squareup.workflow1.TreeSnapshot
 import com.squareup.workflow1.Workflow
 import com.squareup.workflow1.WorkflowAction
 import com.squareup.workflow1.WorkflowExperimentalApi
+import com.squareup.workflow1.WorkflowExperimentalRuntime
 import com.squareup.workflow1.WorkflowIdentifier
 import com.squareup.workflow1.WorkflowInterceptor
 import com.squareup.workflow1.WorkflowInterceptor.WorkflowSession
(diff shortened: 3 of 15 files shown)


workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/WorkflowOperatorsTest.kt
```diff
@@ -207,23 +207,37 @@ class WorkflowOperatorsTest {
...
       renderProps: Unit,
+      renderState: T,
       context: RenderContext
     ): T {
       // Listen to the flow to trigger a re-render when it updates.
-      context.runningWorker(rerenderWorker as Worker<Any?>) { WorkflowAction.noAction() }
-      return flow.value
+      context.runningWorker(rerenderWorker) { output: T ->
+        action("rerenderUpdate") {
+          state = output
```
steve-the-edwards, line 233: prior to this state was not actually changing - so this would not work in the optimized runtime.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
```diff
@@ -211,12 +216,14 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
    */
   private fun updateCachedWorkflowInstance(
     workflow: StatefulWorkflow<PropsT, StateT, OutputT, RenderingT>
-  ) {
+  ): Boolean {
```
zach-klippenstein, line 219: Nit: Add a note to the kdoc explaining this new return value

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
```diff
@@ -227,39 +234,59 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
...
+    val didUpdateCachedInstance = updatePropsAndState(props, workflow)
 
-    baseRenderContext.unfreeze()
-    val rendering = interceptedWorkflowInstance.render(props, state, context)
-    baseRenderContext.freeze()
+    if (!runtimeConfig.contains(PARTIAL_TREE_RENDERING) ||
+      !lastRendering.isInitialized ||
+      subtreeStateDidChange
+    ) {
+      if (!didUpdateCachedInstance) {
```
zach-klippenstein, line 243: Would it be simpler to just call `updateCachedWorkflowInstance` unconditionally here? It's idempotent, right?

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
```diff
@@ -289,4 +322,17 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
       SideEffectNode(key, job)
     }
   }
+
+  @JvmInline
+  internal value class Box<T>(private val _value: Any? = Uninitialized) {
```
zach-klippenstein, line 327: Nit: This seems like a super generic and potentially useful helper, I'd pull it out.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
```diff
@@ -289,4 +322,17 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
...
+    val isInitialized: Boolean get() = _value !== Uninitialized
+
+    @Suppress("UNCHECKED_CAST")
+    fun getOrThrow(): T {
+      check(isInitialized)
```
zach-klippenstein, line 332: Nit: Give this a descriptive error message.

workflow-testing/src/test/java/com/squareup/workflow1/WorkflowsLifecycleTests.kt
```diff
@@ -19,7 +20,9 @@ class WorkflowsLifecycleTests {
     RuntimeConfigOptions.RENDER_PER_ACTION,
     setOf(RENDER_ONLY_WHEN_STATE_CHANGES),
     setOf(CONFLATE_STALE_RENDERINGS),
-    setOf(CONFLATE_STALE_RENDERINGS, RENDER_ONLY_WHEN_STATE_CHANGES)
+    setOf(CONFLATE_STALE_RENDERINGS, RENDER_ONLY_WHEN_STATE_CHANGES),
+    setOf(RENDER_ONLY_WHEN_STATE_CHANGES, PARTIAL_TREE_RENDERING),
+    setOf(CONFLATE_STALE_RENDERINGS, RENDER_ONLY_WHEN_STATE_CHANGES, PARTIAL_TREE_RENDERING),
   ).asSequence()
```
zach-klippenstein, line 26: Not in scope but… why isn't this just `sequenceOf(…)`?

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
```diff
@@ -272,7 +299,13 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
...
       stateChanged = actionApplied.stateChanged || (childResult?.stateChanged ?: false)
     )
-    return if (actionApplied.output != null) {
+    // Our state changed or one of our children's state changed.
+    subtreeStateDidChange = aggregateActionApplied.stateChanged
+    return if (actionApplied.output != null ||
+      runtimeConfig.contains(PARTIAL_TREE_RENDERING)
+    ) {
+      // If we are using the optimization, always return to the parent, so we carry a path that
+      // notes that the subtree did change all the way to the root.
```
zach-klippenstein, line 308: Thanks for this comment, but I'm missing the middle reason why the optimization requires propagating change status all the way up but we don't need that without the optimization.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
steve-the-edwards: good point, we check equality there, so we could call it again

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
steve-the-edwards: No longer returning as we just call this relying on its idempotency.

workflow-testing/src/test/java/com/squareup/workflow1/WorkflowsLifecycleTests.kt
steve-the-edwards: i'm not sure tbh. the way it was written originally is likely just a mistake?

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
steve-the-edwards: Extracted into `workflow-core` as `NullableInitBox` with its own tests.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RuntimeConfig.kt
```diff
@@ -38,6 +38,22 @@ public enum class RuntimeConfigOptions {
   @WorkflowExperimentalRuntime
   RENDER_ONLY_WHEN_STATE_CHANGES,
 
+  /**
+   * Only re-render each active Workflow node if:
+   * 1. It's own state changed, OR
+   * 2. One of it's descendant's state has changed.
```
rjrjr, line 44:
```suggestion
   * Only re-render each active Workflow node if:
   * 1. Its own state changed, OR
   * 2. One of its descendant's state has changed.
```

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RuntimeConfig.kt
```diff
@@ -38,6 +38,22 @@ public enum class RuntimeConfigOptions {
...
+   *
+   * Otherwise return the cached rendering (as there is no way it could have changed).
+   *
+   * Note however that you must be careful using this because there may be external
+   * state that your Workflow's draw in and re-render and if that is not explicitly
+   * tracked within that Workflow's state, then it will not re-render. In this case,
+   * make sure that the state is tracked within the Workflow's state (even through
+   * an artificial token) in some way.
```
rjrjr, line 52:
```suggestion
   * state that your Workflow draws in and re-renders, and if that is not explicitly
   * tracked within that Workflow's state then the Workflow will not re-render.
   * In this case  make sure that the implicit state is tracked within the Workflow's
   * `StateT` in some way, even if only via a hash token.
```

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
```diff
@@ -260,6 +285,7 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
    * Applies [action] to this workflow's [state] and then passes the resulting [ActionApplied]
    * via [emitAppliedActionToParent] to the parent, with additional information as to whether or
    * not this action has changed the current node's state.
+   *
```
rjrjr, line 288: snip

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
```diff
@@ -272,7 +298,13 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
...
       stateChanged = actionApplied.stateChanged || (childResult?.stateChanged ?: false)
     )
-    return if (actionApplied.output != null) {
+    // Our state changed or one of our children's state changed.
+    subtreeStateDidChange = aggregateActionApplied.stateChanged
+    return if (actionApplied.output != null ||
+      runtimeConfig.contains(PARTIAL_TREE_RENDERING)
+    ) {
+      // If we are using the optimization, always return to the parent, so we carry a path that
+      // notes that the subtree did change all the way to the root.
```
rjrjr, line 307:
```suggestion
      // If we are using the optimization, always return to the parent, so we carry a path that
      // notes that the subtree did change all the way to the root.
      //
      // We don't need that without the optimization because there is nothing
      // to output from the root of the runtime -- the output has propagated
      // as far as it needs to causing all corresponding state changes.
      //
      // However, the root and the path down to the changed nodes must always
      // re-render now, so this is the implementation detail of how we get
      // subtreeStateDidChange = true on that entire path to the root.
```

rjrjr (approved):

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/NullableInitBox.kt
```diff
@@ -0,0 +1,28 @@
+package com.squareup.workflow1
+
+import kotlin.jvm.JvmInline
+
+/**
+ * Used to wrap immutable nullable values whose holder may not yet be initialized.
+ * Check [isInitialized] to see if the value has been assigned.
+ */
+@JvmInline
+public value class NullableInitBox<T>(private val _value: Any? = Uninitialized) {
```
rjrjr, line 10: I didn't take in at first that this is a `value class`. Cool!

steve-the-edwards: That was a @zach-klippenstein special when we were noodling on this.

PR #1250, opened 2025-01-20
1249: Cache Intercepted Workflow Instance

### LINKED ISSUE #1249: Cache Ids and interceptor instances
To prevent object recreation where possible.

steve-the-edwards:
We cache the instance of the workflow before and after it is intercepted.

We check for instance equality any time a new workflow instance is passed in (`render()` and `snapshot()`) and we update our cached instances when that happens. This will allow us to continue to use the same `WorkflowNode` with different instances of the workflow.

Closes #1249

### DIFF
--- a/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
+++ b/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
@@ -71,6 +71,8 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
   override val identifier: WorkflowIdentifier get() = id.identifier
   override val renderKey: String get() = id.name
   override val sessionId: Long = idCounter.createId()
+  private var cachedWorkflowInstance: StatefulWorkflow<PropsT, StateT, OutputT, RenderingT>
+  private var interceptedWorkflowInstance: StatefulWorkflow<PropsT, StateT, OutputT, RenderingT>
 
   private val subtreeManager = SubtreeManager(
     snapshotCache = snapshot?.childTreeSnapshots,
@@ -99,8 +101,9 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
   init {
     interceptor.onSessionStarted(this, this)
 
-    state = interceptor.intercept(workflow, this)
-      .initialState(initialProps, snapshot?.workflowSnapshot, this)
+    cachedWorkflowInstance = workflow
+    interceptedWorkflowInstance = interceptor.intercept(cachedWorkflowInstance, this)
+    state = interceptedWorkflowInstance.initialState(initialProps, snapshot?.workflowSnapshot, this)
   }
 
   override fun toString(): String {
@@ -132,10 +135,10 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
   fun snapshot(workflow: StatefulWorkflow<*, *, *, *>): TreeSnapshot {
     @Suppress("UNCHECKED_CAST")
     val typedWorkflow = workflow as StatefulWorkflow<PropsT, StateT, OutputT, RenderingT>
+    updateCachedWorkflowInstance(typedWorkflow)
     return interceptor.onSnapshotStateWithChildren({
       val childSnapshots = subtreeManager.createChildSnapshots()
-      val rootSnapshot = interceptor.intercept(typedWorkflow, this)
-        .snapshotState(state)
+      val rootSnapshot = interceptedWorkflowInstance.snapshotState(state)
       TreeSnapshot(
         workflowSnapshot = rootSnapshot,
         // Create the snapshots eagerly since subtreeManager is mutable.
@@ -203,6 +206,20 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
   }
 
   /**
+   * Call this after we have been passed any workflow instance, in [render] or [snapshot]. It may
+   * have changed and we should check to see if we need to update our cached instances.
+   */
+  private fun updateCachedWorkflowInstance(
+    workflow: StatefulWorkflow<PropsT, StateT, OutputT, RenderingT>
+  ) {
+    if (workflow !== cachedWorkflowInstance) {
+      // The instance has changed.
+      cachedWorkflowInstance = workflow
+      interceptedWorkflowInstance = interceptor.intercept(cachedWorkflowInstance, this)
+    }
+  }
+
+  /**
    * Contains the actual logic for [render], after we've casted the passed-in [Workflow]'s
    * state type to our `StateT`.
    */
@@ -210,11 +227,11 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
     workflow: StatefulWorkflow<PropsT, StateT, OutputT, RenderingT>,
     props: PropsT
   ): RenderingT {
-    updatePropsAndState(workflow, props)
+    updateCachedWorkflowInstance(workflow)
+    updatePropsAndState(props)
 
     baseRenderContext.unfreeze()
-    val rendering = interceptor.intercept(workflow, this)
-      .render(props, state, context)
+    val rendering = interceptedWorkflowInstance.render(props, state, context)
     baseRenderContext.freeze()
 
     workflowTracer.trace("UpdateRuntimeTree") {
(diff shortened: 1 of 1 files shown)


zach-klippenstein:
This seems like a breakage of very intentional semantics, semantics which that test exists to enforce. The new behavior seems unintuitive and wrong, and Stephen even calls this out in the PR description so I'm a bit confused as to how anyone sees this as a good thing?

`mapWorkflow` was intended to be usable as a utility to compose workflows. If someone writes two different functions that return a workflow by mapping another workflow, then the caller shouldn't need to be aware of whether those functions happen to use the same underlying source workflow.

rjrjr:
```kotlin
class ChildRendering(val counted: Int)
```

```kotlin
class ChildWorkflow : StatefulWorkflow<Unit, Int, Nothing, ChildRendering>() {
  override fun initialState(props: Unit, snapshot: Snapshot?) = 0

  override fun render(
    renderProps: Unit,
    renderState: Int,
    context: RenderContext
  ): ChildRendering {
    context.runningWorker(oncePerSecond) { action("") { state = state + 1 } }
    return ChildRendering(renderState)
  }
```

```kotlin
  when (props) {
    0 -> renderChild(childWorkflow.mapRendering { "rendering1: ${it.counted}" })
    1 -> renderChild(childWorkflow.mapRendering { "rendering2: ${it.counted}" })
    else -> fail()
  }
```

Today if `props` changes after ~two seconds I see:

```
rendering1: 0
rendering1: 1
rendering2: 2
rendering2: 3
```

…and that's mindblowing.

With Steve's fix I see:

```
rendering1: 0
rendering1: 1
rendering1: 2
rendering1: 3
```

…and now that I've read Zach's response that also seems like a shock -- although I still think less of one.

I'm wondering it's practical to see…

```
rendering1: 0
rendering1: 1
rendering2: 0
rendering2: 1
```

…without having to provide the key.

And also wondering if this variation would work, where only a _stable_ lambda passed to separate `mapRendering` calls produces identical workflows.

```kotlin
  fun stamp(c: ChildRendering): String = "stamped: $counted"

  when (props) {
    0 -> renderChild(childWorkflow.mapRendering(::stamp))
    1 -> renderChild(childWorkflow.mapRendering(::stamp))
    else -> fail()
  }
```

```
stamped: 0
stamped: 1
stamped: 2
stamped: 3
```

steve-the-edwards:
> This seems like a breakage of very intentional semantics, semantics which that test exists to enforce.

I agree this breaks intentional semantics with tests enforcing it! What I don't understand is why those exist, or if they are useful? I went searching for how this was first decided and I only got as far as the rewrite from Reactor where this seemed to be the case from the beginning.

In my opinion the problem with the API exists with the combination of `.mapRendering()` and `ImpostorWorkflow` using _strictly_ the identifier of the wrapped workflow in `realIdentifier`.

I do not think we should every allow a `WorkflowSession` (state and all) to continue with a different workflow instance (or render a different workflow instance). <- that seems like a problem waiting to happen.
The fundamental question (if anyone can remember) is why we allow that ^^. A `WorkflowNode` should (in my mind) only be able to render one workflow ever.

As I noted in the comment, if we want to just `.mapRendering()` that should be done on the _rendering_ not on the Workflow class (then we would have none of these problems). But that would be a different thing, `.mapRendering()` as it stands is designed to produce a new Workflow - which is fine, but it has the same identifier as the previous workflow, which, imo, is not fine.

FWIW, I do think this test is a pretty unique case (changing the `.mapRendering()`'ed instance one after the other), as if these were ever rendered at the same time it would crash (given they have the same ID).

With respect to @rjrjr 's suggestion of this behavior:

> I'm wondering if it's practical to see…
> 
> rendering1: 0
> rendering1: 1
> rendering2: 0
> rendering2: 1

That is actually what I was hoping for initially (I thought of adding a random int, or monotonically increasing int to every use of `mapRendering()`). Where I got caught there is:
1. There is a reason we expose `realIdentifier` as we sometimes need to get at whatever that wraps.
2. So then we would have to have a 'composed' `WorkflowIdentifier` from which we can still access the original - but it is decorated to make it unique. (<-- I decided this was a larger change than I wanted to make here, so I settled for trying to enforce the intended behavior (what Ray shows above) by suggesting to use a different rendering key. Then I was going to see how much this actually comes up in our own use to see how big of a problem that is).

zach-klippenstein:
Ok I misread the tests. That trigger value is props, not an internal thing. Yes Ray's suggestion seems reasonable: source workflow session restarts.

This whole discussion might be moot though if `mapRendering` is never used in Register – do you know what the current usages require?

rjrjr:
> Ok I misread the tests. That trigger value is props, not an internal thing. Yes Ray's suggestion seems reasonable: source workflow session restarts.
> 
> This whole discussion might be moot though if `mapRendering` is never used in Register – do you know what the current usages require?

We use it like crazy to re-shape root workflows to meet the requirements of their containers -- which are very unlikely to trigger this case.

The "well you should just map the rendering" advice isn't really practical. This is very useful glue for spots where you have a `workflow: Workflow<*, RoundHole, *>` param that you need to fill with `Workflow<*, SquarePeg, *>`, and we have a lot of those.

steve-the-edwards:
> The "well you should just map the rendering" advice isn't really practical. This is very useful glue for spots where you have a `workflow: Workflow<*, RoundHole, *>` param that you need to fill with `Workflow<*, SquarePeg, *>`, and we have a lot of those.

I think we rely on `proxiedIdentifier` (which is the `ImpostorWorkflow::realIdentifier`) in order to determines [equality](https://github.com/square/workflow-kotlin/blob/3b56a157e157725a924cc10b7af0a701196a7d27/workflow-core/src/commonMain/kotlin/com/squareup/workflow1/WorkflowIdentifier.kt#L122). If we were add in `hashCode()` of the *workflow* instance alongside the proxiedIdentifier (in other words, two Impostors wrapping the same `realIdentifier` would be differentiated) I don't think that would be a problem for the container case.

i.e., 
```
public val Workflow<*, *, *>.computedIdentifier: WorkflowIdentifier
  get() {
    val maybeImpostor = this as? ImpostorWorkflow
    return WorkflowIdentifier(
      type = Snapshottable(this::class),
      proxiedIdentifier = maybeImpostor?.realIdentifier,
      workflowHashCode = hashCode(),
      description = maybeImpostor?.let { it::describeRealIdentifier }
    )
  }

// In WorkflowIdentifier

  override fun equals(other: Any?): Boolean = when {
    this === other -> true
    other !is WorkflowIdentifier -> false
    else -> type.typeName == other.type.typeName &&
      (proxiedIdentifier == other.proxiedIdentifier && workflowHashCode == other.workflowHashCode)
  }

```

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
```diff
@@ -202,6 +205,18 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
     coroutineContext.cancel(cause)
   }
 
+  // Call this after we have been passed any workflow instance, in [render] or [snapshot]. It may
+  // have changed and we should check to see if we need to update our cached instances.
```
zach-klippenstein, line 209: Nit: Make this a kdoc comment so it's visible directly at callsites.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
```diff
@@ -202,6 +205,18 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
...
+  private fun updateCachedWorkflowInstance(
+    workflow: StatefulWorkflow<PropsT, StateT, OutputT, RenderingT>
+  ) {
+    if (workflow !== cachedWorkflowInstance) {
+      // instance has changed.
+      interceptedWorkflowInstance = interceptor.intercept(workflow, this)
+      cachedWorkflowInstance = workflow
```
zach-klippenstein, line 216: Why do you update `cachedWorkflowInstance` first in `init`, but second here? Is there some case where `intercept` needs the cached workflow to not be updated yet?

zach-klippenstein (approved):

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
steve-the-edwards: Done.

PR #1262, opened 2025-02-04
Improve the kdoc on the newly-public collectToSink.

zach-klippenstein:


workflow-core/src/commonMain/kotlin/com/squareup/workflow1/Sink.kt
```diff
@@ -40,18 +40,34 @@ public fun <T1, T2> Sink<T1>.contraMap(transform: (T2) -> T1): Sink<T2> = Sink {
 
 /**
  * Collects from a [Flow] by converting each item into a [WorkflowAction] and then sending them
- * to the [actionSink]. This operator propagates back pressure from the workflow runtime, so if there
- * is a lot of contention on the workflow runtime the flow will be suspended while the action is
- * queued.
+ * to the [actionSink]. This may be used as an alternative to a [asWorker] for certain [Flow]s.
```
rjrjr, line 43:
```suggestion
 * to the [actionSink]. This may be used as a lighter weight alternative to
 * [asWorker] for certain [Flow]s.
```

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/Sink.kt
```diff
@@ -40,18 +40,34 @@ public fun <T1, T2> Sink<T1>.contraMap(transform: (T2) -> T1): Sink<T2> = Sink {
...
+ *       }
+ *     }
  *   }
  * }
  * ```
+ *
+ * > **Warning:** Be careful using this from `runningSideEffect` when the source [Flow] comes from
+ * > state or props. The side effect will capture those values when it first runs, and if the instance
+ * > of the flow changes in a future render pass, the side effect will still be collecting the stale
+ * > [Flow]. In this case
```
rjrjr, line 70: In this case…?

rjrjr (approved):

PR #1263, opened 2025-02-12
Fix Conflate Stale Rendering Logic for Short Circuit

steve-the-edwards:
When we kept rendering for `CONFLATE_STALE_RENDERINGS` we were not also checking if we should short circuit. We need to do both.

If any of the actions applied produces output from the _root_ runtime though, then we will stop applying actions, update the rendering, and then send the output.

### DIFF
--- a/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
+++ b/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
@@ -161,19 +161,12 @@ public fun <PropsT, OutputT, RenderingT> renderWorkflowIn(
   }
 
   /**
-   * If [runtimeConfig] contains [RuntimeConfigOptions.RENDER_ONLY_WHEN_STATE_CHANGES] then
-   * send any output, but return true which means restart the runtime loop and process another
-   * action.
+   * If [runtimeConfig] contains [RuntimeConfigOptions.RENDER_ONLY_WHEN_STATE_CHANGES] and
+   * we have not changed state, then return true to short circuit the render loop.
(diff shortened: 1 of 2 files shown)


workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
```diff
@@ -1180,6 +1182,248 @@ class RenderWorkflowInTest {
...
+          runtimeConfig = runtimeConfig,
+          workflowTracer = workflowTracer,
+        ) {}
+
+        launch {
+          trigger.emit("changed state")
+        }
+        advanceUntilIdle()
+
+        assertEquals(2, renderCount)
```
rjrjr, line 1254:
```suggestion
        // 2 renderings (initial and then the update.) Not *3* renderings.
        assertEquals(2, renderCount)
```

rjrjr (approved):

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
```diff
@@ -183,34 +176,40 @@ public fun <PropsT, OutputT, RenderingT> renderWorkflowIn(
       // launched.
       var actionResult: ActionProcessingResult = runner.processAction()
 
-      if (shortCircuitForUnchangedState(actionResult)) continue
+      if (shouldShortCircuitForUnchangedState(actionResult)) {
```
zach-klippenstein, line 179: This is much better.

zach-klippenstein (approved):

PR #1265, opened 2025-02-19
Clean up RenderWorkflowIn Tests

steve-the-edwards:
In the course of https://github.com/square/workflow-kotlin/pull/1263 I encountered some issues with these tests.

1. We had two tests - `onOutput_called_after_rendering_emitted` and `output_is_emitted_before_next_render_pass` which were stating that they were testing the opposite things. In fact, there was a bug in `onOutput_called_after_rendering_emitted` where the assertion that should have failed the test was uncaught and only killed the coroutine `TestScope` rather than bubbling up to fail the test. So that test was 'failing' silently.
2. The stated intention of the API however, _is_ to call `onOutput` _after_ the rendering is emitted.
3. How do I know that? We tried to switch them to be more inline with the internal behaviour of a single Workflow's output  - https://github.com/square/workflow-kotlin/issues/54 , and then https://github.com/square/workflow-kotlin/pull/68 - and that created a problem in the internal square usage with snapshot ordering - https://github.com/square/workflow-kotlin/pull/151. However, when that was reverted, the name of the test did not change, so its name did not match the actual test assertion, until we updated the coroutine usage in the KMM migration (https://github.com/square/workflow-kotlin/pull/799) where the assertion changed to be output _then_ rendering (in line with the name of the test, but changed from the previous behaviour).
4. Now, the reality is more nuanced. We _do_ update the rendering before we call output, but we do so by assigning the `.value` on the `StateFlow` of `RenderingAndSnapshot`s. This is a non-suspending operation, whereas calling `onOutput` is suspending. So really, `onOutput` will be called before any collectors of the renderings `StateFlow` can be dispatched to handle the update, but if the `.value` of that `StateFlow` of renderings was fetched in `onOutput` then it would be "up-to-date".
5. Turns out this is likely good enough behaviour for both coherence/consistency of the renderings wrt handling of the output.

Anyway, as a result of discovering this I have gone through and audited the `RenderWorklfowInTest` tests, updating them to:
1. Use `runTest` so that by default anything cancelling the `CoroutineScope` provided will also fail the test.
2. Remove the incorrect test.
3. Remove the `testScope` and `pausedTestScope` in favour of using the scope provided by `runTest` (and the `backgroundScope` when we have an operation - like the runtime that we don't care if its stopped before the test is done).

### DIFF
--- a/workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
+++ b/workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
@@ -6,43 +6,29 @@ import com.squareup.workflow1.internal.ParameterizedTestRunner
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.CoroutineExceptionHandler
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Dispatchers.Unconfined
 import kotlinx.coroutines.ExperimentalCoroutinesApi
-import kotlinx.coroutines.FlowPreview
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.flow.MutableStateFlow
-import kotlinx.coroutines.flow.consumeAsFlow
-import kotlinx.coroutines.flow.launchIn
+import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.map
-import kotlinx.coroutines.flow.onEach
-import kotlinx.coroutines.flow.produceIn
+import kotlinx.coroutines.flow.receiveAsFlow
 import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.plus
 import kotlinx.coroutines.suspendCancellableCoroutine
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.test.StandardTestDispatcher
 import kotlinx.coroutines.test.TestScope
 import kotlinx.coroutines.test.UnconfinedTestDispatcher
 import kotlinx.coroutines.test.advanceUntilIdle
-import kotlinx.coroutines.test.runCurrent
+import kotlinx.coroutines.test.runTest
 import okio.ByteString
 import kotlin.test.Test
 import kotlin.test.assertEquals
 import kotlin.test.assertTrue
 
-@OptIn(ExperimentalCoroutinesApi::class, FlowPreview::class, WorkflowExperimentalRuntime::class)
+@OptIn(ExperimentalCoroutinesApi::class, WorkflowExperimentalRuntime::class)
 class RenderWorkflowInTest {
-
-  /**
-   * A [TestScope] that will not run until explicitly told to.
-   */
-  private lateinit var pausedTestScope: TestScope
-
-  /**
-   * A [TestScope] that will automatically dispatch enqueued routines.
-   */
-  private lateinit var testScope: TestScope
 
   private val traces: StringBuilder = StringBuilder()
   private val testTracer: WorkflowTracer = object : WorkflowTracer {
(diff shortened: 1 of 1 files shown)


workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
```diff
@@ -58,42 +44,48 @@ class RenderWorkflowInTest {
...
+    setOf(CONFLATE_STALE_RENDERINGS),
+    setOf(CONFLATE_STALE_RENDERINGS, RENDER_ONLY_WHEN_STATE_CHANGES)
+  )
+
+  private val tracerOptions = setOf<WorkflowTracer?>(
+    null,
+    testTracer
+  )
+
+  private val runtimeOptions: Sequence<Pair<RuntimeConfig, WorkflowTracer?>> = cartesianProduct(
```
rjrjr, line 59: Neat

rjrjr (approved):
LGTM, but it would be wise to get @zach-klippenstein or someone else with stronger Coroutine chops to LG.

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
```diff
@@ -377,85 +341,168 @@ class RenderWorkflowInTest {
...
-          emittedRenderings += rendering.rendering
+  /**
+   * A different form of [onOutput_called_after_rendering_emitted]. Here we launch the workflow
+   * runtime on its own [TestScope] with a [StandardTestDispatcher], which will be paused until
+   * told to advance.
+   *
+   * We *collect* emitted renderings on the [UnconfinedTestDispatcher] of the [runTest].
+   * The point here is that when the runtime sets the value on the StateFlow - a non-suspending
+   * operation - and then it calls [onOutput] - a suspending operation - the [onOutput] handler
+   * will not be immediately dispatched (it is waiting for dispatch from the scheduler), but the
```
zach-klippenstein, line 439: > the [onOutput] handler will not be immediately dispatched (it is waiting for dispatch from the scheduler)

The workflow runtime calls `onOutput` directly, it's just a function, so the dispatcher is not involved at all. The reason this timing is the way it is is because the runtime sets the MSF's `value` property before calling `onOutput`. Or am I misreading this explanation?

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
```diff
@@ -377,85 +341,168 @@ class RenderWorkflowInTest {
...
+   * collector of the renderings [StateFlow] will be dispatched and update the 'emitted' renderings.
+   * Then when we let the runtime's scheduler go ahead, it will have already been populated.
+   */
+  @Test fun onOutput_called_after_rendering_emitted_and_collected() {
```
zach-klippenstein, line 443: Maybe I'm missing something but it looks like what this test is really testing is the behavior of `Unconfined` dispatchers. The behavior it validates is implied by the previous test when the renderings flow is collected with an unconfined dispatcher. Doesn't seem worth writing our own tests for core coroutine features.

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
```diff
@@ -377,85 +341,168 @@ class RenderWorkflowInTest {
...
+        val emittedRenderings = mutableListOf<String>()
+        val receivedOutputs = mutableListOf<String>()
+        val renderings: StateFlow<RenderingAndSnapshot<String>> = renderWorkflowIn(
+          workflow = workflow,
+          scope = testScope,
+          props = MutableStateFlow(Unit),
+          runtimeConfig = runtimeConfig,
+          workflowTracer = workflowTracer,
+        ) { it: String ->
+          // The list collecting the renderings already contains it by the time onOutput is fired.
```
zach-klippenstein, line 477: Having expressed my doubt that this test is really worth having at all, if we _do_ want it, it might make the test more clear and explicit to use a pattern that is often used to validate synchronization expectations in concurrent code. I can't remember if we use this elsewhere in workflow tests, but basically you just do this (in the test method, the class, or an external helper class, doesn't matter):
```kotlin
val actual = AtomicInteger(0)
fun expect(expected: Int) {
  val actual = actual.getAndIncrement()
  assertEquals(expected, actual, "Expected $expected, was $actual")
}
```

And then you pepper the concurrent code with calls something like this:
```kotlin
onOutput = { it: String ->
  expect(1)
}
…
launch {
  renderings.collect {
    expect(0)
  }
}
```

It's easy to see what you're testing because the code says explicitly what order the `expect` calls are expected to occur in.

zach-klippenstein (approved):
Glad we're finally using the official test scopes for this!

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
steve-the-edwards: `onOutput` is a `suspend fun`, so it will go through dispatch. It's a good point though that perhaps we don't need it to be a `suspend fun`

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
steve-the-edwards: That's a nice pattern! I will see if I can make use of in future refactors.

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
steve-the-edwards: heh, ya you're point is valid. I was trying to make a point with the test more than actually test anything.

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
steve-the-edwards: I'm thinking we made `onOutput` a `suspend fun` because we wanted it to be able to cancel the runtime's scope? cc @rjrjr if you remember.

rjrjr: No idea. I doubt we did it by accident.

PR #1267, opened 2025-02-21
*BREAKING* Introduces `BaseRenderContext.remember` and stable `eventHandlers`.

rjrjr:
Workflow makes it very convenient to render view models with anonymous lambdas as their event handler functions. Compose hates that.

To address that mismatch without forcing everyone to retrofit their apps to use event objects instead of lambdas (it's a little late for that!) we introduce support for stable event handlers: anonymous lambdas whose identity looks the same to Compose across updates.

In order to do this we're breaking the existing `eventHandler` and `safeEventHandler` functions a bit.

- We introduce a new optional `remember: Boolean? = null` parameter. Set that true to get the new stability. If you leave it to the default `null` we look for a new `STABLE_EVENT_HANDLER : RuntimeConfigOption` to decide what to do. If you set that config option on an existing app and make no other changes, all of your existing `eventHandler` functions will be stable.

- When `remember` is true, the existing `name` parameter becomes weight bearing. It's no longer just a logging aid, it's part of a key identifying your stable lambda. The other parts of the key are its return type, and the types of any of its parameters. Duplicating a key within a particular `render()` call is a runtime error, similar to the rules for `renderChild`, `runningWorker`, and `runningSideEffect`.

- To make it easier to find fix and prevent those new runtime errors, `testRender` and `WorkflowTestParams` now accept optional `RuntimeConfig` parameters, and throw appropriate errors if `STABLE_EVENT_HANDLER` is set. `testRender` also now honors `JvmTestRuntimeConfigTools.getTestRuntimeConfig()`.

- Most of the `eventHandler` functions have also been changed to `inline` -- necessary so that we can reify their parameter types for the key scheme described above

All of this is built on a new `BaseRenderContext.remember` primitive, which provides a light weight mechanism to save a bit of state across a workflow session without having to find room for it in `StateT`.

`BaseRenderContext` also now provides `val runtimeConfig: RuntimeConfig` in support of all of the above.

### DIFF
--- a/samples/compose-samples/src/main/java/com/squareup/sample/compose/inlinerendering/InlineRenderingWorkflow.kt
+++ b/samples/compose-samples/src/main/java/com/squareup/sample/compose/inlinerendering/InlineRenderingWorkflow.kt
@@ -3,7 +3,6 @@
 package com.squareup.sample.compose.inlinerendering
 
 import androidx.compose.animation.AnimatedContent
-import androidx.compose.animation.ExperimentalAnimationApi
 import androidx.compose.animation.SizeTransform
 import androidx.compose.animation.fadeIn
 import androidx.compose.animation.fadeOut
@@ -38,12 +37,15 @@ object InlineRenderingWorkflow : StatefulWorkflow<Unit, Int, Nothing, Screen>() 
     renderProps: Unit,
     renderState: Int,
     context: RenderContext
-  ) = ComposeScreen {
-    Box {
-      Button(onClick = context.eventHandler("increment") { state += 1 }) {
-        Text("Counter: ")
-        AnimatedCounter(renderState) { counterValue ->
-          Text(counterValue.toString())
+  ): ComposeScreen {
+    val onClick = context.eventHandler("increment") { state += 1 }
+    return ComposeScreen {
+      Box {
+        Button(onClick = onClick) {
+          Text("Counter: ")
+          AnimatedCounter(renderState) { counterValue ->
+            Text(counterValue.toString())
+          }
         }
       }
     }
@@ -68,7 +70,6 @@ internal fun InlineRenderingWorkflowPreview() {
   InlineRenderingWorkflowRendering()
 }
 
-@OptIn(ExperimentalAnimationApi::class)
 @Composable
 private fun AnimatedCounter(
   counterValue: Int,
@@ -79,6 +80,7 @@ private fun AnimatedCounter(
     transitionSpec = {
       ((slideInVertically() + fadeIn()).togetherWith(slideOutVertically() + fadeOut()))
         .using(SizeTransform(clip = false))
-    }
+    },
+    label = ""
   ) { content(it) }
 }
--- a/samples/dungeon/timemachine/src/test/java/com/squareup/sample/timemachine/TimeMachineWorkflowTest.kt
+++ b/samples/dungeon/timemachine/src/test/java/com/squareup/sample/timemachine/TimeMachineWorkflowTest.kt
@@ -25,7 +25,10 @@ class TimeMachineWorkflowTest {
     val delegateWorkflow = Workflow.stateful<String, Nothing, DelegateRendering>(
       initialState = "initial",
       render = { renderState ->
-        DelegateRendering(renderState, setState = eventHandler("") { s -> state = s })
+        DelegateRendering(
+          renderState,
+          setState = eventHandler("setState") { s -> state = s }
+        )
       }
     )
     val clock = TestTimeSource()
--- a/workflow-core/src/commonMain/kotlin/com/squareup/workflow1/BaseRenderContext.kt
+++ b/workflow-core/src/commonMain/kotlin/com/squareup/workflow1/BaseRenderContext.kt
@@ -1,8 +1,7 @@
-// Type variance issue: https://github.com/square/workflow-kotlin/issues/891
 @file:Suppress(
-  "EXPERIMENTAL_API_USAGE",
+  "ktlint:standard:indent",
   "ktlint:standard:parameter-list-spacing",
-  "ktlint:standard:parameter-wrapping"
+  "ktlint:standard:parameter-wrapping",
 )
 @file:JvmMultifileClass
 @file:JvmName("Workflows")
@@ -48,6 +47,8 @@ import kotlin.reflect.typeOf
  * See [renderChild].
  */
 public interface BaseRenderContext<out PropsT, StateT, in OutputT> {
+
+  public val runtimeConfig: RuntimeConfig
 
   /**
    * Accepts a single [WorkflowAction], invokes that action by calling [WorkflowAction.apply]
@@ -117,7 +118,6 @@ public interface BaseRenderContext<out PropsT, StateT, in OutputT> {
    * [Job][kotlinx.coroutines.Job] can be extracted from that and used to get guaranteed to be
    * executed lifecycle hooks, e.g. via [Job.invokeOnCompletion][kotlinx.coroutines.Job.invokeOnCompletion].
    *
-   *
    * @param key The string key that is used to distinguish between side effects.
    * @param sideEffect The suspend function that will be launched in a coroutine to perform the
    * side effect.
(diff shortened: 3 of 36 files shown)


workflow-core/src/commonMain/kotlin/com/squareup/workflow1/BaseRenderContext.kt
```diff
@@ -127,221 +126,250 @@ public interface BaseRenderContext<out PropsT, StateT, in OutputT> {
...
+  name: String,
+  crossinline update: WorkflowAction<PropsT, StateT, OutputT>.Updater.(
+    E1,
+    E2
+  ) -> Unit
+): (E1, E2) -> Unit {
+  val handler: (E1, E2) -> Unit = { e1, e2 ->
+    actionSink.send(action("eH:$name") { update(e1, e2) })
   }
+  val key = "$name+${E1::class.simpleName}+${E2::class.simpleName}"
```
rjrjr, line 232: Using `simpleName` here b/c KMP & Javascript.

- [ ] Find a way to use `expect` / `actual` to use `qualifiedName` when it's available

rjrjr:
Yeah, we don't need the `eventHandler1` nonsense.

rjrjr:
@steve-the-edwards, @zach-klippenstein: I've updated this with a `STABLE_EVENT_HANDLERS` config option that is passed to `RealRenderContext` as `val stableEventHandlers`. That's working nicely for `eventHandler`. I've done the first three variations: 0 arg, 1 arg, 2 arg. 

- `BaseRenderContext` interface declares methods with explicit `eventType: KClass<EventT>` parameters
- `RealRenderContext` implements those, consulting `stableEventHandlers`
- `StatefulWorkflow.RenderContext` and `StatelessWorkflow.RenderContext` declare `inline` overloads that use reified param types and call the new `BaseRenderContext` methods.

`safeEventHandler` is trickier. Seems like I can go two ways:
- define `safeEventHandler` methods on `BaseRenderContext` -- they will be visible in the IDE to `StatelessWorkflow` authors, which seems unfortunate. 
- make `BaseRenderContext` provide `val runtimeConfig: RuntimeConfig` so that `StatefulWorkflow.RenderContext` can honor it. A bit of a wart, but not so bad? 

I'm leaning toward the latter. Do you have any better ideas?

rjrjr:
> I'm leaning toward the latter. Do you have any better ideas?

Oof, bit of a non-starter, requires a dependency on `workflow-runtime`. So maybe option three, add `val stableEventHandlers: Boolean` to `BaseRenderContext` until the migration is done.

I kind of hate all of these options, but that's what I'll do just to experiment.

rjrjr:
Oh! `safeEventHandler` calls `eventHandler` directly, so it should just work as is!

rjrjr:
Looking promising:

```kotlin
  override fun eventHandler(
    name: String,
    update: WorkflowAction<
      @UnsafeVariance PropsT,
      StateT,
      @UnsafeVariance OutputT
      >.Updater.() -> Unit
  ): () -> Unit {
    val handler = { actionSink.send(action("eventHandler:$name", update)) }
    return if (stableEventHandlers) {
      class Box0 {
        lateinit var delegate: () -> Unit
        fun fire() = delegate()
      }

      val box = remember("stableEventHandler:$name") { Box0() }
      box.delegate = handler
      box::fire
    } else {
      handler
    }
  }
```

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/RealRenderContext.kt
```diff
@@ -70,6 +84,96 @@ internal class RealRenderContext<out PropsT, StateT, OutputT>(
...
+    val handler: (EventT) -> Unit = { event ->
+      actionSink.send(action("eventHandler:$name") { update(event) })
+    }
+    return if (stableEventHandlers) {
+      class Box1<E> {
+        lateinit var delegate: (E) -> Unit
+        fun fire(e: E) = delegate(e)
+      }
+
+      val key = "stableEventHandler:$name+${eventType.simpleName}"
```
rjrjr, line 138: - [ ] Find a way to use `expect` / `actual` so that we can use `qualifiedName` on non-JavaScript platforms.

samples/dungeon/timemachine/src/test/java/com/squareup/sample/timemachine/TimeMachineWorkflowTest.kt
```diff
@@ -25,7 +26,10 @@ class TimeMachineWorkflowTest {
     val delegateWorkflow = Workflow.stateful<String, Nothing, DelegateRendering>(
       initialState = "initial",
       render = { renderState ->
-        DelegateRendering(renderState, setState = eventHandler("") { s -> state = s })
+        DelegateRendering(
+          renderState,
+          setState = eventHandler("setState", typeOf<String>()) { s -> state = s }
```
steve-the-edwards, line 31: any reason not to use the `inline` version here?

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/RealRenderContext.kt
```diff
@@ -70,6 +84,544 @@ internal class RealRenderContext<out PropsT, StateT, OutputT>(
...
+
+  override fun eventHandler(
+    name: String,
+    update: WorkflowAction<
+      @UnsafeVariance PropsT,
+      StateT,
+      @UnsafeVariance OutputT
+      >.Updater.() -> Unit
+  ): () -> Unit {
+    val handler = { actionSink.send(action("eventHandler:$name", update)) }
```
steve-the-edwards, line 105: nit: can we make this prefix `eH:`? There is very limited room in the perfetto trace char limit.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
```diff
@@ -162,6 +170,29 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
...
+  ): ResultT {
+    remembered.forEachStaging {
+      require(key != it.key) {
+        "Expected remember keys to be unique: \"$key\""
+      }
+    }
+
+    val result = remembered.retainOrCreate(
+      predicate = {
+        key == it.key && it.resultType == resultType && inputs.contentEquals(it.inputs)
```
steve-the-edwards, line 187: Is `contentEquals` on a vararg just working on the array? How expensive is it?

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt
```diff
@@ -259,6 +260,356 @@ public interface WorkflowInterceptor {
         handler: (CO) -> WorkflowAction<P, S, O>
       ) -> CR
     ): CR = proceed(child, childProps, key, handler)
+
+    public fun <CResult> onRemember(
```
steve-the-edwards, line 264: I think these need to be implemented for wrapped `RenderContext` in `ChainedWorkflowInterceptor` https://github.com/square/workflow-kotlin/blob/56e6ee061fa7bc508e7e4e0d5188b9945f61d24f/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/ChainedWorkflowInterceptor.kt#L130

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/RealRenderContext.kt
rjrjr: Done.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
rjrjr: The IDE warned me to use it instead of `==`, I don't think that's reliable on `Array`. Not sure where to find the implementations:

```
/**
 * Returns `true` if the two specified arrays are *structurally* equal to one another,
 * i.e. contain the same number of the same elements in the same order.
 * 
 * The elements are compared for equality with the [equals][Any.equals] function.
 * For floating point numbers it means that `NaN` is equal to itself and `-0.0` is not equal to `0.0`.
 */
@SinceKotlin("1.4")
public expect infix fun <T> Array<out T>?.contentEquals(other: Array<out T>?): Boolean
```

rjrjr: From the kdoc it looks legit.

rjrjr: And yes, the compiler says I'm using the Array version.

rjrjr: Screenshot 2025-02-27 at 4 24 09 PM

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/BaseRenderContext.kt
```diff
@@ -49,6 +48,8 @@ import kotlin.reflect.typeOf
  */
 public interface BaseRenderContext<out PropsT, StateT, in OutputT> {
 
+  public val runtimeConfig: RuntimeConfig
```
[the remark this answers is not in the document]
rjrjr: Nice thing about putting it here is it paves the way to adding a `runtimeConfig` line to `Workflow.testRender`, which we'll need as a tool for migrating to the new behavior:

```kotlin
public fun <PropsT, StateT, OutputT, RenderingT>
  StatefulWorkflow<PropsT, StateT, OutputT, RenderingT>.testRender(
  props: PropsT,
  initialState: StateT,
  runtimeConfig = JvmTestRuntimeConfigTools.getTestRuntimeConfig()
): RenderTester<PropsT, StateT, OutputT, RenderingT> =
```

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/WorkflowAction.kt
```diff
@@ -8,6 +8,11 @@ import kotlin.jvm.JvmMultifileClass
 import kotlin.jvm.JvmName
 import kotlin.jvm.JvmOverloads
 
+/**
+ * Convenience alias for working with [WorkflowAction.Updater].
+ */
+public typealias Updater<P, S, O> = WorkflowAction<P, S, O>.Updater
```
zach-klippenstein, line 14: Nice

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/StatelessWorkflow.kt
```diff
@@ -136,7 +327,11 @@ public fun <PropsT, OutputT, RenderingT> RenderContext(
  * their own internal state.
  */
 public inline fun <PropsT, OutputT, RenderingT> Workflow.Companion.stateless(
```
zach-klippenstein, line 329: Not related to this change, but did we end up discouraging the use of these functions?

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/BaseRenderContext.kt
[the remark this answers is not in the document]
zach-klippenstein: Seems like a reasonable place to expose this, I don't know where else would even make sense.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/RealRenderContext.kt
```diff
@@ -85,7 +109,8 @@ internal class RealRenderContext<out PropsT, StateT, OutputT>(
     frozen = false
   }
 
-  private fun checkNotFrozen() = check(!frozen) {
-    "RenderContext cannot be used after render method returns."
+  private fun checkNotFrozen(reason: () -> String = { "" }) = check(!frozen) {
+    "RenderContext cannot be used after render method returns" +
+      "${reason().takeUnless { it.isBlank() }?.let { " ($it)" }}"
```
zach-klippenstein, line 114: Nice

zach-klippenstein (approved):

samples/compose-samples/src/main/java/com/squareup/sample/compose/inlinerendering/InlineRenderingWorkflow.kt
```diff
@@ -79,6 +80,7 @@ private fun AnimatedCounter(
     transitionSpec = {
       ((slideInVertically() + fadeIn()).togetherWith(slideOutVertically() + fadeOut()))
         .using(SizeTransform(clip = false))
-    }
+    },
+    label = ""
```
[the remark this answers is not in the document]
rjrjr: Nope, seems to serve no purpose, will revert.

samples/compose-samples/src/main/java/com/squareup/sample/compose/inlinerendering/InlineRenderingWorkflow.kt
```diff
@@ -38,12 +37,15 @@ object InlineRenderingWorkflow : StatefulWorkflow<Unit, Int, Nothing, Screen>()
...
     context: RenderContext
-  ) = ComposeScreen {
-    Box {
-      Button(onClick = context.eventHandler("increment") { state += 1 }) {
-        Text("Counter: ")
-        AnimatedCounter(renderState) { counterValue ->
-          Text(counterValue.toString())
+  ): ComposeScreen {
+    val onClick = context.eventHandler("increment") { state += 1 }
+    return ComposeScreen {
```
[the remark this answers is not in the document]
rjrjr: Oh, I had to break out the `val onClick` to prevent calling `eventHandler` outside of the render pass.

PR #1268, opened 2025-02-26
WIP Sketching Compose-based workflows.

zach-klippenstein:
See https://www.notion.so/block-xyz/Compose-based-Workflows-1a6be352683880b98600dde8e4413fb9?pvs=4


zach-klippenstein:
Here's a prototype for a way to do skipping with non-unit-returning composables: https://github.com/squareup/market/pull/11218

PR #1269, opened 2025-03-05
`DRAIN_EXCLUSIVE_ACTIONS` implementation

steve-the-edwards:
If there are multiple actions that are prepared to be applied that are in separate (exclusive) areas of the tree, then apply them all before rendering again.

Also ensure that we track if any of the chained actions have changed state. If they have, then pass the new rendering to the UI!

Update tests to test this behavior.

### DIFF
--- a/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
+++ b/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
@@ -1,6 +1,7 @@
 package com.squareup.workflow1
 
 import com.squareup.workflow1.RuntimeConfigOptions.CONFLATE_STALE_RENDERINGS
+import com.squareup.workflow1.RuntimeConfigOptions.DRAIN_EXCLUSIVE_ACTIONS
 import com.squareup.workflow1.RuntimeConfigOptions.RENDER_ONLY_WHEN_STATE_CHANGES
 import com.squareup.workflow1.WorkflowInterceptor.RenderPassSkipped
 import com.squareup.workflow1.WorkflowInterceptor.RenderPassesComplete
@@ -14,6 +15,7 @@ import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.yield
 
 /**
  * Launches the [workflow] in a new coroutine in [scope] and returns a [StateFlow] of its
@@ -167,13 +169,18 @@ public fun <PropsT, OutputT, RenderingT> renderWorkflowIn(
   /**
    * If [runtimeConfig] contains [RuntimeConfigOptions.RENDER_ONLY_WHEN_STATE_CHANGES] and
    * we have not changed state, then return true to short circuit the render loop.
+   *
+   * @param actionResult: Whether this action has changed state.
+   * @param actionChainingHasChangedState: Whether any of the chained actions has changed state.
    */
   fun shouldShortCircuitForUnchangedState(
     actionResult: ActionProcessingResult,
+    actionChainingHasChangedState: Boolean = false
   ): Boolean {
     return runtimeConfig.contains(RENDER_ONLY_WHEN_STATE_CHANGES) &&
       actionResult is ActionApplied<*> &&
-      !actionResult.stateChanged
+      !actionResult.stateChanged &&
+      !actionChainingHasChangedState
   }
 
   scope.launch {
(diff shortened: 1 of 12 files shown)


workflow-testing/src/test/java/com/squareup/workflow1/WorkflowsLifecycleTests.kt
```diff
@@ -143,19 +156,20 @@ class WorkflowsLifecycleTests {
...
+          !runtimeConfig.contains(DRAIN_EXCLUSIVE_ACTIONS)
+        ) {
           // 2 rendering or 1 depending on runtime config.
           awaitNextRendering()
         }
 
-        assertEquals(1, started, "Side Effect not started 1 time.")
-        assertEquals(1, cancelled, "Side Effect not cancelled 1 time.")
+        assertEquals(0, started, "Side Effect not started.")
+        assertEquals(0, cancelled, "Side Effect not cancelled.")
```
steve-the-edwards, line 172: This test was just wrong before. The behavior it is supposed to test is the current behavior of the runtime where if we start and stop rendering a workflow in the same main thread message its side effect never gets executed, because a LAZY start means it needs a dispatch to happen.

We were avoiding this before by using just `runCurrent()` rather than `advanceUntilIdle()` after causing the 2 different actions to be invoked.

Now we are actually testing what we are saying we intended to test.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
```diff
@@ -186,11 +187,42 @@ public fun <PropsT, OutputT, RenderingT> renderWorkflowIn(
...
+      var drainingActionResult = actionResult
+      if (runtimeConfig.contains(DRAIN_EXCLUSIVE_ACTIONS)) {
+        drain@ while (isActive &&
+          drainingActionResult is ActionApplied<*> &&
+          drainingActionResult.output == null
+        ) {
+          // We start by yielding, because if we are on an Unconfined dispatcher, we want to give
+          // other signals (like Workers listening to the same result) a chance to get dispatched
+          // and queue their actions.
+          yield()
```
zach-klippenstein, line 199: Passes the sniff test: This is a "tight" loop that doesn't necessarily suspend (I think `sendOutput` is the only suspending call?), so yielding in the middle seems reasonable even if we weren't using an unconfined dispatcher.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
```diff
@@ -186,11 +187,42 @@ public fun <PropsT, OutputT, RenderingT> renderWorkflowIn(
       // we don't surprise anyone with an unexpected rendering pass. Show's over, go home.
       if (!isActive) return@launch
 
+      var drainingActionResult = actionResult
+      if (runtimeConfig.contains(DRAIN_EXCLUSIVE_ACTIONS)) {
+        drain@ while (isActive &&
```
zach-klippenstein, line 192: Do we still need this `isActive` check since `yield` should throw if the job is cancelled?

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
```diff
@@ -186,11 +187,42 @@ public fun <PropsT, OutputT, RenderingT> renderWorkflowIn(
...
+          }
+        }
+      }
+
       // Next Render Pass.
       var nextRenderAndSnapshot: RenderingAndSnapshot<RenderingT> = runner.nextRendering()
 
       if (runtimeConfig.contains(CONFLATE_STALE_RENDERINGS)) {
-        conflate@ while (isActive && actionResult is ActionApplied<*> && actionResult.output == null) {
+        conflate@ while (isActive &&
```
zach-klippenstein, line 222: Same question about `isActive`.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
```diff
@@ -96,7 +96,12 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
   private val eventActionsChannel =
     Channel<WorkflowAction<PropsT, StateT, OutputT>>(capacity = UNLIMITED)
   private var state: StateT
-  private var subtreeStateDidChange: Boolean = true
+
+  // Our state or that of one of our descendants changed.
```
zach-klippenstein, line 100: Kdoc

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
```diff
@@ -96,7 +96,12 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
...
+  private var subtreeStateDirty: Boolean = true
+
+  // Our state changed.
```
zach-klippenstein, line 103: Kdoc

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
```diff
@@ -209,16 +214,27 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
...
+      true
+    }
 
     empty = empty && (eventActionsChannel.isEmpty || eventActionsChannel.isClosedForReceive)
 
-    // Listen for any events.
-    with(selector) {
-      eventActionsChannel.onReceive { action ->
-        return@onReceive applyAction(action)
+    if (!skipChangedNodes || !selfStateDirty) {
```
zach-klippenstein, line 232: Nit: Can either of these properties change between line 221 and here? If not, consider assignin to a local val to make it more clear that both these ifs use the same condition.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
```diff
@@ -96,7 +96,12 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
   private val eventActionsChannel =
     Channel<WorkflowAction<PropsT, StateT, OutputT>>(capacity = UNLIMITED)
   private var state: StateT
-  private var subtreeStateDidChange: Boolean = true
+
+  // Our state or that of one of our descendants changed.
+  private var subtreeStateDirty: Boolean = true
+
+  // Our state changed.
+  private var selfStateDirty: Boolean = true
```
zach-klippenstein, line 104: Maybe i missed it in the code, but it looks like we never set this on its own (without also setting `subtreeStateDirty`)? Do we really need it?

zach-klippenstein (approved):

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
steve-the-edwards: Yes, lines 349-350 we set them using different conditions when the action is being applied.

```
// Our state changed.
    selfStateDirty = actionApplied.stateChanged
    // Our state changed or one of our children's state changed.
    subtreeStateDirty = aggregateActionApplied.stateChanged
```

steve-the-edwards: that's key for drain exclusive because `subtreeStateDirty` is true always on a path all the way to the root no matter what changed in the tree. We use `selfStateDirty` to determine whether the node is exclusive to the previous action applied because otherwise there would never be any other exclusive nodes that we can process actions on.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
steve-the-edwards: No, its technically not needed. I like it for explicitness sake /readability though. is there an impact of this that i'm not considering and we should remove it?

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
steve-the-edwards: no they can't. and sure I can do that!

steve-the-edwards:
- [ ] Update to add `onUiUpdate()` to the `WorkflowInterceptor` interface
- [ ] Update to check for `conflationHasChangedState` across both DEA and CSR

steve-the-edwards:
Closed in favour of https://github.com/square/workflow-kotlin/pull/1355

PR #1272, opened 2025-03-12
Adds `STABLE_EVENT_HANDLERS` to runtime config tools

rjrjr:
Overhauls the config calls and in particular eliminates `baseline-`. (Still parses `baseline` as a no-op for ease of migration.)

Doesn't actually use the new config option yet, incremental change due to CI problems.


rjrjr:
See #1270

PR #1275, opened 2025-03-12
Eliminates top level subprojects, allprojects gradle blocks

rjrjr:
Which suddenly have become a gradle race (?) with version catalog init.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -31,17 +31,6 @@ plugins {
 
 shardConnectedCheckTasks(project)
 
-subprojects {
-
-  afterEvaluate {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,6 +7,8 @@
 
 Workflow is an application framework that provides architectural primitives.
 
+## fnord
```
zach-klippenstein, line 10: Intentional?

zach-klippenstein (approved):

README.md
rjrjr: Nnnnnnoooope

PR #1285, opened 2025-03-20
Make `testRender()` mimic `RealRenderContext.frozen`

rjrjr:
Allows unit tests to capture errors around using the `RenderContext` at illegal times:

- Calling `remember()` or `runningChild` outside of the `render()` block
- Calling `send()` while rendering


steve-the-edwards (approved):
Love it!

PR #1289, opened 2025-04-09
Introduces `RenderTester.expectRemember()`.

rjrjr:


workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/RealRenderContext.kt
```diff
@@ -109,8 +109,7 @@ internal class RealRenderContext<out PropsT, StateT, OutputT>(
     frozen = false
   }
 
-  private fun checkNotFrozen(reason: () -> String = { "" }) = check(!frozen) {
-    "RenderContext cannot be used after render method returns" +
-      "${reason().takeUnless { it.isBlank() }?.let { " ($it)" }}"
+  private fun checkNotFrozen(reason: () -> String) = check(!frozen) {
```
rjrjr, line 112: Drive by cleanup. We really don't want empty reasons, they defeat BugSnag grouping.

workflow-testing/src/main/java/com/squareup/workflow1/testing/RealRenderTester.kt
```diff
@@ -57,16 +58,6 @@ internal class RealRenderTester<PropsT, StateT, OutputT, RenderingT>(
    * property stores the [WorkflowAction] that was specified to handle that output.
    */
   private var processedAction: WorkflowAction<PropsT, StateT, OutputT>? = null,
-  /**
```
rjrjr, line 60: These were never passed in as parameters, moved them out of the constructor.

workflow-testing/src/main/java/com/squareup/workflow1/testing/RealRenderTester.kt
```diff
@@ -101,25 +92,41 @@ internal class RealRenderTester<PropsT, StateT, OutputT, RenderingT>(
...
+  private val renderedChildren: MutableList<Pair<WorkflowIdentifier, String>> = mutableListOf()
 
-  private var rememberSet = mutableSetOf<TestRememberKey>()
+  /**
+   * Tracks the keys of all calls to [runningSideEffect], so it can emulate the behavior of the real
+   * runtime and throw if a side effects is ran twice in the same pass.
+   */
+  private val ranSideEffects: MutableList<String> = mutableListOf()
+
+  /** TODO */
```
rjrjr, line 128: Ooops.

workflow-testing/src/main/java/com/squareup/workflow1/testing/RenderTester.kt
```diff
@@ -129,7 +130,7 @@ public fun <PropsT, StateT, OutputT, RenderingT>
  *   )
  *   .expectWorker(
  *     workerClass = SubmitLoginWorker::class
- *     key = "signin",
+ *     key = "signIn",
```
rjrjr, line 133: The verb phrase is two words!

workflow-testing/src/main/java/com/squareup/workflow1/testing/RealRenderTester.kt
```diff
@@ -101,25 +92,44 @@ internal class RealRenderTester<PropsT, StateT, OutputT, RenderingT>(
     ) : Expectation<Nothing>() {
       override fun describe(): String = description
     }
+
+    data class ExpectedRemember(
+      val matcher: (RememberInvocation) -> Boolean,
+      val exactMatch: Boolean,
+      val description: String,
+    ) : Expectation<Any?>() {
```
Plastix, line 100: Why is this `Expectation<Any?>`? Remembered expectations are closer to side effects so I'd expect `Expectation<Nothing>`. This matches the workflow expectation format which is maybe ok too? I don't really know the distinction here

Plastix (approved):

workflow-testing/src/main/java/com/squareup/workflow1/testing/RealRenderTester.kt
rjrjr: Oh right. That was right around the time I realized I couldn't provide a mock value. Thanks, good catch.

PR #1294, opened 2025-04-16
Update CSR to yield() and update tests

steve-the-edwards:
As I was testing `DRAIN_EXCLUSIVE_ACTIONS` and ensuring that our tests would mirror the way that runtime is usually run (that is, with `Dispatchers.Main.immediate` - which is an unconfined dispatcher), I realized that we should always be testing with `UnconfinedTestDispatcher` in the runtime in order to accurately represent that behavior vis-a-vis these optimizations.

That testing change caused me to recognize that if we do not `yield()` before processing extra actions then depending on where the coroutines for processing the updated signals are dispatched, they may not be able to actually process them and send their actions to the sink. (like multiple Workers listening to the same signal).

So here we add `yield()` before processing extra actions. 

There was also a bug where we did not output if we were short-circuiting after an action processed as part of CSR. I fixed this bug and updated the test to ensure it was tested.

I also updated `RenderWorkflowInTest` and `WorkflowLifecycleTests` tests to use the `UnconfinedTestDispatcher`. There was a confusing `@Ignore`d test there that I updated the kdoc on to show why it was ignored.

UPDATE: I made the dispatcher one of the parameters for the test and use both `UnconfinedTestDispatcher` and `StandardTestDispatcher`.

More discussion of this and alternatives: https://square.slack.com/archives/C07UHTZKAPJ/p1744896778454249

### DIFF
--- a/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
+++ b/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
@@ -12,6 +12,7 @@ import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.yield
 
 /**
  * Launches the [workflow] in a new coroutine in [scope] and returns a [StateFlow] of its
@@ -170,7 +171,7 @@ public fun <PropsT, OutputT, RenderingT> renderWorkflowIn(
   }
 
   scope.launch {
-    while (isActive) {
+    outer@ while (isActive) {
       // It might look weird to start by processing an action before getting the rendering below,
       // but remember the first render pass already occurred above, before this coroutine was even
       // launched.
@@ -178,7 +179,7 @@ public fun <PropsT, OutputT, RenderingT> renderWorkflowIn(
 
       if (shouldShortCircuitForUnchangedState(actionResult)) {
         sendOutput(actionResult, onOutput)
-        continue
+        continue@outer
       }
 
       // After resuming from runner.processAction() our coroutine could now be cancelled, check so
@@ -189,15 +190,22 @@ public fun <PropsT, OutputT, RenderingT> renderWorkflowIn(
       var nextRenderAndSnapshot: RenderingAndSnapshot<RenderingT> = runner.nextRendering()
 
       if (runtimeConfig.contains(CONFLATE_STALE_RENDERINGS)) {
-        while (isActive && actionResult is ActionApplied<*> && actionResult.output == null) {
+        conflate@ while (isActive && actionResult is ActionApplied<*> && actionResult.output == null) {
+          // We start by yielding, because if we are on an Unconfined dispatcher, we want to give
+          // other signals (like Workers listening to the same result) a chance to get dispatched
+          // and queue their actions.
+          yield()
           // We may have more actions we can process, this rendering could be stale.
           actionResult = runner.processAction(waitForAnAction = false)
 
           // If no actions processed, then no new rendering needed. Pass on to UI.
-          if (actionResult == ActionsExhausted) break
+          if (actionResult == ActionsExhausted) break@conflate
 
           // Skip rendering if we had unchanged state, keep draining actions.
-          if (shouldShortCircuitForUnchangedState(actionResult)) continue
+          if (shouldShortCircuitForUnchangedState(actionResult)) {
+            sendOutput(actionResult, onOutput)
+            continue@outer
+          }
 
           // Make sure the runtime has not been cancelled from runner.processAction()
           if (!isActive) return@launch
--- a/workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
+++ b/workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
@@ -20,11 +20,12 @@ import kotlinx.coroutines.launch
 import kotlinx.coroutines.suspendCancellableCoroutine
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.test.StandardTestDispatcher
+import kotlinx.coroutines.test.TestDispatcher
 import kotlinx.coroutines.test.TestScope
 import kotlinx.coroutines.test.UnconfinedTestDispatcher
 import kotlinx.coroutines.test.advanceUntilIdle
-import kotlinx.coroutines.test.runCurrent
 import kotlinx.coroutines.test.runTest
+import kotlinx.coroutines.yield
 import okio.ByteString
 import kotlin.test.Test
 import kotlin.test.assertEquals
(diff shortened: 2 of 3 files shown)


workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
```diff
@@ -190,14 +191,22 @@ public fun <PropsT, OutputT, RenderingT> renderWorkflowIn(
...
 
           // If no actions processed, then no new rendering needed. Pass on to UI.
           if (actionResult == ActionsExhausted) break
 
           // Skip rendering if we had unchanged state, keep draining actions.
-          if (shouldShortCircuitForUnchangedState(actionResult)) continue
+          if (shouldShortCircuitForUnchangedState(actionResult)) {
+            // Emit the Output
+            sendOutput(actionResult, onOutput)
+            continue
```
steve-the-edwards, line 208: This needs to continue at the outer loop.

steve-the-edwards: Done.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
```diff
@@ -170,15 +171,15 @@ public fun <PropsT, OutputT, RenderingT> renderWorkflowIn(
   }
 
   scope.launch {
-    while (isActive) {
+    outer@ while (isActive) {
```
zach-klippenstein, line 174: If this code is getting complex enough that we're starting to need labels, it's probably a sign that it should be factored down a bit.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
```diff
@@ -189,15 +190,23 @@ public fun <PropsT, OutputT, RenderingT> renderWorkflowIn(
...
           actionResult = runner.processAction(waitForAnAction = false)
 
           // If no actions processed, then no new rendering needed. Pass on to UI.
-          if (actionResult == ActionsExhausted) break
+          if (actionResult == ActionsExhausted) break@conflate
 
           // Skip rendering if we had unchanged state, keep draining actions.
-          if (shouldShortCircuitForUnchangedState(actionResult)) continue
+          if (shouldShortCircuitForUnchangedState(actionResult)) {
+            // Emit the Output
```
zach-klippenstein, line 206: Nit: This comment is pretty redundant, "emit" is just another word for "send" and that's literally the function name.

zach-klippenstein, line 206: Nit: This comment is pretty redundant, "emit" is just another word for "send" and that's literally the function name.

zach-klippenstein, line 206: Nit: This comment is pretty redundant, "emit" is just another word for "send" and that's literally the function name.

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
```diff
@@ -1543,20 +1557,24 @@ class RenderWorkflowInTest {
...
-        ) {}
-
-        launch {
-          trigger.emit("changed state")
+        ) {
+          // Yield in output so that we ensure that we let the collector of the renderings
+          // collect each of them before processing the next action.
+          yield()
         }
-        val collectionJob = launch(UnconfinedTestDispatcher(testScheduler)) {
```
zach-klippenstein, line 1551: Yay for consistency within the test!

zach-klippenstein (commented):
So this is great in that it now actually runs the tests the way we run the code in Register. However, it removes test coverage for normal dispatchers. Are we officially dropping support for non-immediate dispatchers? I thought we had some headless ones in Register, so we probably can't do that. It looks like we already have some kind of parameterized testing infra, can we use that to test on both types of dispatchers?

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
steve-the-edwards: I thought the labels were so nice and clean :D. but yes, perhaps refactoring some functions might be better?

steve-the-edwards:
> So this is great in that it now actually runs the tests the way we run the code in Register. However, it removes test coverage for normal dispatchers. Are we officially dropping support for non-immediate dispatchers? I thought we had some headless ones in Register, so we probably can't do that. It looks like we already have some kind of parameterized testing infra, can we use that to test on both types of dispatchers?

The parameterized testing is my own hand-rolled one here due to `RuntimeConfig` 😅 . 
We could extend it for dispatchers yes.

However, i don't think it would be that simple as the test shape will have to change slightly for a `StandardTestDispatcher` (in the least it would include `runCurrent()` etc.).

I think your point that "We should test with non-immediate dispatchers" is a good one.
I disagree that this 'takes away' that testing as the way it was currently was by no means robust or comprehensive. I think I would take your suggestion and make it into a separate PR that would add the same behavior testing for `RenderWorkflow` with non-immediate dispatchers (likely just as a separate suite of tests).

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
steve-the-edwards: Was this some kind of inception? 
 .... Was this some kind of inception? 
    ..... Was this some kind of inception?

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
steve-the-edwards: i don't want to do this right now as it should be done very carefully watching for its own behavior impacts because the methods we refactor out will need to handle the different cases of killing the scope vs. restarting outer loop vs continuing inner loop.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
zach-klippenstein: Wtf is going on lol 🫥

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
```diff
@@ -61,12 +62,21 @@ class RenderWorkflowInTest {
     testTracer
   )
 
-  private val runtimeOptions: Sequence<Pair<RuntimeConfig, WorkflowTracer?>> = cartesianProduct(
-    runtimes.asSequence(),
-    tracerOptions.asSequence()
+  private val myStandardTestDispatcher = StandardTestDispatcher()
+  private val dispatcherOptions = setOf<TestDispatcher>(
+    UnconfinedTestDispatcher(),
+    myStandardTestDispatcher
```
zach-klippenstein, line 68: Sweet!

zach-klippenstein (approved):

PR #1296, opened 2025-04-22
Pass Rendering if any Conflations changed state

steve-the-edwards:
We were short circuiting if the *last* action handled did not change state - but that's wrong. We cannot short circuit unless _all_ of the actions processed so far have not changed state (otherwise we might not pass the updated Rendering to the UI ever).

Found this in register, added test to red-green repro the scenario.

### DIFF
--- a/workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
+++ b/workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
@@ -1393,7 +1393,6 @@ class RenderWorkflowInTest {
               trigger.asWorker()
             ) {
               action("") {
-                state = it
                 setOutput(it)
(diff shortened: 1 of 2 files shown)


workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
```diff
@@ -1632,6 +1630,90 @@ class RenderWorkflowInTest {
     }
   }
 
+  @Test
+  fun for_conflate_and_render_only_when_state_changed_we_do_not_conflate_stacked_actions_into_one_rendering_if_previous_rendering_changed() {
```
zach-klippenstein, line 1634: Couldn't get goose to summarize this paragraph for you? 😜

zach-klippenstein (approved):

PR #1297, opened 2025-04-23
Add Interceptor Methods for Skip and Update Rendering

steve-the-edwards:
This is necessary in order to do effective tracing with optimizations enabled.

These are also helpful and quite logical methods for the interceptor.

Specifically `onRenderingUpdated(rendering)` is powerful as you could decorate every rendering passed out of the runtime via an interceptor.


workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt
```diff
@@ -12,10 +12,10 @@ import kotlin.reflect.KType
  * Provides hooks into the workflow runtime that can be used to instrument or modify the behavior
  * of workflows.
  *
- * This interface's methods mirror the methods of [StatefulWorkflow]. It also has one additional
- * method, [onSessionStarted], that is notified when a workflow is started. Each method returns the
- * same thing as the corresponding method on [StatefulWorkflow], and receives the same parameters
- * as well as two extra parameters:
+ * This interface's methods mirror the methods of [StatefulWorkflow], and three additional methods,
+ * explained below.
+ * Each method returns the same thing as the corresponding method on [StatefulWorkflow], and\
```
pyricau, line 17: `\` ?

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt
```diff
@@ -129,6 +142,21 @@ public interface WorkflowInterceptor {
     session: WorkflowSession
   ): Snapshot? = proceed(state)
 
+  /**
+   * Called when a [RenderingAndSnapshot] is updated from the runtime (i.e. when the [StateFlow]
+   * is provided a new value - this may still be de-duplicated).
```
pyricau, line 147: "this may still be de-duplicated" => by consumers of the state flow, based on some other things?

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt
```diff
@@ -129,6 +142,21 @@ public interface WorkflowInterceptor {
...
+   * @return [RenderingAndSnapshot]: a possibly modified [RenderingAndSnapshot].
+   */
+  public fun <R> onRenderingUpdated(
+    rendering: RenderingAndSnapshot<R>
+  ): RenderingAndSnapshot<R> = rendering
+
+  /**
+   * Called when the rendering is skipped because no state has changed after action cascade(s).
+   */
+  public fun onRenderingSkipped(): Unit = Unit
```
pyricau, line 158: if the outcome of an action loop is always either "onRenderingSkipped" or "onRenderingUpdated", then I would actually merge those and pass in a sealed class of "Updated" vs "Skipped", which would make this behavior a lot clearer.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt
steve-the-edwards: by the `StateFlow` itself which de-duplicates emissions downstream.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt
steve-the-edwards: ya I think that sounds reasonable!

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt
```diff
@@ -129,6 +142,21 @@ public interface WorkflowInterceptor {
     session: WorkflowSession
   ): Snapshot? = proceed(state)
 
+  /**
+   * Called when a [RenderingAndSnapshot] is updated from the runtime (i.e. when the [StateFlow]
+   * is provided a new value - this may still be de-duplicated).
+   *
+   * @return [RenderingAndSnapshot]: a possibly modified [RenderingAndSnapshot].
+   */
+  public fun <R> onRenderingUpdated(
```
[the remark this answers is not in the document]
steve-the-edwards: yes I can rename this.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt
```diff
@@ -28,6 +28,19 @@ import kotlin.reflect.KType
...
+ * There are 3 more methods in this interface:
+ *
+ *  1. [onSessionStarted] - called when a new [WorkflowSession] is created the first time a
+ *     workflow is rendered with the [CoroutineScope] for that session.
+ *  1. [onRenderingSkipped] - called when the rendering is short-circuited because no state
+ *     change was detected as part of the action cascade(s) and the
+ *     [RuntimeConfig.RENDER_ONLY_WHEN_STATE_CHANGES] is enabled.
+ *  1. [onRenderingUpdated] - called when a [RenderingAndSnapshot] is passed to the [StateFlow]
+ *     returned by [renderWorkflowIn]. It is passed the [RenderingAndSnapshot] which could be
+ *     decorated.
```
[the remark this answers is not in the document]
steve-the-edwards: `onSessionStarted` actually existed before these changes (I just moved the documentation and made it clear).

Yes, agreed that we are violating SRP and there is an opening here for a higher/orthogonal tracing dedicated API. I also don't want to tackle that with this work if possible and the interceptor can handle this without growing any more unwieldy than it already is, I think.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt
```diff
@@ -129,6 +142,29 @@ public interface WorkflowInterceptor {
...
+   */
+  public fun onRuntimeLoopIteration(
+    outcome: RuntimeLoopOutcome
+  ): Unit = Unit
+
+  public sealed interface RuntimeLoopOutcome
+  public data object RenderPassSkipped : RuntimeLoopOutcome
+  public data class RenderPassesComplete<R>(
+    val renderingAndSnapshot: RenderingAndSnapshot<R>
+  ) : RuntimeLoopOutcome
```
rjrjr, line 166: We've avoided `data class` and `data object` in public types.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt
rjrjr, line 42: Stale kdoc, there is only one new method now.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt
```diff
@@ -129,6 +142,29 @@ public interface WorkflowInterceptor {
     session: WorkflowSession
   ): Snapshot? = proceed(state)
 
+  /**
+   * Called when a the runtime loop has "iterated." Typically this the application of 1 action
+   * and 1 render pass.
```
rjrjr, line 147:
```suggestion
  * Called to report the [outcome] of each tick of the runtime loop. In the simplest case 
  * this is the application of one action and one render pass, but optimizations can
  * change that:
```

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt
```diff
@@ -129,6 +142,29 @@ public interface WorkflowInterceptor {
...
+   *
+   * If there is no state change with the `RENDER_ONLY_WHEN_STATE_CHANGES` optimization,
+   * it might not be a render pass at all, in which case [RenderPassSkipped] is the outcome.
+   * In the case of the `CONFLATE_STALE_RENDERINGS` optimization it could be multiple render
+   * passes.
+   * If there is at least one render pass, then [RenderPassesComplete] is passed as the outcome,
+   * which includes the actual [RenderingAndSnapshot] returned from the runtime.
```
rjrjr, line 154: Maybe make this a bullet list.

rjrjr (approved):

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt
steve-the-edwards: Thanks for catching that. Will fix.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt
steve-the-edwards: Will update.

PR #1302, opened 2025-05-05
Better crash reporter error grouping from RenderContext assertions

rjrjr:
Normally crash reporters group errors by hashing against the elements of the top stack frame. This causes problems for bottleneck code that is enforcing invariants that should be implemented by client code: we wind up with "kitchen sink" groups that hold every crash of a particular type, rather than distinct groups for each specific way that mistake is made.

We fix this for a few spots in `RenderContext` by inserting a fake top stack frame based on some hashable key provided by client code -- the child workflow, the "key" param for runningSideEffect, etc.

So far this is a JVM-specific hack.

### DIFF
--- a/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/Throwables.kt
+++ b/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/Throwables.kt
@@ -1,8 +1,66 @@
 package com.squareup.workflow1.internal
 
-import kotlinx.coroutines.CancellationException
+import kotlin.contracts.ExperimentalContracts
+import kotlin.contracts.contract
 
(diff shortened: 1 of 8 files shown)


workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/Throwables.kt
```diff
@@ -1,8 +1,66 @@
 package com.squareup.workflow1.internal
 
-import kotlinx.coroutines.CancellationException
+import kotlin.contracts.ExperimentalContracts
+import kotlin.contracts.contract
 
-internal tailrec fun Throwable.unwrapCancellationCause(): Throwable? {
```
rjrjr, line 5: This was unused.

steve-the-edwards (approved):

PR #1310, opened 2025-05-12
Better choices for `stackTraceKey`

rjrjr:
Found a few spots where `stackTraceKey` was including id hashes, resulting in separate crash reporter error groups per process. Fixed that and over-documented the rules. Glad this isn't public API.

### DIFF
--- a/workflow-core/src/commonMain/kotlin/com/squareup/workflow1/WorkerWorkflow.kt
+++ b/workflow-core/src/commonMain/kotlin/com/squareup/workflow1/WorkerWorkflow.kt
@@ -40,7 +40,8 @@ internal class WorkerWorkflow<OutputT>(
       unsnapshottableIdentifier(workerType)
     }
 
-  override fun describeRealIdentifier(): String = workerType.toString()
+  override fun describeRealIdentifier(): String =
+    workerType.toString().replace(" (Kotlin reflection is not available)", "")
 
   override fun initialState(
     props: Worker<OutputT>,
--- a/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/RealRenderContext.kt
+++ b/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/RealRenderContext.kt
@@ -72,7 +72,9 @@ internal class RealRenderContext<out PropsT, StateT, OutputT>(
     key: String,
     handler: (ChildOutputT) -> WorkflowAction<PropsT, StateT, OutputT>
   ): ChildRenderingT {
-    checkNotFrozen(child) { "renderChild(${child.identifier})" }
+    checkNotFrozen(child.identifier) {
+      "renderChild(${child.identifier})"
+    }
     return renderer.render(child, props, key, handler)
   }
 
@@ -111,6 +113,7 @@ internal class RealRenderContext<out PropsT, StateT, OutputT>(
 
   /**
    * @param stackTraceKey ensures unique crash reporter error groups.
+   * It is important that keys are stable across processes, avoid system hashes.
    *
    * @see checkWithKey
    */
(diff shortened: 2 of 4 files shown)


rjrjr:
```
  at com.squareup.workflow1.internal.RealRenderContext.fakeMethodForCrashGrouping(com.squareup.workflow1.Worker<[REDACTED]> (Kotlin reflection is not available)-key:237304290)
```

- Not thrilled about `Kotlin reflection is not available` coming from the identifier but that's just a nit
- That "key" is okay this time but I think it's a mistake to use it. Remembering too many UUID calls for that.

rjrjr:
> ```
>   at com.squareup.workflow1.internal.RealRenderContext.fakeMethodForCrashGrouping(com.squareup.workflow1.Worker<[REDACTED]> (Kotlin reflection is not available)-key:237304290)
> ```
> 
> * Not thrilled about `Kotlin reflection is not available` coming from the identifier but that's just a nit
> * That "key" is okay this time but I think it's a mistake to use it. Remembering too many UUID calls for that.

Actually that kotlin reflection thing is pretty bad, the type is the only unique bit.

rjrjr:
> > ```
> >   at com.squareup.workflow1.internal.RealRenderContext.fakeMethodForCrashGrouping(com.squareup.workflow1.Worker<[REDACTED]> (Kotlin reflection is not available)-key:237304290)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * Not thrilled about `Kotlin reflection is not available` coming from the identifier but that's just a nit
> > * That "key" is okay this time but I think it's a mistake to use it. Remembering too many UUID calls for that.
> 
> Actually that kotlin reflection thing is pretty bad, the type is the only unique bit.

Ah, but the redacted bit had a very nice name. Okay, I think this will work. Going to look a bit harder to see if I can get the "Kotlin reflection" noise out of there, every time it shows up is irritating.

rjrjr:
> Ah, but the redacted bit had a very nice name. Okay, I think this will work. Going to look a bit harder to see if I can get the "Kotlin reflection" noise out of there, every time it shows up is irritating.

No wait, that's in the message, not the stack key. So the stack key should be using whatever is in the message…

rjrjr:
That's the `toString()` of `WorkflowIdentifier`, which looks like it was crazy carefully over engineered. I was avoiding it, going to go ahead and use it.

rjrjr:
> That's the `toString()` of `WorkflowIdentifier`, which looks like it was crazy carefully over engineered for exactly this kind of case. I was avoiding it, going to go ahead and use it.

- I confused myself, nothing in the line I quoted is part of the message, it really was all the key. So yes, uniqueness looks good
- that damn "Kotlin reflection is not available" bit is still in that ID to string and you know what? I'm just going to filter it the hell out right at the source.

rjrjr:
> > That's the `toString()` of `WorkflowIdentifier`, which looks like it was crazy carefully over engineered for exactly this kind of case. I was avoiding it, going to go ahead and use it.
> 
> * I confused myself, nothing in the line I quoted is part of the message, it really was all the key. So yes, uniqueness looks good
> * that damn "Kotlin reflection is not available" bit is still in that ID to string and you know what? I'm just going to filter it the hell out right at the source.

Ah, bliss!

```
  at com.squareup.workflow1.internal.RealRenderContext.fakeMethodForCrashGrouping(com.squareup.workflow1.Worker<com.squareup.squareone.core.Sq…us>:1728925657)
```

rjrjr:
Bah, backing out the reflection tweak b/c of a broken unit test, `WorkflowNodeTest.sideEffect_coroutine_is_named()`. But I'm coming back for it!

steve-the-edwards (approved):

Issue #1311, opened 2025-05-13
On Android, process all updates from actions from an event before the next frame (Choreographer.doFrame)


steve-the-edwards:
We use an inner loop for `CONFLATE_STALE_RENDERINGS` and `DRAIN_EXCLUSIVE_ACTIONS` ([here](https://github.com/square/workflow-kotlin/pull/1269)) to apply multiple actions in one render pass (or in the case of CSR, in one 'UI update').

When the runtime is executed on the `Main.immediate` dispatcher, we handle one side effect of a worker update, and then render and try to process actions _before_ the other workers (and their side effects underneath) have been able to be dispatched in order to send the other actions that need processing to the action sink.

This means that the optimizations like `CONFLATE_STALE_RENDERINGS` are much less effective with the `Main.immediate` dispatcher as we are not able to populate the actions that need processing as fast as we can process them.

We need to fix that! We have enshrined that desire in a currently ignored android test that uses the `Main.immediate` dispatcher.

steve-the-edwards:
The reason we could not use `yield()` in the middle of a runtime loop update is we want to keep the semantics of "1 main thread message for 1 workflow update." 

This does not meant that we can't `yield()` _before_ processing the first action after resuming and knowing that it is ready.

See #1344

steve-the-edwards:
> Which would take us out of the tick of the click event, wouldn't it? 

Yes this is a great point. I need to _only_ do this for Workers and/or possibly side effects if we give them a special action type.

rjrjr:
> and/or possibly side effects if we give them a special action type

If you were able to pull that off I think it might cover workers as well. But pragmatically, almost no one is posting actions from side effects.

steve-the-edwards:
The update here is that the important part is not when or if we use `yield()`. The important part is the following:

1. We should process _all_ updates in response to an event before the next 'frame'. E.g., on Android, this would be before the next `Choreographer.doFrame()`

That ^ is the guarantee we want to satisfy.

PR #1312, opened 2025-05-13
#1311: Start side effects `DEFAULT` instead of `LAZY`.

steve-the-edwards:
See #1311 for more information on the problem.

The first commit is a red test that will test that behaviour on Android with the `Main.immediate` dispatcher.

The second commit makes that test green!! 

It does so by launching side effects with `DEFAULT` instead of `LAZY`. That means that with an eager dispatcher like `Main.immediate` your workers and side effects will be eagerly entered as they are created during `render`. We want to maintain the distinction that we cannot send to the sink _in render()_ but we need to be able to do so in a worker/side effect. So we freeze the `RenderContext` whenever we dispatch the coroutine for that side effect (and then unfreeze it as necessary when that is complete) using a `ContinuationInterceptor`.

Note that this is a pretty significant change to the contract for workers/side effects; we've explored this before without making any significant changes, preferring instead to provide the `CoroutineScope` of the Workflow with a `SessionWorkflow` ( https://github.com/square/workflow-kotlin/pull/1102, https://github.com/square/workflow-kotlin/issues/1093 ).

In this instance our motivation is to enable further optimization and it is a pretty good motivation!

We will want to reflect on ramifications of this semantics change.

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -64,6 +64,7 @@ include(
   ":workflow-config:config-jvm",
   ":workflow-core",
   ":workflow-runtime",
+  ":workflow-runtime-android",
   ":workflow-rx2",
   ":workflow-testing",
   ":workflow-tracing",
--- a/workflow-runtime-android/README.md
+++ b/workflow-runtime-android/README.md
@@ -1,0 +1,4 @@
+# Module Workflow Runtime Android
+
+This module is an Android library that is used to test the Workflow Runtime with Android specific
+coroutine dispatchers. These are headless android-tests that run on device without UI.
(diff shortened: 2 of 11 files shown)


steve-the-edwards:
- [ ]  I need to use a delegating `ContinuationInterceptor` so that we still use whatever dispatcher is in the context - just adding the layer of freeze/unfreeze. E.g., see https://github.com/Kotlin/kotlinx.coroutines/issues/2478

steve-the-edwards:
Not needed. see https://github.com/square/workflow-kotlin/pull/1353

PR #1313, opened 2025-05-13
Completes our navigation story

rjrjr:
#### Optimize central `unwrap` implementations.

The default implementation of `Unwrappable.unwrap` creates a `Sequence` that is almost never necessary (but it is at least fail-safe so I think we should keep it). Here we provide improved implementations for `Wrapper` and `BackStackScreen`.

#### Introduces NavigationMonitor

Takes advantage of `unwrap()` and `Compatible.keyFor()` to provide navigation logging by reporting the top (read: last-most inner-most) sub-rendering, which conventionally is the one that is visible and accessible to the user. Reports each time the `Compatible.keyFor` the top is unequal to the previous one, which conventionally indicates that a new view object will replace the previous one.

### DIFF
--- a/workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/navigation/NavigationMonitor.kt
+++ b/workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/navigation/NavigationMonitor.kt
@@ -1,0 +1,52 @@
+package com.squareup.workflow1.ui.navigation
+
+import com.squareup.workflow1.ui.Compatible
+import com.squareup.workflow1.ui.unwrap
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.onEach
+
+/**
+ * Reports navigation across a series of calls to [update], probably made
+ * for each rendering posted by
+ * [renderWorkflowIn][com.squareup.workflow1.renderWorkflowIn].
+ *
+ * Takes advantage of [unwrap()] and [Compatible.keyFor] to provide navigation
+ * logging by reporting the top (read: last-most, inner-most) sub-rendering,
+ * which conventionally is the one that is visible and accessible to the user.
+ *
+ * Reports each time the [Compatible.keyFor] the top is unequal to the previous one,
+ * which conventionally indicates that a new view object will replace the previous one.
+ */
+public class NavigationMonitor(
+  skipFirstScreen: Boolean = false,
+  private val onNavigate: (Any) -> Unit = { println(it) }
+) {
+  @Volatile
+  private var lastKey: String? = if (skipFirstScreen) null else ""
+
+  /**
+   * Uses [unwrap] to find the topmost element of [rendering] and
+   * reports it with [onNavigate] if [Compatible.keyFor] reveals that
+   * it is of a different kind from the previous top.
+   */
+  public fun update(rendering: Any) {
+    val unwrapped = rendering.unwrap()
+
+    Compatible.keyFor(unwrapped).takeIf { it != lastKey }?.let { newKey ->
+      if (lastKey != null) onNavigate(unwrapped)
+      lastKey = newKey
+    }
+  }
+}
+
+/**
+ * Creates a [NavigationMonitor] and [updates it][NavigationMonitor.update]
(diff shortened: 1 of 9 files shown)


workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/navigation/NavigationMonitor.kt
```diff
@@ -0,0 +1,51 @@
...
+ * which conventionally is the one that is visible and accessible to the user.
+ *
+ * Reports each time the `Compatible.keyFor` the top is unequal to the previous one,
+ * which conventionally indicates that a new view object will replace the previous one.
+ */
+public class NavigationMonitor(
+  skipFirstScreen: Boolean = false,
+  private val onNavigate: (Any) -> Unit = { println(it) }
+) {
+  private var lastKey: String? = if (skipFirstScreen) null else ""
```
rjrjr, line 24: @pyricau @steve-the-edwards We have only ever used our flavor of this on the main thread with the Main.Immediate dispatcher, so this simple implementation has been just fine. Should we stick with what we know? Should this field be marked `@Volatile`?

rjrjr: My gut is to leave it this way. It's such a trivial class that anyone doing something more interesting and finding flaws could so easily fork it.

Should I put something in the kdoc warning of its limitations?

pyricau: Enforce that it's single threaded.

rjrjr: I think that's the sanity check I was hoping for, "`@Volatile` is just fine."

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/navigation/NavigationMonitor.kt
```diff
@@ -0,0 +1,52 @@
...
+   * Uses [unwrap] to find the topmost element of [rendering] and
+   * reports it with [onNavigate] if [Compatible.keyFor] reveals that
+   * it is of a different kind from the previous top.
+   */
+  public fun update(rendering: Any) {
+    val unwrapped = rendering.unwrap()
+
+    Compatible.keyFor(unwrapped).takeIf { it != lastKey }?.let { newKey ->
+      if (lastKey != null) onNavigate(unwrapped)
+      lastKey = newKey
```
rjrjr, line 37: @steve-the-edwards this is the order of these operations in house. Wondering if I should assign `lastKey` before calling `onNavigate`:

```kotlin
lastKey.let { lastKeyOrNull ->
  lastKey = newKey
  if (lastKeyOrNull != null) onNavigate(unwrapped)
}
```

Only rationale is to better handle any terrible choices a dev makes that leads a workflow update as a side effect of `onNavigate()`, so that a recursive call won't have its `lastKey` value stomped.

On the one hand, why even pretend to support that. On the other hand, it'll be maddening to track down if it happens.

rjrjr: @pyricau WDYT?

pyricau: I don't think you should aim to recover right if a dev triggers a workflow update as a side effect of `onNavigate()`. That just seems wrong?

So, two options:
- A) Don't care, if it happens all bets are off
- B) Make sure it can't happen: enforce single threadness, set a boolean field while calling onNavigate to throw if we reenter.

pyricau (approved):

workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/navigation/NavigationMonitor.kt
rjrjr: I'm going to err on the side of under- rather than over-engineering.

PR #1322, opened 2025-05-22
Create module for a new Workflow Trace Viewer app

wardellbagby:
- Creates a new module that when ran, runs a Compose for Desktop app. The plan is for that module to host a new desktop app for viewing Workflow traces.


wardellbagby:
Bah, I broke Kotlin. Gimmie a bit here.

workflow-trace-viewer/build.gradle.kts
```diff
@@ -0,0 +1,50 @@
...
+    jvmMain {
+      dependencies {
+        implementation(compose.runtime)
+        implementation(compose.foundation)
+        implementation(compose.material)
+        implementation(compose.ui)
+        implementation(compose.components.resources)
+        implementation(compose.components.uiToolingPreview)
+        implementation(libs.androidx.lifecycle.viewmodel.core)
+        implementation(libs.androidx.lifecycle.compose)
```
steve-the-edwards, line 22: I think a lot of these are forward looking what might be useful, right? But will these `androidx` ones be usable?

steve-the-edwards (approved):

workflow-trace-viewer/build.gradle.kts
wardellbagby: Yeah, they're KMP now and are expected to used with Compose for Desktop.

PR #1326, opened 2025-05-23
Update Publishing Plugin

steve-the-edwards:
For #1314

Already approved here - https://github.com/square/workflow-kotlin/pull/1317 - but Github was choking somehow on auto-merge.


steve-the-edwards:
Note that I am considering this approved because https://github.com/square/workflow-kotlin/pull/1317 was approved but it was broken for merge (just infinitely spinning)

ijwhelan (approved):

PR #1328, opened 2025-05-23
Kotlin 2.0.21 & Return 'out RenderingT' variance

### LINKED ISSUE #1327: Add back covariance for `RenderingT`
Turns out we do make use of covariant types for Workflow based on subclasses of `RenderingT`, such as any subclass of `Screen`. We need to find a way to get this back in, even with Kotlin 2+

steve-the-edwards:
Fixes #1327 

We do rely on covariance for `RenderingT` for Workflow subtypes that subtype the `RenderingT`. Also we don't need to remove it due to UnsafeVariance.

Also dropping back down to `2.0.21` from `2.1.21` since that is what we currently use internally and this way we don't have to worry about matching binary compatibility.


.npmrc
```diff
@@ -0,0 +1 @@
+registry=https://registry.yarnpkg.com
```
steve-the-edwards, line 1: Another attempt to try and prevent local config from changing the registry specified by `yarn.lock` when we regenerate it for new versions. The previous `.yarnrc` did not work 🤔 . Either way I've manually corrected the `yarn.lock` again.

ijwhelan (approved):

PR #1330, opened 2025-05-29
Give EnvironmentScreen a toString to help when debugging.

zach-klippenstein:


workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/EnvironmentScreen.kt
```diff
@@ -14,6 +14,11 @@ public class EnvironmentScreen<out C : Screen>(
 ) : Wrapper<Screen, C>, Screen {
   override fun <D : Screen> map(transform: (C) -> D): EnvironmentScreen<D> =
     EnvironmentScreen(transform(content), environment)
+
+  override fun toString(): String = "EnvironmentScreen(" +
+    "content type=${content::class}, " +
```
rjrjr, line 19: This will be more informative and less noisy.

```suggestion
    "content key=${Compatible.keyFor(content)}, " +
```

PR #1335, opened 2025-06-05
WIP: wenli/workflow viewer

wenli-cai:
Given snapshots of all workflows, the visualizer shows a tree-like structure to allow developers to have an easier time of debugging through state changes and render passes

### Current progress
1. Basic sandbox-y background to allow free movement
2. Given a simple json data to represent each `WorkflowNode`, a basic tree structure can be drawn; these nodes can also be opened and closed upon clicking
3. Arrows to show relationship between nodes, however this is currently bugged and not in the right location


workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
```diff
@@ -5,5 +5,12 @@ import androidx.compose.runtime.Composable
 
 @Composable
 fun App() {
-  Text("Hello world!")
+  val jsonString = object {}.javaClass.getResource("/workflow-simple.json")?.readText()
```
japplin: I think getClass().getResource should work here

workflow-trace-viewer/build.gradle.kts
```diff
@@ -23,6 +23,8 @@ kotlin {
         implementation(compose.desktop.currentOs)
         implementation(libs.kotlinx.coroutines.swing)
         implementation(compose.materialIconsExtended)
+        implementation(libs.moshi.kotlin)
+        implementation("io.github.vinceglb:filekit-dialogs-compose:0.10.0-beta03")
```
steve-the-edwards: Add this to `gradle/libs.versions.toml` and then reference it here rather than the direct include (eases maintenance).

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/SandboxBackground.kt
```diff
@@ -0,0 +1,83 @@
...
+  content: @Composable () -> Unit,
+) {
+  var scale by remember { mutableFloatStateOf(1f) }
+  var offset by remember { mutableStateOf(Offset.Zero) }
+
+  Box(
+    modifier
+      .fillMaxSize()
+      .pointerInput(Unit) {
+        // this allows for user's panning to view different parts of content
```
steve-the-edwards:
```suggestion
        // This allows for user's panning to view different parts of content.
```

All comments should be sentences and use punctuation as such.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/SandboxBackground.kt
```diff
@@ -0,0 +1,83 @@
...
+        awaitEachGesture {
```
steve-the-edwards, line 48: Is this from a blog post or ref article or something? If so, it can be valuable to post the link in a comment as then we can follow up during maintenance.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/SandboxBackground.kt
steve-the-edwards: Here and elsewhere.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/SandboxBackground.kt
```diff
@@ -0,0 +1,83 @@
...
+            event.changes.forEach { it.consume() }
+          }
+
+          // panning: this tracks multiple events within one gesture to see what the user is doing,
+          // then calculates the offset and pans the screen accordingly
+          val drag = event.changes.firstOrNull()
+          if (drag != null && drag.pressed) {
+            var prev = drag.position
+            while (true) {
+              val nextEvent = awaitPointerEvent()
```
steve-the-edwards: there is no way to get stuck here right? As in, is it possible we already captured event.changes but those were all of them and there is nothign to `await()`? or is this `await()`ing changes we already know we have?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/UploadFile.kt
```diff
@@ -0,0 +1,51 @@
...
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import io.github.vinceglb.filekit.PlatformFile
+import io.github.vinceglb.filekit.dialogs.FileKitType
+import io.github.vinceglb.filekit.dialogs.compose.rememberFilePickerLauncher
+
+@Composable
+public fun UploadFile(
```
steve-the-edwards, line 25: Needs kdoc.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/WorkflowJsonParser.kt
```diff
@@ -0,0 +1,29 @@
...
+import com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterFactory
+import java.io.IOException
+
+/**
+ * Parses a JSON string into [WorkflowNode] with Moshi adapters
+ *
+ * All the caught exceptions should be handled by the caller, and appropriate UI feedback should be
+ * provided to user
+ */
+public fun fetchRoot(
```
steve-the-edwards: Maybe change name to `parseTrace` or something? `fetchRoot` does not communicate all that it is doing, even though the root `WorkflowNode?` is what is returned.

wenli-cai: Done

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/UploadFile.kt
wenli-cai: Done

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/SandboxBackground.kt
wenli-cai: Done and fixed all other grammatical errors in comments.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/SandboxBackground.kt
wenli-cai: Added note that code is AI generated.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/SandboxBackground.kt
wenli-cai: Changed functionality with higher-level `onDragGestures` API from Android

workflow-trace-viewer/build.gradle.kts
wenli-cai: Done

gradle/libs.versions.toml
```diff
@@ -3,7 +3,10 @@
 agpVersion = "8.8.0"
 
 compileSdk = "34"
+# Any version above 0.10.0-beta03 requires Compose 1.8.0 or higher.
+filekitDialogsComposeVersion = "0.10.0-beta03"
```
wardellbagby: Let's stick with the same naming convention we have here. Android SDK versions get camel case but dependency versions use kebab-case. In that same vein, dependencies should be named without the `version` suffix.

So this would be `filekit-dialogs-compose`, Moshi would be `moshi-kotlin`, and uiGraphicsAndroidVersion would be `androidx-ui-graphics-android`.

This is also is roughly alphabetical order, with spacings denoting group separations for when we have many dependencies all related to the same thing, like `androidx`, so this and the other new versions should be listed in that same order.

It's ultimately a small thing admittedly, but it's always good to be consistent 'cause then things tend to stay how you'd expect them to be when you come back later!

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/SandboxBackground.kt
```diff
@@ -0,0 +1,72 @@
...
+  var offset by remember { mutableStateOf(Offset.Zero) }
+
+  Box(
+    modifier
+      .fillMaxSize()
+      .pointerInput(Unit) {
+        // Panning capabilities: watches for drag gestures and applies the translation
+        detectDragGestures { _, translation->
+          offset += translation
+        }
```
steve-the-edwards, line 43: Nice!

steve-the-edwards (approved):
LGTM. A couple things still before merge:

- [ ] Add the `.klib` to `.gitignore` as @wardellbagby was suggesting
- [x] get sign off from someone more well versed with Compose then me (@wardellbagby , @zach-klippenstein, etc.)

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
```diff
@@ -1,9 +1,44 @@
...
+ * Main composable that provides the different layers of UI.
+ */
 @Composable
-fun App() {
-  Text("Hello world!")
+public fun App(
+  modifier: Modifier = Modifier
+) {
+  Box {
+    val selectedFile = remember { mutableStateOf<PlatformFile?>(null) }
```
wardellbagby: Prefer using Kotlin delegation here.

```kotlin
var selectedFile by remember { mutableStateOf<PlatformFile?>(null) }
```

That'll let you do simple reassignments without having to do the whole `.value` stuff to access or change the value. Android Studio will try to prompt you to import the `getValue` and `setValue` extension methods that make that possible, but if it doesn't, you may need to manually import them. 

An example of this syntax is here: https://github.com/squareup/android-emulator-runner/blob/92d600d5e9cb215e3509ee215881e55a42fb6fac/desktop-app/src/jvmMain/kotlin/com/wardellbagby/aer/ui/App.kt#L192

The imports you _may_ need to add manually are:

```kotlin
import androidx.compose.runtime.getValue
import androidx.compose.runtime.setValue 
```

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/WorkflowJsonParser.kt
```diff
@@ -0,0 +1,29 @@
...
+import com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterFactory
+import java.io.IOException
+
+/**
+ * Parses a JSON string into [WorkflowNode] with Moshi adapters.
+ *
+ * All the caught exceptions should be handled by the caller, and appropriate UI feedback should be
+ * provided to user.
+ */
+public fun parseTrace(
```
wardellbagby, line 14: JSON parsing _can_ be long running. This would be better as a `suspend fun` that wraps this work in a `withContext(Dispatchers.IO)` block. Then, instead of having a `jsonString` being used as state in your `WorkflowContent` Composable, you'd have a nullable `WorkflowNode`.

That would make `WorkflowContent` look something like this:

```kotlin
@Composable
private fun WorkflowContent(file: PlatformFile?) {
  var rootNode by remember { mutableStateOf<WorkflowNode?>(null) }
  LaunchedEffect(file) {
     if(file != null) { 
       rootNode = parseTrace(file?.readString()) // in the future, a try-catch here to switch to an error state!
    }
  }

  if (rootNode != null) {
    DrawWorkflowTree(root)
  } else {
    Text("Empty data or failed to parse data") // TODO: proper handling of error
  }
}
```

For the future, when you get to the point where you might wanna show an error on parse failures, you'd instead just let this throw and expect it to be caught in the `LaunchedEffect` you call this from so that the consumer could switch to some error state. But that's looking ahead so don't stress that now!

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
```diff
@@ -1,9 +1,44 @@
...
+    if (selectedFile.value != null) {
+      SandboxBackground { WorkflowContent(selectedFile.value) }
+    }
+
+    UploadFile(onFileSelect = { selectedFile.value = it })
+  }
+}
+
+@Composable
+private fun WorkflowContent(file: PlatformFile?) {
```
wardellbagby: Nit: No need for `file` to be nullable here. I imagine it's because Kotlin thinks `.value` can change but we know it won't so it's okay to use `!!` here to force it to be non-null.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/WorkflowTree.kt
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * Since the workflow nodes present a tree structure, we utilize a recursive function to draw the tree.
+ * The Column holds a subtree of nodes, and the Row holds all the children of the current node.
+ */
+@Composable
+public fun DrawWorkflowTree(
+  node: WorkflowNode,
+  modifier: Modifier = Modifier,
+) {
+  Column(
```
wardellbagby, line 44: This works fine for now but eventually, we're probably gonna have to get _lazy_ with it. I'm not familiar with it but I recently learned that Compose has a `LazyLayout` Composable that, similar to `LazyRow` and `LazyColumn`, lets you render a bunch of stuff but is smart enough to only render things that are actually in view. Since these traces are gonna get big, we'll start to see lagging when scrolling. We might already with your `workflow-300` trace!

I can probably bully one of the UI System folks for more info on how we could potentially use that here, since the documentation is really sparse and `LazyLayout` is still experimental.

wardellbagby (commented):
Left a few comments for you but all in all, this does look great!

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/WorkflowJsonParser.kt
wenli-cai: Working on this in another branch!

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/WorkflowTree.kt
wenli-cai: `workflow-300` seems to be fine when scrolling for now... but I will keep this in mind in the future once the traces get very large.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt

gradle/libs.versions.toml
wenli-cai: Done

PR #1339, opened 2025-06-10
1338: Clean up caches after WorkflowNode torn down

### LINKED ISSUE #1338: `canonicalStatefulRenderContext` is leaked after the WorkflowNode is no longer using it; also leaks `lastRendering`

steve-the-edwards:
Fixes #1338

### DIFF
--- a/workflow-core/src/commonMain/kotlin/com/squareup/workflow1/StatelessWorkflow.kt
+++ b/workflow-core/src/commonMain/kotlin/com/squareup/workflow1/StatelessWorkflow.kt
@@ -227,7 +227,7 @@ public abstract class StatelessWorkflow<PropsT, OutputT, out RenderingT> :
    * Class type returned by [asStatefulWorkflow].
    * See [statefulWorkflow] for the instance.
    */
-  private inner class StatelessAsStatefulWorkflow :
+  inner class StatelessAsStatefulWorkflow :
     StatefulWorkflow<PropsT, Unit, OutputT, RenderingT>() {
 
     /**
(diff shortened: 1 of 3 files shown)


workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
```diff
@@ -234,11 +235,10 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
    * after calling this method.
    */
   fun cancel(cause: CancellationException? = null) {
-    // No other cleanup work should be done in this function, since it will only be invoked when
-    // this workflow is *directly* discarded by its parent (or the host).
-    // If you need to do something whenever this workflow is torn down, add it to the
-    // invokeOnCompletion handler for the Job above.
```
steve-the-edwards, line 240: This comment was old. We now `cancel` anytime we stop actively rendering a node.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/StatelessWorkflow.kt
```diff
@@ -268,6 +268,23 @@ public abstract class StatelessWorkflow<PropsT, OutputT, out RenderingT> :
     }
 
     override fun snapshotState(state: Unit): Snapshot? = null
+
+    /**
+     * When we are finished with at least one node that holds on to this workflow instance,
+     * then we clear the cache. The reason we do that every time is that it *might* be the last
```
zach-klippenstein, line 274: Just gonna say, this sounds like a problem for refcounting, although if this is a temporary bandaid while we prep for a bigger, more comprehensive fix, who cares.

PR #1341, opened 2025-06-10
1340: Changing `RenderContext` from an inner class to a nested class with its own generics

### LINKED ISSUE #1340: `RenderingT` still has unsafe variance
I brought back the `out RenderingT` variance declaration because we use the covariance across `RenderingT` for workflows quite a bit (https://github.com/square/workflow-kotlin/pull/1328). I thought we did not need to remove it for `RenderingT`.

However, while not immediately clear (in fact I don't think my IDE was even showing me the error at first), it still has `UnsafeVariance` when we create the inner class `RenderContext<PropsT, StateT, OutputT>` :(.

The reason is that we use the `PropsT`, `StateT`, and `OutputT` that are defined on the `StatefulWorkflow` or `StatelessWorkflow` instance that surrounds the inner class of `RenderContext`, so we are still passing *in* (consuming) the instance that defines `RenderingT` when we create `RenderContext<PropsT, StateT, OutputT>` which has no direct use of `RenderingT` itself. 😢 .

Why did this work? Well I believe this worked because by the time I brought back `out RenderingT` we had specified `kotlinLanguageVersion = 1.9` to the compiler, which means `UnsafeVariance` is still allowed until we use the K2 compiler.

So this still needs to be fixed somehow.

steve-the-edwards:
Yes, this means we have to provide type parameters every time we use it.

However, it also means we won't break the `out RenderingT` variance (which will not be allowed come the K2 compiler).

This will fix #1340 as we will no longer be consuming `RenderingT` (via the Workflow instance to get the `PropsT` and `OutpuT` types) when we pass `RenderContext` into `render()` as its types - for `PropsT` and `OutpuT` will be provided explicitly.

### DIFF
--- a/samples/hello-terminal/hello-terminal-app/src/main/java/com/squareup/sample/helloterminal/HelloTerminalWorkflow.kt
+++ b/samples/hello-terminal/hello-terminal-app/src/main/java/com/squareup/sample/helloterminal/HelloTerminalWorkflow.kt
@@ -37,7 +37,7 @@ class HelloTerminalWorkflow : TerminalWorkflow,
   override fun render(
     renderProps: TerminalProps,
     renderState: State,
-    context: RenderContext
+    context: StatefulWorkflow.RenderContext<TerminalProps, State, ExitCode>
   ): TerminalRendering {
     val (rows, columns) = renderProps.size
     val header = """
--- a/benchmarks/performance-poetry/complex-poetry/src/main/java/com/squareup/benchmarks/performance/complex/poetry/MaybeLoadingGatekeeperWorkflow.kt
+++ b/benchmarks/performance-poetry/complex-poetry/src/main/java/com/squareup/benchmarks/performance/complex/poetry/MaybeLoadingGatekeeperWorkflow.kt
@@ -27,7 +27,7 @@ class MaybeLoadingGatekeeperWorkflow<T : Any>(
   override fun render(
     renderProps: Unit,
     renderState: IsLoading,
-    context: RenderContext
+    context: RenderContext<Unit, IsLoading, Unit>
   ): MayBeLoadingScreen {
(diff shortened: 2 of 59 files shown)


samples/hello-terminal/hello-terminal-app/src/main/java/com/squareup/sample/helloterminal/HelloTerminalWorkflow.kt
```diff
@@ -37,7 +37,7 @@ class HelloTerminalWorkflow : TerminalWorkflow,
   override fun render(
     renderProps: TerminalProps,
     renderState: State,
-    context: RenderContext
+    context: StatefulWorkflow.RenderContext<TerminalProps, State, ExitCode>
```
rjrjr, line 40: Why the StatefulWorkflow prefix in some spots but not others?

steve-the-edwards: I think just sloppiness in my regex replacing 🤷🏻

steve-the-edwards:
Note also that I can cache `RenderContext` much more effectively and cleanly now as well - https://github.com/square/workflow-kotlin/pull/1342

samples/hello-terminal/hello-terminal-app/src/main/java/com/squareup/sample/helloterminal/HelloTerminalWorkflow.kt
steve-the-edwards: I prefer the `StatefulWorkfow.RenderContext` usage FWIW. but in most cases I think its unambiguous, so no need to be too strict.

PR #1343, opened 2025-06-11
Improve workflow visualizer functionality

wenli-cai:
Changed project structure and included new directories. 

**Two notable for added functionality**
1. Instead of "opening" each node to provide more workflow-specific information, a side bar will open instead. This allows for much more information to be presented and will help to avoid cluttering up the main state diagram (but more basic information can still be tacked on later)

2. Since traces will have different traces to see how render passes affected the responsible workflows, a trace selection bar that allows you to jump between different snapshots will help in visualization.

https://github.com/user-attachments/assets/0e8d8674-c438-452a-b741-016342e1d6c7


workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/model/WorkflowNode.kt
```diff
@@ -0,0 +1,14 @@
+package com.squareup.workflow1.traceviewer.model
+
+/**
+ * Since the logic of Workflow is hierarchical (where each workflow may have parent workflows and/or
+ * children workflows, a tree structure is most appropriate for representing the data rather than
+ * using flat data structures like an array.
+ *
+ * TBD what more metadata should be involved with each node, e.g. (props, states, # of render passes)
+ */
+public data class WorkflowNode(
```
rjrjr: 1. Let's avoid reusing the term `WorkflowNode`, that's a very key object in the workflow runtime. Maybe just `Node`? Or `ViewerNode`?

2. Avoid `data class` in library APIs

3. A package with a single member is odd, unless you're very confident that there will be other classes here soon. Perhaps you should promote this class and just call it `Model` or `NodeModel`.

Be interested to hear @steve-the-edwards's take.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/utils/JsonParser.kt
```diff
@@ -0,0 +1,48 @@
+package com.squareup.workflow1.traceviewer.utils
```
rjrjr: `util` rather than `utils` for the package name.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/utils/JsonParser.kt
```diff
@@ -0,0 +1,48 @@
...
+
+    val workflowList = Types.newParameterizedType(List::class.java, WorkflowNode::class.java)
+    val workflowAdapter: JsonAdapter<List<WorkflowNode>> = moshi.adapter(workflowList)
+    val root = workflowAdapter.fromJson(json) ?: emptyList()
+    root
+  } catch (e: JsonDataException) {
+    throw JsonDataException("Failed to parse JSON: ${e.message}", e)
+  } catch (e: IOException) {
+    throw IOException("Malformed JSON: ${e.message}", e)
+  }
```
rjrjr: This wrap and rethrow behavior isn't really providing any value. I think you just want to let them fly and add an `@throws` line or two in the kdoc.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/utils/JsonParser.kt
```diff
@@ -0,0 +1,48 @@
...
+): List<WorkflowNode> {
+  val jsonString = file?.readString()
+  return jsonString?.let { parseTrace(it) } ?: emptyList()
+}
+
+/**
+ * Parses a JSON string into [WorkflowNode] with Moshi adapters
+ *
+ * All the caught exceptions should be handled by the caller, and appropriate UI feedback should be
+ * provided to user
```
rjrjr: Implicit, especially if you use `@throws` here instead

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
```diff
@@ -19,28 +24,40 @@ import io.github.vinceglb.filekit.readString
...
+          onNodeSelect = { selectedNode = it }
+        )
+      }
     }
 
-    UploadFile(onFileSelect = { selectedFile = it })
-  }
-}
+    // Top trace selector row
+    StateSelectTab(
```
wardellbagby: Considering how many traces there are likely to be, you might consider using a dropdown instead.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/StateSelectTab.kt
```diff
@@ -0,0 +1,51 @@
...
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.unit.dp
+import com.squareup.workflow1.traceviewer.model.WorkflowNode
+
+/**
+ * A trace tab selector that allows devs to switch between different states within the provided trace.
+ */
+@Composable
+public fun StateSelectTab(
+  trace: List<WorkflowNode>,
```
wardellbagby: Let's name this `traces` to make it clear that this is a list.

But also, the name `traces` is kinda confusing since this Composable is named `StateSelectTab`. Those should be consistent. I don't think `trace` is super clear, though. I think @rjrjr and @steve-the-edwards might have a better idea there.

Personally, I'm thinking something like `snapshots`? But `snapshot` also has a defined meaning in Workflow land so maybe not. My thought there is that since all a trace is is an export of a snapshot of a Workflow's state tree over time, naming this `snapshots` helps make it clear that this is effectively a list of all of the "root nodes" that the user has loaded. `rootNodes` is also a potential option there.

Regardless, let's name the Composable and this field the same.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/StateSelectTab.kt
```diff
@@ -0,0 +1,51 @@
...
+  currentIndex: Int,
+  onIndexChange: (Int) -> Unit,
+  modifier: Modifier = Modifier
+) {
+  val state = rememberLazyListState()
+
+  Surface(
+    modifier = modifier
+      .padding(4.dp),
+    color = Color.White,
```
wardellbagby: Nit: And this is really a nit but consider using the pre-existing `MaterialTheme` (that might require a new dependency on Material though...) for pulling colors like this. Just so things can be consistent going forward.

https://developer.android.com/develop/ui/compose/designsystems/material3

I'm saying this is totally optional but if you do, do it in a follow-up PR. In a perfect world, we'd be using existing Square design but _someone_ (....me) still hasn't ported Market (our most recent Square design system) to be usable in Compose for Desktop so...

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowInfoPanel.kt
```diff
@@ -0,0 +1,98 @@
...
+    modifier
+      .fillMaxHeight()
+      .background(Color.LightGray)
+  ) {
+    if (node == null) {
+      Text("No node selected")
+      return@Column
+    }
+
+    Column(
```
wardellbagby: A column with a single column child is a bit weird. These can probably be combined.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt
```diff
@@ -18,27 +19,45 @@ import androidx.compose.ui.Alignment
...
-)
+@Composable
+public fun RenderDiagram(
+  file: PlatformFile,
+  traceInd: Int,
+  onFileParse: (List<WorkflowNode>) -> Unit,
+  onNodeSelect: (WorkflowNode) -> Unit,
+) {
+  var workflowNodes by remember { mutableStateOf<List<WorkflowNode>>(emptyList()) }
+  var isLoading by remember { mutableStateOf(true) }
```
wardellbagby: Either put `file` in the key block for `remember` or set `isLoading` to true in your `LaunchedEffect`. Otherwise, it'll still think it's not loading when you start loading a new file.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
```diff
@@ -19,28 +24,40 @@ import io.github.vinceglb.filekit.readString
 public fun App(
   modifier: Modifier = Modifier
 ) {
-  Box {
-    var selectedFile by remember { mutableStateOf<PlatformFile?>(null) }
+  var selectedFile by remember { mutableStateOf<PlatformFile?>(null) }
+  var selectedNode by remember { mutableStateOf<WorkflowNode?>(null) }
+  var workflowTrace by remember { mutableStateOf<List<WorkflowNode>>(emptyList()) }
```
wardellbagby: Kind of the same comment I make below here. `snapshotIndex` implies this is the index of a list of `snapshot`s but it's really the index for `workflowTrace`, so those should be named similarly.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/model/WorkflowNode.kt
wenli-cai: Could you explain (2)? Do you mean replace it with a normal class like a POJO (I guess POKO) ?

For (3), I'm thinking of taking all the mutable states in `App.kt` out and having them in a wrapper class, which would be inside model. Would it be better if it was also by itself, like a `StatesModel`?

rjrjr: I just mean don't use the word `data`. It generates a bunch of code that makes API evolution really difficult / impossible. If you need them go ahead and implement hash / equals manualy.

I don't have the context to answer WRT `App.kt`, and really the issue I raised is a small one. If you think more models are coming might as well keep the package.

wenli-cai: I see, I've never looked deeply into data class, so thanks for pointing that out.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowInfoPanel.kt
wenli-cai: Done

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt
wenli-cai: Good catch, thanks!

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/utils/JsonParser.kt
wenli-cai: Done

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/utils/JsonParser.kt
wenli-cai: Done

workflow-trace-viewer/README.md
```diff
@@ -10,6 +10,14 @@ It can be run via Gradle using:
...
 ```
 
+### Terminology
+
+**Trace**: A trace is a file — made up of frames — that contains the execution history of a Workflow. It includes information about render passes, how states have changed within workflows, and the specific props being passed through. The data collected to generate these should be in chronological order, and allows developers to step through the process easily.
+
+**Frame**: Essentially a "snapshot" of the current "state" of the whole Workflow tree. It contains relevant information about the changes in workflow states and how props are passed throughout.
+
+- Note that "snapshot" and "state" are different from `snapshotState` and `State`, which are idiomatic to the Workflow library.
+
```
wenli-cai, line 20: Set some standard for naming terminology in this project. Open to changing it though!

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/StateSelectTab.kt
wenli-cai: Changed the naming schemes, let me know if this seems better. See:
https://github.com/square/workflow-kotlin/pull/1343/commits/422b94bd53c23a2f223f2231d82b99ce46bc355d#r2141079704

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
wenli-cai: See: 
https://github.com/square/workflow-kotlin/pull/1343/commits/422b94bd53c23a2f223f2231d82b99ce46bc355d#r2141079704

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
```diff
@@ -19,28 +24,43 @@ import io.github.vinceglb.filekit.readString
...
 
-@Composable
-private fun WorkflowContent(file: PlatformFile) {
-  var jsonString by remember { mutableStateOf<String?>(null) }
-  LaunchedEffect(file) {
-    jsonString = file.readString()
-  }
-  val root = jsonString?.let { parseTrace(it) }
+    // Right side information panel
+    InfoPanel(selectedNode)
```
japplin: nit: instead of a comment here rename InfoPanel to RightInfoPanel.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowInfoPanel.kt
```diff
@@ -0,0 +1,98 @@
...
+        .align(Alignment.Top)
+    ) {
+      Icon(
+        imageVector = if (panelOpen) Filled.KeyboardArrowLeft else Filled.KeyboardArrowRight,
+        contentDescription = if (panelOpen) "Close Panel" else "Open Panel",
+        modifier = Modifier
+      )
+    }
+
+    // based on open/close, display the node details (Column)
```
japplin: This code was already really easy to understand so good job! No need for the comment.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowInfoPanel.kt
```diff
@@ -0,0 +1,98 @@
...
+ * It can be toggled open or closed, and resets when the user selects a new file
+ *
+ * @param selectedNode The currently selected workflow node, or null if no node is selected.
+ */
+@Composable
+public fun InfoPanel(
+  selectedNode: Node?,
+  modifier: Modifier = Modifier
+) {
+  // This row is ordered RTL
```
japplin: What is causing this row to be ordered right to left?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowInfoPanel.kt
```diff
@@ -0,0 +1,98 @@
...
+      .background(Color.LightGray)
+      .padding(8.dp),
+    horizontalAlignment = Alignment.CenterHorizontally
+  ) {
+    if (node == null) {
+      Text("No node selected")
+      return@Column
+    }
+
+    Text("only visible with a node selected")
```
japplin, line 86: Did you mean to leave this Text here?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowInfoPanel.kt
```diff
@@ -0,0 +1,98 @@
...
+    }
+  }
+}
+
+/**
+ * The text details of the selected node. This should be closely coupled with the [Node]
+ * data class to see what information should be displayed.
+ */
+@Composable
+private fun PanelDetails(
```
japplin: nit: Instead of the comment perhaps `NodeDetailsPanel`

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt
```diff
@@ -48,40 +68,40 @@ public fun DrawWorkflowTree(
       .fillMaxSize(),
     horizontalAlignment = Alignment.CenterHorizontally,
   ) {
-    // draws itself
-    DrawNode(node)
+    // draws the node itself
```
japplin: nit: Code is already clear no need for this comment.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/UploadFile.kt
```diff
@@ -35,12 +36,14 @@ public fun UploadFile(
...
     ) {
+      onReset()
       onFileSelect(it)
     }
+
     Button(
       onClick = { launcher.launch() },
       modifier = Modifier
-        .align(Alignment.BottomEnd),
+        .align(Alignment.BottomStart),
```
japplin: I noticed that there is a comment in your App method saying that this is a bottom left button, but if you instead passed set the alignment of the button in the modifier you're already passing in then the comment wouldn't be required. I also think it's more typical that the parent would specify the position of a button rather than the button itself specifying where it is on screen.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/UploadFile.kt
```diff
@@ -23,6 +23,7 @@ import io.github.vinceglb.filekit.dialogs.compose.rememberFilePickerLauncher
  */
 @Composable
 public fun UploadFile(
+  onReset: () -> Unit,
```
japplin: Can onReset be an implicit step executed every time onFileSelect is called instead of a separate param?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/JsonParser.kt
```diff
@@ -0,0 +1,42 @@
...
+import com.squareup.workflow1.traceviewer.model.Node
+import io.github.vinceglb.filekit.PlatformFile
+import io.github.vinceglb.filekit.readString
+import java.io.IOException
+
+/**
+ * Parses the data from the given file and initiates the workflow tree
+ */
+public suspend fun fetchTrace(
+  file: PlatformFile?,
```
japplin: Why allow a nullable PlatformFile?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/JsonParser.kt
```diff
@@ -0,0 +1,42 @@
...
+import com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterFactory
+import com.squareup.workflow1.traceviewer.model.Node
+import io.github.vinceglb.filekit.PlatformFile
+import io.github.vinceglb.filekit.readString
+import java.io.IOException
+
+/**
+ * Parses the data from the given file and initiates the workflow tree
+ */
+public suspend fun fetchTrace(
```
japplin: This method isn't doing much, and it's comment is saying that it parses the file which is what the parseTrace method is doing. Also parseTrace throws exceptions which is commented on parseTrace method but not here if they were one function that wouldn't be an issue.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/JsonParser.kt
```diff
@@ -0,0 +1,42 @@
...
+public fun parseTrace(
+  json: String
+): List<Node> {
+  val moshi = Moshi.Builder()
+    .add(KotlinJsonAdapterFactory())
+    .build()
+
+  val workflowList = Types.newParameterizedType(List::class.java, Node::class.java)
+  val workflowAdapter: JsonAdapter<List<Node>> = moshi.adapter(workflowList)
+  val root = workflowAdapter.fromJson(json) ?: emptyList()
```
japplin: Let's return null on an error loading the list and then show an error message to the user that loading the trace was unsuccessful. That would also allow us to handle all of the exceptions here instead of throwing them.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/model/WorkflowNode.kt
```diff
@@ -0,0 +1,14 @@
+package com.squareup.workflow1.traceviewer.model
+
+/**
+ * Since the logic of Workflow is hierarchical (where each workflow may have parent workflows and/or
+ * children workflows, a tree structure is most appropriate for representing the data rather than
+ * using flat data structures like an array.
+ *
+ * TBD what more metadata should be involved with each node, e.g. (props, states, # of render passes)
+ */
+public class Node(
```
wardellbagby: Don't forget to rename this file as well!

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/FrameSelectTab.kt
```diff
@@ -0,0 +1,51 @@
...
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.unit.dp
+import com.squareup.workflow1.traceviewer.model.Node
+
+/**
+ * A trace tab selector that allows devs to switch between different states within the provided trace.
+ */
+@Composable
+public fun StateSelectTab(
```
wardellbagby: The file was renamed but the Composable doesn't match it.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
wenli-cai: Agreed, maybe even a combination of the 2, where with a dropdown you can instantly jump to a frame, and a scrolling bar could let you traverse locally. I think, though, this could be something I implement down the line as iterative improvements.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/JsonParser.kt
wenli-cai: Mistake, fixed now.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/JsonParser.kt
wenli-cai: True, I've combined them now

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowInfoPanel.kt
wenli-cai: I think RTL was the wrong description, I've changed it now to say `This row is aligned to the right of the screen.` 

The full-width spacer pushes everything to the end.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowInfoPanel.kt
wenli-cai: Understood!

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowInfoPanel.kt
wenli-cai: Done

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowInfoPanel.kt
wenli-cai: Yeah just for now, it would be first to go once actual data starts coming in.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/UploadFile.kt
wenli-cai: Done

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/UploadFile.kt
wenli-cai: Makes sense, also a lot more self-explanatory when looking at parent containers. Fixed

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/JsonParser.kt

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt
```diff
@@ -30,22 +31,36 @@ import io.github.vinceglb.filekit.PlatformFile
...
+  var isLoading by remember(traceFile) { mutableStateOf(true) }
+  var error by remember(traceFile) { mutableStateOf<Throwable?>(null) }
 
   LaunchedEffect(traceFile) {
-    isLoading = true
-    nodes = fetchTrace(traceFile)
-    onFileParse(nodes)
+    val parseResult = parseTrace(traceFile)
+
+    if (parseResult is ParseResult.Failure) {
```
wardellbagby: We should use Kotlin's `when` statement here to make the casting automatic!

```kotlin
when(parseResult) {
  is ParseResult.Failure -> { 
    error = parseResult.error
  }
  is ParseResult.Success -> {
    val parsedFrames = parseResult.trace ?: emptyList()
    // etc
  }
}
```

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/model/WorkflowNode.kt
wenli-cai: Done

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/FrameSelectTab.kt
wenli-cai: Done

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/model/WorkflowNode.kt
steve-the-edwards: In general `data class` can be very useful, but when you are a library author and you are producing versions that require binary compatibility its more pain than gain. See https://jakewharton.com/public-api-challenges-in-kotlin/ for more on this.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/StateSelectTab.kt
steve-the-edwards: Trace and Frame are great I think!

PR #1344, opened 2025-06-13
1311: Enable applying Multiple Actions on immediate dispatcher for `CONFLATE_STALE_RENDERINGS`

### LINKED ISSUE #1311: On Android, process all updates from actions from an event before the next frame (Choreographer.doFrame)

steve-the-edwards:
First commits adds a failing Android test to demonstrate the problem reported in #1311 .

Second commit fixes this for `CONFLATE_STALE_RENDERINGS` by `yield()`ing _before_ applying the first action.

The issue with `yield()` in the middle of CSR's inner loop was that it breaks our workflow runtime stabilizing into two main thread messages (`yield()` always posts to the Handler). We want to keep the guarantee of each Workflow update in response to an event as happening within _1_ main thread message.

This solution satisfies that constraint by `yield()`ing _before_ we apply the first action. With these optimizations enabled, when `select` resumes with a deferrable action ready to be processed, we provide it to the runtime loop as a `Deferred` - then we `yield()` _before_ applying it. This allows any other possible collectors to go ahead and queue actions. We can then drain them manually without `select` in the inner loop for conflate (and later DEA).

Note: We add a field for `isDeferrable` on action. Worker actions are by default deferrable. This allows us to opt actions from UI handlers _out_ of this type of optimization.

We also add tests on `Main.immediate` that confirm that non-deferrable actions are handled in one main thread message from *the callback*. Otherwise they are handled in one main thread message *from the action application*.
 
This also adds new methods for fetching those available actions to apply without `select` (using `tryReceive`) to the `WorkflowRunner` and `WorkflowNode` as well as `WorkflowChildNode`.

Fixes #1311

### DIFF
--- a/workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/AndroidRenderWorkflowInTest.kt
+++ b/workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/AndroidRenderWorkflowInTest.kt
@@ -1,35 +1,39 @@
 package com.squareup.workflow1
 
+import android.os.Handler
+import android.os.Looper
+import android.os.Message
 import com.squareup.workflow1.RuntimeConfigOptions.CONFLATE_STALE_RENDERINGS
+import com.squareup.workflow1.RuntimeConfigOptions.PARTIAL_TREE_RENDERING
+import com.squareup.workflow1.RuntimeConfigOptions.RENDER_ONLY_WHEN_STATE_CHANGES
+import com.squareup.workflow1.RuntimeConfigOptions.STABLE_EVENT_HANDLERS
 import com.squareup.workflow1.WorkflowInterceptor.RenderPassesComplete
 import com.squareup.workflow1.WorkflowInterceptor.RuntimeLoopOutcome
 import kotlinx.coroutines.CoroutineStart.UNDISPATCHED
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.ExperimentalCoroutinesApi
-import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.flow.MutableStateFlow
-import kotlinx.coroutines.flow.receiveAsFlow
+import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.flow.drop
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.plus
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.test.UnconfinedTestDispatcher
 import kotlinx.coroutines.test.runTest
-import org.junit.Ignore
 import org.junit.Test
+import java.util.concurrent.CountDownLatch
 import kotlin.test.assertEquals
+import kotlin.test.assertFalse
 import kotlin.test.assertTrue
 
 @OptIn(WorkflowExperimentalRuntime::class, ExperimentalCoroutinesApi::class)
 class AndroidRenderWorkflowInTest {
 
-  @Ignore("#1311: Does not yet work with immediate dispatcher.")
   @Test
-  fun with_conflate_we_conflate_stacked_actions_into_one_rendering() =
+  fun conflate_renderings_for_multiple_worker_actions_same_trigger() =
     runTest(UnconfinedTestDispatcher()) {
 
-      var childHandlerActionExecuted = false
-      val trigger1 = Channel<String>(capacity = 1)
-      val trigger2 = Channel<String>(capacity = 1)
+      val trigger = MutableStateFlow("unchanged state")
       val emitted = mutableListOf<String>()
       var renderingsPassed = 0
       val countInterceptor = object : WorkflowInterceptor {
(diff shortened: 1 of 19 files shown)


workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/AndroidRenderWorkflowInTest.kt
```diff
@@ -0,0 +1,222 @@
...
+              // Update the rendering in order to show conflation.
+              state = "$it+update1"
+            }
+          }
+          runningWorker(
+            worker = trigger.drop(1).asWorker(),
+            key = "Worker3"
+          ) {
+            action("") {
+              // Update the rendering in order to show conflation.
```
rjrjr: Don't you mean "Update the state"?

rjrjr: Wouldn't the test be more rigorous if you accumulated a bunch of deltas to the state rather than clobbering each time? So at the end you could assert that the rendering = `"initial+1+2+3+4"`

workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/AndroidRenderWorkflowInTest.kt
```diff
@@ -0,0 +1,222 @@
...
+        }
+      }
+
+      trigger.value = "changed state 1"
+
+      renderedMutex.lock()
+
+      collectionJob.cancel()
+
+      // 2 renderings (initial and then the update.) Not *5* renderings.
```
rjrjr, line 134: 😗 👌

workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/AndroidRenderWorkflowInTest.kt
```diff
@@ -0,0 +1,222 @@
...
+      assertEquals(
+        2,
+        renderingsPassed,
+        "Expected only 2 renderings passed to interceptor when conflating actions."
+      )
+      assertEquals("changed state 1+update4", emitted.last())
+    }
+
+  @Test
+  fun with_conflate_we_conflate_stacked_actions_from_side_effects_into_one_rendering() =
```
rjrjr: What made you decide both tests were necessary? (Don't get me wrong, more is better as far as I'm concerned, just curious.)

workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/AndroidRenderWorkflowInTest.kt
```diff
@@ -0,0 +1,222 @@
...
+import kotlinx.coroutines.flow.drop
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.plus
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.test.UnconfinedTestDispatcher
+import kotlinx.coroutines.test.runTest
+import org.junit.Test
+import kotlin.test.assertEquals
+
+@OptIn(WorkflowExperimentalRuntime::class, ExperimentalCoroutinesApi::class)
```
rjrjr, line 28: Should you add a test that simulates having an action enqueued via an event handler lambda on a rendering? End to end demonstration of the primary promise we're trying to make to ourselves. Perhaps it should be parameterized to run w/all the config cominbations.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
```diff
@@ -348,6 +368,59 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
     }
   }
 
+  //
+  // /**
```
rjrjr: Commented out b/c not necessary after all? Or is this the thing you have to add to make it actually work?

workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/AndroidRenderWorkflowInTest.kt
steve-the-edwards: Yes, that's a much better test!

workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/AndroidRenderWorkflowInTest.kt
steve-the-edwards: it was early on when i wondered if the Worker machinery was introducing any impedence in resuming to get the actions queued. Its not.

workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/AndroidRenderWorkflowInTest.kt
steve-the-edwards: Yup!

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
steve-the-edwards: Not necessary. I just forgot to remove this.

workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/AndroidRenderWorkflowInTest.kt
steve-the-edwards: Done.

workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/AndroidRenderWorkflowInTest.kt
steve-the-edwards: I added these tests (both for deferrable and non deferrable actions). You and @pyricau can double check my logic for asserting on the main thread message.

steve-the-edwards:
So this adds the ability to set an action as `deferrable` in order to try and keep the `keyUp` in the _same_ main thread for UI interactions.

Ultimately I think we don't want that. we would rather have _everything_ from Workflow's perspective in one main thread message (including teh click action handling). SO I think eventually I'd prefer to make all actions deferrable. but we can go one step at a time?

steve-the-edwards:
closed in favour of https://github.com/square/workflow-kotlin/pull/1353

PR #1345, opened 2025-06-13
Exception reporter disentangling for Android UI code

rjrjr:
Uses `Throwable.withKey` in a few more choke points, to break up crash reporter kitchen sinks. Had to make a few `internal` core methods public to allow their use from the android modules, but I think that's okay -- they're still in a package named `internal`, our intent should be clear.

### DIFF
--- a/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/Throwables.kt
+++ b/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/Throwables.kt
@@ -4,8 +4,35 @@ import kotlin.contracts.ExperimentalContracts
 import kotlin.contracts.contract
 
 /**
+ * Like Kotlin's [requireNotNull], but uses [stackTraceKey] to create a fake top element
+ * on the stack trace, ensuring that a crash reporter's default grouping will create unique
+ * groups for unique keys.
+ *
+ * @see [withKey]
+ *
+ * @throws IllegalArgumentException if the [value] is false.
+ */
+@OptIn(ExperimentalContracts::class)
+inline fun <T : Any> requireNotNullWithKey(
+  value: T?,
+  stackTraceKey: Any,
+  lazyMessage: () -> Any = { "Required value was null." }
+): T {
+  contract {
+    returns() implies (value != null)
+  }
+  if (value == null) {
+    val message = lazyMessage()
+    val exception: Throwable = IllegalArgumentException(message.toString())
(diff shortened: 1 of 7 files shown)


workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/Throwables.kt
```diff
@@ -3,6 +3,33 @@ package com.squareup.workflow1.internal
 import kotlin.contracts.ExperimentalContracts
 import kotlin.contracts.contract
 
+/**
+ * Like Kotlin's [requireNotNull], but uses [stackTraceKey] to create a fake top element
+ * on the stack trace, ensuring that BugSnag's default grouping will create unique
```
steve-the-edwards: nit: You could generalize this to "any crash reporter that uses stacktrace" rather than particular Bugsnag.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/Throwables.kt
```diff
@@ -3,6 +3,33 @@ package com.squareup.workflow1.internal
...
+ * groups for unique keys.
+ *
+ * @see [withKey]
+ *
+ * @throws IllegalArgumentException if the [value] is false.
+ */
+@OptIn(ExperimentalContracts::class)
+inline fun <T : Any> requireNotNullWithKey(
```
steve-the-edwards, line 16: i think you'll have to mark this `public` for explicit API mode, right?

steve-the-edwards (approved):

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/Throwables.kt
rjrjr: Yup. I wonder why the build didn't catch that?

rjrjr: Oooh, I wonder if we turned off that check for `internal` package. If so I think I can undo the other explicit publics I added in here. Trying it out…

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/Throwables.kt
rjrjr: Thanks, accidental copy paste from the internal method.

PR #1347, opened 2025-06-17
1311: For Android use Compose Dispatcher; Move Android RenderWorkflow from workflow-ui to `workflow-runtime-android`

steve-the-edwards:
The `AndroidUiDispatcher.Main` from Compose gives us the dispatching behaviour we want:
1. greedily dispatches all coroutines with a trampoline dispatcher before a frame callback and ensures they are drained.
2. This will also mean we get these updates in before Compose does its frame callback to start teh Recomposer

Also as part of this I moved the `AndroidRenderWorkflow` code from `workflow-ui/android` to `workflow-runtime-android` since we have code there (including tests) that is about workflow-runtime specifications when running on Android (its bigger than just a UI concern).


steve-the-edwards:
Sadly it looks like the tests I wrote to verify this timing are flaky. Going to need to get stable tests for this or we won't be able to include tests for it.

steve-the-edwards:
Broke this up into several better PRs:
1. https://github.com/square/workflow-kotlin/pull/1350
2. https://github.com/square/workflow-kotlin/pull/1352
3. one to follow.

PR #1349, opened 2025-06-18
Updated ComposeScreen idioms in samples.

rjrjr:
Demonstrates and documents the preferred pattern of keeping Composable functions as static as possible, and fixes a few stray doc spots where `WorkflowRendering` was still being called with its `ViewEnvironment` parameter.


vgonda (approved):
Thanks!!

PR #1350, opened 2025-06-19
1311: Dispatcher Tests for RenderWorkflowIn on Android

steve-the-edwards:
A set of tests for using `renderWorkflowIn` with particular dispatchers on Android. These are used in two ways:

1. To ensure we are satisfying the guarantee we want to keep: Updates from the runtime in response to an event are processed before the next Choreographer frame on Android.
2. We are effectively conflating renderings on Android when using a particular dispatcher (e.g., in this case the `Main.immediate` dispatcher). Now, this 2nd category is currently failing! That's part of the fix that needs to happen for #1311 . So we merge these red tests first.

### DIFF
--- a/workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/android/AndroidDispatchersRenderWorkflowInTest.kt
+++ b/workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/android/AndroidDispatchersRenderWorkflowInTest.kt
@@ -1,0 +1,424 @@
+@file:OptIn(WorkflowExperimentalRuntime::class)
+@file:Suppress("JUnitMalformedDeclaration")
+
+package com.squareup.workflow1.android
+
+import app.cash.burst.Burst
+import com.squareup.workflow1.RuntimeConfig
+import com.squareup.workflow1.RuntimeConfigOptions.CONFLATE_STALE_RENDERINGS
+import com.squareup.workflow1.RuntimeConfigOptions.Companion.RENDER_PER_ACTION
+import com.squareup.workflow1.RuntimeConfigOptions.PARTIAL_TREE_RENDERING
+import com.squareup.workflow1.RuntimeConfigOptions.RENDER_ONLY_WHEN_STATE_CHANGES
+import com.squareup.workflow1.RuntimeConfigOptions.STABLE_EVENT_HANDLERS
+import com.squareup.workflow1.Workflow
+import com.squareup.workflow1.WorkflowExperimentalRuntime
+import com.squareup.workflow1.WorkflowInterceptor
+import com.squareup.workflow1.WorkflowInterceptor.RenderPassesComplete
+import com.squareup.workflow1.WorkflowInterceptor.RuntimeLoopOutcome
+import com.squareup.workflow1.action
+import com.squareup.workflow1.asWorker
+import com.squareup.workflow1.renderChild
+import com.squareup.workflow1.runningWorker
+import com.squareup.workflow1.stateful
+import com.squareup.workflow1.ui.renderWorkflowIn
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.flow.drop
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.plus
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.test.runTest
+import org.junit.Before
+import org.junit.Ignore
+import org.junit.Test
+import papa.Choreographers
+import java.util.concurrent.atomic.AtomicInteger
+import kotlin.test.assertEquals
+
+@OptIn(WorkflowExperimentalRuntime::class)
+@Burst
+class AndroidDispatchersRenderWorkflowInTest(
+  private val runtime: RuntimeOptions = RuntimeOptions.DEFAULT
+) {
+
+  @Before
+  fun setup() {
+    resetOrderCounter()
+  }
+
+  @Ignore("See https://github.com/square/workflow-kotlin/issues/1311")
+  @Test
+  fun conflate_renderings_for_multiple_worker_actions_same_trigger() =
+    runTest {
+
+      val trigger = MutableStateFlow("unchanged state")
+      val emitted = mutableListOf<String>()
+      var renderingsPassed = 0
+      val countInterceptor = object : WorkflowInterceptor {
+        override fun onRuntimeLoopTick(outcome: RuntimeLoopOutcome) {
+          if (outcome is RenderPassesComplete<*>) {
+            renderingsPassed++
+          }
+        }
+      }
+
+      val childWorkflow = Workflow.stateful<String, String, String>(
+        initialState = "unchanged state",
+        render = { renderState ->
+          runningWorker(
+            worker = trigger.drop(1).asWorker(),
+            key = "Worker1"
+          ) {
+            action("") {
+              val newState = "$it+u1"
+              state = newState
+              setOutput(newState)
+            }
+          }
+          renderState
+        }
+      )
+      val workflow = Workflow.stateful<String, String, String>(
+        initialState = "unchanged state",
+        render = { renderState ->
+          renderChild(childWorkflow) { childOutput ->
+            action("childHandler") {
+              state = childOutput
+            }
+          }
+          runningWorker(
+            worker = trigger.drop(1).asWorker(),
+            key = "Worker2"
+          ) {
+            action("") {
+              // Update the state in order to show conflation.
+              state = "$state+u2"
+            }
+          }
+          runningWorker(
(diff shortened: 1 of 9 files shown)


workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/android/AndroidDispatchersRenderWorkflowInTest.kt
```diff
@@ -0,0 +1,415 @@
...
+import org.junit.Ignore
+import org.junit.Test
+import papa.Choreographers
+import java.util.concurrent.atomic.AtomicInteger
+import kotlin.test.assertEquals
+
+@OptIn(WorkflowExperimentalRuntime::class)
+@Burst
+class AndroidDispatchersRenderWorkflowInTest(
+  private val runtime: RuntimeOptions = RuntimeOptions.DEFAULT
```
steve-the-edwards, line 44: Note that burst will create a separate test class for each runtime config combination and run them separately. (i'll need to migrate my hand rolled parameterization in the other tests for runtime config to this soon).

workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/android/AndroidDispatchersRenderWorkflowInTest.kt
```diff
@@ -0,0 +1,415 @@
...
+          if (it == "changed state") {
+            // The rendering we were looking for!
+            expectInOrder(1)
+            completedMutex.unlock()
+          } else {
+            expectInOrder(0)
+            Choreographers.postOnFrameRendered {
+              // We are expecting this to happen last, after we get the rendering!
+              expectInOrder(2)
+            }
```
steve-the-edwards, line 308: Right before the event that triggers the change I add a callback for the end of the frame (through papa) that expects to be done at order position 2 (the last position for this test, critically _after_ we've received the updated rendering we want from the change).

@pyricau - does this look like a reasonable way to test this guarantee?

workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/android/AndroidDispatchersRenderWorkflowInTest.kt
```diff
@@ -0,0 +1,415 @@
...
+        if (it.name == "neverends+neverends") {
+          // The rendering we were looking for after the event!
+          expectInOrder(1)
+          completedMutex.unlock()
+        } else {
+          expectInOrder(0)
+          Choreographers.postOnFrameRendered {
+            // This should be happening last!
+            expectInOrder(2)
+          }
```
steve-the-edwards, line 356: Note same thing here, right before the callback event we add the listener to the end of the next frame.

workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/android/AndroidDispatchersRenderWorkflowInTest.kt
```diff
@@ -0,0 +1,415 @@
...
+
+      testScheduler.advanceUntilIdle()
+
+      launch {
+        trigger.value = "state change"
+      }
+
+      testScheduler.advanceUntilIdle()
+
+      renderedMutex.lock()
```
zach-klippenstein, line 154: Nit: Could we use Turbine to write the assertions on the renderings flow?

workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/android/AndroidDispatchersRenderWorkflowInTest.kt
```diff
@@ -0,0 +1,415 @@
...
+    prefix: String = ""
+  ) {
+    val localActual = orderIndex.getAndIncrement()
+    assertEquals(
+      expected,
+      localActual,
+      "$prefix: This should have happened" +
+        " in a different order position:"
+    )
+  }
```
zach-klippenstein: Nit: Shouldn't these be at the top or bottom of the class, not in the middle of test methods?

workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/android/AndroidDispatchersRenderWorkflowInTest.kt
```diff
@@ -0,0 +1,415 @@
...
+          }
+          renderState
+        }
+      )
+
+      val renderings = renderWorkflowIn(
+        workflow = workflow,
+        scope = backgroundScope +
+          Dispatchers.Main.immediate,
+        props = MutableStateFlow(Unit).asStateFlow(),
```
zach-klippenstein, line 291: Oof, I can't believe there's no `stateFlowOf()`.

workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/android/AndroidDispatchersRenderWorkflowInTest.kt
zach-klippenstein: Can we add an expectInOrder to the end of the test, maybe after the runTest block, to ensure that this callback is also invoked?

workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/android/AndroidDispatchersRenderWorkflowInTest.kt
```diff
@@ -0,0 +1,415 @@
...
+    setOf(CONFLATE_STALE_RENDERINGS, RENDER_ONLY_WHEN_STATE_CHANGES, PARTIAL_TREE_RENDERING)
+  ),
+  RENDER_ONLY_CONFLATE_PARTIAL_STABLE(
+    setOf(
+      CONFLATE_STALE_RENDERINGS,
+      RENDER_ONLY_WHEN_STATE_CHANGES,
+      PARTIAL_TREE_RENDERING,
+      STABLE_EVENT_HANDLERS
+    )
+  )
```
zach-klippenstein, line 423: Holy crap, do we really need to test _all_ these combinations? Are we running (or planning to) all of these in production??

zach-klippenstein (approved):

workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/android/AndroidDispatchersRenderWorkflowInTest.kt
```diff
@@ -0,0 +1,415 @@
...
+            }
+          }
+          renderState
+        }
+      )
+      val props = MutableStateFlow(Unit)
+      val renderings = renderWorkflowIn(
+        workflow = workflow,
+        scope = backgroundScope +
+          Dispatchers.Main.immediate,
```
zach-klippenstein: Consider pulling this out into a file constant or something so that when you change it later it's easier to ensure that all usages are changed. And could probably pull some other common values out of the tests too.

workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/android/AndroidDispatchersRenderWorkflowInTest.kt
steve-the-edwards: You mean as a verification that we actually have a choreographer on this thread we are running on? Ya that's probably good.

workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/android/AndroidDispatchersRenderWorkflowInTest.kt
steve-the-edwards: I tried and this version was more readable/intuitive I thought.

workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/android/AndroidDispatchersRenderWorkflowInTest.kt
steve-the-edwards: no, but I wanted not to have restrictions (or unknown bugs) on what combinations could be used.

workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/android/AndroidDispatchersRenderWorkflowInTest.kt
steve-the-edwards: Done.

workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/android/AndroidDispatchersRenderWorkflowInTest.kt
steve-the-edwards: Done. moved to bottom.

workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/android/AndroidDispatchersRenderWorkflowInTest.kt
steve-the-edwards: Done. Added.

PR #1352, opened 2025-06-19
Move non-ui Android RenderWorkflow concerns to workflow-runtime-android

steve-the-edwards:
Now that we have a `workflow-runtime-android` module (and published artifact), it is the place that makes sense to hold the Android versions of `renderWorkflowIn`, as they are beyond solely UI concerns, so do not need to be in `workflow-ui`.

Yes, I recognize that we _could_ distribute this as an Android library _from_ the `workflow-runtime` KMP module with a particular sourceset as well as particular publishing.

I think that would be a good thing to do, but I don't want to figure out how to set that up right now. I think our balance of jvm, ios, and js artifacts from the KMP modules right now makes sense. I do not think we lose anything from this Android library being a standalone module (after all, it was before when it was in `workflow-ui`).

### DIFF
--- a/samples/containers/app-poetry/src/main/java/com/squareup/sample/poetryapp/PoetryActivity.kt
+++ b/samples/containers/app-poetry/src/main/java/com/squareup/sample/poetryapp/PoetryActivity.kt
@@ -13,10 +13,10 @@ import com.squareup.sample.poetry.RealPoemWorkflow
 import com.squareup.sample.poetry.RealPoemsBrowserWorkflow
 import com.squareup.sample.poetry.model.Poem
 import com.squareup.workflow1.WorkflowExperimentalRuntime
+import com.squareup.workflow1.android.renderWorkflowIn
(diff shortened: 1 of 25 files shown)


samples/containers/app-poetry/src/main/java/com/squareup/sample/poetryapp/PoetryActivity.kt
```diff
@@ -42,7 +42,7 @@ class PoetryActivity : AppCompatActivity() {
 
 class PoetryModel(savedState: SavedStateHandle) : ViewModel() {
   val renderings: Flow<Screen> by lazy {
-    renderWorkflowIn(
+    com.squareup.workflow1.android.renderWorkflowIn(
```
zach-klippenstein: Nit: You're already statically importing this, I think the FQN is just an IDE bug? Search-and-replace to fix?

steve-the-edwards: Done.

Issue #1257, opened 2025-01-24
RA-64264 Make WorkflowLayout sample based on the frame clock


rjrjr:
The `renderings: Flow<Screen>` provided to `WorkflowLayout.take()` can fire a lot more often than is useful, especially in large apps that don't use `initialState` / `StateFlow.value` effectively. Can `take` buffer what it receives, keeping only the latest, and take care to call [show] either once or never for frame clock tick?

Remember that it is absolutely fine to drop renderings; each time a new rendering appears it means that the combined workflow state has changed, and the previous ones are now invalid. We're not an animation tool.

zach-klippenstein:
That would be the default behavior if `WorkflowLayout` were compose. I think for a `View` we could do something like this:
```kotlin
fun <T> Flow<T>.sampleAtFrame(choreographer: Choreographer): Flow<T> = flow {
  var frameScheduled = false
  var mostRecentValue: T? = null
  
  coroutineScope {
    collect { value ->
      mostRecentValue = value
      if (!frameScheduled) {
        frameScheduled = true
        choreography.awaitNextFrame {
          // TODO: Probably violates flow's context invariants, maybe should
          //  just be a non-suspending "collectAtFrame" instead.
          launch(start = CoroutineStart.UNDISPATCHED) {
            frameScheduled = false
            emit(mostRecentValue)
          }
        }
      }
    }
  }
}

private suspend fun Choreographer.awaitNextFrame(block: () -> Unit) =
  suspendCancellableCoroutine { continuation ->
    val frameCallback = FrameCallback { block() }
    postFrameCallback(frameCallback)
    continuation.doOnCancellation { removeFrameCallback(frameCallback) }
  }
```

Caveat is I'm not sure how this will sync with Compose's frame callback.

zach-klippenstein:
```kotlin
suspend fun <T> Flow<T>.collectAtFrame(choreographer: Choreographer, collector: () -> Unit) {
  var frameScheduled = false
  var mostRecentValue: T? = null

  collect { value ->
    mostRecentValue = value
    if (!frameScheduled) {
      frameScheduled = true
      choreography.awaitNextFrame {
        frameScheduled = false
        // Clear mostRecentValue before transferring control to caller.
        val value = mostRecentValue
        mostRecentValue = null
        collector(value)
      }
    }
  }
}
```

zach-klippenstein:
Actually I think we should probably just make `show` debounce itself until frame boundary, not put the logic in `take`.

[...]

PR #1353, opened 2025-06-19
1311:Use Compose's AndroidUiDispatcher.Main to conflate actions in runtime.

### LINKED ISSUE #1311: On Android, process all updates from actions from an event before the next frame (Choreographer.doFrame)

steve-the-edwards:
Note this also includes some adjustments to the runtime loop. After the first suspension to wait and apply an action, with `CONFLATE_STALE_RENDERINGS` we loop without suspending - applying any actions that are already queued up on the action queues.

If no dispatcher is specified on Android, use Compose's `AndroidUiDispatcher.Main` for the runtime (requiring Compose).

As it turns out, Compose's `AndroidUiDispatcher.Main` and its greedy behaviour to drain all dispatches _before_ the next Android Choreographer frame gives us exactly the performant behaviour we want.

In fact, when dispatching with `AndroidUiDispatcher.Main` all the coroutines that would be resumed to apply an action will be greedily run before the next Choreographer frame. This ticks the "full" runtime loop for each one, so the inner loop for `CONFLATE_STALE_RENDERINGS` is not really necessary. The greediness means that the updates all happen before the rendering can be consumed by view code (as well as all happen before the next frame). Because this is not an _immediate_ dispatcher, when we pass a new rendering to the stateflow to be consumed by the view code, it is *not* immediately picked up (it has to be dispatched, even though the `AndroidUiDispatcher.Main` is going to do that greedily *and* before the next frame). This changes ordering.

We keep the inner loop for `CONFLATE_STALE_RENDERINGS` however, as it is still useful for other dispatchers.

Still the code used to drain immediately available actions can be used by `DRAIN_EXCLUSIVE_ACTIONS` to avoid render *passes*.

Fixes #1311 

This should also take care of #1257 if we start using the Compose AndroidUiDispatcher.


zach-klippenstein:
> In fact, CONFLATE_STALE_RENDERINGS is basically a no-op when using the AndroidUiDispatcher.Main to dispatch the runtime and consume the renderings. 
> …
> On Android context where we use this dispatcher, we can likely just not use CONFLATE_STALE_RENDERINGS at all. however, for other dispatchers, it could still have an effect.

Is it _basically_ a no-op, or _actually_ a no-op? Because if it's not identical, we shouldn't treat it as optional. We already have so many confusing workflow flags, if we start requiring various configurations with caveats about things like dispatchers, the messiness just explodes.

workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/android/AndroidDispatchersRenderWorkflowInTest.kt
```diff
@@ -118,58 +138,69 @@ class AndroidDispatchersRenderWorkflowInTest(
       val renderings = renderWorkflowIn(
         workflow = workflow,
         scope = backgroundScope +
-          Dispatchers.Main.immediate,
+          AndroidUiDispatcher.Main,
```
zach-klippenstein: If this is the only dispatcher we're testing, is it now a hard requirement that apps using workflow UI on Android _must_ use Compose? If so we should probably communicate that very clearly and update the docs too. If not, then can we test both?

workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/android/AndroidDispatchersRenderWorkflowInTest.kt
```diff
@@ -118,58 +138,69 @@ class AndroidDispatchersRenderWorkflowInTest(
...
         props = props,
-        runtimeConfig = setOf(CONFLATE_STALE_RENDERINGS),
+        runtimeConfig = runtime.runtimeConfig,
```
zach-klippenstein, line 125: Ah cool, so this change is saying that the behavior this tests validates now happens for all configs, as long as we're using that dispatcher?

workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/android/AndroidDispatchersRenderWorkflowInTest.kt
zach-klippenstein: And furthermore, does this imply the behavior broke for other dispatchers? If not, can we just add a parameter to the parameterization that specifies the dispatcher to use, so we can continue to validate it?

workflow-runtime-android/src/main/java/com/squareup/workflow1/android/AndroidRenderWorkflow.kt
```diff
@@ -272,9 +277,17 @@ public fun <PropsT, OutputT, RenderingT> renderWorkflowIn(
   onOutput: suspend (OutputT) -> Unit = {}
 ): StateFlow<RenderingT> {
   val restoredSnap = savedStateHandle?.get<PickledTreesnapshot>(KEY)?.snapshot
+
+  // Add in Compose's AndroidUiDispatcher.Main by default if none is specified.
```
zach-klippenstein, line 293: Please add this new magical behavior to the kdoc as well.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
```diff
@@ -201,16 +201,14 @@ public fun <PropsT, OutputT, RenderingT> renderWorkflowIn(
...
 
           // If no actions processed, then no new rendering needed. Pass on to UI.
           if (actionResult == ActionsExhausted) break@conflate
 
           // Skip rendering if we had unchanged state, keep draining actions.
-          if (shouldShortCircuitForUnchangedState(
-              actionResult = actionResult,
-            )
-          ) {
+          if (shouldShortCircuitForUnchangedState(actionResult = actionResult)) {
```
zach-klippenstein: Nit: Do we really need this parameter to be named here? When the value passed in has the same name, it doesn't really add any readability.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
```diff
@@ -207,24 +206,35 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
...
+   * children's.
+   *
+   * @return [ActionProcessingResult] of the action processed, or [ActionsExhausted] if there were
+   * none immediately available.
+   */
+  fun applyNextAvailableAction(): ActionProcessingResult {
+    val result = subtreeManager.applyNextAvailableChildAction()
+
+    if (result == ActionsExhausted) {
+      return eventActionsChannel.tryReceive().getOrNull()?.let { action ->
```
zach-klippenstein, line 240: Please confirm if i understand this right: If we hit this branch, it means either no children had actions to run, _or_ some child had an action but it didn't produce an output?

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
```diff
@@ -207,24 +206,35 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
    * a re-render, e.g. my state changed or a child state changed.
    *
    * It is an error to call this method after calling [cancel].
-   *
-   * @return [Boolean] whether or not the queues were empty for this node and its children at the
-   *    time of suspending.
    */
-  @OptIn(ExperimentalCoroutinesApi::class, DelicateCoroutinesApi::class)
-  fun onNextAction(selector: SelectBuilder<ActionProcessingResult>): Boolean {
+  fun onNextAction(selector: SelectBuilder<ActionProcessingResult>) {
```
zach-klippenstein: Nit: Might be time to rename this, to distinguish better from `applyNextAvailableAction`. It's a registration function, so maybe something like `registerSubtreeActionSelectors`? The "on…" naming never really made sense anyway since this is never called in the standard DSL style.

Can be done in a separate PR.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
```diff
@@ -207,24 +206,35 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
...
         return@onReceive applyAction(action)
       }
     }
-    return empty
+  }
+
+  /**
+   * Will try to apply any immediately available actions in this action queue or any of our
+   * children's.
+   *
```
zach-klippenstein, line 228:
```suggestion
   *
   * Contrast this to [onNextAction], which is used to await the next action when none is available right now.
   * Both call [applyAction] when one is available.
```

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
```diff
@@ -207,24 +206,35 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
    * a re-render, e.g. my state changed or a child state changed.
    *
    * It is an error to call this method after calling [cancel].
```
zach-klippenstein, line 207:
```suggestion
   * It is an error to call this method after calling [cancel].
   *
   * Contrast this to [applyNextAvailableAction], which is used to check for an action
   * that is already available without waiting, and then _immediately_ apply it.
   * Both call [applyAction] when one is available.
```

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowRunner.kt
```diff
@@ -83,24 +80,25 @@ internal class WorkflowRunner<PropsT, OutputT, RenderingT>(
    * and resume (breaking ties with order of declaration). Guarantees only continuing on the winning
    * coroutine and no others.
    */
-  @OptIn(WorkflowExperimentalRuntime::class)
-  suspend fun processAction(waitForAnAction: Boolean = true): ActionProcessingResult {
+  suspend fun waitAndProcessAction(): ActionProcessingResult {
```
zach-klippenstein: Nit: Since you're renaming this anyway, "await" is the more idiomatic term for "listen while suspended". In contrast, "wait" canonically implies blocking a thread.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowRunner.kt
```diff
@@ -83,24 +80,25 @@ internal class WorkflowRunner<PropsT, OutputT, RenderingT>(
...
-      val empty = rootNode.onNextAction(this)
-      if (!waitForAnAction && runtimeConfig.contains(CONFLATE_STALE_RENDERINGS) && empty) {
-        // With CONFLATE_STALE_RENDERINGS if there are no queued actions and we are not
-        // waiting for one, then return ActionsExhausted and pass the rendering on.
-        onTimeout(timeMillis = 0) {
-          // This will select synchronously since time is 0.
-          ActionsExhausted
-        }
-      }
+      rootNode.onNextAction(this)
```
zach-klippenstein: (Follow-up to nit): Renaming this method would also make this code pretty much self-documenting and the above line comment redundant.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowRunner.kt
```diff
@@ -83,24 +80,25 @@ internal class WorkflowRunner<PropsT, OutputT, RenderingT>(
...
   }
 
+  /**
+   * Will try to apply any immediately available actions for this runtime (no suspending).
+   *
+   * @return [ActionProcessingResult] of the action processed, or [ActionsExhausted] if there were
+   * none immediately available.
+   */
+  fun applyNextAvailableAction(): ActionProcessingResult {
+    return rootNode.applyNextAvailableAction()
```
zach-klippenstein: Nit: Maybe name this `applyNextAvailableSubtreeAction` to make it more clear when reading callsites that this actually walks the tree? It took me a few read-throughs of the subtree/WN code to remember these were actually depth-first visitors methods (this and `onNextAction`).

zach-klippenstein (commented):
Ok so after reading, I'm still not sure I fully understand what the significant behavior difference is between using Compose's main dispatcher and Main.immediate. 

_Numbering these assertions/questions so you can respond more easily._

1. When the conflation flag is _off_ and multiple actions are available immediately – then with Compose's dispatcher they _may_ result in multiple emissions to the UI layer with redundant updates, but all will still happen before the frame is drawn. `Main.immediate` will process them all immediately but also emit to the UI immediately, so effectively the same as far as the UI is concerned. But that was existing behavior.
1. When no actions are available, but another action is sent before the next Choreographer frame, then the action is still be guaranteed to be processed before the frame is drawn (either by `Main.immediate`'s immediacy or by compose's pre-frame drain). Also existing behavior.
1. When the flag is _on_, then we avoid suspending when multiple actions are available, which is theoretically a performance win internally since not suspending is faster than suspending, even with `Main.immediate`?
    1. Externally, the behavior is the same (only emit last conflated rendering to UI) even with `Main.immediate` because we already weren't emitting on the renderings flow until after the (previously suspending) conflation loop?
1. If all these are true, then this change only affects behavior for multiple immediate actions when the conflation flag is off? Otherwise the we don't suspend in the conflation loop, so the choice of dispatcher doesn't matter.

workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/android/AndroidDispatchersRenderWorkflowInTest.kt
steve-the-edwards: The general logic against both `StandardTestDispatcher` and `UnconfinedTestDispatcher` is tested in the `workflow-runtime` module tests themselves. This is all just for Android. Those tests are all still there. This is testing optimizations with the Android dispatchers.

workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/android/AndroidDispatchersRenderWorkflowInTest.kt
steve-the-edwards: yes.

workflow-runtime-android/src/main/java/com/squareup/workflow1/android/AndroidRenderWorkflow.kt
steve-the-edwards: Will do!

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
steve-the-edwards: no, if a child had an action it would return `ActionApplied` even if there was no `Output`. This is only if there were no child actions.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowRunner.kt
steve-the-edwards: great point.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
steve-the-edwards:
```
   * Contrast this to [registerTreeActionSelectors] which will add select clauses that will await
   * the next action. That will also end up with [applyAction] being called when the clauses is
   * selected.
```

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
steve-the-edwards:
```
   * Contrast this to [applyNextAvailableAction], which is used to check for an action
   * that is already available without waiting, and then _immediately_ apply it.
   * The select clauses added here also call [applyAction] when one of them is selected.
   */
  fun registerTreeActionSelectors(selector: SelectBuilder<ActionProcessingResult>) {
```

steve-the-edwards:
> In fact, when dispatching with `AndroidUiDispatcher.Main` all the coroutines that would be resumed to apply an action will be greedily run before the next Choreographer frame. This ticks the "full" runtime loop for each one, so the inner loop for `CONFLATE_STALE_RENDERINGS` is not really necessary.

I updated the PR description:

> In fact, when dispatching with AndroidUiDispatcher.Main all the coroutines that would be resumed to apply an action will be greedily run before the next Choreographer frame. This ticks the "full" runtime loop for each one, so the inner loop for CONFLATE_STALE_RENDERINGS is not really necessary.

workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/android/AndroidDispatchersRenderWorkflowInTest.kt
```diff
@@ -124,58 +120,73 @@ class AndroidDispatchersRenderWorkflowInTest(
...
+      assertEquals(
+        expected = 2,
+        actual = renderingsConsumed.size,
+        message = "Expected 2 consumed renderings."
+      )
+      // There are 2 attempts to produce a rendering for Conflate (initial and then the update.)
+      // And otherwise there are *5* attempts to produce a new rendering.
+      val expected = if (runtime.runtimeConfig.contains(CONFLATE_STALE_RENDERINGS)) 2 else 5
+      assertEquals(
+        expected = if (runtime.runtimeConfig.contains(CONFLATE_STALE_RENDERINGS)) 2 else 5,
```
rjrjr:
```suggestion
        expected = expected,
```

workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/android/AndroidDispatchersRenderWorkflowInTest.kt
```diff
@@ -124,58 +120,73 @@ class AndroidDispatchersRenderWorkflowInTest(
...
+        actual = renderingsProduced,
+        message = "Expected $expected renderings to be produced (passed signal to interceptor)."
```
rjrjr:
```suggestion
        message = "Expected $expected renderings to be produced with config ${runtime.runtimeConfig} (passed signal to interceptor)."
```

workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/android/AndroidDispatchersRenderWorkflowInTest.kt
```diff
@@ -225,50 +236,67 @@ class AndroidDispatchersRenderWorkflowInTest(
...
+      // There are 2 attempts to produce a rendering for Conflate (initial and then the update.)
+      // And otherwise there are *3* attempts to produce a new rendering.
+      val expected = if (runtime.runtimeConfig.contains(CONFLATE_STALE_RENDERINGS)) 2 else 3
       assertEquals(
-        2,
-        renderingsPassed,
-        "Expected only 2 renderings passed to interceptor when conflating actions."
+        expected = if (runtime.runtimeConfig.contains(CONFLATE_STALE_RENDERINGS)) 2 else 3,
+        actual = renderingsProduced,
+        message = "Expected $expected renderings to be produced (passed signal to interceptor)."
```
rjrjr: Same changes as above

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
```diff
@@ -178,19 +178,19 @@ public fun <PropsT, OutputT, RenderingT> renderWorkflowIn(
...
 
       if (shouldShortCircuitForUnchangedState(actionResult)) {
         chainedInterceptor.onRuntimeLoopTick(RenderPassSkipped())
         sendOutput(actionResult, onOutput)
         continue@outer
       }
 
-      // After resuming from runner.processAction() our coroutine could now be cancelled, check so
-      // we don't surprise anyone with an unexpected rendering pass. Show's over, go home.
+      // After resuming from runner.waitAndProcessAction() our coroutine could now be cancelled,
```
rjrjr: I don't think there is any such method. Did you mean `awaitAndApplyAction()`?

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
```diff
@@ -222,9 +220,7 @@ public fun <PropsT, OutputT, RenderingT> renderWorkflowIn(
             continue@outer
           }
 
-          // Make sure the runtime has not been cancelled from runner.processAction()
-          if (!isActive) return@launch
-
+          // Render pass for the updated state from the action applied.
```
rjrjr, line 223: Seems like you're confident that `applyNextAvailabletreeAction()` cannot make us inactive, how come?

rjrjr (approved):

workflow-runtime-android/src/androidTest/java/com/squareup/workflow1/android/AndroidDispatchersRenderWorkflowInTest.kt
steve-the-edwards: Thanks. made all these changes manually

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
steve-the-edwards: because it does not suspend so there is no way for the scope to be cancelled.

PR #1354, opened 2025-06-20
Migrate Parametrized Tests to Burst

steve-the-edwards:
Start using Burst anywhere we parametrize tests. This will help prep for adding new runtime configs in one place.


workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
```diff
@@ -30,498 +32,407 @@ import kotlinx.coroutines.test.advanceUntilIdle
...
-      runtimes.asSequence(),
-      tracerOptions.asSequence(),
-      dispatcherOptions.asSequence()
-    )
+  private val dispatcherUsed =
+    if (useUnconfined) UnconfinedTestDispatcher() else myStandardTestDispatcher
 
-  private val runtimeTestRunner =
-    ParameterizedTestRunner<Triple<RuntimeConfig, WorkflowTracer?, TestDispatcher>>()
+  private fun advanceIfStandard(dispatcher: TestDispatcher) {
```
zach-klippenstein: It looks like the only thing you ever pass here is `dispatcherUsed`, so don't need the parameter. I believe the only result of passing in a different value would be that it would make the test incorrect, so why allow the foot gun?

PR #1355, opened 2025-06-20
DRAIN_EXCLUSIVE_ACTIONS implementation

steve-the-edwards:
Add an option to get the next action that is mutually exclusive to the dirty nodes. Then we can apply as many of those actions as possible.

Add tests for this as well and ensure existing tests can run with this runtime config on.

### DIFF
--- a/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt
+++ b/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt
@@ -6,6 +6,7 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Job
 import kotlin.coroutines.CoroutineContext
 import kotlin.coroutines.coroutineContext
+import kotlin.jvm.JvmInline
 import kotlin.reflect.KType
 
 /**
@@ -34,9 +35,7 @@ import kotlin.reflect.KType
  *
  *  1. [onSessionStarted] - called when a new [WorkflowSession] is created the first time a
  *     workflow is rendered with the [CoroutineScope] for that session.
- *  1. [onRuntimeLoopTick] - Called to report the [RuntimeLoopOutcome] of each tick of the runtime
- *     loop. In the simplest case this is the application of one action and one render pass, but
- *     optimizations can change that. See [onRenderingUpdated] for more.
+ *  1. [onRuntimeUpdate] - Called to report [RuntimeUpdate]s from the event loop.
  *
  * ## On Profiling
  *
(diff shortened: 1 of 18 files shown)


workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt
```diff
@@ -140,43 +138,56 @@ public interface WorkflowInterceptor {
...
+   * @param update A [RuntimeUpdate] from the event loop.
    */
-  public fun onRuntimeLoopTick(
-    outcome: RuntimeLoopOutcome
+  public fun onRuntimeUpdate(
+    update: RuntimeUpdate
   ): Unit = Unit
 
-  public sealed interface RuntimeLoopOutcome
+  public sealed interface RuntimeUpdate
```
zach-klippenstein, line 169: At the risk of over-optimizing, this could be a value class to avoid allocating `RenderingProduced` objects every pass even when they're ignored.

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
```diff
@@ -1739,6 +1747,298 @@ class RenderWorkflowInTest(
...
+        assertEquals(2, childActionAppliedCount, "Expected each child action applied.")
+        assertEquals(1, parentRenderCount, "Expected parent only rendered once.")
+        assertEquals(1, receivedOutputs.size, "Expected one output.")
+        assertEquals("changed state", receivedOutputs[0])
+      }
+    }
+  }
+
+  @Test
+  fun `for_drain_exclusive_and_render_only_when_state_changes_we_handle_multiple_actions_in_one_render_but_we_do_pass_rendering_if_state_changed_earlier`() {
```
zach-klippenstein: Nit: You shouldn't need the backticks here, you're not using spaces or other illegal characters.

steve-the-edwards: It's actually because this is KMP code that runs on js as well, IIRC. For some reason the kotlin-js compiler can't handle long names that don't have ticks? Or maybe that's why I had to add the `_`. Now I can't remember. But it was something with kotlin-js.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt
steve-the-edwards: nice. will take a look at doing this.

steve-the-edwards:
> I don't understand the interaction between conflation and drainExclusive. It looks like we run the drainExclusive loop first, then render, then conflate, drainExclusive, and render, etc. Wouldn't we want to continuously interleave drainExclusive and conflation passes? Or is that what we're doing and I just can't read code?

the main loop (after the first synchronous render) is:
1. `awaitAndApplyAction()` for the first available action
2. DEA loop: Drain any exclusive actions immediately available - synchronously for any actions already queued (no suspend here).
3. render()
4. CSR loop: Drain any other actions immediately available; `render()` after each one.
5. pass updated `rendering` down the StateFlow.

No interleaving is necessary from what I can tell. The exclusive actions for DEA are just a subset of the actions that can be applied without rendering at all.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt
steve-the-edwards: I'd have to talk more with you about what you are thinking. I am not sure we even need to report the rendering itself to the interceptor. I can start by just removing that?

steve-the-edwards: Well, that object is already allocated (probably why I was not worried originally). So we could start by _just_ making `RenderingProduced` a `value class` I think I was confused originally because this comment was on the `RuntimeUpdate` sealed interface

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
steve-the-edwards: Turns out I was just remembering wrong or had copied this forward when not needed. Removed now!

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt
steve-the-edwards: I made `RenderingProduced` a `value class` now. The others are just objects, so this should all be quite fast / have minimal allocations.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
```diff
@@ -281,7 +296,7 @@ internal class WorkflowNode<PropsT, StateT, OutputT, RenderingT>(
 
     if (!runtimeConfig.contains(PARTIAL_TREE_RENDERING) ||
       !lastRendering.isInitialized ||
-      subtreeStateDidChange
+      subtreeStateDirty
```
mannydelgado, line 299: Should we be checking the `selfStateDirty` here as well?
is there in which the self state might be dirty but the subtree isnt?

steve-the-edwards: no. if self is dirty, subtree is also dirty. Not true the other way around though, subtree may be dirty without the self being dirty.

rjrjr:
> > I don't understand the interaction between conflation and drainExclusive. It looks like we run the drainExclusive loop first, then render, then conflate, drainExclusive, and render, etc. Wouldn't we want to continuously interleave drainExclusive and conflation passes? Or is that what we're doing and I just can't read code?
> 
> the main loop (after the first synchronous render) is:
> 
> 1. `awaitAndApplyAction()` for the first available action
> 2. DEA loop: Drain any exclusive actions immediately available - synchronously for any actions already queued (no suspend here).
> 3. render()
> 4. CSR loop: Drain any other actions immediately available; `render()` after each one.
> 5. pass updated `rendering` down the StateFlow.
> 
> No interleaving is necessary from what I can tell. The exclusive actions for DEA are just a subset of the actions that can be applied without rendering at all.

It'd be good to capture this in a comment in the code.

rjrjr (approved):

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt
zach-klippenstein: There's no point in making a subclass of this a value class, since it will be boxed when referred to as a parent type. We'd need to replace the entire sealed class/interface hierarchy with a single value class.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
zach-klippenstein: Subtree includes self, basically.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt
steve-the-edwards: Oh, interesting!

PR #1356, opened 2025-06-24
Refactor WorkflowNode into AbstractWorkflowNode <- StatefulWorkflowNode. [CLF-436]

zach-klippenstein:
This is to prepare for go/compose-based-workflows.

This is two commits to make git diff recognize the extracted abstract class as the new file instead of the `WorkflowNode -> StatefulWorkflowNode` rename. However, github doesn't get this and shows `StatefulWorkflowNode` as a net new file, so it's must easier to review if you only look at the first commit.

### DIFF
--- a/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/AbstractWorkflowNode.kt
+++ b/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/AbstractWorkflowNode.kt
@@ -1,0 +1,121 @@
+package com.squareup.workflow1.internal
+
+import com.squareup.workflow1.ActionApplied
+import com.squareup.workflow1.ActionProcessingResult
+import com.squareup.workflow1.ActionsExhausted
+import com.squareup.workflow1.NoopWorkflowInterceptor
+import com.squareup.workflow1.RuntimeConfig
+import com.squareup.workflow1.RuntimeConfigOptions
+import com.squareup.workflow1.TreeSnapshot
+import com.squareup.workflow1.Workflow
+import com.squareup.workflow1.WorkflowInterceptor
+import com.squareup.workflow1.WorkflowInterceptor.WorkflowSession
+import com.squareup.workflow1.WorkflowTracer
+import kotlin.coroutines.CoroutineContext
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.CoroutineName
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.selects.SelectBuilder
+
+internal fun <PropsT, OutputT, RenderingT> createWorkflowNode(
+  id: WorkflowNodeId,
+  workflow: Workflow<PropsT, OutputT, RenderingT>,
+  initialProps: PropsT,
+  snapshot: TreeSnapshot?,
+  baseContext: CoroutineContext,
+  // Providing default value so we don't need to specify in test.
+  runtimeConfig: RuntimeConfig = RuntimeConfigOptions.DEFAULT_CONFIG,
+  workflowTracer: WorkflowTracer? = null,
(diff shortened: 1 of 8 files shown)


zach-klippenstein:
- [ ] Check that merging these 2 commits will still cause git to show file history for `WorkflowNode -> StatefulWorkflowNode`. If not, then split into 2 separate PRs.

tcmulcahy (approved):

handstandsam (approved):

steve-the-edwards (approved):

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/AbstractWorkflowNode.kt
```diff
@@ -0,0 +1,121 @@
...
+  protected val emitAppliedActionToParent: (ActionApplied<OutputT>) -> ActionProcessingResult,
+  baseContext: CoroutineContext,
+) {
+
+  /**
+   * Scope that has a job that will live as long as this node and be cancelled when [cancel] is
+   * called. Also adds a debug name to this coroutine based on its ID.
+   */
+  val scope: CoroutineScope =
+    CoroutineScope(baseContext + Job(parent = baseContext[Job]) + CoroutineName(id.toString()))
```
steve-the-edwards: what's the purpose of holding this field as a `scope`? Typically leads to breaking structured concurrency

steve-the-edwards: i get we were storing the CoroutineContext before - including the job- but that would be without the privileged API access to launchign that `scope` provides.

zach-klippenstein: I don't remember why I made it a scope, I initially wrote this code 6 months ago. But I don't understand your concern, how does this break structured concurrency?

steve-the-edwards: saving the *CoroutineScope* as a field allows us to launch coroutines from arbitrary non-suspending functions on this class. that then means that we can easily break structured concurrency by having coroutines scoped beyond the current call stack lifecycle - the lifecycle control of the coroutines is no longer tied to the structure of the program (i.e. call stack) but rather how we use the right cleanup ceremony on the owning class (this node).

Now, we could say that's fine since this object (the node) *is* the lifecycle object so we already need to follow the right ceremony for it. however, there is a change here where `StatefulWorkflowNode` *itself* used to *be* the `CoroutineScope` - at which point there is less concern of footgunning the coroutine lifecycle since its tied to the class instance itself (i forgot this the first time i reviewed the code).

So why can't we just make the abstract node be a `CoroutineScope` itself again? that way we the structure of the concurrency lifecycles is 'clearer' again. Does that make sense?

zach-klippenstein: > there is a change here where StatefulWorkflowNode itself used to be the CoroutineScope - at which point there is less concern of footgunning the coroutine lifecycle since its tied to the class instance itself (i forgot this the first time i reviewed the code)

I don't understand this argument. The node itself being the scope can just as easily allow any code that has access to that node launching coroutines. In fact, we can actually make this field protected, which _actually_ makes what you're concerned about less likely.

- [x] Make `WorkflowNode.scope` protected.

By the way, the `CoroutineScope` [kdoc](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-scope/) itself advises against implementing `CoroutineScope` in your own types:
> ## Obtaining a coroutine scope
> Manual implementations of this interface are not recommended. Instead, a [CoroutineScope](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-scope/index.html) should be obtained in a way that reflects the intended structured concurrency relationships.

PR #1363, opened 2025-07-08
Workflow visualizer prototype

wenli-cai:
First prototype of creating a workflow visualizer.

Changes were made to how a given JSON file is parsed due to a difference in how logs were being stored. 

The Node model also needed to be changed to match any additional information being passed in

- Since each `WorkflowInterceptor.WorkflowSession.sessionId` is unique, and the fields of the same Workflow Nodes can differ, we only use sessionId as a form of comparison.

### DIFF
--- a/workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/JsonParser.kt
+++ b/workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/JsonParser.kt
@@ -27,15 +27,15 @@ public suspend fun parseTrace(
   return try {
     val jsonString = file.readString()
     val workflowAdapter = createMoshiAdapter()
-    val unParsedTrace = workflowAdapter.fromJson(jsonString)
+    val parsedRenderPasses = workflowAdapter.fromJson(jsonString)
 
-    val parsedTrace = mutableListOf<Node>()
-    unParsedTrace?.forEach { renderPass ->
+    val parsedFrames = mutableListOf<Node>()
+    parsedRenderPasses?.forEach { renderPass ->
       val parsed = getFrameFromRenderPass(renderPass)
-      parsedTrace.add(parsed)
+      parsedFrames.add(parsed)
     }
 
-    ParseResult.Success(parsedTrace)
+    ParseResult.Success(parsedFrames)
   } catch (e: Exception) {
     ParseResult.Failure(e)
   }
--- a/workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/JsonParser.kt
+++ b/workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/JsonParser.kt
@@ -42,8 +42,10 @@ private fun createMoshiAdapter(): JsonAdapter<List<List<Node>>> {
   val moshi = Moshi.Builder()
     .add(KotlinJsonAdapterFactory())
     .build()
-  val workflowList = Types.newParameterizedType(List::class.java,
-    Types.newParameterizedType(List::class.java, Node::class.java))
+  val workflowList = Types.newParameterizedType(
+    List::class.java,
+    Types.newParameterizedType(List::class.java, Node::class.java)
+  )
   val adapter: JsonAdapter<List<List<Node>>> = moshi.adapter(workflowList)
   return adapter
 }
--- a/workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/JsonParser.kt
+++ b/workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/JsonParser.kt
@@ -9,7 +9,8 @@ import io.github.vinceglb.filekit.PlatformFile
 import io.github.vinceglb.filekit.readString
 
 /**
- * Parses a given file's JSON String into [Node] with Moshi adapters.
+ * Parses a given file's JSON String into a list of [Node]s with Moshi adapters. Each of these nodes
+ * count as the root of a tree which forms a Frame.
  *
  * @return A [ParseResult] representing result of parsing, either an error related to the
  * format of the JSON, or a success and a parsed trace.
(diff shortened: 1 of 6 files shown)


workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/JsonParser.kt
```diff
@@ -19,18 +20,60 @@ public suspend fun parseTrace(
...
     val jsonString = file.readString()
-    val moshi = Moshi.Builder()
-      .add(KotlinJsonAdapterFactory())
-      .build()
-    val workflowList = Types.newParameterizedType(List::class.java, Node::class.java)
-    val workflowAdapter: JsonAdapter<List<Node>> = moshi.adapter(workflowList)
-    val trace = workflowAdapter.fromJson(jsonString)
-    ParseResult.Success(trace)
+    val workflowAdapter = createMoshiAdapter()
+    val unParsedTrace = workflowAdapter.fromJson(jsonString)
```
steve-the-edwards: nit: this is really `jsonFrames` or something like that right? In other words, the 'list' part of it has been parsed out, just not the tree of each frame.

steve-the-edwards: In fact, i think this has actually done all the parsing into `Node`? It's just we have not built it into a tree. So these need renaming.

really this is the `parsedTrace`, and below `parsedTrace` should be `frames` (the frames as trees?)

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/JsonParser.kt
```diff
@@ -19,18 +20,60 @@ public suspend fun parseTrace(
...
+  val workflowList = Types.newParameterizedType(
+    List::class.java,
+    Types.newParameterizedType(List::class.java, Node::class.java)
+  )
+  val adapter: JsonAdapter<List<List<Node>>> = moshi.adapter(workflowList)
+  return adapter
+}
+
+/**
+ * We take an unparsed render pass and build up a tree structure from it to form a Frame.
```
steve-the-edwards, line 60:
```suggestion
 * We take an unparsed render pass and build up a tree structure from it to form a Frame.
 
 * @return Node the root node of the tree for that frame.
```

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/JsonParser.kt
```diff
@@ -19,18 +20,60 @@ public suspend fun parseTrace(
...
+ */
+private fun getFrameFromRenderPass(renderPass: List<Node>): Node {
+  val childrenByParent = renderPass.groupBy { it.parent }
+  val root = childrenByParent["root"]?.single()
```
steve-the-edwards: let's extract this to a constant so its clear (you can create a `companion object` with a `const val ROOT_ID: String = "root"` or something like that.)

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/JsonParser.kt
```diff
@@ -19,18 +20,60 @@ public suspend fun parseTrace(
...
+  )
+  val adapter: JsonAdapter<List<List<Node>>> = moshi.adapter(workflowList)
+  return adapter
+}
+
+/**
+ * We take an unparsed render pass and build up a tree structure from it to form a Frame.
+ */
+private fun getFrameFromRenderPass(renderPass: List<Node>): Node {
+  val childrenByParent = renderPass.groupBy { it.parent }
```
steve-the-edwards: we are using ids for the parent not names right? Otherwise this won't be unique.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/JsonParser.kt
```diff
@@ -19,18 +20,60 @@ public suspend fun parseTrace(
...
+  val root = childrenByParent["root"]?.single()
+  return buildTree(root!!, childrenByParent)
+}
+
+/**
+ * Recursively builds a tree using each node's children.
+ */
+private fun buildTree(node: Node, childrenByParent: Map<String, List<Node>>): Node {
+  val children = (childrenByParent[node.name] ?: emptyList())
```
steve-the-edwards: i think we need to use `node.id` as the parent identifier, otherwise it won't be unique!

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/JsonParser.kt
wenli-cai: Done

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/JsonParser.kt
wenli-cai: It's directly referencing the parent Node, which are compared by id's, so yes. Unless you are asking about something else?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/JsonParser.kt

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/JsonParser.kt
wenli-cai: this is a good point, but on another branch, I have 3 different structures going on:
`affectedNodes: List<Set<Node>>`: This contains a set for each render pass, so I can easily access and decide which nodes need to be highlighted within the larger tree. If there is too much I'm passing around between functions, I can easily remove this and use a recursive algo to find the nodes I need to highlight.
`parsedTrace: List<Node>`: This would be a formed tree that only contains information of the render pass
`frameTree: List<Node>` (this is poorly named): This would be a continuously growing tree that contains every workflow that was ever started

I would agree that from that everything should be `parsedSomething`, though. Maybe like `parsedRenderPasses` for the moshi call, and then `parsedFramesSet`, `parsedFrames`,  `mainTree` for the 3 data structures? Or maybe I need to change the naming scheme...

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/JsonParser.kt

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/JsonParser.kt

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/JsonParser.kt
```diff
@@ -8,8 +8,15 @@ import com.squareup.workflow1.traceviewer.model.Node
 import io.github.vinceglb.filekit.PlatformFile
 import io.github.vinceglb.filekit.readString
 
+/*
+ The root workflow Node uses an ID of 0, and since we are filtering childrenByParent by the
+ parentId, the root node has a parent of -1 ID. This is reflected seen inside android-register
+ */
+const val ROOT_ID: String = "-1"
```
steve-the-edwards, line 15: My guess is when the parent is null in the `WorkflowSession` you have just given it a default value of -1 in your recording code?

steve-the-edwards (approved):

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/JsonParser.kt
wenli-cai: Yep! I will make note of this on both PR's in workflow library and register

PR #1364, opened 2025-07-09
Introduce WorkStealingDispatcher.

zach-klippenstein:


workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcher.kt
```diff
@@ -0,0 +1,317 @@
...
+    context: CoroutineContext,
+    block: Runnable
+  ) {
+    var continuation: DispatchContinuation? = null
+    lock.withLock {
+      val queue = this.queue
+        ?: ArrayList<Runnable>(initialCapacity = INITIAL_QUEUE_CAPACITY)
+          .also { this.queue = it }
+      queue += block
+      println("OMG dispatch: queue size = ${queue.size}")
```
zach-klippenstein: - [x] Remove logs once the PR is green. Leave them in until then to help debugging.

zach-klippenstein:
Tests in this PR and the runtime one (https://github.com/square/workflow-kotlin/pull/1365) are green, so removed the logs, and this is fully ready to review (and hopefully merge).

build.gradle.kts
```diff
@@ -101,6 +102,15 @@ subprojects {
...
 }
 
+subprojects {
+  tasks.withType(Test::class.java)
+    .configureEach {
+      testLogging {
+        exceptionFormat = FULL
+      }
+    }
+}
```
zach-klippenstein: This makes it a little easier to debug tests from CI logs, but I don't care if this goes in with this PR.

workflow-runtime/build.gradle.kts
```diff
@@ -17,6 +17,8 @@ kotlin {
   if (targets == "kmp" || targets == "js") {
     js(IR) { browser() }
   }
+
+  compilerOptions.freeCompilerArgs.add("-Xexpect-actual-classes")
```
zach-klippenstein, line 23: Without this we get an error about the `internal expect actual class Lock`. It doesn't affect public API, so we don't need to worry about transitive compatibility.

build.gradle.kts
rjrjr: Hard to see a downside. Could use a comment explaining the benefit.

workflow-runtime/build.gradle.kts
rjrjr, line 23: Comment saying why?

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcher.kt
```diff
@@ -0,0 +1,306 @@
...
+  }
+
+  /**
+   * Executes queued work items until there are none left, then returns.
+   *
+   * @param onStartLocked Called while [lock] is held exactly 1 time before any tasks are executed.
+   * @param onFinishedLocked Called while [lock] is held exactly 1 time after all tasks are finished
+   * executing.
+   */
+  private inline fun advanceUntilIdle(
```
rjrjr: Nit: I find that overloads like this make for confusing return trips down the road, how about `doAdvanceUntilIdle()`?

rjrjr (approved):
LGTM but I don't think my review is worth much. Maybe we can get @jingibus to take a look?

build.gradle.kts
steve-the-edwards: nice! I like it.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcher.kt
```diff
@@ -0,0 +1,306 @@
...
+
+/**
+ * A [CoroutineDispatcher] that delegates to another dispatcher but allows stealing any work
+ * scheduled on this dispatcher and performing it synchronously by calling [advanceUntilIdle].
+ *
+ * The easiest way to create one is by calling [wrapDispatcherFrom].
+ *
+ * E.g.
+ * ```
+ * val dispatcher = WorkStealingDispatcher.wrappingDispatcherFrom(scope.coroutineContext)
```
steve-the-edwards:
```suggestion
 * val dispatcher = WorkStealingDispatcher.wrapDispatcherFrom(scope.coroutineContext)
```

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcher.kt
```diff
@@ -0,0 +1,306 @@
...
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.Runnable
+import kotlin.concurrent.Volatile
+import kotlin.coroutines.Continuation
+import kotlin.coroutines.ContinuationInterceptor
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.resume
+
+private typealias RunQueue = ArrayList<Runnable>
```
steve-the-edwards: consider `ArrayDeque` since we only drain from the ends? I only thought about that because that's what `AndroidUiDispatcher` in Compose uses for the trampoline dispatch queue.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcher.kt
```diff
@@ -0,0 +1,306 @@
...
+     * is done, so we don't need to guard this property with a lock.
+     */
+    private var intercepted: Continuation<Unit>? = null
+
+    /**
+     * Resumes this continuation on [delegateInterceptor] by intercepting it and resuming the
+     * intercepted continuation. When a dispatcher returns false from [isDispatchNeeded], then when
+     * continuations intercepted by it are resumed, they may either be ran in-place or scheduled to
+     * a special thread-local queue. The only way to access this queue is to have the dispatcher
+     * intercept a continuation and resume the intercepted continuation.
```
steve-the-edwards: Are you talking about the event loop here? Might as well mention it.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcher.kt
```diff
@@ -0,0 +1,306 @@
...
+ * dispatcher.advanceUntilIdle()
+ * ```
+ *
+ * @param delegateInterceptor The [CoroutineDispatcher] or other [ContinuationInterceptor] to
+ * delegate scheduling behavior to. This can either be a confined or unconfined dispatcher, and its
+ * behavior will be preserved transparently.
+ */
+internal class WorkStealingDispatcher(
+  private val delegateInterceptor: ContinuationInterceptor
+) : CoroutineDispatcher() {
```
steve-the-edwards, line 45: Do we want to consider implementing `Delay` and opting in to the `InternalCoroutinesApi` usage?

I'm concerned that if we do not forward `scheduleResumeAfterDelay` and `invokeOnTimeout` to the `delegateInterceptor` when _it_ implements `delay` then we could end up with some unexpected behaviour!

I guess Compose's `AndroidUiDispatcher` does not concern itself with this becuase it is fine to just let the `DefaultDelay` (which is `Dispatchers.Main`) handle delays for it? I'm not sure we have that luxury here since this is not always wrapping the 'main' dispatcher. I guess we could make it clear that this is only intended to be used to wrap the Main dispatcher?

Does compose mess with the main dispatcher loader at all? i can''t find anything about that.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcher.kt
```diff
@@ -0,0 +1,306 @@
...
+  private fun RunQueue.runAll() {
+    forEach {
+      it.run()
+    }
+  }
+
+  /**
+   * A reusable continuation that is used to access the coroutine runtime's resumption behavior for
+   * both confined and unconfined dispatchers. See [resumeOnDelegateDispatcher] for more information
+   * on how this works.
```
steve-the-edwards:
```suggestion
   * A reusable continuation that is used to access the delegate dispatcher's resumption behavior. This works for
   * both confined and unconfined dispatchers. See [resumeOnDelegateDispatcher] for more information
   * on how this works.
```

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcher.kt
```diff
@@ -0,0 +1,306 @@
...
+   *
+   * [WorkStealingDispatcher] guarantees that only one instance of this class will be created per
+   * dispatcher, and that it will never be resumed more than once concurrently, so it's safe to
+   * reuse.
+   */
+  private inner class DispatchContinuation : Continuation<Unit> {
```
steve-the-edwards:
```suggestion
  private inner class DelegateDispatchContinuation : Continuation<Unit> {
```

Could we call this the `DelegateDispatchContinuation`? It helps remind me that this is here to access the delegate dispatcher's behaviour.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcher.kt
```diff
@@ -0,0 +1,306 @@
...
+      "limitedParallelism is not supported for WorkStealingDispatcher"
+    )
+  }
+
+  /**
+   * "Steals" work that was scheduled on this dispatcher but hasn't had a chance to run yet and runs
+   * it, until there is no work left to do. If the work schedules more work, that will also be ran
+   * before the method returns.
+   *
+   * Some care must be taken with this method: if it is called from a coroutine that is dispatched
```
steve-the-edwards: "Some care"! ha. That's an understatement!

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcher.kt
```diff
@@ -0,0 +1,306 @@
...
+        // handle them.
+        // Note that we could "cancel" the dispatch if this is true here, since we're stealing all
+        // its work, but we can't cancel that task so it will just have to noop.
+        wasDispatchScheduled = dispatchScheduled.also {
+          if (!it) dispatchScheduled = true
+        }
+      },
+      onFinishedLocked = {
+        // If we set this flag above, then clear it now so future dispatch calls schedule normally.
+        dispatchScheduled = wasDispatchScheduled
```
steve-the-edwards: why would we ever need to set it back to true? haven't we just drained the queue?

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcher.kt
```diff
@@ -0,0 +1,306 @@
...
+    return when {
+      queue == null -> {
+        // The next dispatch would allocate a new queue, so recycle one if possible.
+        this.queue = queueToRecycle?.apply { clear() }
+        null
+      }
+
+      queue.isEmpty() -> {
+        // There's nothing to process in an empty queue, so don't return it at all. And since the
+        // next dispatch call already has a queue to use, so just let the recycled one be GC'd and
```
steve-the-edwards:
```suggestion
        // next dispatch call already has a queue to use, just let the recycled one be GC'd and
```

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcher.kt
```diff
@@ -0,0 +1,306 @@
...
+          this.queue = queueToRecycle?.apply { clear() }
+        }
+      }
+    }
+  }
+
+  private fun RunQueue.runAll() {
+    forEach {
+      it.run()
+    }
```
steve-the-edwards: I don't think `forEach` guarantees ordering, does it? My frame of reference for contrast is the trampolining queue in Compose's `AndroidUiDispatcher`. Having the ordering behaviour enforced (FIFO) is helpful because it means that our actions will get queued up before we start handling them.

I think I see how we can get around not locking getting the next item off the queue, because we are using an isolated queue for each draining, and the rest of the tasks build up on a new queue. So that part makes sense.

How do you think this performs compared to using the locked `nextTask()` method of Compose's `AndroidUiDispatcher`?

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcher.kt
```diff
@@ -0,0 +1,306 @@
...
+      // If delegate is a CoroutineDispatcher, intercepted will be a special Continuation that will
+      // check the delegate's isDispatchNeeded to decide whether to call dispatch() or to enqueue it
+      // to the thread-local unconfined queue.
+      intercepted.resume(Unit)
+    }
+
+    /**
+     * DO NOT CALL DIRECTLY! Call [resumeOnDelegateDispatcher] instead.
+     */
+    override fun resumeWith(result: Result<Unit>) {
```
steve-the-edwards, line 230: Ok, so the delegateDispatcher dispatches our special continuation and when its complete it calls `resumeWith` here.

Why do we drain everything else immediately? Like i'm not understanding why we call `advanceUntilIdle` below?

After a single dispatch to the delegate we always drain all of this?

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcherTest.kt
```diff
@@ -0,0 +1,439 @@
...
+    launch(dispatcher) {
+      expect(1)
+      yield()
+      expect(3)
+    }
+
+    testDispatcher.scheduler.advanceUntilIdle()
+    expect(4)
+  }
+
```
steve-the-edwards, line 661: No `delay()` tests here. But again, what do we want do with that? if we're going to ignore, then we need to make that clear.

build.gradle.kts
zach-klippenstein: Tony said it can get too noisy. If there's a lot of failures, maybe but it's not really noise? Worst case we can always remove it.

- [x] Comment

workflow-runtime/build.gradle.kts
zach-klippenstein: - [x] Comment

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcher.kt
zach-klippenstein: ~It's also what `Dispachers.Unconfined` uses, however the only mutations we actually do on this are append and clear. Does `ArrayDeque` actually have any benefit there? If we were messing around with the beginning of the list (like `Unconfined` does), then it clearly would.

I'd actually used Compose's `MutableVector` initially, which isn't available in this module (yet). Then I switched to androidx.collections' `MutableObjectList` which is effectively the same thing, but requires a new dependency that probably isn't worth it if the built-in types are good enough for the built-in dispatchers.~

I guess your point is that there's no reason to do the list swapping at all. Just use `ArrayDeque` and keep a single queue and process one at a time, like the other dispatchers.

I figured this would be better since it doesn't need to lock after running every task, but it still has to lock when enqueuing every task, and a little extra locking might be better than a few extra allocations. For comparison, `Unconfined` doesn't need locking at all since its event loop is always thread local. `AndroidUiDispatcher` does locking, and it just locks every iteration while draining the queue, so that's probably fine. Probably better to do the same as the other dispatcher though, and will be simpler code for us.

- [ ] Use `ArrayDeque`

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcher.kt
zach-klippenstein: I think it might be necessary for tests, since they might have special `delay` behavior. But I would rather not use internal APIs without a real solid reason so I want to validate this first. And I guess if we're running a headless runtime on `Dispatchers.Default` we probably shouldn't waste the main thread for delays. 

Compose doesn't touch the main dispatcher loader AFAIK, it explicitly references the compose dispatcher when creating recomposers for android. It _does_ implement `Delay` for some internal test `ContinuationInterceptors` though.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcher.kt
zach-klippenstein: Well, it's technically fine to do that. It will guarantee that the tasks are processed asap, and no tasks will be lost or anything. They just might not be processed by the end of the call on a particular thread. For our uses, we're only ever calling it from synchronous code so we don't need to worry about adding the extra overhead of making this method fully synchronized as well.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcher.kt
zach-klippenstein: > haven't we just drained the queue?

Yes, and our scheduled drain operation will indeed be a noop if nothing is dispatched in the meantime.

> why would we ever need to set it back to true?

Because if something _is_ dispatched between the end of this operation and when our scheduled drain runs, we don't want to schedule _another_ drain.

I think what you're wondering is why we don't cancel the scheduled drain here instead. The reason is: we can't. The dispatcher API only allows you to schedule `Runnable`s to run, not cancel them. I think the reason for _that_ is that even a cancelled continuation needs to be resumed when cancelled to run any `finally` blocks.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcher.kt
zach-klippenstein: Perhaps it should be named closer to how it works, not what it does? `runAllEnqueuedTasks`?

- [ ] rename private `advanceUntilIdle`

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcher.kt
zach-klippenstein: `forEach` guarantees ordering for the "batch" of tasks, but ordering of batches depends on the underlying dispatcher and how `advanceUntilIdle` is called. It's not guaranteed if the underlying dispatcher is not single-threaded and/or `advanceUntilIdle` is called from multiple threads. E.g. Dispatchers aren't required to guarantee FIFO ordering. `Dispatchers.Default` can't, for example, since it has no control over thread scheduling. The current impl of WSD doesn't provide any _additional_ guarantees on top of whatever the underlying dispatcher does. It weakens ordering guarantees if `advanceUntilIdle` is called from other threads, but as mentioned above this isn't something we do, and would just add extra overhead for no reason.

Discussed performance considerations above. tl;dr: The trade-off is that when there's contention, we can process a whole batch without locking after every task, but might end up allocating a few extra queue list objects. But `AndroidUiDispatcher` locks on every task, so that's probably fine.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcher.kt
zach-klippenstein: I said "coroutine runtime" since the dispatcher itself doesn't have any control over what happens when `isDispatchNeeded` is false, that's built-in to the coroutine runtime itself. Really it's the combination of the runtime and the dispatcher. I'll try to figure out a more thorough wording.

- [ ] Reword

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcher.kt
zach-klippenstein: > Ok, so the delegateDispatcher dispatches our special continuation and when its complete it calls resumeWith here.

Not necessarily. I'm basically just re-typing my code comment here, but if `delegateInterceptor` is a `CoroutineDispatcher`, then `interceptContinuation` will return a `DispatchedContinuation`. That continuation's `resumeWith` will check the dispatcher's `isDispatchNeeded` and either ask the dispatcher to dispatch the call to the original continuation's `resumeWith`, call the original `resumeWith` directly, or dispatch it to the thread local event loop.

Here's some snippets from the coroutines source for the convenience of other reviewers: `DispatchedContinuation.resumeWith`:
```kotlin
    override fun resumeWith(result: Result<T>) {
        val context = continuation.context
        val state = result.toState()
        if (dispatcher.isDispatchNeeded(context)) {
            _state = state
            resumeMode = MODE_ATOMIC
            dispatcher.dispatch(context, this)
        } else {
            executeUnconfined(state, MODE_ATOMIC) {
                withCoroutineContext(this.context, countOrElement) {
                    continuation.resumeWith(result)
                }
            }
        }
    }
```

`DispatchedContinuation` implements `Runnable` so can be passed to `dispatch` directly. It's `run` method looks like:
```kotlin
    final override fun run() {
             …
                    if (exception != null) {
                        continuation.resumeWithException(exception)
                    } else {
                        continuation.resume(getSuccessfulResult(state))
                    }
              …
    }
```

zach-klippenstein: > Why do we drain everything else immediately? Like i'm not understanding why we call advanceUntilIdle below?
>
> After a single dispatch to the delegate we always drain all of this?

The job of this continuation is to run any tasks dispatched to us in the appropriate context. `resumeWith` is our callback into that appropriate context, and once we're there we can drain the queue, which is what `advanceUntilIdle` does. If the public `advanceUntilIdle` is never called, then this is the only way any tasks will ever get ran.

I don't think I understand your question though.

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcherTest.kt
zach-klippenstein: It would be good to have integration tests for `delay` regardless.

- [x] Add tests for `delay`

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcher.kt
steve-the-edwards: Ah thank-you. Your snippets and walk through was helpful for clearing up my understanding here. I missed the step of exactly how the `DispatchedContinuation` that gets dispatched by whatever we are wrapping will always end up calling back here after it is dispatched by the delegate dispatcher.

I guess what I am surprised by is why we don't run just _one_ task off the queue (then call `resumeWithDelegateDispatcher` again) instead of drain them all?

Like in this case, we are delegating to the dispatcher just to decide when to _drain_ the whole queue, rather than for each runnable.

My mental model was originally that _each_ runnable would get dispatched by the delegate dispatcher, _unless_ we call our special `drainTheQueue()` (nee `advanceUntilIdle()`) which will run them all synchronously.

zach-klippenstein: [Talked](https://square.slack.com/archives/CE470RL5T/p1752177424735509?thread_ts=1752107831.230289&cid=CE470RL5T) a bunch with @steve-the-edwards offline, and am now convinced the current behavior is wrong. This impl serializes parallelizable dispatchers (like `Dispatchers.Default`), which it should not do. Stephen's mental model is correct, so I'll rewrite this to do that.

- [x] Support parallel dispatchers.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcher.kt
zach-klippenstein: Added delay tests and support for delay. It's pretty gross though, we have to access `DefaultDelay` with some hacks since it's internal. It's generating a compiler warning that I'm not sure how to suppress.

See https://github.com/Kotlin/kotlinx.coroutines/issues/3758.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcher.kt
zach-klippenstein: Obsolete since rewriting.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcher.kt
zach-klippenstein: Done.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcher.kt
zach-klippenstein: Obsolete since rewriting.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcher.kt
zach-klippenstein: Instead I went with a slightly more complex solution where we subclass WSD with one that implements `Delay`, and then choose which to instantiate based on whether the delegate implements `Delay` or not. This is better since it doesn't "fake" delay support if the delegate doesn't.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcher.kt
zach-klippenstein: Obsolete

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcher.kt
zach-klippenstein: Updated this comment to reflect the new behavior, and added another test to validate.

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcherTest.kt
```diff
@@ -0,0 +1,940 @@
...
+    expect(0)
+    dispatcher.dispatch {
+      expect(2)
+    }
+    expect(1)
+
+    dispatcher.advanceUntilIdle()
+    expect(3)
+
+    baseDispatcher.blocks.single().run()
```
steve-the-edwards, line 460: We're just testing that this does not throw right? That is a little opaque here, a comment may be helpful here and below.

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcherTest.kt
```diff
@@ -0,0 +1,940 @@
...
+    val baseDispatcher = RecordingDispatcher()
+    val dispatcher = WorkStealingDispatcher(baseDispatcher)
+
+    expect(0)
+    dispatcher.dispatch {
+      expect(2)
+    }
+    expect(1)
+
+    baseDispatcher.blocks.single().run()
```
steve-the-edwards, line 474: Could you make this an extension method or helper method on `RecordingDispatcher`, like

`baseDispatcher.runDispatchedTask()` or something like that?  That would _really_ help make what's going on with these calls more clear i think!

workflow-runtime/src/jvmTest/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcherStressTest.kt
```diff
@@ -0,0 +1,265 @@
...
+   */
+  @Suppress("CheckResult")
+  @Test fun stressTestDispatchingFromMultipleThreadsWithAdvanceUntilIdle() {
+    // Use a test dispatcher so we can pause time.
+    val baseDispatcher = StandardTestDispatcher()
+    val dispatcher = WorkStealingDispatcher(baseDispatcher)
+    val scope = CoroutineScope(dispatcher)
+
+    val numThreads = Runtime.getRuntime().availableProcessors().coerceAtLeast(4)
+    // TODO handle when numThreads is odd.
```
steve-the-edwards: Issue for this TODO or its not a big deal?

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcherTest.kt
zach-klippenstein: I mean, most tests implicitly check that their code doesn't throw, that's basically the definition of a test.

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcherTest.kt
zach-klippenstein: How so? This code makes it clear that the test expects only a single task to have been enqeued, and then runs it. A method that communicates that would need to be named more like "assertSingleTaskAndRun" which is just as verbose.

workflow-runtime/src/jvmTest/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcherStressTest.kt
zach-klippenstein: Addressed.

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcherTest.kt
steve-the-edwards: Ok, I can see your point. My point was that I have to go look at the definition of `RecordingDispatcher` to know that `blocks` is the list of dispatched `Runnable`s. Anyway, its nitty and subjective, so fine with either way.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcher.kt
```diff
@@ -0,0 +1,273 @@
...
+    /**
+     * Flag used to avoid checking the queue for the task when this continuation is executed by the
+     * delegate dispatcher after it's already been ran by advancing. This is best-effort – if
+     * there's a race, the losing thread will still lock and check the queue before nooping.
+     *
+     * Access to this property does not need to be synchronized with [lock] or by any other method,
+     * since it's just a write-once hint.
+     */
+    @Volatile
+    private var consumed = false
```
steve-the-edwards, line 197: Nice!

steve-the-edwards (approved):

PR #1365, opened 2025-07-09
Use WorkStealingDispatcher in runtime, behind a flag.

zach-klippenstein:
Installs the `WorkStealingDispatcher` introduced in https://github.com/square/workflow-kotlin/pull/1364 into the workflow runtime behind a new runtime flag.


rjrjr:
So if this works, use it instead of adding the new Compose dependency?

zach-klippenstein:
> So if this works, use it instead of adding the new Compose dependency?

I would still like to get away from Unconfined if there's an opportunity to do so, but if this lets us move faster without shaving that yak, then awesome.

zach-klippenstein:
Changed to drafts while I write tests.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/RuntimeConfig.kt
```diff
@@ -161,6 +169,8 @@ public enum class RuntimeConfigOptions {
       //     DRAIN_EXCLUSIVE_ACTIONS,
       //   )
       // ),
+
+      ALL(RuntimeConfigOptions.entries.toSet())
```
zach-klippenstein: It seems to me like it would be a good idea to have a config that turns on all the flags (for testing, but also for green-field uses of the runtime), but if you'd rather I add explicit configs for this new flag I can do that too.

workflow-runtime/src/commonTest/kotlin/com/squareup/workflow1/RenderWorkflowInTest.kt
```diff
@@ -1494,7 +1501,9 @@ class RenderWorkflowInTest(
 
   @Test
   fun for_conflate_we_do_not_conflate_stacked_actions_into_one_rendering_if_output() {
-    if (runtimeConfig.contains(CONFLATE_STALE_RENDERINGS)) {
+    if (CONFLATE_STALE_RENDERINGS in runtimeConfig &&
+      WORK_STEALING_DISPATCHER !in runtimeConfig
```
zach-klippenstein, line 1513: This tested behavior intentionally changes with WSD on.

zach-klippenstein:
I added a single test for the new ordering behavior. I don't know if we need more than that since `WorkStealingDispatcher` itself is tested extensively in isolation. Happy to add more tests if anyone can think up any other interesting edge cases though.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/RenderWorkflow.kt
```diff
@@ -202,6 +213,13 @@ public fun <PropsT, OutputT, RenderingT> renderWorkflowIn(
           conflationHasChangedState = conflationHasChangedState || actionResult.stateChanged
           // We may have more actions we can process, this rendering could be stale.
           // This will check for any actions that are immediately available and apply them.
+          // We advance the dispatcher first to allow any coroutines that were launched by the last
+          // render pass to start up and potentially enqueue actions.
+          dispatcher?.let {
+            workflowTracer.trace("AdvancingWorkflowDispatcher") {
```
zach-klippenstein: @steve-the-edwards I just added this. I figure since this is intended to improve performance, might as well add a trace section to measure the new work preemptively.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/RuntimeConfig.kt
zach-klippenstein: Added a comment indicating the difference in intention between this and other enum entries that might happen to also contain all entries when they're introduced.

zach-klippenstein:
Ownership ended with @zach-klippenstein, @steve-the-edwards is this PR's owner now.

.github/workflows/kotlin.yml
```diff
@@ -239,7 +239,7 @@ jobs:
         uses: mikepenz/action-junit-report@5f47764eec0e1c1f19f40c8e60a5ba47e47015c5 # v4
         if: always() # always run even if the previous step fails
         with:
-          report_paths: '**/build/test-results/test/TEST-*.xml'
+          report_paths: '**/build/test-results/jvmTest/TEST-*.xml'
```
steve-the-edwards, line 242: This was just a bug with the reporting paths that I found while looking at this.

.github/workflows/kotlin.yml
```diff
@@ -470,7 +470,28 @@ jobs:
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - name: Checkout
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+      - name: Check with Gradle
+        uses: ./.github/actions/gradle-task
+        with:
+          task: jvmTest --continue -Pworkflow.runtime=all
```
steve-the-edwards, line 486: I don't test the work stealing dispatcher combinations separately, just adding the 'all' config here for the jvm tests.

.github/workflows/kotlin.yml
```diff
@@ -578,7 +599,7 @@ jobs:
         ### <start-connected-check-shards>
         shardNum: [ 1, 2, 3 ]
         ### <end-connected-check-shards>
-        runtime: [ conflate, stateChange, drainExclusive, conflate-stateChange, partial, conflate-partial, stable, conflate-drainExclusive, stateChange-drainExclusive, partial-drainExclusive, conflate-partial-drainExclusive ]
+        runtime: [ conflate, stateChange, drainExclusive, conflate-stateChange, partial, conflate-partial, stable, conflate-drainExclusive, stateChange-drainExclusive, partial-drainExclusive, conflate-partial-drainExclusive, all ]
```
steve-the-edwards, line 602: Add 'all' the runtime test shard.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/RuntimeConfig.kt
```diff
@@ -56,6 +56,11 @@ public enum class RuntimeConfigOptions {
 
   /**
    * If we have more actions to process, do so before passing the rendering to the UI layer.
+   *
+   * Note that this is *not effective* when using an immediate or unconfined dispatcher as the
+   * handling of an action happens synchronously after the coroutine sending it into the sink
+   * is resumed. This means it never dispatches multiple coroutines that send actions into the sink
+   * before handling the first, so there is never multiple actions to apply!
```
steve-the-edwards, line 63: Added notes here about how CSR and DEA won't work with an immediate or unconfined dispatcher.

.github/workflows/kotlin.yml
rjrjr: What is `report_paths`?

.github/workflows/kotlin.yml
rjrjr: Which should be fine b/c we plan to ship that way?

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/RuntimeConfig.kt
rjrjr: But `WORK_STEALING_DISPATCHER` fixes that, right? If so should say so. If not, what's our actual plan? Seems like we should suggest to the reader whatever it is that we're planning to do in POS.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/RuntimeConfig.kt
```diff
@@ -98,69 +115,349 @@ public enum class RuntimeConfigOptions {
     enum class RuntimeOptions(
       val runtimeConfig: RuntimeConfig
     ) {
-      DEFAULT(RENDER_PER_ACTION),
+      NONE(RENDER_PER_ACTION),
```
rjrjr, line 132: Holy combinatorial explosion.

rjrjr (approved):

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/RuntimeConfig.kt
steve-the-edwards: No, it does not actually fix that. We thought it might until we realized that immediate dispatchers prevent the _dispatch_ of the coroutines for the other actions from happening, so there is nothing to drain after processing just the first action.

The way to fix that would be to use a non immediate dispatcher - e.g. `Dispatchers.Main` - along with a frame clock listener that explicitly calls `advanceUntilIdle()` for each frame so that we can maintain the guarantee of stably updating the runtime each frame.

I implemented that and tested it here with these tests, but in discussion with @zach-klippenstein we decided that it was not helpful.
1. If you are using Compose, then you should use `AndroidUiDispatcher.Main` for this, otherwise the timing of the two frame callbacks might not be right.
2. If you are not using Compose, we don't really know this and we are OK with just not being able to use CSR or DEA.

.github/workflows/kotlin.yml
steve-the-edwards: 'eventually', yes

.github/workflows/kotlin.yml
steve-the-edwards: what files it tries to pick up test results from to show in the PR directly. (this was just not working before).

.github/workflows/kotlin.yml
steve-the-edwards: but good point that we should make sure each combination we roll out with has been tested here.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/RuntimeConfig.kt
steve-the-edwards: I updated it to say this:

```
   * Note further that [WORK_STEALING_DISPATCHER] does not fix this for immediate dispatchers. In
   * order to use this optimization, a non-immediate dispatcher must be used. If using a non-
   * immediate dispatcher, we recommend that you ensure that the Workflow runtime completes all
   * known updates before the next 'frame.' This can be done using a dispatcher that is integrated
   * with the frame clock on your platform. E.g., on Android we recommend using Compose UI's
   * `AndroidUiDispatcher.Main` in order to access these optimizations.
 ```

rjrjr: Makes me think we should just drop CSR then, since it's already pretty redundant with what's built into `ComposeView` et al.

steve-the-edwards: I thought so at first too, but if the view runner work is not cancellable it does save view runner work by producing fewer renderings when they are stale.

PR #1366, opened 2025-07-10
Workflow visualizer UI improvement

wenli-cai:
### UI changes

The specific nodes that were affected in the render pass now highlighted in context to the whole tree, rather than just showing the specific frame. This makes it easier to see the exact process and layers of workflow that were passed through 

### Code changes

More fields were added to `Node` to support visualization capabilities, and is closely related with how `ActionLogger` records data from workflow runtime. 

All UI functions are now internal, since there's no need to support any consumers using this specific API (as it is meant to be run as a downloadable application).


workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowInfoPanel.kt
```diff
@@ -62,45 +69,120 @@ public fun RightInfoPanel(
...
+      return@LazyColumn
     }
+    item {
+      Text(
+        text = "Workflow Details",
+        style = MaterialTheme.typography.h6,
+        modifier = Modifier.padding(top = 8.dp, bottom = 8.dp)
+      )
+    }
+    val fields = listOf("Name", "Id", "Props", "State", "Rendering")
```
wenli-cai, line 106: is there a way around having to manually list out these fields like this? Final constants only allow primitive types too.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/model/Node.kt
```diff
@@ -7,26 +7,52 @@ package com.squareup.workflow1.traceviewer.model
...
+    }
+  }
+}
+
+internal fun Node.addChild(child: Node): Node {
+  return copy(children = this.children + child)
+}
+
+internal fun Node.replaceChild(child: Node): Node {
+  return copy(children = this.children.map { if (it.id == child.id) child else it })
```
steve-the-edwards: immutable maps can use the plus operator (`.plus` or `+` as infix) to do this without having to map each node. e.g.,

```suggestion
  return copy(children = this.children.plus(child.id to child))
```

That will replace only the value with `child.id` as its key.

Kotlin Playground link: https://pl.kotl.in/8IX2XBth3

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/model/NodeDiff.kt
```diff
@@ -0,0 +1,12 @@
+package com.squareup.workflow1.traceviewer.model
+
+/**
+ * Represents the difference between the current and previous state of a node in the workflow trace.
```
steve-the-edwards: It does not really represent the difference in any way. It just holds both nodes.

It might better be called `NodeUpdate` or something like that?

You might start adding some diffing methods on this class to help with that (if it's useful).

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowInfoPanel.kt
```diff
@@ -62,45 +69,122 @@ public fun RightInfoPanel(
...
```
steve-the-edwards, line 106: If you stick with this method (which I think is fine btw!) it would be good to just have this list as a constant on `Node` so that it keeps it close the code its representing.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/JsonParser.kt
```diff
@@ -83,7 +88,36 @@ private fun buildTree(node: Node, childrenByParent: Map<String, List<Node>>): No
...
+
+  val updatedNode = frame.copy(children = main.children)
+
+  return frame.children.fold(updatedNode) { mergedTree, frameChild ->
+    val mainTreeChild = mergedTree.children.singleOrNull { it.id == frameChild.id }
+    if (mainTreeChild != null) {
+      mergedTree.replaceChild(mergeFrameIntoMainTree(frameChild, mainTreeChild))
+    } else {
+      mergedTree.addChild(frameChild)
+    }
```
steve-the-edwards, line 114: Another option rather than keeping a separate list of `affectedNodes` would be to mark up these nodes as they are parsed within the actual node class itself. I.e. you could have an enum of change types and a field on `Node` for `status`., e.g., status is `SAME`, `NEW`, `PROPS_CHANGED`, `STATE_CHANGED`, `CHILD_STATE_CHANGED` or something like that? That way we can continue to add more colouring information to the UI for different aspects.

workflow-trace-viewer/build.gradle.kts
```diff
@@ -27,6 +27,13 @@ kotlin {
         implementation(libs.filekit.dialogs.compose)
       }
     }
+    jvmTest {
+      dependencies {
+        implementation(kotlin("test"))
+        implementation(kotlin("test-junit5"))
+        implementation(libs.junit.jupiter)
```
steve-the-edwards: What is jupiter for? I could not see something related to this on your unit test that you wrote?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/model/Node.kt
wenli-cai: `.plus` would work on a map, but `children` is a list of Nodes. Would you think a map is a better use case here? I was thinking I could at least keep the insertion order using a List, unless I used something like a LinkedHashMap?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/JsonParser.kt
wenli-cai: Yeah this would be great to have. I'll note this as a TODO in the future.

workflow-trace-viewer/build.gradle.kts
wenli-cai: It wasn't being used for anything, thanks for catching that. I had goose write the tests and it had just included it. Is it cleaning if it's removed?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowInfoPanel.kt
wenli-cai: Fixed with reflection in this PR https://github.com/square/workflow-kotlin/pull/1367

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/model/NodeDiff.kt
wenli-cai: That's true, `NodeUpdate` is good.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
```diff
@@ -25,21 +29,30 @@ public fun App(
...
   var selectedTraceFile by remember { mutableStateOf<PlatformFile?>(null) }
-  var selectedNode by remember { mutableStateOf<Node?>(null) }
+  var selectedNode by remember { mutableStateOf<NodeDiff?>(null) }
   var workflowFrames by remember { mutableStateOf<List<Node>>(emptyList()) }
   var frameIndex by remember { mutableIntStateOf(0) }
+  val sandboxState = remember { SandboxState() }
+
+  LaunchedEffect(frameIndex) {
+    sandboxState.reset()
+  }
```
zach-klippenstein: This is an anti-pattern (written about [multiple](https://www.notion.so/square-seller/Chronicle-Edition-2-2f71d66186b14aba8ae3ca6c4f61884c?source=copy_link#f00dc2a0838d465c95cd88f27f972251) [times](https://www.notion.so/square-seller/Chronicle-Edition-3-10e70293beed809385e3de61602001eb?source=copy_link#15270293beed80f1aa49d03cd1e34e24) in Compose Chronicles).

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/model/Node.kt
zach-klippenstein: Sounds like `LinkedHashMap` would be better for this case.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt
```diff
@@ -29,16 +30,18 @@ import io.github.vinceglb.filekit.PlatformFile
...
-  onNodeSelect: (Node) -> Unit,
+  onNodeSelect: (Node, Node?) -> Unit,
   modifier: Modifier = Modifier
 ) {
-  var frames by remember { mutableStateOf<List<Node>>(emptyList()) }
   var isLoading by remember(traceFile) { mutableStateOf(true) }
   var error by remember(traceFile) { mutableStateOf<Throwable?>(null) }
+  var frames by remember { mutableStateOf<List<Node>>(emptyList()) }
+  var fullTree by remember { mutableStateOf<List<Node>>(emptyList()) }
+  var affectedNodes by remember { mutableStateOf<List<Set<Node>>>(emptyList()) }
```
zach-klippenstein: Nit: Can any of these be `SnapshotStateList`s? 

Or if you want them to be immutable, then consider using `ImmutableList` and friends.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/JsonParser.kt
```diff
@@ -83,7 +88,36 @@ private fun buildTree(node: Node, childrenByParent: Map<String, List<Node>>): No
...
+ *
+ * @return Node the newly formed tree with the frame merged into it.
+ */
+internal fun mergeFrameIntoMainTree(
+  frame: Node,
+  main: Node
+): Node {
+  if (frame.id != main.id) {
+    throw IllegalArgumentException("Frame root ID does not match main tree root ID.")
+  }
```
zach-klippenstein: Nit: Could be `require`

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SandboxBackground.kt
```diff
@@ -49,8 +42,8 @@ public fun SandboxBackground(
           val event = awaitPointerEvent()
           if (event.type == PointerEventType.Scroll) {
             val scrollDelta = event.changes.first().scrollDelta.y
-            scale *= if (scrollDelta < 0) 1.1f else 0.9f
-            scale = scale.coerceIn(0.1f, 10f)
+            val factor = 1f + (-scrollDelta * 0.1f)
+            sandboxState.scale = (sandboxState.scale * factor).coerceIn(0.1f, 10f)
```
zach-klippenstein: Can you add a comment explaining what this math is for and what the constants mean?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowInfoPanel.kt
```diff
@@ -62,45 +69,122 @@ public fun RightInfoPanel(
...
-        modifier = textModifier,
-        style = textStyle
+        text = label,
+        style = MaterialTheme.typography.h6,
+        color = Color.Black,
+        fontWeight = FontWeight.Medium
       )
+      if (!open) {
+        return@Card
+      }
```
zach-klippenstein, line 152: UX nit: Would `AnimatedVisibility` look more polished?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/model/Node.kt
```diff
@@ -7,26 +7,52 @@ package com.squareup.workflow1.traceviewer.model
...
   val parent: String,
   val parentId: String,
-  val props: Any? = null,
-  val state: Any? = null,
-  val renderings: Any? = null,
+  val props: String,
+  val state: String,
+  val rendering: String = "",
   val children: List<Node>,
 ) {
```
zach-klippenstein, line 19: This class looks like it's intended to basically be a database table, where the properties are referenced dynamically by the UI. Can these properties just be stored in a map, with keys defined as constants? No need to use reflection.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
```diff
@@ -25,21 +29,30 @@ public fun App(
...
 
   Box(
     modifier = modifier
   ) {
```
zach-klippenstein, line 46: Not in PR scope, but you probably want `propagateMinConstraints` here.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/model/Node.kt
wenli-cai: Sounds good, doing this would also mean that `addChild` and `replaceChild` are the exact same in implementation. Is it good still keep those as different functions, though, to make the recursive algorithm easier to understand?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowInfoPanel.kt
wenli-cai: Definitely, there's a lot of ui/ux changes I have written down to look more polished, and it's something I'm looking to complete later in the project.

workflow-trace-viewer/build.gradle.kts
japplin: Would be great to remove this if it's not needed.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
wenli-cai: Could you explain this more? As far as I understand, this would be telling that all children containers need to be a specific size within the `Box`?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/model/Node.kt
wenli-cai: These properties are all here for the purpose of moshi being able to parse a file into these objects. Unless there is a way to use moshi in the way you're describing it.

Even then, I'm not sure what the difference is. Could you help me explain the use case of having a data class like this vs using a map and treating it like a database table?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SandboxBackground.kt
wenli-cai: Done

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/JsonParser.kt

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt
wenli-cai: I didn't know about this, this is great to know! The next step in the project would require these being mutable, so thanks for pointing this out

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
zach-klippenstein: Should be
```kotlin
LaunchedEffect(sandboxState) {
  snapshotFlow { frameIndex }.collect {
    sandboxState.reset()
  }
}
```

workflow-trace-viewer/build.gradle.kts
```diff
@@ -50,3 +57,10 @@ compose {
     }
   }
 }
+
+tasks.named<Test>("jvmTest") {
```
japplin, line 60: I also wonder if this is not needed?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt

workflow-trace-viewer/build.gradle.kts
wenli-cai: Done

workflow-trace-viewer/build.gradle.kts
steve-the-edwards: +1 - do we know why this _is_ needed? if not, lets remove.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/model/Node.kt
steve-the-edwards: The map woudl be just key : value where both are strings. Since everything but children here is a string, the idea would be that essentially we are using this `data class` to store a bunch of String properties. Since they are all strings, we could just make it a `nodeProperties = mapOf<String, String>()` for everything but the children. once that is a map then you are able to easily iterate both the 'properties' keys (with `nodeProperties.keys`.) and the values (with `nodeProperties.values`) without needing to use reflection etc.

As for Moshi I imagine they have a default adapter that just maps JSON properties to a map.

Where it gets slightly more complicated  is with the `children` property here, since that is not a String and is why we have this as a full class.

You _could_ do something like:

```
data class Node ( val properties: ImmutableMap<String, String>, val children: LinkedHashMap<String, Node>) {} 
```

and then you'd have to change the parsing. ultimately probably that would be the 'best' fit for the data (and you'd be able to avoid using reflection as the properties list change since you'd always iterate over all the keys).

As for whether you do that now, I'm not too worried either way 🤷🏻

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/model/Node.kt
steve-the-edwards: you can leave them separate as long as one uses the other (so you avoid having to update multiple places with the duplicate code).

steve-the-edwards: WEre you going to do this? Or on a later PR?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/model/Node.kt
```diff
@@ -7,26 +7,52 @@ package com.squareup.workflow1.traceviewer.model
...
+    }
+  }
+}
+
+internal fun Node.addChild(child: Node): Node {
+  return copy(children = LinkedHashMap(this.children.plus(child.id to child)))
+}
+
+internal fun Node.replaceChild(child: Node): Node {
+  return copy(children = LinkedHashMap(this.children.plus(child.id to child)))
```
steve-the-edwards: Make `replaceChild` call `addChild` so you only have 1 place to update/maintain

workflow-trace-viewer/build.gradle.kts
wenli-cai: it seems like it's needed for registering and running junit tests in a multiplatform app? 
https://kotlinlang.org/docs/jvm-test-using-junit.html#add-dependencies

But I havent removed the logging parts afterwards, those aren't necessary.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/model/Node.kt
wenli-cai: Done

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/model/Node.kt
wenli-cai: understood, this makes sense. However, I still don't understand the distinction that @zach-klippenstein made originally about classes and "database table classes". Are data classes not meant to be used in a way that just holds data?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/model/Node.kt
wenli-cai: Fixed in another comment [here](https://github.com/square/workflow-kotlin/pull/1366#discussion_r2216323738)

workflow-trace-viewer/build.gradle.kts
steve-the-edwards: Great. sounds good.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/model/Node.kt
steve-the-edwards: ya they are meant for that. I think the distinction comes down to the _type_ of data. if you have properties of different strongly types vs if you just have a set of `String` properties.

wenli-cai: ah gotcha, I understand now, thanks for clarifying that.

PR #1367, opened 2025-07-11
Workflow Visualizer UX improvements

wenli-cai:
1. Allow each node to be opened and closed to ignore irrelevant information for each frame. Nodes that were directly affected in the specific render pass, though, are all open by default, and everything else is closed by default.

2. Allow for using a mouse's vertical scroll to traverse through the row of frames.

### DIFF
--- a/workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/FrameSelectTab.kt
+++ b/workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/FrameSelectTab.kt
@@ -3,7 +3,6 @@ package com.squareup.workflow1.traceviewer.ui
 import androidx.compose.foundation.MutatePriority
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.gestures.awaitEachGesture
-import androidx.compose.foundation.gestures.scrollBy
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.lazy.LazyRow
 import androidx.compose.foundation.lazy.rememberLazyListState
@@ -11,7 +10,6 @@ import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material.Surface
 import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.graphics.Color
@@ -19,6 +17,7 @@ import androidx.compose.ui.input.pointer.PointerEventType
 import androidx.compose.ui.input.pointer.pointerInput
 import androidx.compose.ui.unit.dp
 import com.squareup.workflow1.traceviewer.model.Node
+import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.launch
 
 /**
@@ -32,7 +31,6 @@ internal fun FrameSelectTab(
   modifier: Modifier = Modifier
 ) {
   val lazyListState = rememberLazyListState()
-  val coroutineScope = rememberCoroutineScope()
 
   Surface(
     modifier = modifier,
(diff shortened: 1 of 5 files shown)


workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowInfoPanel.kt
```diff
@@ -103,14 +103,18 @@ private fun NodePanelDetails(
...
-    val fields = listOf("Name", "Id", "Props", "State", "Rendering")
 
-    items(fields) { field ->
-      DetailCard(
-        label = field,
-        currValue = Node.getNodeField(node.current, field),
-        pastValue = if (node.previous != null) Node.getNodeField(node.previous, field) else null
-      )
+    val fields = Node::class.memberProperties
+    for (field in fields) {
```
steve-the-edwards, line 108: oh nice. Ok i see you moved to reflection here. that's great. I think it makes sense.

steve-the-edwards (approved):
Neat.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/FrameSelectTab.kt
```diff
@@ -24,17 +31,30 @@ internal fun FrameSelectTab(
...
+            if (event.type == PointerEventType.Scroll) {
+              val scrollDeltaY = event.changes.first().scrollDelta.y
+              coroutineScope.launch {
+                lazyListState.scroll(MutatePriority.Default) {
+                  scrollBy(scrollDeltaY * 10f)
+                }
+              }
+            }
+          }
+        },
```
zach-klippenstein: Is there a particular reason you need this animation to outlive the `pointerInput` modifier's lifetime? It would be slightly cleaner to just wrap this with `coroutineScope {}`, i.e.
```suggestion
        .pointerInput(Unit) {
          coroutineScope {
            awaitEachGesture {
              val event = awaitPointerEvent()
              if (event.type == PointerEventType.Scroll) {
                val scrollDeltaY = event.changes.first().scrollDelta.y
                launch {
                  lazyListState.scroll(MutatePriority.Default) {
                    scrollBy(scrollDeltaY * 10f)
                  }
                }
              }
            }
          }
        },
```

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt
```diff
@@ -92,21 +107,42 @@ private fun DrawTree(
     horizontalAlignment = Alignment.CenterHorizontally,
   ) {
     val isAffected = affectedNodes.contains(node)
-    DrawNode(node, previousNode, isAffected, onNodeSelect)
+    // By default, nodes that relevant to this specific frame are expanded. All others are closed.
+    LaunchedEffect(expandedNodes) {
+      expandedNodes[node.id] = isAffected
+    }
```
zach-klippenstein, line 114: This is a weird little bit of backwards data flow. Feels like there should be a presenter or something handling this? 

It also looks like `affectedNodes` and `expandedNodes` effectively represent the same thing, since this code just synchronizes them, or partially synchronizes them? It's possible I'm misunderstanding the code.

wenli-cai: 1. I'm not familiar with using a Presenter as a pattern, if you wouldn't mind explaining a potential solution (or point me to a resource). I can understand that UI should be using callbacks to affect data rather than directly mutating it, but when I experimented hoisting everything out of this Composable and into `RenderDiagram`, the tree would no longer recompose everytime I open/closed a node. 

    But I definitely could've been trying an incorrect way of doing it, leading it to fail.

3. Any `affectedNodes` for the specific renderpass will always be the same, but the nodes that are expanded can be changed. So we can open a non-affected node, and we can also close an affected node

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/FrameSelectTab.kt
wenli-cai: Changed

PR #1371, opened 2025-07-16
WIP Android-specific tests for WorkStealingDispatcher.

zach-klippenstein:
Validates behavior with `Dispatchers.Main` and `.immediate`.

This can't be merged as-is, since it requires the tests to live in the android source set but that's in a different module so it can't see WSD. Need to land https://github.com/square/workflow-kotlin/pull/1370 first.

### DIFF
--- a/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/Synchronization.kt
+++ b/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/Synchronization.kt
@@ -1,5 +1,5 @@
 package com.squareup.workflow1.internal
 
-internal expect class Lock()
+public expect class Lock()
 
-internal expect inline fun <R> Lock.withLock(block: () -> R): R
+public expect inline fun <R> Lock.withLock(block: () -> R): R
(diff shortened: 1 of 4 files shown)


workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkStealingDispatcher.kt
```diff
@@ -38,7 +38,7 @@ import kotlin.coroutines.resume
  * delegate scheduling behavior to. This can either be a confined or unconfined dispatcher, and its
  * behavior will be preserved transparently.
  */
-internal open class WorkStealingDispatcher protected constructor(
+public open class WorkStealingDispatcher protected constructor(
```
zach-klippenstein, line 41: - [ ] Undo this

workflow-runtime/src/jvmMain/kotlin/com/squareup/workflow1/internal/Synchronization.jvm.kt
```diff
@@ -1,5 +1,5 @@
 package com.squareup.workflow1.internal
 
-internal actual typealias Lock = Any
+public actual typealias Lock = Any
 
-internal actual inline fun <R> Lock.withLock(block: () -> R): R = synchronized(this, block)
+public actual inline fun <R> Lock.withLock(block: () -> R): R = synchronized(this, block)
```
zach-klippenstein, line 5: - [ ] Undo this

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/Synchronization.kt
```diff
@@ -1,5 +1,5 @@
 package com.squareup.workflow1.internal
 
-internal expect class Lock()
+public expect class Lock()
 
-internal expect inline fun <R> Lock.withLock(block: () -> R): R
+public expect inline fun <R> Lock.withLock(block: () -> R): R
```
zach-klippenstein, line 5: - [ ] Undo this

PR #1372, opened 2025-07-16
Make RenderingProduced not a value class.

zach-klippenstein:
Value classes get boxed as soon as they're used as a supertype, so it's basically pointless to make sealed class children value classes. I made it a data class instead.

This change is required to move `:workflow-runtime-android` into `:workflow-runtime`, or we get a compiler crash.

I also made the object siblings data objects to fix a compiler suggestion and get better `toString` methods.

### DIFF
--- a/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt
+++ b/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt
@@ -172,13 +172,13 @@ public interface WorkflowInterceptor {
    * A render pass has been skipped by an optimization, multiple actions are applied before
    * the runtime produces a rendering.
    */
-  public object RenderPassSkipped : RuntimeUpdate
+  public data object RenderPassSkipped : RuntimeUpdate
 
   /**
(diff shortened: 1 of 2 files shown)


workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt
```diff
@@ -172,22 +171,21 @@ public interface WorkflowInterceptor {
    * A render pass has been skipped by an optimization, multiple actions are applied before
    * the runtime produces a rendering.
    */
-  public object RenderPassSkipped : RuntimeUpdate
+  public data object RenderPassSkipped : RuntimeUpdate
```
rjrjr, line 174: We avoid using `data` in public API, remember? Seems like this would be fine as just `object` / `class`.

zach-klippenstein: The arguments for not using data classes in public API don't apply to objects, since they contain no data by definition.

And the data class was previously a value class, which is even more impossible to change than a data class, so this is no more constrained.

PR #1374, opened 2025-07-17
Stream emulator data into workflow visualizer app

wenli-cai:
By using Unix Domain Sockets and adb forwarding, the visualizer is able to retrieve logged data directly from the emulator to display the workflow render passes.

The app has also been changed to allow for both File and Live tracing mode through a toggle.


workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
```diff
@@ -16,25 +17,37 @@ import androidx.compose.ui.geometry.Offset
...
   var selectedNode by remember { mutableStateOf<NodeUpdate?>(null) }
-  var workflowFrames by remember { mutableStateOf<List<Node>>(emptyList()) }
+  val workflowFrames = remember { mutableStateListOf<Node>() }
   var frameIndex by remember { mutableIntStateOf(0) }
   val sandboxState = remember { SandboxState() }
 
+  // Default to File mode, and can be toggled to be in Live mode.
+  var traceMode by remember { mutableStateOf<TraceMode>(TraceMode.File(null)) }
+  var selectedTraceFile by remember { mutableStateOf<PlatformFile?>(null) }
+  val socket = remember { SocketClient() }
```
steve-the-edwards: Create this lazily so it only happens if we switch to streaming mode?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
```diff
@@ -16,25 +17,37 @@ import androidx.compose.ui.geometry.Offset
...
+
+  Runtime.getRuntime().addShutdownHook(
+    Thread {
+      socket.close()
+    }
+  )
```
steve-the-edwards: Is this a JVM runtime API that @tcmulcahy was showing you? Not familiar with it.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
```diff
@@ -92,6 +140,11 @@ internal class SandboxState {
 
   fun reset() {
     offset = Offset.Zero
-    scale = 1f
+    // scale = 1f
```
steve-the-edwards: remove?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
```diff
@@ -0,0 +1,69 @@
...
+
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.launch
+import java.net.Socket
+import java.net.SocketException
+
+/**
+ * This is a client that connects to the `ActionLogger` Unix Domain Socket and listens for any new
```
steve-the-edwards: nit: not necessarily `ActionLogger` socket (though it is that in your branch). More importantly, just a publishing socket from an app. (Remember this is in Workflow library which can be used anywhere).

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
steve-the-edwards: Looking at it lazily, it looks liek this is pretty lightweight, so not a big deal.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
steve-the-edwards: Does this need to launch a Thread to do this so that it does not block shutdown? It would be good to have a comment why we need a new Thread for this.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  /**
+   * This will always be called within an asynchronous call, so we do not need to block/launch a
+   * new coroutine here.
+   *
+   * To better separate the responsibility of reading from the socket, we use a channel for the caller
+   * to handle parsing and amalgamating the render passes.
+   */
+  fun beginListen(scope: CoroutineScope) {
```
steve-the-edwards: make this a `suspend fun` so we don't have to pass the `CoroutineScope` on the stack.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
```diff
@@ -0,0 +1,69 @@
...
+  fun close() {
+    if (!initialized) {
+      return
+    }
+    socket.close()
+  }
+
+  /**
+   * This will always be called within an asynchronous call, so we do not need to block/launch a
+   * new coroutine here.
```
steve-the-edwards: But, it does launch a new coroutine?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
```diff
@@ -0,0 +1,69 @@
...
+   *
+   * To better separate the responsibility of reading from the socket, we use a channel for the caller
+   * to handle parsing and amalgamating the render passes.
+   */
+  fun beginListen(scope: CoroutineScope) {
+    scope.launch(Dispatchers.IO) {
```
steve-the-edwards: if you make this a suspend fun, this can just be

```suggestion
    withContext(Dispatchers.IO) {
```

Then make the client handle launching a separate coroutine for this to run in the background.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
steve-the-edwards: This makes the client handle the job/coroutine for this. Otherwise we are giving too much responsibility/power to this function with the `CoroutineScope`

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
```diff
@@ -0,0 +1,69 @@
...
+      val reader = socket.getInputStream().bufferedReader()
+      try {
+        while (true) {
+          val input = reader.readLine()
+          renderPassChannel.trySend(input)
+        }
+      } catch (e: SocketException) {
+        println("Exiting socket listener due to: ${e.message}")
+      }
```
steve-the-edwards, line 76: Don't we need to close the input stream in a `finally` here? Is the BufferedReader handling that behind the scense? I don't see how.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
```diff
@@ -73,16 +99,38 @@ public fun App(
...
-    UploadFile(
-      resetOnFileSelect = {
-        selectedTraceFile = it
-        selectedNode = null
-        frameIndex = 0
-        workflowFrames = emptyList()
+    TraceModeToggleSwitch(
+      onToggle = {
+        resetStates()
+        traceMode = if (traceMode is TraceMode.Live) {
```
ericmaxwell2003, line 95: nit, consider calling the traceMode instance under evaluation as "currentTraceMode"

```
if(currentTraceMode is TraceMode.Live {

} else {

}
```

IMO, it's a bit clearer that the left side is the new state and the right side is existing state.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt
```diff
@@ -26,49 +26,96 @@ import androidx.compose.ui.input.pointer.isPrimaryPressed
...
+    onFileParse(frame)
+  }
 
-    when (parseResult) {
+  // Handles the result of parsing a trace, either from file or live. Live mode includes callback
+  // for when a new frame is received.
+  fun handleParseResult(
+    parseResult: ParseResult,
+    onNewFrame: (() -> Unit)? = null
+  ): Boolean {
```
ericmaxwell2003: Nit, what's teh intent of this boolean?  It seems like it always just returns whatever the value was of `parseResult`, and the result doesn't seem to be used elsewhere in your code anyway.  Thoughts on just making it Unit? (no return type)

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt
```diff
@@ -26,49 +26,96 @@ import androidx.compose.ui.input.pointer.isPrimaryPressed
...
+        onNewFrame?.invoke()
+        true
+      }
+    }
+  }
+
+  LaunchedEffect(traceSource) {
+    when (traceSource) {
+      is TraceMode.File -> {
+        // We guarantee the file is null since this composable can only be called when a file is selected.
```
ericmaxwell2003: Do you mean you guarantee that it's "not" null?  

One nit.. maybe you should do some sort of assertion/error here?

Like 

```
checkNotNull(traceSource.file) { 
  "traceSource.file should never be null here since this composable can only be called when a file is selected."
}
```

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
```diff
@@ -0,0 +1,69 @@
...
+        .trim().toInt()
+    }
+    socket = Socket("localhost", port)
+  }
+
+  fun close() {
+    if (!initialized) {
+      return
+    }
+    socket.close()
```
ericmaxwell2003: Should you check for isClosed first?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
ericmaxwell2003: ➕ 

Fully together
```
suspend fun beginListen() = withContext(Dispatchers.IO) {
   // do all the stuff, now on an IO Thread pool backed coroutines dispatcher.
}
```

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
```diff
@@ -0,0 +1,69 @@
...
+  fun beginListen(scope: CoroutineScope) {
+    scope.launch(Dispatchers.IO) {
+      val reader = socket.getInputStream().bufferedReader()
+      try {
+        while (true) {
+          val input = reader.readLine()
+          renderPassChannel.trySend(input)
+        }
+      } catch (e: SocketException) {
+        println("Exiting socket listener due to: ${e.message}")
```
ericmaxwell2003: Should this be a log with the stack trace.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
```diff
@@ -0,0 +1,69 @@
...
+
+  /**
+   * We use any available ports on the host machine to connect to the emulator.
+   *
+   * `workflow-trace` is the name of the unix socket created, and since Android uses
+   * `LocalServerSocket` -- which creates a unix socket on the linux abstract namespace -- we use
+   * `localabstract:` to connect to it.
+   */
+  fun start() {
+    initialized = true
```
ericmaxwell2003: Are we threadsafe here?   just double checking, you may be

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
```diff
@@ -0,0 +1,69 @@
...
+    val process = ProcessBuilder(
+      "adb", "forward", "tcp:0", "localabstract:workflow-trace"
+    ).start()
+
+    // The adb forward command will output the port number it picks to connect.
+    val port = run {
+      process.waitFor()
+      process.inputStream.bufferedReader().readText()
+        .trim().toInt()
+    }
```
ericmaxwell2003, line 28: nit: thoughts making a function call out instead of this run syntax.

readInt() or something like that 

where readInt == 

```
fun readInt() {
      process.waitFor()
      process.inputStream.bufferedReader().readText()
        .trim().toInt()
    }
```

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt
```diff
@@ -26,49 +26,96 @@ import androidx.compose.ui.input.pointer.isPrimaryPressed
...
+
+      is TraceMode.Live -> {
+        val socket = traceSource.socket
+        socket.beginListen(this)
+        val adapter: JsonAdapter<List<Node>> = createMoshiAdapter<Node>()
+
+        withContext(Dispatchers.IO) {
+          // Since channel implements ChannelIterator, we can for-loop through on the receiver end.
+          for (renderPass in socket.renderPassChannel) {
+            val currentTree = if (fullTree.isEmpty()) null else fullTree.last()
```
ericmaxwell2003: Stylistic nit:   `fullTree.lastOrNull()` could replace this if/else

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
wenli-cai: `util.Log` isn't available for KMP apps, should I look into 3rd party libraries for this?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
wenli-cai: From my understanding, shutdown hooks are registered by JVM and only ran once it starts the shutdown sequence. So it sort of makes sense that it needs to be on its own Thread. I'll make a comment on this

zach-klippenstein: If the runtime is already shutting down, why do we care about blocking? And for that matter, why do we care about closing the socket at all in that case? The JVM/OS will close it for us once the process dies – you can't leak a socket outside a process boundary if you tried.

Also, it looks like this code is running directly in composition, which is extremely bad – it would add a new shutdown hook every time this composable recomposes.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
```diff
@@ -44,18 +57,31 @@ public fun App(
   Box(
     modifier = modifier
   ) {
+    fun resetStates() = run {
```
zach-klippenstein: What is `run` for here? Can't this just be 
```suggestion
    fun resetStates() {
```

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/Main.kt
```diff
@@ -8,6 +8,12 @@ import androidx.compose.ui.window.singleWindowApplication
  * Main entry point for the desktop application, see [README.md] for more details.
  */
 fun main() {
+  Runtime.getRuntime().addShutdownHook(
+    Thread {
+      ProcessBuilder("adb", "forward", "--remove-all")
+        .start().waitFor()
+    }
+  )
```
zach-klippenstein: Why use a shutdown hook instead of just normal control flow, i.e. wrapping `singleWindowApplication` in a `try/finally` and doing the cleanup in finally? You'd need to pass `exitProcessOnExit = false`. Imo that would make this code more straightforward.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt
```diff
@@ -26,49 +26,96 @@ import androidx.compose.ui.input.pointer.isPrimaryPressed
...
+        val parseResult = parseFileTrace(traceSource.file!!)
+        handleParseResult(parseResult)
+      }
+
+      is TraceMode.Live -> {
+        val socket = traceSource.socket
+        socket.beginListen(this)
+        val adapter: JsonAdapter<List<Node>> = createMoshiAdapter<Node>()
+
+        withContext(Dispatchers.IO) {
```
zach-klippenstein: I don't think this code needs to be explicitly ran on `Dispatchers.IO`. I don't think any of it is doing system IO, it's all computation, so it should be on `Dispatchers.Default`.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/JsonParser.kt
```diff
@@ -24,11 +25,11 @@ const val ROOT_ID: String = "-1"
  * @return A [ParseResult] representing result of parsing, either an error related to the
  * format of the JSON, or a success and a parsed trace.
  */
-internal suspend fun parseTrace(
+internal suspend fun parseFileTrace(
   file: PlatformFile,
 ): ParseResult {
   val jsonString = file.readString()
```
zach-klippenstein, line 31: Does this `readString` do system IO directly? If so, it should be ran on `Dispatchers.IO`.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * This is a client that connects to the `ActionLogger` Unix Domain Socket and listens for any new
+ * render passes. Since this app is on JVM and the server is on Android, we use ADB to forward the
+ * port onto the socket.
+ */
+internal class SocketClient {
+  private lateinit var socket: Socket
+  private var initialized = false
+  val renderPassChannel: Channel<String> = Channel(Channel.UNLIMITED)
```
zach-klippenstein: It's a good practice to avoid unlimited channels. See [this post](https://cash.slack.com/archives/C055GSCQKPD/p1743635957748299) for more info.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
zach-klippenstein: Better yet, just inline it, this is a small function and the `readInt` wouldn't be worth factoring out imo.
```kotlin
process.waitFor()
val port = process.inputStream.bufferedReader().readText()
    .trim().toInt()
```

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
zach-klippenstein: Should this check `initialized` to avoid double-initialization?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
zach-klippenstein: I think in general, the start/close contract is unclear. Presumably calling start and close multiple times is not meaningful, so that should be documented in the kdoc, and enforced either by making them idempotent (i.e. calling `start()` again after calling it once does nothing, same for `close()`) or by throwing when multiple calls are detected.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
zach-klippenstein: But then `beginListen` would be a misnomer, since it doesn't just "begin" the listen, it is the whole listen. So a better name might be something like "consumeSocket" or something more descriptive.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
zach-klippenstein: `Throwable.printStackTrace()` is available in KMP.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
zach-klippenstein: Just use the `Closable.use {}` extension.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
zach-klippenstein: I'm assuming @ericmaxwell2003 is referring to `socket.isClosed`, which I don't think you need to check, I would think `socket.close()` would be idempotent.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt
```diff
@@ -26,49 +26,96 @@ import androidx.compose.ui.input.pointer.isPrimaryPressed
...
 /**
  * Access point for drawing the main content of the app. It will load the trace for given files and
  * tabs. This will also all errors related to errors parsing a given trace JSON file.
+ *
+ * This handles either File or Live trace modes, and will parse equally
  */
 @Composable
-internal fun RenderDiagram(
-  traceFile: PlatformFile,
+internal fun RenderTrace(
```
zach-klippenstein: This composable contains lots of logic related to parsing, not views. Might be worth considering factoring it out into a separate class, would be a lot more testable and keep the code in this file to actual view concerns.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/FrameSelectTab.kt
```diff
@@ -31,7 +32,11 @@ internal fun FrameSelectTab(
   modifier: Modifier = Modifier
 ) {
   val lazyListState = rememberLazyListState()
-
+  LaunchedEffect(currentIndex) {
+    if (currentIndex < 0) return@LaunchedEffect
+    lazyListState.animateScrollToItem(currentIndex)
+  }
```
zach-klippenstein: This is quite wasteful, launching a noop effect. Instead check `currentIndex` before even calling `LaunchedEffect`.
```suggestion
  if (currentIndex >= 0) {
    LaunchedEffect(currentIndex) {
      lazyListState.animateScrollToItem(currentIndex)
    }
  }
```

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
wenli-cai: About the point for having threads, it seem like `addShutdownHook` just takes a `Thread` and will run all registered hooks in a random order. I wasn't able to find anything that would allow you to add hooks without it being on a different thread. Do update me know, though, if there are ways to do that (so I know for the future too)!

1. Good point, I am not knowledgeable in this area so I didn't think about this. I had imagined that we ought to free the resources when provided an opportunity to do so, but this makes sense.
2. It's removed now.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/JsonParser.kt
wenli-cai: Does system IO mean something specific here? If not, `PlatformFile.readString()` is already dispatched onto the IO thread, so this should be okay.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
wenli-cai: removed

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
wenli-cai: Changed

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
wenli-cai: Changed, I'm beginning to see how useful `Closeable` is

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt
wenli-cai: Yep, good catch. Removed

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt
wenli-cai: Oops that's a typo, changed to a check now.
Is a check like this for readability? As in we can avoid having a comment here by stating explicitly (with code) that the file cannot be null? Or is it also useful to prevent us from having to use `!!`

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
wenli-cai: Done

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
wenli-cai: Done

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
wenli-cai: I'm not even sure what I was thinking, changed.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt
wenli-cai: Changed

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
wenli-cai: Done

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
wenli-cai: Turns out I dont understand coroutines at all, fixed the comment now.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
wenli-cai: Changed

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
wenli-cai: Changed, changed name to. `pollSocket` as well.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
wenli-cai: Changed, kdoc and the readme all clarifies this now.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
wenli-cai: yep, done

wenli-cai: It should be, `start()` and `close()` are all called on the main thread. Speaking of which I should change the name to be `open()`

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/FrameSelectTab.kt
wenli-cai: Done

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt
```diff
@@ -26,49 +26,108 @@ import androidx.compose.ui.input.pointer.isPrimaryPressed
...
+          try {
+            socket.pollSocket()
+          } catch (e: SocketException) {
+            error = SocketException("Socket has already been closed or is not available: ${e.message}")
+            return@launch
+          }
+        }
+        if (error != null) {
+          return@LaunchedEffect
+        }
```
wenli-cai: Can there be a situation where the if check happens before the launched coroutine is able to return an error? Manually testing it doesn't seem like it's the case, but I'm sure there's a better way of writing this bit of error handling

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt
zach-klippenstein: But I don't mean you need to do it in this PR – just something to think about for later.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
zach-klippenstein: If this class is intended to be single-use, you probably want to also close the channel here so that anything that tries to read from it sees there's no more data coming.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
```diff
@@ -73,16 +93,38 @@ public fun App(
...
+          frameIndex = 0
+          TraceMode.File(null)
+        } else {
+          // TODO: TraceRecorder needs to be able to take in multiple clients if this is the case
+          /*
+          We set the frame to -1 here since we always increment it during Live mode as the list of
+          frames get populated, so we avoid off by one when indexing into the frames.
+           */
+          frameIndex = -1
+          socket.open()
```
zach-klippenstein: I would not open the socket here, since it's not clear who's responsible for closing it.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt
zach-klippenstein: Yea this is definitely a race, the if check is pretty much guaranteed to run before the coroutine even starts since compose's dispatcher will dispatch the coroutine. And `pollSocket` is long-running – an error could happen at any time over the lifetime of the socket.

I think what you want to happen is that if the socket errors out, we just give up, so the loop that's reading from the socket below is cancelled. There are a few ways to do this, but if `SocketClient` is intended to be single-use, then probably the way that makes the most sense is to have `pollSocket()` close the channel after it returns. Then the `for/in` loop on line 119 will detect the socket is closed and complete. You can make `SocketClient` implement `Closeable`, close the channel in `SocketClient.close()`, and then use the `use` extension.

Probably worth rethinking the API of `SocketClient` and how it's used a bit to make something that's a bit more self-contained and safe. Maybe we should pair on it? Feel free to throw something on my calendar.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
zach-klippenstein: It does seem like a bit of a smell to create a network client in composition though, especially since this is a single-use object.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt
wenli-cai: Implementing `Closeable` seems like a good idea, that would mean the responsibility for opening and closing the socket is more clear, right?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
wenli-cai: Changed

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SocketClient.kt
steve-the-edwards: You aren't the only one!

steve-the-edwards (approved):

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/TraceModeToggleSwitch.kt
```diff
@@ -0,0 +1,58 @@
...
+
+  Column(
+    modifier = modifier.padding(16.dp),
+    horizontalAlignment = Alignment.CenterHorizontally
+  ) {
+    Switch(
+      checked = checked,
+      onCheckedChange = {
+        checked = it
+        onToggle()
```
zach-klippenstein: This is an anti-pattern, since you now have two sources of truth for `traceMode` and violates UDF. You don't need the `checked` property at all, just check the `traceMode` parameter every time. Or do something like
```kotlin
val isLive = traceMode is TraceMode.Live
```
since you check it below as well.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/TraceParser.kt
```diff
@@ -0,0 +1,107 @@
...
+      }
+
+      is TraceMode.Live -> {
+        val adapter: JsonAdapter<List<Node>> = createMoshiAdapter<Node>()
+        streamRenderPassesFromDevice { renderPass ->
+          val currentTree = fullTree.lastOrNull()
+          val parseResult = parseLiveTrace(renderPass, adapter, currentTree)
+          handleParseResult(parseResult, onNewFrame)
+        }
+        error = "Socket has already been closed or is not available."
```
zach-klippenstein, line 87: Beautiful!

zach-klippenstein (approved):
One last Compose UI thing to clean up, unrelated to the network stuff we discussed.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/TraceModeToggleSwitch.kt
wenli-cai: Fixed

PR #1375, opened 2025-07-18
Fix workflow freezing thread safety.

zach-klippenstein:


workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/RealRenderContext.kt
```diff
@@ -100,7 +100,6 @@ internal class RealRenderContext<PropsT, StateT, OutputT>(
    * Freezes this context so that any further calls to this context will throw.
    */
   fun freeze() {
-    checkNotFrozen("freeze") { "freeze" }
```
zach-klippenstein, line 103: This was just a defensive programming check, since this method is only called by internal code. It's more useful for it to be idempotent now.

workflow-runtime/src/appleMain/kotlin/com/squareup/workflow1/internal/Synchronization.apple.kt
```diff
@@ -12,3 +18,35 @@ internal actual inline fun <R> Lock.withLock(block: () -> R): R {
...
   }
 }
+
+/**
+ * Implementation of [ThreadLocal] that works in a similar way to Java's, based on a thread-specific
+ * map/dictionary.
+ */
+internal actual class ThreadLocal<T>(
+  private val initialValue: () -> T
+) : NSObject(), NSCopyingProtocol {
```
rjrjr, line 33: @jamieQ maybe you could give this a sanity check?

workflow-runtime/src/jvmTest/kotlin/com/squareup/workflow1/StressTestHelpers.kt
```diff
@@ -0,0 +1,27 @@
+package com.squareup.workflow1
+
+import java.util.concurrent.CountDownLatch
+
+/**
+ * Returns the maximum number of threads that can be ran in parallel on the host system, rounded
```
rjrjr:
```suggestion
 * Returns the maximum number of threads that can be run in parallel on the host system, rounded
```

:D

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/RealRenderContext.kt
```diff
@@ -51,8 +51,12 @@ internal class RealRenderContext<PropsT, StateT, OutputT>(
    * Used to:
    *  - prevent modifications to this object after [freeze] is called.
    *  - prevent sending to sinks before render returns.
+   *
+   * This is a [ThreadLocal] since we only care about preventing calls during rendering from the
```
rjrjr, line 56: And we're confident that we only ever call `freeze()` from the render thread? I can't think of any violations.

rjrjr (approved):
You have my LG for what it's worth. Not my strong suit.

workflow-runtime/src/jvmTest/kotlin/com/squareup/workflow1/WorkflowRuntimeMultithreadingStressTest.kt
```diff
@@ -0,0 +1,92 @@
...
+      render = { _, count ->
+        val action = action<Unit, Int, Nothing>("countChild") { this.state++ }
+        repeat(childCount) { childIndex ->
+          renderChild(child, props = childIndex, key = "child-$childIndex", handler = { action })
+        }
+        return@stateful count
+      })
+
+    val testDispatcher = newFixedThreadPoolContext(nThreads = testThreadCount, name = "test")
+    testDispatcher.use {
```
steve-the-edwards, line 69: What is this API and why do we need it? is this like a resource API for the threads underneath this disaptcher?

workflow-runtime/src/appleMain/kotlin/com/squareup/workflow1/internal/Synchronization.apple.kt
steve-the-edwards: yes please.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/RealRenderContext.kt
steve-the-edwards: Ya I can't think of `RenderContext` methods that would be called off main thread? will keep thinking.

workflow-runtime/src/iosTest/kotlin/com/squareup/workflow1/internal/ThreadLocalTest.kt
```diff
@@ -0,0 +1,82 @@
+package com.squareup.workflow1.internal
+
+import platform.Foundation.NSLock
+import platform.Foundation.NSThread
+import kotlin.concurrent.Volatile
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class ThreadLocalTest {
```
steve-the-edwards, line 9: @jamieQ and whoever else to review this as well then

steve-the-edwards (approved):

workflow-runtime/src/appleMain/kotlin/com/squareup/workflow1/internal/Synchronization.apple.kt
steve-the-edwards: cc @mjohnson12 , @square-tomb , @watt

workflow-runtime/src/iosTest/kotlin/com/squareup/workflow1/internal/ThreadLocalTest.kt
```diff
@@ -0,0 +1,82 @@
...
+    }
+    thread.start()
+    thread.join()
+
+    assertEquals(42, valueFromThread)
+  }
+
+  @Test fun set_fromDifferentThreads_doNotConflict() {
+    val threadLocal = ThreadLocal(initialValue = { 0 })
+    val threadStartedLatch = NSLock().apply { lock() }
```
jamieQ: what does `apply { ... }` do?

---

edit: the AI tells me it runs the closure on `self` and returns the instance – so, effectively is a configuration block

workflow-runtime/src/iosTest/kotlin/com/squareup/workflow1/internal/ThreadLocalTest.kt
```diff
@@ -0,0 +1,82 @@
...
+    val firstReadLatch = NSLock().apply { lock() }
+    val firstReadDoneLatch = NSLock().apply { lock() }
+    val secondReadLatch = NSLock().apply { lock() }
+
+    val thread = NSThread {
+      threadStartedLatch.unlock()
```
jamieQ: i'm a bit surprised this works without some assertion failure, as the lock API typically require that unlocks happen on the same thread that acquired the lock.

i'd like to offer some possible alternatives, but i'm having trouble figuring out where these 'platform' imports are defined, so am not sure what's available (e.g. semaphores might be the way to go here). anyone know where i might try looking?

workflow-runtime/src/iosTest/kotlin/com/squareup/workflow1/internal/ThreadLocalTest.kt
steve-the-edwards: yup, config block.

workflow-runtime/src/iosTest/kotlin/com/squareup/workflow1/internal/ThreadLocalTest.kt
steve-the-edwards: `platform` is just the prefix used for what is bundled by default as a prebuilt framework with the kotlin native interop.

So this is the `Foundation` framework accessed through those prebuilts.  (https://www.jetbrains.com/help/kotlin-multiplatform-dev/multiplatform-ios-dependencies.html)

There are a bunch of others available, e.g.:
image


Though your point about `Semaphore` does make me think - why don't we just use the `kotlinx.coroutines` `Semaphore` or `Mutex` that are already in common kotlin code shared on platforms.

I looked those up and they use `kotlinx.atomicfu` for synchronization - e.g., https://github.com/Kotlin/kotlinx-atomicfu?tab=readme-ov-file#locks for Locks.

I think that's likely better tested and supported tthen rolling our own?

Was there another reason to roll our own Zach?

workflow-runtime/src/jvmTest/kotlin/com/squareup/workflow1/StressTestHelpers.kt
zach-klippenstein: After August 1 there's gonna be so many more "ran" in the codebases :P

zach-klippenstein: Bonus points if one of the last things you do is write a lint check that catches this.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/RealRenderContext.kt
zach-klippenstein: I can add another comment to `freeze` saying that's a requirement, but yea we only call it from one place.

workflow-runtime/src/jvmTest/kotlin/com/squareup/workflow1/WorkflowRuntimeMultithreadingStressTest.kt
zach-klippenstein: It's a generic helper on `Closeable`: https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.io/use.html

This thread pool is such, and the [kdoc is extremely insistent](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/new-fixed-thread-pool-context.html) about the fact that there are resources owned by this dispatcher and it must be closed to avoid leaks.

workflow-runtime/src/iosTest/kotlin/com/squareup/workflow1/internal/ThreadLocalTest.kt
zach-klippenstein: Equivalent to this:
```kotlin
val threadStartedLatch = NSLock()
threadStartedLatch.lock()
```

workflow-runtime/src/iosTest/kotlin/com/squareup/workflow1/internal/ThreadLocalTest.kt
zach-klippenstein: Should we pair on this? Anything in Cocoa should be available I think, but I'm not sure what the exact rules are. I will also make this test fail to ensure that it's actually running, although when I run it from the IDE it says it runs and passes.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/RealRenderContext.kt
zach-klippenstein: And to @steve-the-edwards reply to this comment (which doesn't show up in the thread for some reason, fucking github), it doesn't matter if we call it on the main thread or not, or even (to Ray's question) if they're called only from render, we just need to make sure freeze and unfreeze are always called symmetrically from the same thread (i.e. when one thread unfreezes, the same thread later freezes).

If arbitrary RenderContext methods are called from various threads, and that code isn't already thread-safe, then it's already broken since (1) `frozen` is not a thread synchronization mechanism and (2) even if it were, we mutate internal data structures outside of the unfrozen section. `actionSink.send` is thread-safe since it just sends to a channel. But i think we were catching that case with the default value of `frozen` before, which we're not now, so I need to rethink that. Maybe a separate flag for whether sending to a sink is allowed.

workflow-runtime/src/iosTest/kotlin/com/squareup/workflow1/internal/ThreadLocalTest.kt
zach-klippenstein: This isn't coroutines code, we need actual thread-level synchronization here. I will try using atomicfu but it was a headache last time I tried, it's a compiler plugin so it's a bit more complex but maybe they've ironed out the kinks.

Also AFAIK atomicfu doesn't have anything like `ThreadLocal`, so we'll have to roll our own for that anyway — I think. @jamieQ from what I found, Cocoa doesn't have `ThreadLocal` either, just this `threadDictionary` thing, but I would be very happy to be wrong about this.

steve-the-edwards: good poitn about not beign able to use the coroutine synchronization 🤦🏻 . 

But yes, something lower level like `atomicfu` might be good. if not, and you come up with better `actual`s with @jamieQ then that's great!

zach-klippenstein: If there's no native `ThreadLocal`, then TBH i'm not too worried about using (the correct) explicit Apple type for a lock over atomicfu - a lock is a lock, ultimately it's not that complicated. The tests here are for the `ThreadLocal` implementation, I didn't write (and don't intend to) write any tests for the lock expect/actuals.

I looked at Compose and they use every one of these techniques – KMP uses atomicfu SynchronizedObject, but there's unix-specific code in AOSP that even manually builds a lock type on atomics. I really don't want to do that, ever.

workflow-runtime/src/jvmTest/kotlin/com/squareup/workflow1/StressTestHelpers.kt
rjrjr: Oh my God.

Oh no.

zach-klippenstein: Done.

workflow-runtime/src/iosTest/kotlin/com/squareup/workflow1/internal/ThreadLocalTest.kt
zach-klippenstein: We paired and decided to use `NSCondition` for the tests, which is actually designed for exactly what I was trying to shoehorn `NSLock` into doing. Going to keep using `NSLock` for the `actual Lock` since in that case it's being used correctly.

PR #1376, opened 2025-07-21
Update API Design Link Syntax in README File

theck13:
Update the [***API Design*** link](https://github.com/square/workflow-kotlin/blob/main/workflow-ui/compose/README.md#api-design) syntax in the [`workflow-ui/compose/README.md` file](https://github.com/square/workflow-kotlin/blob/main/workflow-ui/compose/README.md) to link to the ***API Design*** section.


rjrjr:
Derp! Thanks.

PR #1379, opened 2025-07-23
Remove Compose Dependency from Android Runtime

steve-the-edwards:


workflow-runtime-android/src/main/java/com/squareup/workflow1/android/AndroidRenderWorkflow.kt
```diff
@@ -141,10 +140,12 @@ public fun <OutputT, RenderingT> renderWorkflowIn(
  * in any workflows, after the initial render pass, will be handled by this scope, and cancelling
  * this scope will cancel the workflow runtime and any running workers. Note that any dispatcher
  * in this scope will _not_ be used to execute the very first render pass (which happens
- * synchronously).
- * Also note that if there is no [CoroutineDispatcher] in this scope then Compose UI's
- * [AndroidUiDispatcher.Main] will be used as this is the most performant Android UI dispatcher
- * for Workflow.
+ * synchronously). We recommend using a `CoroutineDispatcher` in this scope's `CoroutineContext`
+ * that will ensure that all dispatched coroutines are run before the next Choreographer frame.
+ * Compose UI's `AndroidUiDispatcher.Main` provides this behavior in a performant way!
```
rjrjr, line 145: Should update all the compose samples to demonstrate.

rjrjr: * Could also mention that compose apps can use `Workflow.renderAsState` to sidestep thinking about this. Too much?
* Did we ever extend `renderAsState` to allow a config to be passed in?

workflow-runtime-android/src/main/java/com/squareup/workflow1/android/AndroidRenderWorkflow.kt
```diff
@@ -291,16 +293,9 @@ public fun <PropsT, OutputT, RenderingT> renderWorkflowIn(
 ): StateFlow<RenderingT> {
   val restoredSnap = savedStateHandle?.get<PickledTreesnapshot>(KEY)?.snapshot
 
-  // Add in Compose's AndroidUiDispatcher.Main by default if none is specified.
```
rjrjr, line 294: Would it be too magical to use reflection to look for `AndroidUiDispatcher.Main` and default to it if available?

rjrjr (approved):

workflow-runtime-android/src/main/java/com/squareup/workflow1/android/AndroidRenderWorkflow.kt
steve-the-edwards: That's a good point about the samples!

steve-the-edwards: > Did we ever extend renderAsState to allow a config to be passed in?

I don't think so!

workflow-runtime-android/src/main/java/com/squareup/workflow1/android/AndroidRenderWorkflow.kt
steve-the-edwards: I think probably?

workflow-runtime-android/src/main/java/com/squareup/workflow1/android/AndroidRenderWorkflow.kt
steve-the-edwards: > Did we ever extend renderAsState to allow a config to be passed in?


Turns out we did. It takes the scope by default from `rememberCoroutineScope()` which will take it from Compose which will likely be `AndroidUiDispatcher.Main` on Android.

PR #1382, opened 2025-07-24
Improve visualizer usability

wenli-cai:
Currently 
1. Color code each Node in this frame based on its events, i.e. it was a new node, its states were changed, nothing changed but a child was changed. Also provides a color legend on hover of each node.
2. Improve node detail panel UI
3. Change tree UI to hide unaffected children
4. Allow file dump after live tracing
5. Show text diff for nodes' states
6. Allow device selection for live mode
7. Change selector style for selecting frames


workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/DiffUtils.kt
```diff
@@ -0,0 +1,225 @@
...
+
+  fun buildStringWithStyle(
+    style: SpanStyle,
+    text: String,
+    builder: AnnotatedString.Builder
+  ) {
+    builder.pushStyle(style)
+    builder.append(text)
+    builder.pop()
+  }
```
wenli-cai: Having a helper function like this makes it very easy to understand what these three lines are doing, but having to list out the arguments (line 38 vs 43) ends up taking 5 lines instead of 3. Is this a worth it trade off?

workflow-core/dependencies/jvmMainRuntimeClasspath.txt
```diff
@@ -1,11 +1,11 @@
...
-org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm:1.7.3
-org.jetbrains.kotlinx:kotlinx-coroutines-core:1.7.3
+org.jetbrains.kotlin:kotlin-bom:2.2.0
+org.jetbrains.kotlin:kotlin-stdlib-common:2.2.0
+org.jetbrains.kotlin:kotlin-stdlib-jdk7:2.2.0
+org.jetbrains.kotlin:kotlin-stdlib-jdk8:2.2.0
+org.jetbrains.kotlin:kotlin-stdlib:2.2.0
+org.jetbrains.kotlinx:kotlinx-coroutines-bom:1.9.0
+org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm:1.9.0
+org.jetbrains.kotlinx:kotlinx-coroutines-core:1.9.0
```
zach-klippenstein: I don't think we can update these dependencies to 2.2.0 until Register is on it. At any rate, it's better to do major dependency updates separately from feature changes.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
```diff
@@ -47,47 +63,78 @@ internal fun App(
   }
 
   Box(
-    modifier = modifier
+    modifier = modifier.onGloballyPositioned {
+      appWindowSize = it.size
+    }
```
zach-klippenstein: When you only need the size, use `onSizeChanged`. `onGloballyPositioned` is very expensive. More [here](https://www.notion.so/square-seller/Chronicle-Edition-4-15470293beed80e8a104c14a25bf06d9?source=copy_link#18270293beed80268132ebab4a04725d).

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
```diff
@@ -31,14 +41,20 @@ import io.github.vinceglb.filekit.PlatformFile
...
   modifier: Modifier = Modifier
 ) {
+  var appWindowSize by remember { mutableStateOf(IntSize(0, 0)) }
   var selectedNode by remember { mutableStateOf<NodeUpdate?>(null) }
-  val workflowFrames = remember { mutableStateListOf<Node>() }
+  var frameSize by remember { mutableIntStateOf(0) }
+  var rawRenderPass by remember { mutableStateOf("") }
   var frameIndex by remember { mutableIntStateOf(0) }
   val sandboxState = remember { SandboxState() }
+  val nodeLocations = remember { mutableListOf<SnapshotStateMap<Node, Offset>>() }
```
zach-klippenstein, line 50: It's usually a bad practice to use `mutableListOf` in composition, should this be `mutableStateListOf`?

zach-klippenstein: `nodeLocations` is read in composition, so it needs to be snapshot state backed. 
```suggestion
  val nodeLocations = remember { mutableStateListOf<SnapshotStateMap<Node, Offset>>() }
```

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
```diff
@@ -47,47 +63,78 @@ internal fun App(
...
-        .align(Alignment.TopEnd)
-    )
+        .align(Alignment.TopCenter)
+        .padding(top = 8.dp),
+      verticalArrangement = Arrangement.spacedBy(8.dp),
+      horizontalAlignment = Alignment.CenterHorizontally
+    ) {
+      if (active) {
+        // Since we can jump from frame to frame, we fill in the map during each recomposition
+        if (nodeLocations.getOrNull(frameInd) == null) {
```
zach-klippenstein, line 115:
```suggestion
        if (nodeLocations[frameInd] == null) {
```

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
```diff
@@ -47,47 +63,78 @@ internal fun App(
...
+          // frameSize has not been updated yet, so on the first frame, frameSize = nodeLocations.size = 0,
+          // and it will append a new map
+          while (nodeLocations.size <= frameSize) {
+            nodeLocations.add(mutableStateMapOf())
```
zach-klippenstein: Nit: 
```suggestion
            nodeLocations += mutableStateMapOf()
```

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
```diff
@@ -47,47 +63,78 @@ internal fun App(
...
+          }
+        }
+
+        SearchBox(
+          nodes = nodeLocations[frameInd].keys.toList(),
```
zach-klippenstein: Nit: Since `nodeLocations[frameInd]` is used multiple times in this function, would be a bit cleaner to pull it into a val.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
```diff
@@ -47,47 +63,78 @@ internal fun App(
...
+          onSearch = { name ->
+            sandboxState.scale = 1f
+            val node = nodeLocations[frameInd].keys.firstOrNull { it.name == name }
+            val newX = sandboxState.offset.x - nodeLocations[frameInd][node]!!.x + appWindowSize.width / 2
```
zach-klippenstein: Nit: `.getValue(key)` is a bit more idiomatic than `[key]!!`
```suggestion
            val newX = sandboxState.offset.x - nodeLocations[frameInd].getValue(node).x + appWindowSize.width / 2
```

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/model/Node.kt
```diff
@@ -31,6 +31,20 @@ internal data class Node(
   override fun hashCode(): Int {
     return id.hashCode()
   }
+
+  companion object {
+    fun getNodeFields(): List<String> {
+      return listOf("Props", "State")
+    }
```
zach-klippenstein: This can just be a val, and i'd just statically allocate it.
```suggestion
    val nodeFields: List<String> = listOf("Props", "State")
```

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/model/Node.kt
```diff
@@ -31,6 +31,20 @@ internal data class Node(
...
+
+    fun getNodeData(node: Node, field: String): String {
```
zach-klippenstein: It would be more idiomatic for this to be an extension fun on `Node`.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt
```diff
@@ -53,33 +70,208 @@ internal fun DrawTree(
...
-      isAffected,
-      isExpanded,
-      onNodeSelect,
-      onExpandToggle = { expandedNodes[node.id] = !expandedNodes[node.id]!! }
+      node = node,
+      nodePast = previousFrameNode,
+      isAffected = isAffected,
+      isExpanded = isExpanded,
+      onNodeSelect = onNodeSelect,
+      onExpandToggle = { expandedNodes[node.id] = !expandedNodes[node.id]!! },
```
zach-klippenstein:
```suggestion
      onExpandToggle = { expandedNodes[node.id] = !expandedNodes.getValue(node.id) },
```

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt
```diff
@@ -53,33 +70,208 @@ internal fun DrawTree(
...
+  previousFrameNode: Node?,
+  affectedNodes: Set<Node>,
+  expandedNodes: MutableMap<String, Boolean>,
+  onNodeSelect: (NodeUpdate) -> Unit,
+  storeNodeLocation: (Node, Offset) -> Unit
+) {
+  if (children.isEmpty()) return
+
+  val groupKey = "${node.id}_unaffected_group"
+  LaunchedEffect(Unit) {
```
zach-klippenstein: Nit: This doesn't need a coroutine, it would be slightly more efficient to use `DisposableEffect`.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt
```diff
@@ -53,33 +70,208 @@ internal fun DrawTree(
...
       ) {
-        /*
-        We pair up the current node's children with previous frame's children.
-        In the edge case that the current frame has additional children compared to the previous
-        frame, we replace with null and will check before next recursive call.
-         */
-        node.children.forEach { (index, childNode) ->
-          val prevChildNode = previousNode?.children?.get(index)
+        nestedChildren.forEach { childNode ->
           DrawTree(
```
zach-klippenstein, line 271: Can these lists (`group` and `nestedChildren`) partially change contents or order between recompositions?

If so, it would be good to wrap the `forEach` bodies in a `key() { … }` function and pass some kind of identifier for the node.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt
```diff
@@ -94,35 +286,88 @@ internal fun DrawTree(
...
+  val isHovered by interactionSource.collectIsHoveredAsState()
+
+  val nodeUpdate = NodeUpdate.create(
+    current = node,
+    past = nodePast,
+    isAffected = isAffected
+  )
+
+  Box {
+    Box(
```
zach-klippenstein: It looks like this Box just has a single child (the Column). Just move the modifier to the column directly and append a `.wrapContentSize()`.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt
```diff
@@ -94,35 +286,88 @@ internal fun DrawTree(
...
+
+/**
+ * A tooltip that appears on hover showing the node state type
+ */
+@Composable
+private fun NodeTooltip(
+  nodeState: NodeState,
+  modifier: Modifier = Modifier
+) {
+  Box(
```
zach-klippenstein: Another single-child Box. Just pass the modifier to `Text` directly with a `wrapContentSize()`.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/control/DisplayDevices.kt
```diff
@@ -0,0 +1,61 @@
...
+      Text(
+        text = "No device available. Boot up a new device and restart the visualizer",
+        modifier = Modifier.align(Alignment.Center)
+      )
+      return@Box
+    }
+
+    val emulatorRegex = Regex("""\bemulator-\d+\b""")
+    Column {
+      devices.forEach { device ->
```
zach-klippenstein, line 45: Another `forEach`, does it need keys?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/control/FileDump.kt
```diff
@@ -0,0 +1,61 @@
...
+import java.time.LocalDateTime
+import java.time.format.DateTimeFormatter
+
+@Composable
+internal fun FileDump(
+  trace: String,
+  modifier: Modifier = Modifier
+) {
+  var clicked by remember { mutableStateOf(false) }
+  Button(
```
zach-klippenstein, line 30: Should this button be disabled when `clicked == true`?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/control/SearchBox.kt
```diff
@@ -0,0 +1,83 @@
...
+        }
+      )
+    },
+    colors = SearchBarColors(Color.White, Color.Black),
+    expanded = expanded,
+    onExpandedChange = { expanded = it },
+  ) {
+    val relevantNodes = nodes.filter { it.name.contains(searchText, ignoreCase = true) }
+    Column {
+      relevantNodes.take(5).forEach { node ->
```
zach-klippenstein, line 68: Key?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SandboxBackground.kt
```diff
@@ -41,14 +44,29 @@ internal fun SandboxBackground(
         awaitEachGesture {
           val event = awaitPointerEvent()
           if (event.type == PointerEventType.Scroll) {
-            val scrollDelta = event.changes.first().scrollDelta.y
+            val pointerInput = event.changes.first()
+            val pointerOffsetToCenter = Offset(
+              // For some reason using 1.5 made zooming more natural than 2
+              x = pointerInput.position.x - appWindowSize.width / (3 / 2),
```
zach-klippenstein, line 50: These `pointerInput`s use parameters from the composable. They should either be put in `rememberUpdatedState` or passed as keys to the modifier so the coroutine is restarted if the params change (e.g. if the window size changes).

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/parser/DiffUtils.kt
```diff
@@ -0,0 +1,222 @@
...
+      )
+    }
+  }
+}
+
+/**
+ * Parses the full diff within Tag.CHANGED to give back a list of operations to perform
+ */
+private fun parseChangedDiff(fullDiff: String): List<Pair<SpanStyle, String>> {
+  val operations: MutableList<Pair<SpanStyle, String>> = mutableListOf()
```
zach-klippenstein: Nit: Can use `buildList`.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/parser/DiffUtils.kt
```diff
@@ -0,0 +1,222 @@
...
+    i++
+  }
+
+  // Just append whatever is left, since there are no trailing commas
+  if (currentField.isNotBlank()) fields += currentField.toString().trim()
+  return fields
+}
+
+private fun extractTypeName(field: String): String {
+  val stateRegex = Regex("""^(\w+)\(""")
```
zach-klippenstein: Nit: Pull this out into a constant so it doesn't have to get recompiled every time this function is called.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/parser/TraceParser.kt
```diff
@@ -25,30 +33,32 @@ import com.squareup.workflow1.traceviewer.ui.DrawTree
...
   modifier: Modifier = Modifier
 ) {
   var isLoading by remember(traceSource) { mutableStateOf(true) }
   var error by remember(traceSource) { mutableStateOf<String?>(null) }
-  val frames = remember { mutableStateListOf<Node>() }
-  val fullTree = remember { mutableStateListOf<Node>() }
-  val affectedNodes = remember { mutableStateListOf<Set<Node>>() }
+  val frames = remember(traceSource) { mutableStateListOf<Node>() }
+  val fullTree = remember(traceSource) { mutableStateListOf<Node>() }
+  val affectedNodes = remember(traceSource) { mutableStateListOf<Set<Node>>() }
```
zach-klippenstein: Nit: Since `traceSource` is used as a key for all this state, it might be cleaner to just wrap this function body in a `key(traceSource) { … }`.

zach-klippenstein (changes requested):

workflow-core/dependencies/jvmMainRuntimeClasspath.txt
wenli-cai: Yeah I missed some things when reverting the dependencies, fixed.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
wenli-cai: Hm, the original goal was that I would only want compositions to track what's happening to each `SnapshotStateMap` since that's where all the locations are stored, and I would just index into the list. Would my current code not work for that?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
wenli-cai: I think using square brackets to index could throw an IndexOutOfBoundsException right? Having `.getOrNull` was for the purpose when we go from frame 0 immediately to frame 10, so we would need to fill in the gap of frames with empty maps.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/parser/DiffUtils.kt
wenli-cai: Done

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/parser/TraceParser.kt
wenli-cai: Done.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/parser/DiffUtils.kt
wenli-cai: Done, cool feature.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
wenli-cai: True, done.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/model/Node.kt
wenli-cai: Done

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/model/Node.kt
wenli-cai: Changed, is this because I'm interacting directly with a Node object?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt
wenli-cai: Fixed.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt
wenli-cai: hm, I don't think so since no new nodes are ever added, and the list could only ever change by switching frames. The only time any of these lists here can change is when a new frame is selected, but that would mean the composable is thrown out and restarted right?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt
wenli-cai: There's also `NodeToolTip`, which shows on the top right of the node box, would it still need to be pulled out?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/control/DisplayDevices.kt
wenli-cai: Done, I think so since a new device could finish booting up when we get to this point in the code.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/control/FileDump.kt
wenli-cai: I think the option to keep writing to disk should be kept open, since they can continuously interact with the app, add more frames, and keep saving more files rather than having to restart the whole process.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/control/SearchBox.kt
wenli-cai: Added

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SandboxBackground.kt
wenli-cai: Like this `.pointerInput(appWindowSize)` on line 41?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/model/NodeUpdate.kt
```diff
@@ -1,12 +1,38 @@
...
+      return NodeUpdate(current, past, state)
+    }
+  }
+}
+
+internal enum class NodeState(val color: Color) {
+  NEW(Color(0x804CAF50)), // green
+  STATE_CHANGED(Color(0xFFE57373)), // red
+  PROPS_CHANGED(Color(0xFFFF8A65)), // orange
+  CHILDREN_CHANGED(Color(0x802196F3)), // blue
```
steve-the-edwards: I think we just call this "RENDERED" or something like that?
When PTR is _on_ then this will represent "Children Changed", but otherwise, that's not always true as all nodes will be rendered.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowInfoPanel.kt
```diff
@@ -55,6 +58,9 @@ internal fun RightInfoPanel(
     IconButton(
       onClick = { panelOpen = !panelOpen },
       modifier = Modifier
+        .size(40.dp)
+        .clip(CircleShape)
+        .background(Color.White)
         .padding(8.dp)
         .size(40.dp)
```
steve-the-edwards: size specified twice? Does order matter?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowInfoPanel.kt
```diff
@@ -98,19 +104,33 @@ private fun NodePanelDetails(
     }
     item {
       Text(
-        text = "Workflow Details",
-        style = MaterialTheme.typography.h6,
-        modifier = Modifier.padding(top = 8.dp, bottom = 8.dp)
+        text = "${node.current.parent} (ID: ${node.current.parentId})",
+        style = MaterialTheme.typography.subtitle2,
+        color = Color.Gray,
+        modifier = Modifier.padding(top = 8.dp)
```
steve-the-edwards, line 109: We should extract constants/styles for these padding values etc.

(If that needs to be done separately/later, that's fine too!)

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/control/DisplayDevices.kt
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    Column {
+      devices.forEach { device ->
+        key(device) {
+          Card(
+            onClick = {
+              emulatorRegex.find(device)?.value?.let { emulator ->
+                onDeviceSelect(emulator)
+              }
```
steve-the-edwards, line 51: DOn't we only want to show this card if we find the emulator? Othrewise its a noop.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/control/FileDump.kt
```diff
@@ -0,0 +1,61 @@
...
+    modifier = modifier.padding(16.dp),
+    shape = CircleShape,
+    colors = buttonColors(Color.Black),
+    onClick = {
+      clicked = true
+      writeToFile(trace)
+    }
+  ) {
+    val text = if (clicked) {
+      "Trace saved to Downloads"
```
steve-the-edwards: would be great to show the filepath here.

steve-the-edwards (approved):
Resolve the Compose questions/comments from @zach-klippenstein and let's change the example .gifs to a dummy app , but otherwise LGTM.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowInfoPanel.kt
wenli-cai: This would all be the kind of stuff that goes in resources? I may not have time to pull out all the styles.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowInfoPanel.kt
wenli-cai: Nope! Removed one now

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/control/FileDump.kt
wenli-cai: Done

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/control/DisplayDevices.kt
wenli-cai: What do you mean by "find" the emulator? I have a early return above if that's what you're asking?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/model/NodeUpdate.kt
wenli-cai: Fixed

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
zach-klippenstein: That's easy to miss, might be nice to add a comment about that intention.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/model/Node.kt
zach-klippenstein: I think it's because the node is the subject of this query operation. Typically that sort of thing is done with an extension function.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/control/FileDump.kt
zach-klippenstein: Then should the text change back to "Save trace to file" after a while, to indicate the button is still functional?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
zach-klippenstein: No, because when you clear (line 79) or put a new entry in the list (line 100), line 114 needs to know about it since the presence of that index in the list might change.

It's fine to keep the `SnapshotStateMap` as well though so you invalidate the map without invalidating the list. Nested state collections are fine.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt
zach-klippenstein: > when a new frame is selected, but that would mean the composable is thrown out and restarted right?

I don't know, it would be if the composable that renders the frame is keyed on the frame. Either way though, sounds like this is fine.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt
zach-klippenstein: If I'm reading it correctly, the structure is:

```
Box {
  Box {
    Column
  }
  Tooltip
}
```

You only need the outer box to layer the column and the tooltip, the inner box only wraps the column and can be eliminated.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowInfoPanel.kt
zach-klippenstein: I wouldn't use Android resources for it, if that's what you mean. Just kotlin constants.

zach-klippenstein (approved):

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
wenli-cai: Fixed

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
wenli-cai: ah okay, that makes a lot of sense. Fixed

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/ui/WorkflowTree.kt
wenli-cai: Ahh okay I misinterpreted what you were saying at first, fixed now.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/App.kt
```diff
@@ -96,13 +149,12 @@ internal fun App(
...
           TraceMode.File(null)
         } else {
-          // TODO: TraceRecorder needs to be able to take in multiple clients if this is the case
           /*
           We set the frame to -1 here since we always increment it during Live mode as the list of
           frames get populated, so we avoid off by one when indexing into the frames.
            */
           frameIndex = -1
-          TraceMode.Live
+          TraceMode.Live()
```
mannydelgado, line 157: If we dont provide the device ID here does it select one for us? was working through and noticed we require deviceId from Live trace

mannydelgado (approved):

PR #1402, opened 2025-07-28
Fix PhoneWindow.superDispatchTouchEvent NullPointerException

tcmulcahy:
PhoneWindow.superDispatchTouchEvent will crash with NullPointerException if DecorView is null: https://cs.android.com/android/platform/superproject/+/android-latest-release:frameworks/base/core/java/com/android/internal/policy/PhoneWindow.java;l=2021

We should only dispatch touch events if DecorView is non-null. Note that we already checking for non-null DecorView below. This commit just expands that check.


tcmulcahy:
This is my first contribution to workflow - I verified this change compiles via `./gradlew -Pworkflow.targets=jvm build`. Please advise if there are other tests I should run.

PR #1406, opened 2025-08-01
Upstreaming Runtime Monitoring and Tracing

steve-the-edwards:
This is the evolution of a class we've long used in production internally. This takes a lot of the bookkeeping that we had there and makes it available to other pluggable `WorkflowRuntimeTracer`s.

There is also an update to the `RuntimeUpdate` type that is clearer about when the runtime loop is finished (regardless of if the rendering is skipped or not).

First commit removes the old `TracingWorkflowInterceptor` and the corresponding `trace-encoder` module.

The second commit adds the `WorkflowRuntimeMonitor` as well as a bunch of reflection-less logging utilities that we've used for some time.


workflow-tracing/src/main/java/com/squareup/workflow1/tracing/Loggable.kt
```diff
@@ -0,0 +1,9 @@
+package com.squareup.workflow1.tracing
+
+/**
+ * Optional interface implemented by workflow `PropsT`, `StateT` `OutputT` classes
+ * to customize their logging names for [WorkflowRuntimeMonitor] output.
+ */
+public interface Loggable {
+  public fun toLogString(): String = toString().wfStripSquarePackage()
```
rjrjr: When thinking about upstreaming this I always felt bad about how `com.squareup.com` is hard coded into `wfStripSquarePackage()`. Wonder if we should expose a static property or something to allow customization. I don't think that should derail this initial commit, just food for thought.

But in the meantime, it does seem sketchy to have a default implementation that relies on a non-public function.

Is this call redundant here anyway? Do we also make downstream calls?

workflow-tracing/src/main/java/com/squareup/workflow1/tracing/RuntimeLoggingUtils.kt
```diff
@@ -0,0 +1,208 @@
...
+import com.squareup.workflow1.WorkflowInterceptor.WorkflowSession
+import com.squareup.workflow1.tracing.WorkflowLoggingConstants.WORKER_ACTION_NAME
+import kotlin.reflect.KClass
+import kotlin.reflect.KType
+
+public object WorkflowLoggingConstants {
+  /**
+   * Used as a key to detect actions originating from the Workflow that implements a worker.
+   */
+  // Keep this up-to-date with the action in com.squareup.workflow1.WorkerWorkflow.
```
rjrjr: Do we have a unit test that will break if we screw that up?

workflow-tracing/src/main/java/com/squareup/workflow1/tracing/RuntimeLoggingUtils.kt
```diff
@@ -0,0 +1,208 @@
...
+}
+
+internal fun KClass<*>.toWfLoggingName(): String {
+  return getWfHumanClassName(this)
+}
+
+/**
+ * Alternative to [toString] used by Workflow logging.
+ */
+internal fun getWfLogString(log: Any?): String {
```
rjrjr: Making this function non-public defeats a major use case. We want feature code implementing `Loggable` to be able to call this on members so that the opt-in scheme works recursively. I'd be inclined to most or all of this file public -- making `Loggable` public without this invites serious footguns.

workflow-tracing/src/main/java/com/squareup/workflow1/tracing/RuntimeUpdateLogLine.kt
```diff
@@ -0,0 +1,131 @@
+package com.squareup.workflow1.tracing
+
+import com.squareup.workflow1.WorkflowOutput
+import com.squareup.workflow1.tracing.ActionAppliedLogLine.WorkflowActionLogType.CASCADE
+import com.squareup.workflow1.tracing.ActionAppliedLogLine.WorkflowActionLogType.RENDERING_CALLBACK
+import com.squareup.workflow1.tracing.ActionAppliedLogLine.WorkflowActionLogType.WORKER_OUTPUT
+
+/**
+ * PLEASE NOTE: these logs lines are turned into strings in production, all the time, and there
```
rjrjr: "log lines"

workflow-tracing/src/main/java/com/squareup/workflow1/tracing/RuntimeTraceContext.kt
```diff
@@ -0,0 +1,40 @@
...
+   * A list of all causes for the current runtime loop processing (it can be multiple in the case
+   * of some optimizations).
+   */
+  public val renderIncomingCauses: MutableList<RenderCause>
+
+  public var previousRenderCause: RenderCause?
+  public var currentRenderCause: RenderCause?
+
+  /**
+   * Add an update into the [RuntimeUpdates] tracked by [WorkflowRuntimeMonitor].
```
rjrjr: - The kdoc on `RuntimeUpdateLogLine` is more informative, should probably say a lot of that here instead / too.
- We can suggest a useful spot to call this from (see below).
- Be good to update the kdoc of `reportNavigation` to draw attention to this, or if dependencies allow it add this to the default `onNavigate` param value.

```kotlin
* Consider calling this from the `onNavigate` function you provide to
* `reportNavigation`, e.g.
*
*         val renderings: Flow<Screen> by lazy {
*           renderWorkflowIn(
*             workflow = RootNavigationWorkflow,
*             scope = viewModelScope,
*             savedStateHandle = savedState,
*             runtimeConfig = RuntimeConfigOptions.ALL
*           ).reportNavigation {
*             runtimeMonitor.addRuntimeUpdate(
*               UiUpdateLogLine(getWfLogString(it))
*             )
*           }
```

workflow-tracing/src/main/java/com/squareup/workflow1/tracing/WorkflowRuntimeTracer.kt
```diff
@@ -0,0 +1,106 @@
...
+  public open fun onRuntimeUpdateEnhanced(
+    runtimeUpdate: RuntimeUpdate,
+    currentActionHandlingChangedState: Boolean,
+    configSnapshot: ConfigSnapshot,
+  ) = Unit
+
+  /** SECTION: [WorkflowInterceptor] overrides. */
+
+  /**
+   * [WorkflowRuntimeTracer]s should not override this method, using [onWorkflowSessionStarted]
```
rjrjr: It can't b/c this is `final`, so "should not" is an odd thing to say.

workflow-tracing/build.gradle.kts
```diff
@@ -1,18 +1,29 @@
...
 
+android {
+  namespace = "com.squareup.workflow1.tracing"
+}
+
 dependencies {
+  api(libs.androidx.collection)
   api(libs.kotlin.jdk8)
   api(libs.kotlinx.coroutines.core)
+  api(libs.squareup.papa)
```
rjrjr: Is this direct tie to `papa` going to cause dependency hell down the road? Can we introduce a separate module that ties to `papa`, similar to what we did for `radiography`?

workflow-tracing/src/main/java/com/squareup/workflow1/tracing/Loggable.kt
steve-the-edwards: we can remove the default.

workflow-tracing/src/main/java/com/squareup/workflow1/tracing/RuntimeLoggingUtils.kt
steve-the-edwards: I can make it public, but my general idea was that you would implement your own `getLogString()` to handle more cases than what we can include here.

workflow-tracing/src/main/java/com/squareup/workflow1/tracing/WorkflowRuntimeTracer.kt
steve-the-edwards: I guess i was just diary-ing *my thoughts* rather than making clear kdoc. Will change.

workflow-tracing/build.gradle.kts
steve-the-edwards: sure, we could

workflow-tracing/src/main/java/com/squareup/workflow1/tracing/RuntimeLoggingUtils.kt
rjrjr: Better to err on the side of privacy probably. Easier to open up later.

workflow-tracing/src/main/java/com/squareup/workflow1/tracing/Loggable.kt
rjrjr: Wise, it's a terrible default anyway. Shouldn't be making it so easy to use `toString()` by accident, defeats the purpose.

workflow-tracing/src/main/java/com/squareup/workflow1/tracing/RuntimeLoggingUtils.kt
steve-the-edwards: No. but we should. I will add.

steve-the-edwards: Did one better and just made it a constant in Worker.kt shared in both places

workflow-tracing/src/main/java/com/squareup/workflow1/tracing/RuntimeLoggingUtils.kt
steve-the-edwards: the `getLogString` implementation is less broad than ours in register, because there are a lot of internal types that we don't want to lift here that we handle.

My idea was that we could keep using our Loggable as `SquareLoggable` for those, if desired.

workflow-tracing/src/main/java/com/squareup/workflow1/tracing/Loggable.kt
steve-the-edwards: > Wonder if we should expose a static property or something to allow customization. I don't think that should derail this initial commit, just food for thought.


I thought about this too, but its a bit confusing to add and is a lot of power to munge the logs that i landed on just keeping it simple.

workflow-tracing/src/main/java/com/squareup/workflow1/tracing/RuntimeLoggingUtils.kt
rjrjr: That's a good idea. If it feels good once you've tried it out could suggest it in the kdoc here.

rjrjr: I wonder how many of our internal types we could get out of the big if / else if they implemented `Unwrappable`?

workflow-tracing/src/main/java/com/squareup/workflow1/tracing/RuntimeTraceContext.kt
steve-the-edwards: btw, TIL about `NavigationMonitor`!

rjrjr: Yeah, I snuck that out when updating the tutorials. I'm proud to report that we use it in prod, it's solid.

workflow-tracing/src/main/java/com/squareup/workflow1/tracing/RuntimeUpdateLogLine.kt
steve-the-edwards: Done.

workflow-tracing/build.gradle.kts
steve-the-edwards: Done. separate out into workflow-tracing-papa

workflow-tracing-papa/src/main/java/com/squareup/workflow1/tracing/papa/WorkflowPapaTracer.kt
```diff
@@ -0,0 +1,489 @@
...
+import com.squareup.workflow1.tracing.workerKey
+import kotlinx.coroutines.CoroutineScope
+import kotlin.reflect.KType
+
+/**
+ * [WorkflowRuntimeTracer] plugin to add [SafeTraceInterface] traces.
+ * By default this uses [PapaSafeTrace] which will use [androidx.tracing.Trace] calls that
+ * will be received by the system and included in Perfetto traces.
+ */
+class WorkflowPapaTracer(
```
mannydelgado, line 34: This has the potential to be more of a generalized Tracer, where anything that has extended SafeTraceInterface can work here.

SafeTraceInterface has the workings of what could be a generalized approach with its current functions.
Might need renaming but that could be done in the future.

steve-the-edwards: ya, we can iterate. not many users (only us) of this for now.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt
```diff
@@ -181,13 +181,14 @@ public interface WorkflowInterceptor {
...
-  public data class RenderingProduced<R>(
-    public val renderingAndSnapshot: RenderingAndSnapshot<R>
-  ) : RuntimeUpdate
+  public data object RenderingProduced : RuntimeUpdate
+
+  /**
+   * The runtime has finished its work and is stable again - either skipping rendering because
+   * of no change ([RenderPassSkipped]), or having passed a new rendering ([RenderingProduced]).
+   */
+  public data object RuntimeLoopTick : RuntimeUpdate
```
zach-klippenstein, line 191: Nit: The name could be more descriptive, maybe `RuntimeSettled`?

workflow-tracing/src/main/java/com/squareup/workflow1/tracing/ConfigSnapshot.kt
```diff
@@ -0,0 +1,25 @@
...
+ * Snapshot of the current [RuntimeConfig]
+ */
+@OptIn(WorkflowExperimentalRuntime::class)
+public class ConfigSnapshot(config: RuntimeConfig) {
+  val shortCircuitConfig = config.contains(RENDER_ONLY_WHEN_STATE_CHANGES)
+  val csrConfig = config.contains(CONFLATE_STALE_RENDERINGS)
+  val ptrConfig = config.contains(PARTIAL_TREE_RENDERING)
+  val deaConfig = config.contains(DRAIN_EXCLUSIVE_ACTIONS)
+  val sehConfig = config.contains(STABLE_EVENT_HANDLERS)
+  val wsdConfig = config.contains(WORK_STEALING_DISPATCHER)
```
zach-klippenstein, line 22: These properties are public API, they should probably not use abbreviations.

zach-klippenstein (approved):

PR #1407, opened 2025-08-12
Follow-up cleanups from Runtime Tracing Upstream

steve-the-edwards:
Cleanups from some of @zach-klippenstein's comments on https://github.com/square/workflow-kotlin/pull/1406

### DIFF
--- a/workflow-tracing/src/main/java/com/squareup/workflow1/tracing/ConfigSnapshot.kt
+++ b/workflow-tracing/src/main/java/com/squareup/workflow1/tracing/ConfigSnapshot.kt
@@ -14,12 +14,32 @@ import com.squareup.workflow1.WorkflowExperimentalRuntime
  */
 @OptIn(WorkflowExperimentalRuntime::class)
 public class ConfigSnapshot(config: RuntimeConfig) {
-  val shortCircuitConfig = config.contains(RENDER_ONLY_WHEN_STATE_CHANGES)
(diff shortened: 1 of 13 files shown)


workflow-tracing/src/main/java/com/squareup/workflow1/tracing/ConfigSnapshot.kt
```diff
@@ -14,12 +14,32 @@ import com.squareup.workflow1.WorkflowExperimentalRuntime
...
+        append("PTR, ")
+      }
+      if (config.contains(DRAIN_EXCLUSIVE_ACTIONS)) {
+        append("DEA, ")
+      }
+      if (config.contains(STABLE_EVENT_HANDLERS)) {
+        append("SEH, ")
+      }
+      if (config.contains(WORK_STEALING_DISPATCHER)) {
+        append("SEH, ")
```
mannydelgado: Should this be SEH as well? if so should we combine STABLE || WORK_STEALING?

steve-the-edwards: oops. Should be WSD. Good catch!

steve-the-edwards: Fixed.

PR #1408, opened 2025-08-13
Make Logging Utils Public

steve-the-edwards:
We upstreamed these utils from internal repo, but they are still used there and we don't want to duplicate.


workflow-tracing/src/main/java/com/squareup/workflow1/tracing/RuntimeLoggingUtils.kt
```diff
@@ -9,7 +9,7 @@ import com.squareup.workflow1.WorkflowInterceptor.WorkflowSession
 import kotlin.reflect.KClass
 import kotlin.reflect.KType
 
-internal fun WorkflowSession.toWfLoggingName(): String {
+public fun WorkflowSession.toWfLoggingName(): String {
```
mannydelgado, line 16: Since these are public should we include some sort of kdoc or example of an expected output.

workflow-tracing/src/main/java/com/squareup/workflow1/tracing/RuntimeLoggingUtils.kt
```diff
@@ -18,7 +18,7 @@ internal fun WorkflowSession.toWfLoggingName(): String {
   }
 }
 
-internal fun WorkflowIdentifier.toWfLoggingName(): String {
+public fun WorkflowIdentifier.toWfLoggingName(): String {
```
mannydelgado, line 29: same.

mannydelgado (approved):
nit: small items.

workflow-tracing/src/main/java/com/squareup/workflow1/tracing/RuntimeLoggingUtils.kt
steve-the-edwards: Good point. Will do.

PR #1412, opened 2025-08-20
Link README.md Swift mention

rjrjr:
(But actually I'm just checking to see if I can still put up PRs here. Not so much.)


steve-the-edwards:
https://github.com/square/workflow-kotlin/issues/1413 needed for the fixup bot to be able to correctly auth

PR #1416, opened 2025-08-27
Use actions/setup-gradle

steve-the-edwards:


steve-the-edwards:
So much for a simple migration to this new action.

steve-the-edwards:
Closed as obselete. Put on hold for now.

PR #1418, opened 2025-08-28
wip trying to implement jwilson's idea in workflow

zach-klippenstein:
Very rough sketch, not complete.


swankjesse:
This is exciting!

In case you missed it, the toy version of this from the talk is here:
https://github.com/swankjesse/coroutines-party-tricks/blob/main/code/lib/src/main/kotlin/com/publicobject/tricks/navigation/RealDisplay.kt

samples/containers/thingy/src/main/java/com/squareup/sample/thingy/BackStackWorkflow.kt
```diff
@@ -0,0 +1,325 @@
...
+   *       MyScreen(message = prop)
+   *     }
+   *   }
+   * }
+   * ```
+   * This example shows the props received from the parent to the user via `MyScreen`. Every time
+   * the parent passes a new props, the `showScreen` call is cancelled and called again with the
+   * new props, replacing the old instance of `MyScreen` in the backstack with a new one. Since
+   * both instances of `MyScreen` are compatible, this is not a navigation event but just updates
+   * the `MyScreen` view factory.
```
swankjesse, line 189: So cool

PR #1419, opened 2025-08-28
Clean up Test Specification in our Github Actions

steve-the-edwards:
We were using `jvmTest` previously for the tests where we specify the runtime via the environment arg, but the `jvmTest` task is only for KMP modules. Those KMP modules (namely `workflow-runtime`) already accounted for multiple runtimes using `Burst`, so we were just running the same tests multiple times (it was not even using the `JvmTestRuntimeConfigTools` to pull the arg).

We should be running the regular java/android `test` task with the runtime specification, then the arg will be picked up.

Also some other fixes for the proper targets - e.g., `iosSimulatorArm64` and `jsBrowserTest`.

AND fixes the `failure-path-upload` usage so that we upload the test report successfully (the html one) when there is a failure (whether JVM or Android tests). (I've tested this when the tests were failing that I fixed).

I had to modify two tests to check for particular runtime configs that are incompatible with the test structure, but everything else is working. yay!

### DIFF
--- a/.github/actions/gradle-task/action.yml
+++ b/.github/actions/gradle-task/action.yml
@@ -23,6 +23,9 @@ inputs:
   failure-path-upload:
     description: 'The relative path to a desired log for upload if the task fails.'
     default: 'null'
+  failure-upload-name:
+    description: 'The name for the upload of failure reports.'
+    default: 'specified-upload'
 
 runs:
   using: 'composite'
@@ -149,6 +152,6 @@ runs:
       if: failure() && inputs.failure-path-upload != 'null'
       uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
       with:
-        name: specified-upload
+        name: ${{inputs.failure-upload-name}}
         path: ${{github.workspace}}/${{inputs.failure-path-upload}}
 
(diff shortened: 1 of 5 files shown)


.github/actions/gradle-task/action.yml
```diff
@@ -149,6 +152,6 @@ runs:
       if: failure() && inputs.failure-path-upload != 'null'
       uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
       with:
-        name: specified-upload
+        name: ${{inputs.failure-upload-name}}
```
steve-the-edwards, line 155: @zach-klippenstein I also fixed test report upload on failure - so you can download the typical `.html` report and see the reasons for all your different test failures in CI.

.github/workflows/kotlin.yml
```diff
@@ -201,49 +217,54 @@ jobs:
             --continue
           restore-cache-key: build-logic
           write-cache-key: main-build-artifacts
+          failure-path-upload: '**/build/reports/tests/*[tT]est'
+          failure-upload-name: 'unit-tests-report'
```
steve-the-edwards, line 221: Here is where we specify `failure-path-upload` (what to upload when there is a failure on the job).

zach-klippenstein (approved):

PR #1421, opened 2025-09-05
Use telephoto for panning and zooming

saket:
This PR removes all custom pointer handlers in `SandboxBackground()` so that `Modifier.zoomable()` can take care of both panning and zooming. 

One change to call out is that zooming with the mouse scroll now behaves differently. Previously, you were using scroll alone, but telephoto uses `Option` + scroll by default to match the behavior of other macOS apps. If you prefer the old behavior, you can change this if you want by writing a custom [HardwareShortcutsSpec](https://github.com/saket/telephoto/blob/2c4a5fbf7124eb890987f4ef0ee1ea46500fe9bd/zoomable/src/commonMain/kotlin/me/saket/telephoto/zoomable/ZoomableState.kt#L42). 

It might be a good idea to display all shortcuts somewhere in the UI. Here's what telephoto uses: https://saket.github.io/telephoto/zoomable/#keyboard-shortcuts

Your app actually helped uncover a few bugs in telephoto, which historically hasn't been well optimized for non-Android targets. Some worth mentioning:

- Zoom using keyboard and mouse shortcuts is currently broken because telephoto is [miscalculating zoom factors](https://github.com/saket/telephoto/blob/2c4a5fbf7124eb890987f4ef0ee1ea46500fe9bd/zoomable/src/commonMain/kotlin/me/saket/telephoto/zoomable/RealZoomableState.kt#L457). I'll push a fix soon.

- Overzoom doesn't look great right now. I'll disable it for hardware shortcuts in a future release.

### DIFF
--- a/workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/FileTraceViewer.kt
+++ b/workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/FileTraceViewer.kt
@@ -2,10 +2,10 @@ package com.squareup.workflow1.traceviewer
 
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.padding
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.key
 import androidx.compose.runtime.mutableIntStateOf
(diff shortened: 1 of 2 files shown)


workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/FileTraceViewer.kt
```diff
@@ -40,53 +43,52 @@ internal fun TraceViewerWindow(
...
-    snapshotFlow { frameIndex }.collect {
-      sandboxState.reset()
-    }
-  }
+  Box(modifier) {
+    val zoomableState = key(frameInd) {
+      rememberZoomableState(
+        zoomSpec = ZoomSpec(maxZoomFactor = 1f)
+      ).also {
+        it.contentScale = ContentScale.Fit
```
saket, line 62: It might be worth exploring `ContentScale.Crop` for large traces

PR #1422, opened 2025-09-09
Use telephoto for panning and zooming

japplin:
Re-opened: https://github.com/square/workflow-kotlin/pull/1421

### DIFF
--- a/workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/FileTraceViewer.kt
+++ b/workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/FileTraceViewer.kt
@@ -2,10 +2,10 @@ package com.squareup.workflow1.traceviewer
 
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.padding
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.key
 import androidx.compose.runtime.mutableIntStateOf
@@ -13,14 +13,13 @@ import androidx.compose.runtime.mutableStateListOf
 import androidx.compose.runtime.mutableStateMapOf
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
+import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.runtime.setValue
-import androidx.compose.runtime.snapshotFlow
 import androidx.compose.runtime.snapshots.SnapshotStateMap
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.geometry.Offset
-import androidx.compose.ui.layout.onSizeChanged
-import androidx.compose.ui.unit.IntSize
+import androidx.compose.ui.layout.ContentScale
 import androidx.compose.ui.unit.dp
 import com.squareup.workflow1.traceviewer.model.Node
 import com.squareup.workflow1.traceviewer.model.NodeUpdate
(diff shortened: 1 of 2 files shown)


workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/FileTraceViewer.kt
```diff
@@ -108,15 +110,19 @@ internal fun TraceViewerWindow(
...
             val node = frameNodeLocations.keys.first { it.name == name }
-            val newX = (sandboxState.offset.x - frameNodeLocations.getValue(node).x
-              + appWindowSize.width / 2)
-            val newY = (sandboxState.offset.y - frameNodeLocations.getValue(node).y
-              + appWindowSize.height / 2)
-            sandboxState.offset = Offset(x = newX, y = newY)
+            scope.launch {
+              // TODO: this doesn't work super well.
+              //  Probably because of https://github.com/saket/telephoto/issues/135?
+              zoomableState.zoomTo(
```
japplin, line 121: From local testing this isn't adjusting the pan at all when selecting a node.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/FileTraceViewer.kt
```diff
@@ -40,53 +43,52 @@ internal fun TraceViewerWindow(
...
+    val zoomableState = key(frameInd) {
+      rememberZoomableState(
+        zoomSpec = ZoomSpec(maxZoomFactor = 1f)
+      ).also {
+        it.contentScale = ContentScale.Fit
 
-  Box(
-    modifier = modifier.onSizeChanged {
-      appWindowSize = it
+        // TODO: do we want to draw behind the search bar?
```
japplin, line 64: I think drawing behind the search looks correct, which is what it's doing already, if we didn't allow it there would be a jarring cutoff.

japplin (commented):
The zoom behaviour and locking of content to the bounds of the window feels much better now as does the double to click to zoom. The scroll wheel zooming is buggy and honestly I don't think needed at all given how well double click to zoom works.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/FileTraceViewer.kt
saket: In that case, do you want to synchronize the content padding below with the search box's height?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/FileTraceViewer.kt
saket: This should be fixed by https://github.com/saket/telephoto/commit/68fcaefe0308514febf65ac4d4393f58059b254d. Do you mind if I update `telephoto` to its latest snapshot version?

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/util/SandboxBackground.kt
```diff
@@ -26,51 +28,29 @@ import me.saket.telephoto.zoomable.zoomable
...
-  sandboxState: SandboxState,
+  zoomableState: ZoomableState,
   modifier: Modifier = Modifier,
   content: @Composable () -> Unit,
 ) {
-  val zoomableState = rememberZoomableState()
-
+  val focusRequester = remember { FocusRequester() }
+  LaunchedEffect(zoomableState) {
+    // Request focus to receive keyboard shortcuts.
```
saket, line 37: This is turning out to be more complex than I expected. The content starts in a focused state but loses focus when the search box is clicked. We'll have to restore focus when a search result is selected. Let me try fixing this.

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/FileTraceViewer.kt
japplin: Nope go for it

workflow-trace-viewer/src/jvmMain/kotlin/com/squareup/workflow1/traceviewer/FileTraceViewer.kt
japplin: I'm not sure, I think it was looking good how it was now.

PR #1423, opened 2025-09-11
Prototype of a stepper composable helper for making backstacks.

zach-klippenstein:


workflow-ui/compose/src/main/java/com/squareup/workflow1/compose/Stepper.kt
```diff
@@ -0,0 +1,218 @@
...
+@Suppress("NOTHING_TO_INLINE")
+@Composable
+public inline fun <R> stepper(
+  noinline content: @Composable Stepper<() -> Unit, R>.() -> R
+): List<R> = stepper(advance = { it() }, content = content)
+
+public interface Stepper<T, R> {
+
+  /** The (possibly empty) stack of steps that came before the current one. */
+  val previousSteps: List<Step<R>>
```
vRallev, line 64: I think you added this one later and it wasn't part of the initial revision and I meant to call out that something like this is needed. Likely, I would even go a step further and argue that `stepper()` should not return a `List<R>`, but rather a type that expresses how the backstack was changed. 

Here is my prototype that Amazon uses in production now: https://github.com/amzn/app-platform/blob/main/recipes/common/impl/src/commonMain/kotlin/software/amazon/app/platform/recipes/backstack/PresenterBackstackScope.kt#L19-L32 (it's not part of the official API, because I wanted get  an idea first how it works out in production). Here it runs in the browser: https://amzn.github.io/app-platform/#web-recipe-app

The problem I wanted to solve is to make it easy for the UI layer to play animations for changes in the backstack, like the cross-slide animation in the demo above. Exposing only a stack makes this more challenging for the UI layer. That's why I introduced the [`BackstackChange`](https://github.com/amzn/app-platform/blob/main/recipes/common/impl/src/commonMain/kotlin/software/amazon/app/platform/recipes/backstack/PresenterBackstackScope.kt#L35) type.

workflow-ui/compose/src/main/java/com/squareup/workflow1/compose/Stepper.kt
```diff
@@ -0,0 +1,218 @@
...
+   * Pushes a new frame onto the backstack with the current state and then runs [toState].
+   */
+  fun advance(toState: T)
+
+  /**
+   * Pops the last frame off the backstack and restores its state.
+   *
+   * @return False if the stack was empty (i.e. this is a noop).
+   */
+  fun goBack(): Boolean
```
vRallev, line 76: If this is a stack, why not call types and functions accordingly? `push()` and `pop()` feel more natural to me.

PR #1424, opened 2025-10-08
Collect renderings on context specified in WorkflowLayout.take

steve-the-edwards:


workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLayout.kt
```diff
@@ -104,16 +105,15 @@ public class WorkflowLayout(
...
-    val lifecycleDispatcher = lifecycle.coroutineScope.coroutineContext[CoroutineDispatcher.Key]
     // Just like https://medium.com/androiddevelopers/a-safer-way-to-collect-flows-from-android-uis-23080b1f8bda
-    return lifecycle.coroutineScope.launch(contextWithoutDispatcher) {
+    return lifecycle.coroutineScope.launch {
       lifecycle.repeatOnLifecycle(repeatOnLifecycle) {
-        require(coroutineContext[CoroutineDispatcher.Key] == lifecycleDispatcher) {
-          "Collection dispatch should happen on the lifecycle's dispatcher."
+        withContext(collectionContext) {
+          require(Looper.myLooper() == Looper.getMainLooper()) {
+            "Collection dispatch should happen on the main thread!"
```
tcmulcahy: s/should/must

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLayout.kt
```diff
@@ -89,8 +90,8 @@ public class WorkflowLayout(
    * @param [repeatOnLifecycle] the lifecycle state in which renderings should be actively
    * updated. Defaults to STARTED, which is appropriate for Activity and Fragment.
    * @param [collectionContext] additional [CoroutineContext] we want for the coroutine that is
-   * launched to collect the renderings. This should not override the [CoroutineDispatcher][kotlinx.coroutines.CoroutineDispatcher]
-   * but may include some other instrumentation elements.
+   * launched to collect the renderings, can include a different dispatcher - but it should be
```
tcmulcahy: Suggested wording: "can include a different dispatcher - but we verify that it's a main thread dispatcher" And maybe explain why this restriction exists (naively I thought all of this stuff was immutable so it'd be safe to access them from any thread)

workflow-ui/core-android/src/test/java/com/squareup/workflow1/ui/WorkflowLayoutTest.kt
```diff
@@ -92,23 +90,6 @@ internal class WorkflowLayoutTest {
     unoriginal.show(BScreen(), env)
   }
 
-  @Test fun usesLifecycleDispatcher() {
```
tcmulcahy, line 95: Do we have any test coverage now? If you took out the withContext statement and just ran the block directly would any test fail? (If not, I think we should test that)

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/WorkflowLayout.kt
steve-the-edwards: This part is not for the runtime actually, this is where we collect the emitted renderings from the runtime and update the views that show them.

so we need to get main thread here to update those views.

The wording change is good though!

steve-the-edwards:
```
...but we verify that it is a main thread dispatcher since we are updating views here!
```

workflow-ui/core-android/src/test/java/com/squareup/workflow1/ui/WorkflowLayoutTest.kt
steve-the-edwards: Added test coverage for this, and testing also that we throw if its not a main thread dispatcher in the context.

steve-the-edwards: Whoops. Firebender had its own branch for the work we were doing and it did not get added.... stay tuned.

steve-the-edwards: Ok, added correctly now.

PR #1425, opened 2025-10-14
Add Dispatch Tracking

steve-the-edwards:
Add the runtime's `CoroutineContext` to the `WorkflowSession` info and then also forward the `CoroutineDispatcher` through the `ConfigSnapshot` that is used by the `WorkflowRuntimeTracer`. This can help us get definitive logging about the dispatcher being used in the runtime.

### DIFF
--- a/workflow-runtime/api/workflow-runtime.api
+++ b/workflow-runtime/api/workflow-runtime.api
@@ -124,6 +124,7 @@ public abstract interface class com/squareup/workflow1/WorkflowInterceptor$Workf
 	public abstract fun getParent ()Lcom/squareup/workflow1/WorkflowInterceptor$WorkflowSession;
 	public abstract fun getRenderKey ()Ljava/lang/String;
 	public abstract fun getRuntimeConfig ()Ljava/util/Set;
+	public abstract fun getRuntimeContext ()Lkotlin/coroutines/CoroutineContext;
 	public abstract fun getSessionId ()J
 	public abstract fun getWorkflowTracer ()Lcom/squareup/workflow1/WorkflowTracer;
 	public abstract fun isRootWorkflow ()Z
--- a/workflow-tracing/src/main/java/com/squareup/workflow1/tracing/ConfigSnapshot.kt
+++ b/workflow-tracing/src/main/java/com/squareup/workflow1/tracing/ConfigSnapshot.kt
@@ -8,13 +8,18 @@ import com.squareup.workflow1.RuntimeConfigOptions.RENDER_ONLY_WHEN_STATE_CHANGE
 import com.squareup.workflow1.RuntimeConfigOptions.STABLE_EVENT_HANDLERS
 import com.squareup.workflow1.RuntimeConfigOptions.WORK_STEALING_DISPATCHER
 import com.squareup.workflow1.WorkflowExperimentalRuntime
+import kotlinx.coroutines.CoroutineDispatcher
 
 /**
  * Snapshot of the current [RuntimeConfig]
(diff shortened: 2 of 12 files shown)


workflow-runtime/api/workflow-runtime.api
```diff
@@ -124,6 +124,7 @@ public abstract interface class com/squareup/workflow1/WorkflowInterceptor$Workf
 	public abstract fun getParent ()Lcom/squareup/workflow1/WorkflowInterceptor$WorkflowSession;
 	public abstract fun getRenderKey ()Ljava/lang/String;
 	public abstract fun getRuntimeConfig ()Ljava/util/Set;
+	public abstract fun getRuntimeContext ()Lkotlin/coroutines/CoroutineContext;
```
tcmulcahy, line 127: This is a breaking change for anyone implementing WorkflowInterceptor. I don't know if that's a concern....

workflow-tracing/src/main/java/com/squareup/workflow1/tracing/ConfigSnapshot.kt
```diff
@@ -40,6 +45,7 @@ public class ConfigSnapshot(config: RuntimeConfig) {
       if (config.isEmpty()) {
         append("Base, ")
       }
+      append("Dispatch: ${runtimeDispatch?.toString()?.take(6)}")
```
tcmulcahy: This won't always work. e.g. see the documentation from https://github.com/Kotlin/kotlinx.coroutines/blob/5d4954b17ac18c572e22a9a6e7bb4e4368ec568c/kotlinx-coroutines-core/common/src/MainCoroutineDispatcher.kt#L45 :

```
    /**
     * Returns a name of this main dispatcher for debugging purposes. This implementation returns
     * `Dispatchers.Main` or `Dispatchers.Main.immediate` if it is the same as the corresponding
     * reference in [Dispatchers] or a short class-name representation with address otherwise.
     */
    override fun toString(): String = toStringInternalImpl() ?: "$classSimpleName@$hexAddress"
```

The IO dispatcher - https://github.com/Kotlin/kotlinx.coroutines/blob/5d4954b17ac18c572e22a9a6e7bb4e4368ec568c/kotlinx-coroutines-core/native/src/Dispatchers.kt#L44 - is similar

Maybe you should drop the `Dispatchers.` from the beginning of the string if it exists?

workflow-runtime/api/workflow-runtime.api
steve-the-edwards: New method is fine, no?

workflow-tracing/src/main/java/com/squareup/workflow1/tracing/ConfigSnapshot.kt
steve-the-edwards: good idea.

steve-the-edwards: Changed it to
```
append("Dispatch: ${runtimeDispatch?.toString()?.substringAfter("Dispatchers.")?.take(8)}")
```

substringAfter returns original string by default if does not contain the substring.

workflow-runtime/api/workflow-runtime.api
steve-the-edwards: Oh IC what you are meaning. This is actually in the `WorkflowSession` interface, which we implement via `WorkflowNode`. so this will not ask more of `WorkflowInterceptor` implementors.

Issue #1152, opened 2024-01-03
Deprecate one/disambiguate between launchForTestingFromStart/StateWith and headlessIntegrationTest


steve-the-edwards:
One was for the original integration tests, the other was stood up with coroutines (via `runTest` and `Turbine`) to do pull based testing of multiple Workflows.

We should pick a path and support it.

Right now launchForTestingFromStart/StateWith... does not support specifying `RuntimeConfig` as a parameter - it must be done via a System Property and JvmRuntimeConfigTools (usually shuttled there via a gradle property - handy, but not clear).

steve-the-edwards:
Renaming `headlessIntegrationTest` to something less focused makes sense as well.

PR #1427, opened 2025-10-19
Add missing features to renderForTest Api

Zikstar:
### Description
Addresses issue #1152 : The `launchForTesting` apis are deprecated and there's some extra work to bring all the missing features from `launchForTesting` to the `renderForTest` api. This PR addresses addresses the first two of four left over work streams to fully deprecate the old api namely:
- restoring from Snapshot
- testing for idempotency

It also adds the functionality for 
- awaitForNextSnapshot() and awaitForNextOutput()
which were important parts of the usage of the old api.

I still plan to do some work use the Adapter/Facade pattern to make `launchForTesting` use `renderForTest` under the hood in order to achieve goal 3, but that would be a separate PR.

### DIFF
--- a/workflow-testing/src/test/java/com/squareup/workflow1/testing/WorkflowTurbineTest.kt
+++ b/workflow-testing/src/test/java/com/squareup/workflow1/testing/WorkflowTurbineTest.kt
@@ -5,6 +5,7 @@ import com.squareup.workflow1.StatefulWorkflow
 import com.squareup.workflow1.TreeSnapshot
 import com.squareup.workflow1.Workflow
 import com.squareup.workflow1.action
+import com.squareup.workflow1.parse
 import com.squareup.workflow1.stateful
 import kotlin.test.Test
 import kotlin.test.assertEquals
@@ -44,6 +45,8 @@ class WorkflowTurbineTest {
       // First snapshot should exist
       val firstSnapshot = awaitNextSnapshot()
       assertNotNull(firstSnapshot)
+      val actualSnapshotValue = firstSnapshot.readIntValue()
+      assertEquals(42, actualSnapshotValue)
     }
   }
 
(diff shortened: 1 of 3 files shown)


workflow-testing/src/test/java/com/squareup/workflow1/testing/WorkflowTurbineTest.kt
```diff
@@ -0,0 +1,318 @@
...
+      val (value1, _) = awaitNextRendering()
+      assertEquals(1, value1)
+
+      // Now consume snapshots - they should be in sync
+      val snapshot0 = awaitNextSnapshot()
+      assertNotNull(snapshot0)
+
+      val snapshot1 = awaitNextSnapshot()
+      assertNotNull(snapshot1)
+      assertNotEquals(snapshot0, snapshot1)
```
steve-the-edwards: What does it mean for them to be synchronized here? I'm not sure I understand.

steve-the-edwards (approved):
I think we could improve the snapshot testing part of `WorkflowTurbineTest` as we could check the values inside the snapshot. Also i'm confused by one of the tests, but otherwise LGTM!

workflow-testing/src/test/java/com/squareup/workflow1/testing/WorkflowTurbineTest.kt
Zikstar: Yeah that seems like an erroneous comment. Will delete

Zikstar:
> I think we could improve the snapshot testing part of `WorkflowTurbineTest` as we could check the values inside the snapshot. Also i'm confused by one of the tests, but otherwise LGTM!

Sure! PR is now updated with your comments address @steve-the-edwards

PR #1428, opened 2025-11-07
Make PapaSafeTrace isTraceable configurable

tcmulcahy:
Change PapaSafeTrace from delegating to SafeTrace.isTraceable to accepting isTraceable as a constructor parameter. This gives clients direct control over whether tracing is enabled, removing the implicit dependency on SafeTrace.isTraceable, which will be removed soon.

Update WorkflowPapaTracer's default to explicitly pass isTraceable=false to avoid overhead. Add tests to verify the new configuration capability for both PapaSafeTrace and WorkflowPapaTracer.

### DIFF
--- a/.github/workflows/kotlin.yml
+++ b/.github/workflows/kotlin.yml
@@ -203,7 +203,7 @@ jobs:
     name: General Unit Tests (not KMP)
     runs-on: macos-latest
     needs: build-all
-    timeout-minutes: 40
+    timeout-minutes: 60
     steps:
       - name: Checkout
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2


workflow-tracing-papa/api/workflow-tracing-papa.api
```diff
@@ -1,5 +1,7 @@
 public final class com/squareup/workflow1/tracing/papa/PapaSafeTrace : com/squareup/workflow1/tracing/SafeTraceInterface {
 	public fun <init> ()V
+	public fun <init> (Z)V
+	public synthetic fun <init> (ZILkotlin/jvm/internal/DefaultConstructorMarker;)V
```
tcmulcahy, line 4: Kotlin generates this when you have parameters with default values - see https://stackoverflow.com/a/53912965 for details

PR #1429, opened 2025-11-07
Replace Papa SafeTrace with androidx.tracing

tcmulcahy:
Remove the dependency on papa.SafeTrace and replace it with androidx.tracing.Trace. This simplifies the dependency tree by using the standard AndroidX tracing library instead of the third-party Papa library. (Note: we have a test-only dependency on Papa still for papa.Choreographers)

The PapaSafeTrace class now delegates to androidx.tracing.Trace methods which provide equivalent functionality. The isCurrentlyTracing property now uses Trace.isEnabled() instead of
SafeTrace.isCurrentlyTracing, and logSection() now uses the trace {} inline function from androidx.tracing.ktx.


workflow-tracing-papa/src/main/java/com/squareup/workflow1/tracing/papa/PapaSafeTrace.kt
```diff
@@ -14,31 +15,31 @@ class PapaSafeTrace(
 ) : SafeTraceInterface {
```
steve-the-edwards, line 15: Let's rename this class then as it no longer uses `Papa`

workflow-tracing-papa/src/main/java/com/squareup/workflow1/tracing/papa/PapaSafeTrace.kt
```diff
@@ -14,31 +15,31 @@ class PapaSafeTrace(
...
 
   override val isCurrentlyTracing: Boolean
-    get() = SafeTrace.isCurrentlyTracing
+    get() = Trace.isEnabled()
```
steve-the-edwards, line 18: Does `isEnabled` go off and on based on whether we are actively recording a trace?

tcmulcahy: Yes!

workflow-tracing-papa/src/main/java/com/squareup/workflow1/tracing/papa/PapaSafeTrace.kt
tcmulcahy: Yep, totally. I'll prepare a separate PR to do so. How about `s/PapaSafeTrace/WorkflowTrace` and `s/SafeTraceInterface/TraceInterface`?

PR #1430, opened 2025-11-07
Extend timeout for general unit tests to 60 min

tcmulcahy:
The general unit tests job was timing out at 40 minutes. This change increases the timeout to 60 minutes to allow the test suite to complete successfully.

e.g. https://github.com/square/workflow-kotlin/actions/runs/19176422684/job/54823085604?pr=1429#logs


steve-the-edwards (approved):
Sad that we just keep taking longer and longer, but makes sense to do this.

PR #1431, opened 2025-11-11
Remove Papa/Safe prefixes from names

tcmulcahy:
We are no longer using papa safe trace, so this commit renames SafeTraceInterface to TraceInterface, PapaSafeTrace to WorkflowTrace, WorkflowPapaTracer to WorkflowTracer, and FakeSafeTrace to FakeTrace to better reflect their purpose without legacy naming conventions.

Maintains backward compatibility by keeping deprecated versions of the old class names that delegate to the new implementations. All test files updated to use the new naming convention.

workflow-tracing-papa is misleading, but it's the name of a jar published to maven, so I've preserved it.


steve-the-edwards:
@codex review

steve-the-edwards (approved):

PR #1432, opened 2025-11-13
Add `onSessionCancelled` API

steve-the-edwards:
We want observability into any actions that have already been queued but not yet applied at the time that a workflow node stops rendering in the tree. This is the 'end' of that `WorkflowSession` and we mark that with an optional `onSessionCancelled` API that receives the un-handled actions.

We also include the exception that caused the cancellation if there is one, as well as the workflow session instance itself.


ijwhelan (approved):
Wow, this is a great idea! 😉

PR #1441, opened 2025-12-22
Upgrade Kotlin to 2.3.0

vRallev:
Upgrade Kotlin and other required libraries to the latest version.

This still fails due to [KT-83361](https://youtrack.jetbrains.com/issue/KT-83361).


steve-the-edwards:
Already past this version.

PR #1442, opened 2026-01-14
Implement an alternative runtime for Workflow on top of Compose.

zach-klippenstein:
This runtime is optional and can be activated in the entire tree either using the `COMPOSE_RUNTIME` runtime option, or in a subtree by wrapping a workflow with a `ComposeRuntimeSwizzlerWorkflow`.

See [go/compose-based-workflows](https://go/compose-based-workflows).


handstandsam (approved):

samples/compose-samples/src/main/java/com/squareup/sample/compose/nestedrenderings/NestedRenderingsActivity.kt
```diff
@@ -46,7 +47,8 @@ class NestedRenderingsActivity : AppCompatActivity() {
         workflow = RecursiveWorkflow.mapRendering { it.withEnvironment(viewEnvironment) },
         scope = viewModelScope + AndroidUiDispatcher.Main,
         savedStateHandle = savedState,
-        runtimeConfig = AndroidRuntimeConfigTools.getAppWorkflowRuntimeConfig(),
```
tcmulcahy, line 49: Can we undo the hardcoding? Instead we should specify compose_runtime via the runtime config in CI.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/RuntimeConfig.kt
```diff
@@ -98,7 +98,15 @@ public enum class RuntimeConfigOptions {
...
    */
-  @WorkflowExperimentalRuntime INDEXED_ACTIVE_STAGING_LISTS;
+  @WorkflowExperimentalRuntime INDEXED_ACTIVE_STAGING_LISTS,
+
+  /**
+   * Replaces the traditional Workflow runtime with the Compose runtime at the root of the workflow
+   * tree. To only use the runtime for a subtree, use `ComposeRuntimeSwizzlerWorkflow` from the
+   * `workflow-runtime` library.
+   */
+  @WorkflowExperimentalRuntime COMPOSE_RUNTIME,
```
tcmulcahy, line 108: Can you add these to https://github.com/square/workflow-kotlin/blob/c5eb922923dc14a64c5b8168d1df7dcb1450e51c/workflow-config/config-android/src/main/java/com/squareup/workflow1/config/AndroidRuntimeConfigTools.kt and https://github.com/square/workflow-kotlin/blob/c5eb922923dc14a64c5b8168d1df7dcb1450e51c/workflow-config/config-jvm/src/main/java/com/squareup/workflow1/config/JvmTestRuntimeConfigTools.kt

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/compose/ComposeRenderContext.kt
```diff
@@ -0,0 +1,342 @@
...
+          )
+        }
+      }
+    return interceptedWorkflow
+      .render(
+        renderProps = props as P,
+        renderState = currentState,
+        context = RenderContext(this, interceptedWorkflow),
+      )
+      .also { trapdoor = null }
```
tcmulcahy, line 149: should we use try/finally to ensure that we always set trapdoor to null?

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/compose/ComposeRenderContext.kt
```diff
@@ -0,0 +1,342 @@
...
+  }
+
+  override fun <ResultT> remember(
+    key: String,
+    resultType: KType,
+    vararg inputs: Any?,
+    calculation: () -> ResultT,
+  ): ResultT =
+    requireTrapdoor("remember").inMovableGroup(GROUP_KEY, key) {
+      // RememberStore keys the lifetime of the memoization off the result type and inputs in
```
tcmulcahy: RememberStore?

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/compose/WorkflowComposableRuntimeConfig.kt
```diff
@@ -0,0 +1,47 @@
...
+  val workflowInterceptor: WorkflowInterceptor? = null,
+  val idCounter: IdCounter? = null,
+) {
+
+  override fun toString(): String =
+    "WorkflowComposableRuntimeConfig(" +
+      "runtimeConfig=$runtimeConfig, " +
+      "workflowTracer=$workflowTracer)"
+
+  override fun equals(other: Any?): Boolean {
```
tcmulcahy: Either remove equals/hashCode or make sure they reflect all of the fields.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/ComposeRuntimeSwizzlerWorkflow.kt
```diff
@@ -0,0 +1,121 @@
...
+  }
+}
+
+/**
+ * A holder for the real [State] that only supports referential equality. This allows a workflow
+ * action to set the state to a "new" value while still referring to the same actual state.
+ *
+ * This MUST NOT be a data class or implement equals by comparing state values.
+ */
+private class StateHolder<P, O, R>(val state: State<P, O, R>)
```
tcmulcahy: Dead code?

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/ComposeRuntimeSwizzlerWorkflow.kt
```diff
@@ -0,0 +1,121 @@
...
+
+private class State<P, O, R>(
```
tcmulcahy: Dead code?

tcmulcahy (approved):

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/compose/ComposeRenderContext.kt
zach-klippenstein: This is a comparative explanation to the traditional runtime. Made that more clear.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/compose/ComposeRenderContext.kt
zach-klippenstein: If the render call fails we cancel the whole runtime so there's no point.

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/RuntimeConfig.kt
zach-klippenstein: I was totally unaware of these, thanks!

samples/compose-samples/src/main/java/com/squareup/sample/compose/nestedrenderings/NestedRenderingsActivity.kt
zach-klippenstein: Meant to do that, looks like I missed a few others as well. That's what I get for squashing and pushing on a Friday.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/ComposeRuntimeSwizzlerWorkflow.kt
zach-klippenstein: Removed.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/ComposeRuntimeSwizzlerWorkflow.kt
zach-klippenstein: Removed.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/compose/WorkflowComposableRuntimeConfig.kt
zach-klippenstein: Removed them and updated the stale comment.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/compose/ComposeRenderContext.kt
```diff
@@ -0,0 +1,339 @@
...
+  ): ResultT =
+    requireTrapdoor("remember").inMovableGroup(GROUP_KEY, key) {
+      // RememberStore (the traditional runtime's implementation of this call) keys the lifetime of
+      // the memoization off the result type and inputs in addition to the explicit key. The data
+      // key is only used to keep the memoized state consistent as the order and number of calls to
+      // this context change. The resultType and inputs only need to reset the state for this
+      // particular remember call, so they don't need to be part of the data key. We use a separate
+      // key call since it's simpler than correctly folding multiple objects into a single data key
+      // (which requires calling Composer.joinKey).
+      key(resultType, *inputs) { remember(*inputs, calculation = calculation) }
```
zach-klippenstein, line 190: Does this `key` call need to exist? Just pass the keys to `remember`.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/compose/ComposeRenderContext.kt
```diff
@@ -0,0 +1,339 @@
...
+            new = props as P,
+            state = oldState,
+          )
+        }
+      }
+    return interceptedWorkflow
+      .render(
+        renderProps = props as P,
+        renderState = currentState,
+        context = RenderContext(this, interceptedWorkflow),
```
zach-klippenstein, line 157: Don't allocate this every time?

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/compose/WorkflowSnapshotState.kt
```diff
@@ -0,0 +1,142 @@
...
+    return newState
+  }
+
+  /**
+   * Applies a [WorkflowAction] to the workflow's state, writing the new state and invoking the
+   * output handler if necessary.
+   */
+  inline fun applyAction(action: WorkflowAction<Any?, Any?, Any?>, onNewState: () -> Unit) {
+    // TODO should this be done in a snapshot to ensure all the updates get to the workflow runtime
+    //  at the same time?
```
zach-klippenstein, line 107: Delete TODO

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/compose/RenderWorkflow.kt
```diff
@@ -0,0 +1,251 @@
...
+      parentSession = parentSession,
+      renderKey = renderKey,
+      callerRecomposeScope = recomposeScope,
+    )
+
+  if (onSessionAvailable != null) {
+    DisposableEffect(onSessionAvailable, renderContext) {
+      onSessionAvailable(renderContext)
+      onDispose {}
+    }
```
zach-klippenstein, line 88: I think we can get rid of this effect in the follow-up PR that makes the context a `RememberObserver`.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/compose/RenderWorkflow.kt
```diff
@@ -0,0 +1,251 @@
...
+  invalidateCallerOnNewValue: Boolean,
+  callerRecomposeScope: RecomposeScope,
+  composer: Composer,
+  changed: Int,
+): Any? {
+  // Outer group is restartable: This should wrap the entire body of this function (except the
+  // actual
+  // return statement) and is what defines the recompose scope for producer.
+  // Key chosen "randomly" by mashing on my keyboard.
+  composer.startRestartGroup(23975234)
```
zach-klippenstein, line 141: Make these keys less arbitrary.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/compose/RenderWorkflow.kt
```diff
@@ -0,0 +1,251 @@
...
+  // render. But it's required for restartability. We could try making it a no-op everywhere except
+  // the root node, since we're already invalidating all the way up the tree every time anyway.
+  composer.endRestartGroup()?.updateScope { composer, changed ->
+    // This lambda is called when producer is invalidated. The lambda must create a restartable
+    // group with the same key to preserve positional identity.
+    renderWorkflowRestartableImpl(
+      props = props,
+      onOutput = onOutput,
+      renderContext = renderContext,
+      invalidateCallerOnNewValue = true,
```
zach-klippenstein, line 242: I think this parameter can go away since we invalidate all the way up all the time anyway.

steve-the-edwards (approved):
LFG

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/compose/WorkflowSnapshotState.kt
```diff
@@ -0,0 +1,142 @@
...
+    if (
+      propsChanged ||
+        stateChanged ||
+        (didOnOutputChange == true || (didOnOutputChange == null && oldOnOutput != newOnOutput))
+    ) {
+      record.writable(this) {
+        props = newProps
+        onOutput = newOnOutput
+        state = newState
+      }
```
steve-the-edwards, line 92: 🤖 **[P1] Serialize action writes with composition, or handle snapshot conflicts.** `updateAndGetState()` writes this record inside the Recomposer's mutable snapshot, while `applyAction()` can write the same state object from another thread. `applyActionLock` only serializes actions; it does not cover composition, and this `StateObject` inherits `mergeRecords() = null`. I reproduced the interleaving against this PR's state class and Compose 1.9.4: take a mutable snapshot, update props inside it, apply an increment action outside it, then apply the snapshot; `apply().succeeded` is false. Compose's `Recomposer.applyAndCheck()` turns that into `Unsupported concurrent change during composition`, terminating recomposition. Please cover this interleaving with a regression test and establish serialization or a correct conflict strategy. Source: [Compose 1.9.4 sources, Recomposer.kt:1505–1532 and Snapshot.kt:1332–1336](https://dl.google.com/dl/android/maven2/androidx/compose/runtime/runtime-desktop/1.9.4/runtime-desktop-1.9.4-sources.jar).

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/compose/StateSaving.kt
```diff
@@ -0,0 +1,94 @@
...
+import com.squareup.workflow1.writeByteStringWithLength
+import com.squareup.workflow1.writeList
+import com.squareup.workflow1.writeUtf8WithLength
+import okio.ByteString
+
+internal fun createSaveableStateRegistryForTreeSnapshot(
+  treeSnapshot: TreeSnapshot?
+): SaveableStateRegistry {
+  val snapshot = treeSnapshot?.workflowSnapshot
+  return createSaveableStateRegistryForSnapshot(snapshot)
```
steve-the-edwards, line 21: 🤖 **[P1] Handle saved state when the runtime flag changes.** If an existing `TreeSnapshot` survives a flag change, this treats the legacy workflow's root snapshot bytes as the Compose registry format and drops the legacy child snapshot tree. I reproduced `EOFException` by restoring a valid legacy snapshot containing a length-prefixed string (`"abc"`). In the reverse direction, the legacy workflow receives the Compose registry bytes as its own snapshot. The cross-runtime save/restore test currently skips these transitions. Before rollout, please provide compatibility/migration or an explicit integration contract that detects and resets incompatible saved state in both directions; simply changing the runtime flag is not sufficient.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/compose/ComposeRenderContext.kt
```diff
@@ -0,0 +1,339 @@
...
+      session = this,
+    )
+  }
+
+  private fun applyAction(action: WorkflowAction<P, Any?, O>) {
+    check(trapdoor == null) { "Cannot send to action sink from render pass." }
+    // Send can be called from any thread so wrap non-atomic reads/writes in a critical section.
+    applyActionLock.withLock {
+      @Suppress("UNCHECKED_CAST")
+      state.applyAction(action as WorkflowAction<Any?, Any?, Any?>, onNewState = this::invalidate)
```
steve-the-edwards, line 256: 🤖 **[P1] Stop accepting actions after the workflow session is disposed.** `onDisposed()` reports cancellation but never disables this sink or clears the stored output handler. If a parent removes a child and a callback retained from that child's old rendering later sends an output-only action, this still applies the action and invokes the captured parent handler. That lets a removed child mutate its former parent; the legacy runtime stops selecting that child's action queue after removal. Please coordinate disposal with action application so callbacks from a disposed session cannot propagate outputs. A regression test can retain a child callback, remove the child, invoke the callback, and assert the parent is unchanged.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/WorkflowNode.kt
```diff
@@ -33,19 +35,31 @@ internal fun <PropsT, OutputT, RenderingT> createWorkflowNode(
...
-  )
+  if (workflow is ComposeRuntimeSwizzlerWorkflow) {
+    ComposeWorkflowNode(
+      id = id,
+      workflow = workflow,
+      initialProps = initialProps,
+      snapshot = snapshot,
+      baseContext = baseContext,
+      emitAppliedActionToParent = emitAppliedActionToParent,
+      interceptor = interceptor,
```
steve-the-edwards, line 46: 🤖 **[P2] Forward the runtime configuration and session metadata into the Compose node.** This constructor call omits `runtimeConfig`, `workflowTracer`, `parent`, and `idCounter`, so they silently take their defaults. In particular, `COMPOSE_RUNTIME_SKIPPING` never reaches `WorkflowComposableRuntimeConfig` through the public `renderWorkflowIn()` entry point: both Compose runtime options execute the non-skipping path, including benchmarks labeled `ComposeSkipping`. Subtree sessions also lose their parent and shared ID-counter input, and tracing loses the supplied tracer. Please forward these arguments as the legacy branch does, and test the skipping option through the public entry point.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/compose/ComposeRenderContext.kt
```diff
@@ -0,0 +1,339 @@
...
+  init {
+    @Suppress("UNCHECKED_CAST")
+    val statefulWorkflow = workflow.asStatefulWorkflow() as StatefulWorkflow<P, Any?, O, R>
+    val interceptor = config.workflowInterceptor
+
+    interceptor?.onSessionStarted(workflowScope = workflowScope, session = this)
+
+    interceptedWorkflow =
+      interceptor?.intercept(workflow = statefulWorkflow, workflowSession = this)
+        ?: statefulWorkflow
```
steve-the-edwards, line 116: 🤖 **[P2] Refresh the workflow implementation when its instance changes.** The context is remembered by workflow identity, but `workflow` and `interceptedWorkflow` permanently capture the first instance. When a parent renders a new instance of the same workflow type/key, subsequent `render`, `onPropsChanged`, and `snapshotState` calls still use the old object, potentially retaining old constructor dependencies or captured values. The legacy `StatefulWorkflowNode.maybeUpdateCachedWorkflowInstance()` explicitly handles this while preserving session state. Please update the implementation/interceptor when the instance changes without resetting the session.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/compose/WorkflowSnapshotState.kt
```diff
@@ -0,0 +1,142 @@
...
+      onNewState()
+    }
+
+    // Propagate the output up the workflow tree. Propagation doesn't touch our state but still
+    // should be part of the critical section: For an intermediate node, while it's propagating
+    // an action from one part of its subtree, if another action comes in from a different part
+    // of the subtree, the second action won't propagate until the first one is done. If we
+    // moved this out of the lock, then the propagations of two actions from two subtrees could
+    // be interleaved to their common ancestors.
+    onOutput?.let { onOutput -> applied.output?.value?.let(onOutput) }
```
steve-the-edwards, line 128: 🤖 **[P2] Preserve an explicitly emitted null output.** `WorkflowOutput` is the box that distinguishes no output from `setOutput(null)`, but `applied.output?.value?.let(onOutput)` drops the latter. I reproduced zero output-handler invocations for an action that calls `setOutput(null)`. Check the wrapper for null and then forward its value, e.g. `applied.output?.let { onOutput(it.value) }`. The Compose node's `outputs.tryReceive().getOrNull() ?: return ActionsExhausted` also needs to distinguish a successfully received null from an empty channel.

workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/internal/compose/ComposeWorkflowNode.kt
```diff
@@ -0,0 +1,169 @@
...
+      dispatcher.advanceUntilIdle()
+
+      // Consume the requested frame so it doesn't get fired again from the side effect.
+      recomposeRequests.tryReceive()
+
+      // Send the frame to perform recomposition and effects.
+      // Hard-code unchanging frame time since there's no actual frame time code shouldn't rely on
+      // this value.
+      clock.sendFrame(0L)
+      return rendering!!
```
steve-the-edwards, line 124: 🤖 **[P2] Allow null as a valid rendering.** `R` has no non-null bound, so a workflow may legitimately return a nullable rendering. `rendering!!` throws for that valid result, including on initial render. Please distinguish 'no rendering produced yet' from 'produced null' with an initialization sentinel or `NullableInitBox`, as the legacy runtime does, and cover both an initial null and a transition to null.

steve-the-edwards (commented):
🤖 Follow-up review of `f84fc467`: three P1 and four P2 findings are attached inline. The snapshot conflict, legacy-to-Compose restoration failure, and nullable-output loss were reproduced with isolated PR code against Compose 1.9.4; the remaining findings are based on source inspection. These concern enabling the Compose runtime and preserving the existing runtime contract.

PR #1443, opened 2026-01-14
Make nested renderings flash when clicked, to help test output propagation.

zach-klippenstein:
https://github.com/user-attachments/assets/66339ba2-2155-4d2e-b7c3-ea98bc97b2a1

### DIFF
--- a/samples/compose-samples/src/main/java/com/squareup/sample/compose/nestedrenderings/RecursiveViewFactory.kt
+++ b/samples/compose-samples/src/main/java/com/squareup/sample/compose/nestedrenderings/RecursiveViewFactory.kt
@@ -6,10 +6,7 @@ import androidx.compose.animation.Animatable
 import androidx.compose.animation.core.FastOutLinearInEasing
 import androidx.compose.animation.core.LinearOutSlowInEasing
 import androidx.compose.animation.core.keyframes
-import androidx.compose.foundation.gestures.awaitEachGesture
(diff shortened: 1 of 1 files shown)


samples/compose-samples/src/main/java/com/squareup/sample/compose/nestedrenderings/RecursiveViewFactory.kt
```diff
@@ -44,7 +55,29 @@ val RecursiveComposableFactory = ScreenComposableFactory<Rendering> { rendering
...
   }
 
-  Card(backgroundColor = color) {
+  var lastFlashedTrigger by remember { mutableIntStateOf(rendering.flashTrigger) }
+  val flashAlpha = remember { Animatable(Color(0x00FFFFFF)) }
+
+  // Flash the card white when asked.
+  LaunchedEffect(rendering.flashTrigger) {
+    if (rendering.flashTrigger != 0) {
+      lastFlashedTrigger = rendering.flashTrigger
```
steve-the-edwards, line 64: I don't see what we are using `lastFlashedTrigger` for?

zach-klippenstein: Fixed: https://github.com/square/workflow-kotlin/pull/1444

PR #1445, opened 2026-01-27
Move workflow-runtime-android into android source set in workflow-runtime module. (2nd attempt)

zach-klippenstein:
First attempt was https://github.com/square/workflow-kotlin/pull/1370, but got reverted in https://github.com/square/workflow-kotlin/pull/1378 due to it having broken something.

I need this in order to correctly consume Compose Multiplatform for the work migrating the runtime to compose (https://github.com/square/workflow-kotlin/pull/1442).

This reverts commit 58c1d40ed94a931c414004bf4765690313b28d2c.


.github/workflows/kotlin.yml
```diff
@@ -694,7 +694,62 @@
...
+          api-level: ${{ matrix.api-level }}
+#          prepare-task: prepareConnectedCheckShard${{matrix.shardNum}}
+          test-task: connectedAndroidTest
+          write-cache-key: androidDeviceTest-build-artifacts-${{matrix.shardNum}}
+          restore-cache-key: kmp-instrumentation-test-build-artifacts
+          failure-path-upload: '**/build/reports/androidTests/connected'
+          failure-upload-name: kmp-instrumentation-tests-report-${{matrix.shardNum}}
+
+  # Doesn't use sharding because there are very few Android device tests in KMP.
+  kmp-runtime-instrumentation-tests:
```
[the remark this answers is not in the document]
zach-klippenstein: I just copy-pasted this from other tasks, if we add this we should do it everywhere.

zach-klippenstein:
It looks like benchmarks are no longer being excluded, that should be the last thing to fix.

workflow-runtime/build.gradle.kts
```diff
@@ -1,34 +1,97 @@
+import com.android.build.api.dsl.androidLibrary
 import com.squareup.workflow1.buildsrc.iosWithSimulatorArm64
-import org.jetbrains.kotlin.gradle.ExperimentalKotlinGradlePluginApi
 
 plugins {
+  // This is the new/future plugin for Android in KMP. com.android.library is going away.
+  // See https://developer.android.com/kotlin/multiplatform/plugin
+  id("com.android.kotlin.multiplatform.library")
+  // This is our own convention plugin, not the standard one.
```
steve-the-edwards, line 8: Ultimately we probably want to integrate this into our own `android-multiplatform` convention plugin. Is there a reason that is too difficult to do now?

zach-klippenstein: I also realized we'd probably want to move it to a plugin, but I believe this is the only KMP module we have that targets android right now so I was thinking rule-of-three. The configuration is pretty straightforward so I don't think it should be too hard to clone the existing android one and modify it for KMP.

PR #1490, opened 2026-04-14
Separate key and name in eventHandler

pyricau:
## Summary

- `eventHandler` now takes a `key` parameter (after `name`) used for `remember()`, while `name` is used for the action debug label — with the `"eH: "` prefix removed
- The public `eventHandler` overloads (0-arg and 1-arg) in `StatefulWorkflow.RenderContext` and `StatelessWorkflow.RenderContext` gain a `key: String = name` parameter.
- Motivation: when uniqueness requires adding dynamic content (e.g. an id or uuid) to the remember key, the debug name was previously polluted by that content. Now callers can set `key` and `name` independently — `key` gets the unique suffix, `name` stays readable

### DIFF
--- a/workflow-core/src/commonMain/kotlin/com/squareup/workflow1/HandlerBox.kt
+++ b/workflow-core/src/commonMain/kotlin/com/squareup/workflow1/HandlerBox.kt
@@ -56,11 +56,12 @@ internal class HandlerBox2<E1, E2> {
 internal inline fun <P, S, O, reified E1, reified E2> BaseRenderContext<P, S, O>.eventHandler2(
   name: String,
   remember: Boolean,
+  key: String,
   noinline update: Updater<P, S, O>.(E1, E2) -> Unit
 ): (E1, E2) -> Unit {
-  val handler = { e1: E1, e2: E2 -> actionSink.send(action("eH: $name") { update(e1, e2) }) }
+  val handler = { e1: E1, e2: E2 -> actionSink.send(action(name) { update(e1, e2) }) }
   return if (remember) {
-    val box = remember(name, typeOf<E1>(), typeOf<E2>()) { HandlerBox2<E1, E2>() }
+    val box = remember(key, typeOf<E1>(), typeOf<E2>()) { HandlerBox2<E1, E2>() }
     box.handler = handler
     box.stableHandler
   } else {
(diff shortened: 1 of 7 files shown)


workflow-core/src/commonMain/kotlin/com/squareup/workflow1/HandlerBox.kt
```diff
@@ -9,12 +9,13 @@ internal class HandlerBox0 {
 
 internal fun <P, S, O> BaseRenderContext<P, S, O>.eventHandler0(
   name: String,
+  key: String,
   remember: Boolean,
   update: Updater<P, S, O>.() -> Unit
 ): () -> Unit {
-  val handler = { actionSink.send(action("eH: $name", update)) }
```
pyricau, line 15: I figured `eH: ` provides little value, I'd rather see / search for the full strings passed in.

steve-the-edwards: I think i have some perfetto trace analysis tools built on `eH`. So i'm fine to make the change - but you may have to follow up with other internal tooling changes too!

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/StatefulWorkflow.kt
```diff
@@ -167,22 +170,24 @@ public abstract class StatefulWorkflow<
...
       remember: Boolean? = null,
+      key: String = name,
       update: Updater<PropsT, StateT, OutputT>.() -> Unit
-    ): () -> Unit = eventHandler0(name, remember ?: stableEventHandlers, update)
+    ): () -> Unit = eventHandler0(name, remember ?: stableEventHandlers, key, update)
 
     public inline fun <reified EventT> eventHandler(
       name: String,
       remember: Boolean? = null,
+      key: String = name,
```
steve-the-edwards, line 187: why only add this to the single arg and no arg?

- [x] add `key = name` to the rest of the `eventHandler` overloads

steve-the-edwards (approved):

Plastix (approved):

workflow-core/src/commonMain/kotlin/com/squareup/workflow1/StatefulWorkflow.kt
pyricau: Thanks Claude.

PR #1499, opened 2026-05-04
🤖 [codex] Add renderForTest teardown policy

steve-the-edwards:
👨🏻 : Provides a config API to have `renderForTest` wait for teardown - this allows us to use TestRule hooks to do things like check for leaks, etc., without having to wait for runtime teardown manually.

The default is `CancelAndAwait`, which cancels the runtime, drains the test scheduler, awaits runtime job completion with a bounded timeout, then drains again. The teardown path also preserves the original test failure if cleanup itself times out or fails.


steve-the-edwards:
@codex review

steve-the-edwards:
@builderbot0 review

tcmulcahy (approved):

PR #1553, opened 2026-07-10
Restore KeyedSavedStateRegistryOwners just in time, before consumers can see them [UISA-95]

### LINKED ISSUE #570: Occasional "IllegalStateException: Restarter must be created only during owner's initialization stage"
I can't reproduce this, but we're seeing a steady stream in production.

```
java.lang.RuntimeException: Unable to start activity ComponentInfo{com.squareup.apos/com.squareup.ui.main.MainActivity}: java.lang.IllegalStateException: Restarter must be created only during owner's initialization stage
        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3031)
        at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3166)
        at android.app.ActivityThread.handleRelaunchActivityInner(ActivityThread.java:4929)
        at android.app.ActivityThread.handleRelaunchActivity(ActivityThread.java:4838)
        at android.app.servertransaction.ActivityRelaunchItem.execute(ActivityRelaunchItem.java:69)
        at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:108)
        at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:68)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1883)
        at android.os.Handler.dispatchMessage(Handler.java:106)
        at android.os.Looper.loop(Looper.java:193)
        at android.app.ActivityThread.main(ActivityThread.java:6971)
        at java.lang.reflect.Method.invoke(Method.java:-2)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:865)

Caused by: java.lang.IllegalStateException: Restarter must be created only during owner's initialization stage
        at androidx.savedstate.SavedStateRegistryController.performRestore(SavedStateRegistryController.java:58)
        at com.squareup.workflow1.ui.backstack.StateRegistryAggregator.restoreRegistryControllerIfReady(StateRegistryAggregator.kt:195)
        at com.squareup.workflow1.ui.backstack.ViewStateCache$registryHolder$2.invoke(ViewStateCache.kt:54)
        at com.squareup.workflow1.ui.backstack.ViewStateCache$registryHolder$2.invoke(ViewStateCache.kt:52)
        at com.squareup.workflow1.ui.backstack.StateRegistryAggregator.restoreFromBundle(StateRegistryAggregator.kt:228)
        at com.squareup.workflow1.ui.backstack.StateRegistryAggregator.access$restoreFromBundle(StateRegistryAggregator.kt:39)
        at com.squareup.workflow1.ui.backstack.StateRegistryAggregator$lifecycleObserver$1.onStateChanged(StateRegistryAggregator.kt:87)
        at androidx.lifecycle.LifecycleRegistry.forwardPass(LifecycleRegistry.java:265)
        at androidx.lifecycle.LifecycleRegistry.sync(LifecycleRegistry.java:307)
        at androidx.lifecycle.LifecycleRegistry.moveToState(LifecycleRegistry.java:148)
        at androidx.lifecycle.LifecycleRegistry.handleLifecycleEvent(LifecycleRegistry.java:134)
        at androidx.lifecycle.ReportFragment.dispatch(ReportFragment.java:68)
        at androidx.lifecycle.ReportFragment.dispatch(ReportFragment.java:144)
        at androidx.lifecycle.ReportFragment.onActivityCreated(ReportFragment.java:102)
        at android.app.Fragment.performActivityCreated(Fragment.java:2531)
        at android.app.FragmentManagerImpl.moveToState(FragmentManager.java:1318)
        at android.app.FragmentManagerImpl.moveFragmentToExpectedState(FragmentManager.java:1576)
        at android.app.FragmentManagerImpl.moveToState(FragmentManager.java:1637)
        at android.app.FragmentManagerImpl.dispatchMoveToState(FragmentManager.java:3046)
        at android.app.FragmentManagerImpl.dispatchActivityCreated(FragmentManager.java:2998)
        at android.app.FragmentController.dispatchActivityCreated(FragmentController.java:182)
        at android.app.Activity.performCreate(Activity.java:7249)
        at android.app.Activity.performCreate(Activity.java:7233)
        at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1287)
        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3011)
        at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3166)
        at android.app.ActivityThread.handleRelaunchActivityInner(ActivityThread.java:4929)
        at android.app.ActivityThread.handleRelaunchActivity(ActivityThread.java:4838)
        at android.app.servertransaction.ActivityRelaunchItem.execute(ActivityRelaunchItem.java:69)
        at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:108)
        at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:68)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1883)
        at android.os.Handler.dispatchMessage(Handler.java:106)
        at android.os.Looper.loop(Looper.java:193)
        at android.app.ActivityThread.main(ActivityThread.java:6971)
        at java.lang.reflect.Method.invoke(Method.java:-2)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:865)
```

zach-klippenstein:
## Problem

`WorkflowSavedStateRegistryAggregator` restores its `KeyedSavedStateRegistryOwner` children from an observer on the parent `SavedStateRegistryOwner`'s lifecycle. Observer dispatch order on a `LifecycleRegistry` follows registration order, and `WorkflowLifecycleOwner` lifecycles advance in a synchronous depth-first cascade on window attach — so a descendant's lifecycle observers can run **before** the aggregator's restore observer ever fires.

Concretely: a nested `ComposeView`'s `WrappedComposition` consumes the view-tree `SavedStateRegistry` on `ON_CREATE`. If that dispatch wins the race against the aggregator's restore, the consumer reads an unrestored registry and throws:

```
IllegalStateException: You can 'consumeRestoredStateForKey' only after the corresponding
component has moved to the 'CREATED' state
```

On androidx.compose.ui 1.9.5 this is fatal in a second way: the exception is thrown mid-way through `LayoutNode.attach()`'s recursion during a lazy-item/pager subcompose, which leaves ancestor `LayoutNode`s half-attached (`markAsAttached` ran, `runAttachLifecycle` skipped). The tree then detonates at composition dispose with:

```
IllegalStateException: Must run runDetachLifecycle() once after runAttachLifecycle() and before markAsDetached()
```

Compose 1.7.x tolerated the poisoned state; 1.9.5 crashes the process on teardown. Reproduced deterministically (4/4) in Square's android-register with a workflow-hosted Compose screen containing a lazy list (`DemoV2HomeTest#payTabAndButtonAppears_withBillPayEnabled`); this crash currently blocks register's Compose 1.9.5 upgrade. (A minimal, workflow-free Compose repro of the downstream detonation exists and is being filed against androidx separately — but the state-restoration race fixed here is workflow's own bug regardless.)

## Fix

Two cooperating parts, both local to `workflow-ui/core-android`'s androidx integration:

1. **Lifecycle clamp.** `KeyedSavedStateRegistryOwner` no longer delegates its `Lifecycle` to the view's `WorkflowLifecycleOwner`. It now exposes its own `LifecycleRegistry` that mirrors the delegate's, but is held at `INITIALIZED` until the owner's registry has been restored. This enforces the androidx savedstate contract *by construction*: lifecycle-driven consumers (like Compose's `DisposableSaveableStateRegistry`) cannot observe an unrestored registry, and `performRestore()` remains legal no matter how far the delegate lifecycle has advanced (savedstate requires lifecycle < `STARTED`).

2. **Just-in-time restore.** When the delegate delivers `ON_CREATE` and the owner has not been restored yet, the owner synchronously asks the aggregator to restore it right then (`restoreChildNow`) — before the local lifecycle advances and any downstream consumer can see the registry. If the aggregator's own restore observer hasn't fired yet, it pulls its state from the parent registry on demand (`consumeRestoredStateForKey` only requires the parent registry to be restored, not the parent lifecycle to be `CREATED`). Children pruned from the aggregator restore empty rather than stealing a successor's saved state.

Because ancestors' lifecycles always advance before descendants', an owner's JIT restore can always find its parent's state — restoration order is preserved top-down and no state is lost.

This also structurally fixes #570: a child whose lifecycle is past `INITIALIZED` can now always be restored, because its local lifecycle *cannot* pass `INITIALIZED` unrestored.

### API note

`installChildRegistryOwnerOn` now registers a lifecycle observer at install time, which `LifecycleRegistry` requires to happen on the main thread. This was already true of every production caller (container views during render); the KDoc now documents it, and `ViewStateCacheTest` routes its `update()` calls through `runOnMainSync` accordingly.

## Testing

- New JVM regression tests in `WorkflowSavedStateRegistryAggregatorTest` covering: the dispatch-order race (child advancing before parent `ON_CREATE` is restored just in time *with* saved state), the lifecycle clamp, the detached-parent fallback (restores empty without throwing), and the pruned-child case (does not consume a replacement's state).
- `:workflow-ui:core-android:connectedDebugAndroidTest`: 39/39 pass (incl. `BackStackContainerPersistenceLifecycleTest`, `ViewStateCacheTest`, `WorkflowViewStubLifecycleTest`).
- `:workflow-ui:compose:connectedDebugAndroidTest`: 56/56 pass (incl. `ComposeViewTreeIntegrationTest` save/restore integration).
- Validated against the android-register repro via composite build: previously 4/4 crash, now 3/3 pass with zero occurrences of either crash signature in logcat.

### DIFF
--- a/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/KeyedSavedStateRegistryOwner.kt
+++ b/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/KeyedSavedStateRegistryOwner.kt
@@ -1,6 +1,12 @@
 package com.squareup.workflow1.ui.androidx
 
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.Lifecycle.Event.ON_CREATE
+import androidx.lifecycle.Lifecycle.Event.ON_DESTROY
+import androidx.lifecycle.Lifecycle.State.DESTROYED
+import androidx.lifecycle.LifecycleEventObserver
 import androidx.lifecycle.LifecycleOwner
+import androidx.lifecycle.LifecycleRegistry
 import androidx.savedstate.SavedStateRegistry
 import androidx.savedstate.SavedStateRegistryController
 import androidx.savedstate.SavedStateRegistryOwner
@@ -18,17 +24,83 @@ import androidx.savedstate.findViewTreeSavedStateRegistryOwner
  *
  * To create an instance, call [WorkflowSavedStateRegistryAggregator.installChildRegistryOwnerOn].
  *
+ * ## Lifecycle
+ *
+ * This owner exposes its own [Lifecycle] rather than simply delegating to [lifecycleOwner]'s
+ * (typically the [WorkflowLifecycleOwner] of the view it is installed on). The local lifecycle
+ * mirrors [lifecycleOwner]'s, with one crucial difference: it is held at
+ * [INITIALIZED][Lifecycle.State.INITIALIZED] until this owner's [savedStateRegistry] has been
+ * restored, and only then advances.
+ *
+ * This enforces the androidx savedstate contract that a [SavedStateRegistry] is always restored
+ * before its owner's lifecycle moves to [CREATED][Lifecycle.State.CREATED]:
+ *
+ *  - [SavedStateRegistryController.performRestore] throws if the owner's lifecycle is at least
+ *    `STARTED`, so restoration stays legal no matter how far [lifecycleOwner]'s lifecycle has
+ *    advanced by the time the aggregator's state arrives.
+ *  - Consumers such as Compose UI's `DisposableSaveableStateRegistry` call
+ *    [SavedStateRegistry.consumeRestoredStateForKey], which throws if the registry has not been
+ *    restored. They are driven by lifecycle signals, so refusing to advance until restoration
+ *    makes it impossible for them to observe an unrestored registry.
+ *
+ * [WorkflowLifecycleOwner] lifecycles advance in a synchronous, depth-first cascade when a view
+ * tree attaches to a window. Observer dispatch order on those lifecycles depends on registration
+ * order, so a descendant's observers (e.g. a nested `ComposeView`'s composition) can run before
+ * the aggregator's own restoration observer ever fires. To close that race, when [lifecycleOwner]
+ * delivers `ON_CREATE` and this owner has not been restored yet, it synchronously asks the
+ * aggregator to restore it *now* via [onRestoreNeeded] — before the local lifecycle advances and
+ * any downstream consumer can see the registry. [installObserverOn] is called by the aggregator
+ * before the view can possibly attach, so this observer is always registered — and thus always
+ * dispatched — ahead of any consumer's.
+ *
  * @param key The key used to save and restore this controller from a [SavedStateRegistry].
- * @param lifecycleOwner The [LifecycleOwner] that will be delegated to by this instance.
- * (Required because [SavedStateRegistryOwner] extends [LifecycleOwner] for no clear reason.)
+ * @param lifecycleOwner The [LifecycleOwner] this owner's lifecycle follows. (Required because
+ * [SavedStateRegistryOwner] extends [LifecycleOwner] for no clear reason.)
+ * @param onRestoreNeeded Called at most once, when [lifecycleOwner]'s lifecycle is leaving
+ * `INITIALIZED` but this owner's registry has not been restored yet. The callee must restore
+ * [controller] (possibly with a null state) before returning.
  */
 internal class KeyedSavedStateRegistryOwner internal constructor(
   val key: String,
-  lifecycleOwner: LifecycleOwner
-) : SavedStateRegistryOwner, LifecycleOwner by lifecycleOwner {
+  private val lifecycleOwner: LifecycleOwner,
+  private val onRestoreNeeded: (KeyedSavedStateRegistryOwner) -> Unit
+) : SavedStateRegistryOwner {
+  private val localLifecycle = LifecycleRegistry(this)
+
+  override val lifecycle: Lifecycle get() = localLifecycle
+
   internal val controller: SavedStateRegistryController = SavedStateRegistryController.create(this)
   override val savedStateRegistry: SavedStateRegistry
     get() = controller.savedStateRegistry
+
+  private val lifecycleObserver = object : LifecycleEventObserver {
+    override fun onStateChanged(
+      source: LifecycleOwner,
+      event: Lifecycle.Event
+    ) {
+      // The delegate lifecycle is leaving INITIALIZED, so consumers keyed on our lifecycle are
+      // about to be able to read our registry. Restoration must happen first, and it is still
+      // legal because localLifecycle is INITIALIZED until we advance it below.
+      if (event == ON_CREATE && !savedStateRegistry.isRestored) {
+        onRestoreNeeded(this@KeyedSavedStateRegistryOwner)
+      }
+      localLifecycle.handleLifecycleEvent(event)
+      if (event == ON_DESTROY) {
+        source.lifecycle.removeObserver(this)
+      }
+    }
+  }
+
+  /**
+   * Starts mirroring [lifecycleOwner]'s lifecycle. Must be called after the aggregator has
+   * registered this owner, and before the view this owner is installed on can attach to a window.
+   */
+  internal fun installObserver() {
+    // A DESTROYED lifecycle will never advance, so nothing can ever legally consume this
+    // registry through lifecycle signals; observing would just throw.
+    if (lifecycleOwner.lifecycle.currentState == DESTROYED) return
+    lifecycleOwner.lifecycle.addObserver(lifecycleObserver)
+  }
 
   override fun toString(): String {
     return "KeyedSavedStateRegistryOwner(key='$key', controller=$controller)"
--- a/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/KeyedSavedStateRegistryOwner.kt
+++ b/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/KeyedSavedStateRegistryOwner.kt
@@ -49,7 +49,7 @@ import androidx.savedstate.findViewTreeSavedStateRegistryOwner
  * the aggregator's own restoration observer ever fires. To close that race, when [lifecycleOwner]
  * delivers `ON_CREATE` and this owner has not been restored yet, it synchronously asks the
  * aggregator to restore it *now* via [onRestoreNeeded] — before the local lifecycle advances and
- * any downstream consumer can see the registry. [installObserverOn] is called by the aggregator
+ * any downstream consumer can see the registry. [installObserver] is called by the aggregator
  * before the view can possibly attach, so this observer is always registered — and thus always
  * dispatched — ahead of any consumer's.
  *
@@ -83,6 +83,9 @@ internal class KeyedSavedStateRegistryOwner internal constructor(
       // legal because localLifecycle is INITIALIZED until we advance it below.
       if (event == ON_CREATE && !savedStateRegistry.isRestored) {
         onRestoreNeeded(this@KeyedSavedStateRegistryOwner)
+        check(savedStateRegistry.isRestored) {
+          "onRestoreNeeded contract violation: registry for key '$key' was not restored"
+        }
       }
       localLifecycle.handleLifecycleEvent(event)
       if (event == ON_DESTROY) {
--- a/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/WorkflowSavedStateRegistryAggregator.kt
+++ b/workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/WorkflowSavedStateRegistryAggregator.kt
@@ -1,6 +1,7 @@
 package com.squareup.workflow1.ui.androidx
 
 import android.os.Bundle
+import android.util.Log
 import android.view.View
 import androidx.lifecycle.Lifecycle.Event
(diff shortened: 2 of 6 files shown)


zach-klippenstein:
🤖 **Validation complete — marking ready for review.**

Evidence this fix is safe and effective, gathered against Square's android-register monorepo (where the crash was found):

1. **Original crash repro**: `DemoV2HomeTest#payTabAndButtonAppears_withBillPayEnabled` crashed 4/4 before this change (`Must run runDetachLifecycle() once after runAttachLifecycle() and before markAsDetached()` at teardown, Compose 1.9.5); with this branch it passes, along with the rest of its suite (6/6, composite build against register green-master).
2. **No state-restoration regressions**: `workflow-ui:core-android` device suites pass (`ViewStateCacheTest`, `WorkflowSavedStateRegistryAggregatorTest` instrumented: 39/39 and 56/56), plus the 4 new JVM regression tests in `WorkflowSavedStateRegistryAggregatorTest` covering the dispatch-order race, the lifecycle clamp, detached-parent fallback, and pruned-child restore.
3. **Full register CI**: a snapshot build of this branch (`1.28.1-SNAPSHOT`) ran register's entire suite — 1575/1576 parts green, zero test failures (the one red part is register's policy gate that forbids SNAPSHOT dependency versions, by design). Validation PR: squareup/android-register#159654.

Root-cause writeup is in the PR description; the underlying Compose-side issue (exception during `LayoutNode.attach()` recursion permanently corrupting the node tree) is being filed upstream separately.

handstandsam (approved):

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/KeyedSavedStateRegistryOwner.kt
```diff
@@ -18,18 +24,84 @@ import androidx.savedstate.findViewTreeSavedStateRegistryOwner
...
+    override fun onStateChanged(
+      source: LifecycleOwner,
+      event: Lifecycle.Event
+    ) {
+      // The delegate lifecycle is leaving INITIALIZED, so consumers keyed on our lifecycle are
+      // about to be able to read our registry. Restoration must happen first, and it is still
+      // legal because localLifecycle is INITIALIZED until we advance it below.
+      if (event == ON_CREATE && !savedStateRegistry.isRestored) {
+        onRestoreNeeded(this@KeyedSavedStateRegistryOwner)
+      }
```
steve-the-edwards, line 89: 🤖  The class KDoc says the local lifecycle \"is held at INITIALIZED until this owner's registry has been restored\" — but that invariant is enforced by *forcing* restore here, not by withholding advancement. If `onRestoreNeeded` ever fails its contract (returns without restoring — say a future refactor breaks the `children.put`-before-`installObserver` ordering in `installChildRegistryOwnerOn`), the local lifecycle advances unrestored and we're back to the original downstream crash, far from the cause. A cheap check makes the clamp self-enforcing and fails loudly at the source:

suggestion
```
if (event == ON_CREATE && !savedStateRegistry.isRestored) {
   onRestoreNeeded(this@KeyedSavedStateRegistryOwner)
   check(savedStateRegistry.isRestored) {
      "onRestoreNeeded contract violation: registry for key '$key' was not restored"
   }
}
```

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/KeyedSavedStateRegistryOwner.kt
```diff
@@ -18,18 +24,84 @@ import androidx.savedstate.findViewTreeSavedStateRegistryOwner
...
+ *    restored. They are driven by lifecycle signals, so refusing to advance until restoration
+ *    makes it impossible for them to observe an unrestored registry.
+ *
+ * [WorkflowLifecycleOwner] lifecycles advance in a synchronous, depth-first cascade when a view
+ * tree attaches to a window. Observer dispatch order on those lifecycles depends on registration
+ * order, so a descendant's observers (e.g. a nested `ComposeView`'s composition) can run before
+ * the aggregator's own restoration observer ever fires. To close that race, when [lifecycleOwner]
+ * delivers `ON_CREATE` and this owner has not been restored yet, it synchronously asks the
+ * aggregator to restore it *now* via [onRestoreNeeded] — before the local lifecycle advances and
+ * any downstream consumer can see the registry. [installObserverOn] is called by the aggregator
```
steve-the-edwards: 🤖 Nit: broken KDoc reference — the method is `installObserver`, not `installObserverOn`.

```
* any downstream consumer can see the registry. [installObserver] is called by the aggregator
```

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/WorkflowSavedStateRegistryAggregator.kt
```diff
@@ -193,7 +197,9 @@ public class WorkflowSavedStateRegistryAggregator {
    * with its [LifecycleOwner]. (Use [WorkflowLifecycleOwner] to ensure
    * one is properly installed.)
    *
-   * **This method must be called before [view] is attached to a window.**
+   * **This method must be called on the main thread, before [view] is attached to a window.**
+   * (It observes [view]'s `ViewTreeLifecycleOwner` lifecycle, and
+   * [androidx.lifecycle.LifecycleRegistry] requires observer registration on the main thread.)
```
steve-the-edwards, line 203: 🤖  Nit: `ViewStateCache.update` is the public entry point most custom-container authors actually call, and it now inherits this main-thread requirement (as `ViewStateCacheTest`'s `runOnMainSync` change shows). Worth a sentence in its KDoc too.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/WorkflowSavedStateRegistryAggregator.kt
```diff
@@ -243,9 +249,58 @@ public class WorkflowSavedStateRegistryAggregator {
...
+      restoreFromBundle(consumeFromParent())
+    }
+
+    // restoreFromBundle above restores every unrestored child, including this one.
+    if (child.savedStateRegistry.isRestored) return
+
+    val states = states
+    if (states != null && children[child.key] === child) {
+      child.controller.performRestore(states.remove(child.key))
+    } else {
```
steve-the-edwards, line 294: 🤖 Nit: this branch trades the old crash for silent state loss in the (theoretically unreachable in production) case where a child advances with no attached/restored parent. That's the right trade, but a debug-level log here would help diagnose any container that ever hits it.

steve-the-edwards (approved):

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/KeyedSavedStateRegistryOwner.kt
zach-klippenstein: 🤖 Done in 74c72b34e — added the `check(savedStateRegistry.isRestored)` contract assertion right after the `onRestoreNeeded` call, plus a unit test (`KeyedSavedStateRegistryOwnerTest`) covering a callee that violates the contract.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/KeyedSavedStateRegistryOwner.kt
zach-klippenstein: 🤖 Fixed in 74c72b34e — `[installObserverOn]` → `[installObserver]`.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/WorkflowSavedStateRegistryAggregator.kt
zach-klippenstein: 🤖 Done in 74c72b34e — `ViewStateCache.update`'s KDoc now states the main-thread / before-attach requirement and points at `installChildRegistryOwnerOn` as the source of it.

workflow-ui/core-android/src/main/java/com/squareup/workflow1/ui/androidx/WorkflowSavedStateRegistryAggregator.kt
zach-klippenstein: 🤖 Done in 74c72b34e — the restore-empty fallback now logs at debug level (tag `Workflow`, matching the module's existing logging), including the child key and which of the two conditions was hit, so a container that reaches this branch leaves a breadcrumb for the lost state.

PR #1559, opened 2026-08-14
Drop deprecated rememberSaveable(key = …) from RenderAsState [CLF-279]

zach-klippenstein:
Follow-up to #1558 (Compose 1.9.5). `renderAsState` was carrying a `@Suppress("DEPRECATION")` for `rememberSaveable(key = …)`, which Compose deprecated because it bypasses positional scoping ([r.android.com/3610053](https://r.android.com/3610053)). At Compose 1.9.5 that call is a hard build failure under `-Werror`, so the suppression is load-bearing on that upgrade.

## `key(…) { }` is not a workaround

The plan in CLF-278 was to wrap the call in `key(…)` instead. That does not work. `rememberSaveable` derives its registry key from `currentCompositeKeyHashCode.toString(36)` when no explicit `key` is passed, so `key("workflow-snapshot")` only perturbs that hash — the string never reaches the `SaveableStateRegistry`. Both tests that relied on the explicit key fail against it:

```
savesSnapshot     java.util.NoSuchElementException: Key workflow-snapshot is missing in the map.
restoresSnapshot  expected: foo but was an empty string
```

The second is the dangerous one: state is silently not restored.

## What this does instead

The `snapshotKey` parameter existed only to give tests a known registry key, and no variant of it survives the deprecation, so it is removed along with the two private helpers it needed. `renderAsState` now calls plain positional `rememberSaveable`, and the tests no longer depend on the key:

- `savesSnapshot` takes the registry’s single saved value instead of looking it up by name.
- `restoresSnapshot` hands back the golden snapshot for whatever key is asked for, so it still exercises the real restore path with the golden bytes.
- Added `savesAndRestoresSnapshotsOfSiblingRuntimesIndependently`, covering two runtimes in one composition.

The public API is unchanged; only the `internal` overload lost a parameter.

## Verification

Instrumented tests on an API 28 emulator, 57/57 in `:workflow-ui:compose` green, plus `assembleDebug` and `ktlintCheck`.

Because state persistence fails quietly, I ran two negative controls rather than trusting a green run:

- Forcing `initialSnapshot = null` fails all four restore-dependent tests, so they genuinely exercise restore.
- Compiling against Compose 1.9.5 (BOM 2025.11.01) with the old call and no suppression reproduces the `-Werror` failure; with this change the module compiles clean at 1.9.5.

One thing worth knowing: a shared explicit key did *not* break sibling restore, because `SaveableStateRegistry` keeps a list per key and consumes it in registration order. It works by ordering luck — exactly the fragility the deprecation is about.

### DIFF
--- a/workflow-ui/compose/src/androidTest/java/com/squareup/workflow1/ui/compose/RenderAsStateTest.kt
+++ b/workflow-ui/compose/src/androidTest/java/com/squareup/workflow1/ui/compose/RenderAsStateTest.kt
@@ -127,8 +127,7 @@ internal class RenderAsStateTest {
           scope = scope,
           props = Unit,
           interceptors = emptyList(),
-          onOutput = {},
-          snapshotKey = SNAPSHOT_KEY
+          onOutput = {}
         ).value
       }
     }
@@ -143,11 +142,12 @@ internal class RenderAsStateTest {
 
     composeRule.runOnIdle {
       val savedValues = savedStateRegistry.performSave()
-      println("saved keys: ${savedValues.keys}")
-      // Relying on the int key across all runtimes is brittle, so use an explicit key.
+
+      // rememberSaveable generates its key from the call site's position in the composition, so
+      // the test can't know it. renderAsState saves exactly one value, so just take that one.
       @Suppress("UNCHECKED_CAST")
       val snapshot =
-        ByteString.of(*((savedValues.getValue(SNAPSHOT_KEY).single() as State<ByteArray>).value))
+        ByteString.of(*((savedValues.values.single().single() as State<ByteArray>).value))
       println("snapshot: ${snapshot.base64()}")
       assertThat(snapshot).isEqualTo(EXPECTED_SNAPSHOT)
     }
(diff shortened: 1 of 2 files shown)


zach-klippenstein:
🤖 Closing: the explicit `snapshotKey` in `RenderAsState` is treated as a contract we are not changing, so the `@Suppress("DEPRECATION")` stays for now.

Recording the finding for whoever picks this up next: wrapping the call in `key(…) { }` is **not** a viable substitute. `rememberSaveable` derives its registry key from `currentCompositeKeyHashCode.toString(36)` when no explicit `key` is passed, so `key("workflow-snapshot")` only perturbs that hash and never puts the string into the `SaveableStateRegistry`. Verified on an emulator: `savesSnapshot` fails with `NoSuchElementException: Key workflow-snapshot is missing in the map`, and `restoresSnapshot` silently restores nothing.

Also worth knowing: at Compose 1.9.5 the deprecated call is a hard build failure under `-Werror`, so the suppression is load-bearing for #1558.

PR #1562, opened 2026-08-18
Suppress deprecated Kotlin Android sourceSets warnings [CLF-287]

zach-klippenstein:
Configuring the build logged 52 copies of this warning, one per Android module:

```
w: ⚠️ sourceSets collection in Kotlin Android is deprecated
Kotlin Source Sets collection in Android extension should not be used and is deprecated now.
Solution: Use source set alternative provided by Android Gradle Plugin: https://kotl.in/b2vftz
See https://youtrack.jetbrains.com/issue/KT-74451 for more details.
```

## Where they come from

All 52 have an identical stack trace (`./gradlew tasks --all --no-configuration-cache --stacktrace`):

```
at org.jetbrains.kotlin.gradle.dsl.KotlinAndroidProjectExtension.getSourceSets(KotlinProjectExtension.kt:357)
at com.rickbusarow.ktlint.KtLintPlugin.getSourceDirectories(KtLintPlugin.kt:205)
at com.rickbusarow.ktlint.KtLintPlugin.doApply(KtLintPlugin.kt:125)
```

`KtLintPlugin.getSourceDirectories()` finds source dirs with `extensions.findByType(KotlinSourceSetContainer::class.java)?.sourceSets`. On an Android module that resolves to `KotlinAndroidProjectExtension.sourceSets`, the getter KGP deprecated in [KT-74451](https://youtrack.jetbrains.com/issue/KT-74451). KGP reports the diagnostic once per project, hence one per Android module.

Nothing in this repo reads the deprecated collection — our only `kotlin.sourceSets` use is in `KotlinMultiPlatformConventionPlugin`, which configures `KotlinMultiplatformExtension`.

## Why suppression

`com.rickbusarow.ktlint` 0.2.2 is the newest release on both Maven Central and the Gradle Plugin Portal, so there's no upgrade that fixes it, and no change on our side can stop a third-party plugin from calling that getter. That leaves suppressing the diagnostic or replacing the ktlint plugin outright; this PR takes the first, with a comment saying when it can come back out.

The property targets that one diagnostic ID, so it won't hide anything else — including the same warning if our own build logic ever introduces it, which is the one real downside worth knowing about.

## Verification

- `./gradlew tasks --all --no-configuration-cache`: 52 warnings before, 0 after.
- Diffed the full set of `w:` lines before and after — the only difference is the removed line, no other warning changed.
- `./gradlew :workflow-ui:core-android:compileDebugKotlin :workflow-ui:core-android:ktlintCheck` succeeds and still registers the per-source-set tasks (`ktlintCheckMain`, `ktlintCheckAndroidTest`), confirming the plugin's source discovery is unaffected.

## Update: ktlint is not the only consumer — Dokka reads the getter too

While migrating ktlint to ktfmt (#1565) we found that KGP reports this diagnostic **once per project**, so ktlint (which hit the deprecated getter first during configuration) was masking a second consumer: Dokka's [`KotlinAdapter`](https://github.com/Kotlin/dokka/blob/v2.0.0/dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt) reads `kotlinExtension.sourceSets` on every Android module it documents. With ktlint gone, the 14 Dokka-published Android modules each still warn. Even the latest Dokka (2.2.0) still reads the deprecated getter, and no upstream Dokka issue tracks it yet.

So this suppression is needed regardless of the ktlint→ktfmt migration, and the two PRs are independent — either can merge first. The second commit here updates the `gradle.properties` comment to name both consumers so it stays accurate in any merge order.


handstandsam (approved):

zach-klippenstein:
🤖 Update from testing the ktfmt migration (#1565): this suppression is needed even after the ktlint plugin is removed. KGP reports the `SourceSetsAccessInAndroidExtension` diagnostic once per project, and ktlint was hitting the deprecated getter first — masking a second consumer, Dokka's `KotlinAdapter` ([KotlinAdapter.kt:92](https://github.com/Kotlin/dokka/blob/v2.2.0/dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt), still present in 2.2.0, the latest release). Without the suppression, the 14 Dokka-published Android modules each still warn. So this PR stands on its own regardless of merge order with #1565 — its second commit updates the `gradle.properties` comment to name both consumers.

PR #1564, opened 2026-08-19
Add Compose Multiplatform to the workflow-runtime module.

zach-klippenstein:
This is the first step toward a Compose-based workflow runtime: wiring up Compose Multiplatform (CMP) in `workflow-runtime` without breaking existing consumers.

## What changed

- Applied the `jetbrains.compose` and `compose.compiler` Gradle plugins to `workflow-runtime`.
- Added `jetbrains.compose.runtime` and `jetbrains.compose.runtime-saveable` as `commonMain` dependencies so they resolve to the correct platform artifact on each target (Android, JVM desktop, JS, iOS).
- Removed the Android Compose BOM and direct `androidx.compose.ui` dependency from `androidMain`; the CMP plugin now manages those via `jetbrains.compose.ui`.
- Registered catalog aliases for the new CMP libraries and plugins.

## Key design decision

When the JetBrains Compose plugin is applied to a KMP module, Gradle's variant resolution can become ambiguous for JVM configurations because CMP publishes both Android and desktop variants. To fix this, resolvable-only configurations (i.e. `isCanBeResolved && !isCanBeConsumed`) are attributed with `KotlinPlatformType.jvm` so Gradle consistently picks the desktop variant on JVM. Consumable configurations are intentionally excluded to avoid breaking consumers that don't request a platform type (e.g. SwiftPM lockfile tasks).

## Side effects

- Applying the Compose compiler plugin adds `$stable` fields to `@Stable`/`@Immutable` annotated classes; the Android and JVM `.api` files are updated accordingly.
- Transitive dependency snapshots updated to reflect CMP 1.9.3 artifacts and the JetBrains lifecycle/savedstate libraries it pulls in. Android Compose runtime downgraded from 1.9.5 → 1.9.4 (the version CMP 1.9.3 bundles).

## Testing

Verified that the build and dependency guard baselines pass. No behavioral changes — no new runtime code is introduced yet.


zach-klippenstein:
@codex review

_Comment posted automatically by [outer-loop](https://github.com/squareup/outer-loop)_

tcmulcahy (approved):

PR #1569, opened 2026-09-04
Wire test source sets into workflow-runtime's manual KMP hierarchy [CLF-453]

zach-klippenstein:
## Problem

`workflow-runtime/build.gradle.kts` configures the KMP source-set hierarchy manually (f30439f1f, "Configure source sets manually, so we can share one for JVM and Android"), which disables Kotlin's default hierarchy template. It wires the main side (`nativeMain → appleMain → iosMain → ios*Main`) but never the test side.

So `workflow-runtime/src/iosTest`, which contains `com.squareup.workflow1.internal.ThreadLocalTest`, is an orphan source set that no target compiles:

```
$ nm workflow-runtime/build/bin/iosSimulatorArm64/debugTest/test.kexe | grep -c ThreadLocalTest
0
```

Those tests have silently never run since February. `workflow-core` is unaffected because it still uses the default template.

## Fix

Mirror the hierarchy for tests: `nativeTest → appleTest → iosTest → ios*Test`, with `nativeTest` depending on `commonTest`. Because `iosTest` is now compiled, ktfmt also picks it up for the first time, so the second commit is the formatter's reformat of `ThreadLocalTest.kt` (no logic changes).

## Verification

`:workflow-runtime:iosSimulatorArm64Test` now produces a `ThreadLocalTest` result with all 4 tests passing, alongside the rest of the suite.

Linear: [CLF-453](https://linear.app/squareup/issue/CLF-453)

### DIFF
--- a/workflow-runtime/build.gradle.kts
+++ b/workflow-runtime/build.gradle.kts
@@ -111,6 +111,15 @@ kotlin {
     maybeCreate("iosX64Main").apply { dependsOn(iosMain) }
     maybeCreate("iosSimulatorArm64Main").apply { dependsOn(iosMain) }
 
+    // Test source sets mirror the main hierarchy. Without this, intermediate test source sets like
+    // iosTest are not compiled by any target and their tests silently never run.
+    val nativeTest = maybeCreate("nativeTest").apply { dependsOn(commonTest.get()) }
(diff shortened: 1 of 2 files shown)


zach-klippenstein:
🤖 Superseded by #1568, which now carries this change together with the other iOS test fixes as a single commit.

